PR #1, opened 2021-03-07
Fix off-by-one in equals (`=`)

alimpfard:
This would lead to `(= :whatever :something-else)` being `true`.
Also would lead to pretty much all tests passing regardless of correctness :P

### DIFF
--- a/pkg/rt/lang.go
+++ b/pkg/rt/lang.go
@@ -90,11 +90,13 @@ func installLangNS() {
 	})
 
 	equals, err := vm.NativeFnType.Wrap(func(vs []vm.Value) vm.Value {
-		if len(vs) < 1 {
+		length := len(vs)
+		if length < 1 {
 			// FIXME error out
 			return vm.NIL
 		}
-		for i := range vs[1:] {
+
+		for i := 1; i < length; i++ {
 			if vs[0] != vs[i] {
 				return vm.FALSE
 			}


pkg/rt/lang.go
```diff
@@ -95,7 +95,7 @@ func installLangNS() {
 			return vm.NIL
 		}
 		for i := range vs[1:] {
-			if vs[0] != vs[i] {
+			if vs[0] != vs[i + 1] {
```
alimpfard, line 98: Is this the go way? I'm a total go noob lol

nooga: I think this should probably be a normal `for i:=1; i < len ; i++` loop. See here: https://play.golang.org/p/hiXkN537a-v

nooga:
Thanks for the first PR @alimpfard 🎉 

These functions are here temporarily anyway since we don't have multi-arity fns yet. They will be replaced by primitive 2-arg operators and reductions over arguments will be written in let-go itself at some point.

alimpfard:
> first PR

lol it has to start _somewhere_.

> reduce over the args

what's stopping that from happening right now? just a reduction op?

pkg/rt/lang.go
alimpfard: I see...interesting.
sounds kinda silly to have "slices" and `range slice` but treat it as a normal array/list - one more go weirdness I suppose.

nooga:
> what's stopping that from happening right now? just a reduction op?

`reduce` is pretty important for almost everything in Clojure so will be adding some implementation soon.  Another thing is multi arity fns, see how `=` is defined in Clojure:

```clojure
(defn =
  ([x] true)
  ([x y] (clojure.lang.Util/equiv x y))
  ([x y & more]
   (if (clojure.lang.Util/equiv x y)
     (if (next more)
       (recur y (first more) (next more))
       (clojure.lang.Util/equiv y (first more)))
     false)))
```

It's using this syntax:
```clojure
(fn name
  ([] ...)
  ([a] ...)
  ([a b & rest] ...))
```
whic is currently missing in let-go :)

alimpfard:
Clojure has builtin pattern matching on function arguments? TIL!

nooga (approved):

PR #14, opened 2026-05-13
fix(require): match value-receiver vm.ArrayVector (#13)

### LINKED ISSUE #13: require invocation with vector arg fails
Thank you for working on and sharing this project!

Here's what I'm seeing:

```
;; Running ./lg
user=> (require '[clojure.string :as str])
[1;31merror:[0m require expected Symbol or Vector, got let-go.lang.ArrayVector
  [1;34m-->[0m REPL:1:1
   [1;34m|[0m
 [1;34m1[0m [1;34m|[0m (require '[clojure.string :as str])
   [1;34m|[0m [1;31m^^^[0m

[1mstack trace:[0m
  at require (REPL:1:1)
user=> (ns user (:require [clojure.string :as str]))
nil
user=> (str/split "a,b,c" ",")
("a" "b" "c")
```

So `(require '[clojure.string :as str])` fails while the `ns` handling works.

**Agent feedback**

If helpful, I used pi with Claude Sonnet 4.6 and asked for a repro/fix and it provided the following:

 ArrayVector is a plain Go slice type:

 ```go
   // pkg/vm/vector.go
   type ArrayVector []Value
 ```

 All its methods — RawCount, ValueAt, Type, etc. — are defined on the value receiver ArrayVector, not on *ArrayVector. So ArrayVector satisfies vm.Value directly.

 The require function is written with a type switch that expects a pointer:

 ```go
   // pkg/rt/lang.go
   switch arg := v.(type) {
   case vm.Symbol:
       NS(string(arg))
   case *vm.ArrayVector:         // ← pointer receiver
       // ... :as / :refer handling
   default:
       return vm.NIL, fmt.Errorf("require expected Symbol or Vector, got %s", v.Type().Name())
   }
 ```

 But the reader produces value receivers. When the reader sees '[string :as s], the quote expands to (quote [string :as s]), and [string :as s] is built by:

 ```go
   // pkg/compiler/reader.go
   result := vm.ArrayVector(ret)   // value type, not a pointer
 ```

 So v.(type) is vm.ArrayVector (value), which falls through the *vm.ArrayVector (pointer) case and hits default, producing the error message "expected Symbol or Vector, got let-go.lang.ArrayVector" — where the confusing part is that
 let-go.lang.ArrayVector is what it says it wants, just without the *.

 The ns macro form ((ns x (:require [string :as s]))) works because it's expanded by the reader/compiler at parse time and never goes through the runtime require function at all.

 The fix is a one-liner in lang.go — add case vm.ArrayVector: alongside (or instead of) case *vm.ArrayVector: and adjust the body to not dereference arg:

 ```go
   case vm.ArrayVector:
       if arg.RawCount() < 1 { ... }
       nsName := arg.ValueAt(vm.Int(0)).(vm.Symbol)
       // ... rest unchanged
 ```

SAY-5:
Fixes #13. The `require` runtime function type-switched on `*vm.ArrayVector` but the reader produces a value-receiver `vm.ArrayVector` (`type ArrayVector []Value`), so every vector form fell through to the `default` branch and returned `require expected Symbol or Vector, got let-go.lang.ArrayVector`.

Per @nooga in https://github.com/nooga/let-go/issues/13#issuecomment-4435057291 ("The fix looks good ... Happy to accept a PR for this one"), this PR drops the pointer cases and adds a regression test in `test/bugfix_test.lg`.

### DIFF
--- a/pkg/rt/lang.go
+++ b/pkg/rt/lang.go
@@ -3401,7 +3401,7 @@ func installLangNS() {
 			switch arg := v.(type) {
 			case vm.Symbol:
 				NS(string(arg)) // triggers autoloading
-			case *vm.ArrayVector:
+			case vm.ArrayVector:
 				// Vector form: [ns-name :as alias] or [ns-name :refer [syms...]]
 				if arg.RawCount() < 1 {
 					return vm.NIL, fmt.Errorf("require: empty vector")
@@ -3423,7 +3423,7 @@ func installLangNS() {
 					case vm.Keyword("refer"):
 						if val == vm.Keyword("all") {
 							cns.Refer(target, "", true)
-						} else if vec, ok := val.(*vm.ArrayVector); ok {
+						} else if vec, ok := val.(vm.ArrayVector); ok {
 							syms := make([]vm.Symbol, vec.RawCount())
 							for j := 0; j < vec.RawCount(); j++ {
 								syms[j] = vec.ValueAt(vm.Int(int64(j))).(vm.Symbol)
--- a/test/bugfix_test.lg
+++ b/test/bugfix_test.lg
@@ -43,3 +43,14 @@
       (doseq [x [1 2 3]]
         (swap! result conj x))
       (is (= [1 2 3] @result)))))
+
+;; Bug #13: vector form of (require ...) failed at runtime
+(deftest require-vector-form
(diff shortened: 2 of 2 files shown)


nooga:
Thanks @SAY-5!

nooga (approved):

PR #15, opened 2026-05-14
fix(bytecode): serialize #uuid constants for AOT compilation

olivergeorge:
UUIDType was on the compiler's const-emit whitelist (pkg/compiler/compiler.go) but the bytecode encoder and decoder had no case for *vm.UUID, so AOT-compiling any source containing a #uuid literal failed with "unsupported value type for serialization: *vm.UUID":

    echo '(prn #uuid "00000000-0000-0000-0000-000000000000")' > /tmp/x.lg
    lg -c /tmp/x.lgb /tmp/x.lg

Allocates TagUUID = 0x0B in the scalar tag block, adds the matching encoder/decoder arms and the *vm.UUID case in internStringsForValue. UUIDs are encoded as their canonical lowercase string via the existing string table -- the form already exposed by (*UUID).Unbox(). The tag is an append-only addition under bytecode format v1; older decoders error on encountering it.

Tests: TestLGBParity gains a "uuid-literal" case (compile -> encode ->
decode -> run -> compare to source-direct run), TestUUIDRoundtrip locks in the value codec, TestUUIDRoundtripCanonicalisesUppercase asserts the reader's lowercasing survives the round-trip.

### DIFF
--- a/pkg/bytecode/bytecode_test.go
+++ b/pkg/bytecode/bytecode_test.go
@@ -267,6 +267,43 @@ func TestVoidRoundtrip(t *testing.T) {
 	}
 }
 
+func TestUUIDRoundtrip(t *testing.T) {
+	cases := []string{
+		"550e8400-e29b-41d4-a716-446655440000",
+		"00000000-0000-0000-0000-000000000000",
+		"ffffffff-ffff-ffff-ffff-ffffffffffff",
+	}
+	for _, tc := range cases {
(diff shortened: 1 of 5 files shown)


nooga (approved):

nooga:
@olivergeorge could you rebase please? I'm happy to merge this.

olivergeorge:
I think this is merged now as 497e68976f on main

PR #18, opened 2026-05-14
feat(var): n-arg alter-var-root + read root, not dynamic binding

olivergeorge:
Completes `alter-var-root` with two changes, split across two commits for review hygiene:

**Commit 1 — `fix(var): alter-var-root reads root, not dynamic binding`**

The existing 2-arg native read `v.Deref()`, which returns the top of the dynamic binding stack when one is pushed. Clojure semantics require `alter-var-root` to read and write the *root* binding directly, regardless of any current dynamic binding. Adds `(*Var).Root()` accessor and switches the native to use it.

**Commit 2 — `feat(var): n-arg alter-var-root`**

Accepts `(alter-var-root v f & args)`, applying `(f root & args)` and setting the result as the new root. The 2-arg form continues to work unchanged. Matches Clojure's contract and unlocks idioms like `(alter-var-root #'global-hierarchy derive tag parent)` without falling back to atoms.

## Tests

New `test/alter_var_root_test.lg` covers 2-arg and n-arg forms, return value, type and arity errors, exception bubbling (root unchanged on throw), argument order (via a non-commutative op), and dynamic-binding bypass under both shapes.

## Note on atomicity

Clojure's `alter-var-root` guarantees an atomic read-modify-write. Because let-go evaluates synchronously today, this patch implements the semantic shape (bypassing dynamic bindings) but deliberately omits var-level synchronization. The read/apply/write in `pkg/rt/lang.go` will need a lock if/when let-go adds concurrent evaluation; an in-source TODO marks the spot.

## Out of scope

- README "Not implemented" list edit — left for a separate docs audit.
- `sync.Mutex` on `Var` — not added speculatively; let-go is single-threaded today.
- Switching `global-hierarchy` from atom to var-root — separate follow-up.

## Verification

- `go test ./...` green at HEAD and at the intermediate commit
- `go build ./...` clean

### DIFF
--- a/pkg/rt/lang.go
+++ b/pkg/rt/lang.go
@@ -5434,14 +5434,14 @@ func installLangNS() {
 	})
 	ns.Def("array?", isArrayf)
 
-	// alter-var-root — alter a var's root binding via (f root).
+	// alter-var-root — alter a var's root binding via (f root & args).
 	// Reads/writes the root, bypassing any current dynamic binding.
 	// TODO: the read/apply/write is not atomic. let-go evaluates synchronously
 	// today, so contention does not arise. If concurrent evaluation is added,
 	// centralize the RMW under Var-level synchronization.
 	alterVarRoot, _ := vm.NativeFnType.Wrap(func(vs []vm.Value) (vm.Value, error) {
-		if len(vs) != 2 {
-			return vm.NIL, fmt.Errorf("alter-var-root expects 2 args")
+		if len(vs) < 2 {
+			return vm.NIL, fmt.Errorf("alter-var-root expects at least 2 args")
 		}
 		v, ok := vs[0].(*vm.Var)
 		if !ok {
@@ -5452,7 +5452,7 @@ func installLangNS() {
 			return vm.NIL, fmt.Errorf("alter-var-root expects a function")
 		}
 		old := v.Root()
-		result, err := fn.Invoke([]vm.Value{old})
+		result, err := fn.Invoke(append([]vm.Value{old}, vs[2:]...))
 		if err != nil {
 			return vm.NIL, err
 		}
--- a/pkg/rt/lang.go
+++ b/pkg/rt/lang.go
@@ -5434,7 +5434,11 @@ func installLangNS() {
 	})
 	ns.Def("array?", isArrayf)
 
-	// alter-var-root — atomically alter a var's root binding
+	// alter-var-root — alter a var's root binding via (f root).
+	// Reads/writes the root, bypassing any current dynamic binding.
+	// TODO: the read/apply/write is not atomic. let-go evaluates synchronously
+	// today, so contention does not arise. If concurrent evaluation is added,
+	// centralize the RMW under Var-level synchronization.
 	alterVarRoot, _ := vm.NativeFnType.Wrap(func(vs []vm.Value) (vm.Value, error) {
 		if len(vs) != 2 {
 			return vm.NIL, fmt.Errorf("alter-var-root expects 2 args")
@@ -5447,7 +5451,7 @@ func installLangNS() {
 		if !ok {
 			return vm.NIL, fmt.Errorf("alter-var-root expects a function")
 		}
-		old := v.Deref()
+		old := v.Root()
 		result, err := fn.Invoke([]vm.Value{old})
 		if err != nil {
 			return vm.NIL, err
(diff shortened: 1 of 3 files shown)


nooga:
Go does not have the concept of thread locality which makes bindings and vars a little bit funny. I'm going to have a closer look at this one.

olivergeorge:
Parking this as draft.  I think it needs a revision.

olivergeorge:
While iterating on this PR I noticed `Var.bindings` looked like the
next domino in the same direction as `Var.root`, and took a side branch
to see where it led. 

Experience report below. Totally fine if the answer is "interesting,
parked."

## Aim

Move `Var.bindings` off `*Var` onto a per-execution context (sci-style /
Starlark `*Thread`-style) without fighting Go's runtime or churning
every `Fn` implementer.

A per-execution `*Execution` threaded through call frames and natives
gives JVM-Clojure-shaped semantics (sibling invisibility, explicit
conveyance, exception-safe `binding`, owner-checked `set!`) without
goroutine-ID hacks.

## Method

Prototype on
[`feat/dynamic-var-binding-conveyance`](https://github.com/olivergeorge/let-go/tree/feat/dynamic-var-binding-conveyance):
new `Execution` / `BindingFrame` / `TBox` types in `pkg/vm/exec.go`,
`*Var.DerefIn` / `SetIn` with a legacy fallback during transition,
`Frame.exec *Execution` threaded through `OP_LOAD_VAR`, four contract
tests in `pkg/vm/exec_test.go`. Code is the artefact.

## Results

### Code

6 commits ahead of `upstream/main`. Full `go test -race ./...` green,
including the existing `binding` integration tests.

| Contract test | Result |
| --- | --- |
| Sibling invisibility (32 goroutines × 500 iters) | pass |
| Throw-safe pop (defer-based) | pass |
| `set!` owner check (non-owner errors) | pass |
| Conveyance via `Fork()` | pass |

Plus four regression anchors (shared-TBox visibility, non-dynamic-var
rejection, concurrent-set isolation, threadBound fast-path) — all pass.

### Bugs surfaced

- `Var.bindings` is a shared `[]Value` mutated under `binding` from any
  goroutine → slice-header race → Go runtime crash (same shape as the
  `Var.root` race this PR addresses, not just a lost update).
- `binding` macro at `pkg/rt/core/core.lg:444-456` lacks `try/finally`;
  a throw inside the body leaks the binding globally.
- `bound-fn` / `bound-fn*` shimmed as `identity` / plain `fn` in the
  compat suite (`portability.lg:48-49`); conveyance to `future` not
  wired up.


### Decisions worth flagging

- **Compat over churn.** Added an `ExecAwareFn` extension interface
  alongside `Fn` rather than changing `Fn.Invoke` itself. The 21
  existing implementers don't churn. If `Fn` is internal to let-go,
  hoisting `InvokeWithExec` onto `Fn` is a tidier final shape; if it's
  a stable extension surface, the extension interface is permanent.
  Your call later.
- **Dual-truth window, intentional.** `Var.bindings` and the new
  context coexist via a fallback path in `Var.DerefIn`. Keeps the
  existing `binding` macro working while the new machinery proves
  itself. Has to collapse in a single commit (drop `Var.bindings` +
  drop fallback together) — otherwise the dual-truth window widens.
- **Fast path preserved.** `*Var.threadBound atomic.Bool` (sticky) lets
  var-deref skip the binding-frame walk when nothing's ever been bound.
  Same trick JVM Clojure uses.
- **Logical next steps, not yet done.** Rewire `push-binding!` /
  `pop-binding!` onto the new context (the point of no easy return);
  capture-and-convey at `go` / `future` / `pkg/rt/async.go` spawn sites
  (~10-12 sites); real `bound-fn` / `bound-fn*` + remove the compat
  shims.

## Conclusion

Direction is mechanically viable; the contract question is open. Two
apparent options:

- **Continue.** Land the migration commit, wire `go` / `future`
  conveyance, drop the `bound-fn` shims. Aligns let-go with JVM Clojure.
- **Park / lighter path.** Document `bound-fn` as unsupported
  (ClojureScript does this), fix the `Var.bindings` race + macro
  `try/finally` only, stop there.

AI-assisted work, as you'd expect by now — the code and tests are the
source of truth, not the write-up. Happy to take your lead. Easy to
park till you're ready. Goal is to match your taste for let-go if/when
you are — much better than pushing a shape you'd be working around
later.

nooga (approved):

PR #21, opened 2026-05-17
feat(core): data-side ident-tag hierarchies + non-class isa?

olivergeorge:
**Builds on:** the n-arg `alter-var-root` + root-binding read/write from #18 (merged).

---

Lands the data side of `clojure.core` hierarchies plus the non-class branch of `isa?`. The class-isa? branches (`isAssignableFrom`, `supers` walk) are out of scope — let-go has no Java class model. Multimethod-dispatch integration is also out of scope; that's the natural follow-up.

Two commits:

- `feat(core): data-side hierarchies + non-class isa?` — `make-hierarchy`, `derive`, `underive`, `parents`, `ancestors`, `descendants`, `isa?` as native let-go functions; replaces the test-side stubs at `test/compat/clojure/core-test/portability.lg`.
- `chore(rt): regenerate core_compiled.lgb` — mechanical AOT bytecode regen.

## Implementation notes

**Storage of the global hierarchy.** Held as a `^:private` non-dynamic var root (`(def ^:private global-hierarchy (make-hierarchy))`), mutated through `alter-var-root` — matches Clojure JVM directly. Uses the n-arg `alter-var-root` with root-binding read/write semantics that landed in #18. Thread-level atomicity is bounded by let-go's single-threaded evaluator today, same caveat #18 carries.

**Validation.** `when-not` + `throw` rather than `assert`, so thrown-ness is predictable independent of let-go's assertion semantics. Error messages match Clojure's wording where they can be exercised from compat tests, but no test asserts exact wording.

**Tag policy.** Keywords and symbols only. The `(class? tag)` branch in Clojure's source is dropped — let-go has no Java classes, no compat test asks for it, and the same rationale applies to `isa?`. The 2-arg `derive`/`underive` forms enforce namespace-qualified idents because the global hierarchy is shared mutable state across libraries; namespacing prevents collisions. The 3-arg forms operate on a local hierarchy map the caller owns, so the choice is theirs (matching Clojure's behaviour — see <https://ask.clojure.org/index.php/14759/>).

**`underive` rebuild.** Uses a nested reduce over the parents map rather than Clojure's `flatten` / `interpose` / `partition` trick. Behaviourally identical on the diamond case; avoids any future hazard if seqable tag types are added.

**`isa?` semantics.** `(= child parent)` OR `parent ∈ (ancestors child)` in the given (or global) hierarchy, OR — when both arguments are same-length vectors — element-wise tuple `isa?` (empty tuples are `isa?` to each other). The class-related branches in Clojure's `isa?` (direct `isAssignableFrom` and the `supers` walk for child classes) are not implemented; let-go has no host class model. Real Clojure libs that do `(derive Object ::any)` and rely on `(isa? "abc" ::any)` won't work here, and would need that host-type story before they could.

## Tests

`test/hierarchies_test.lg` — 31 deftests / 133 assertions covering:

- `make-hierarchy` shape; freshness
- Linear chains, diamonds, multiple parents
- `derive` idempotency, cycle detection, "already has X as ancestor" rejection
- `underive` round-trip; diamond preservation when removing one of two parents
- 1-arg lookup garbage inputs (`nil`, numbers, strings, vectors, maps) for `parents`/`ancestors`/`descendants`
- Global-hierarchy round-trips under both `derive`/`underive`, using namespace-disjoint test tags so a mid-test failure cannot poison the next test
- `isa?`: identity, direct edge, transitive ancestor, diamond, unrelated tags, global form, vector-tuple element-wise (including empty tuples, nested tuples, and asymmetric nested-vs-scalar shapes that fall through to the equality branch), garbage input

## Compat impact

| File | Status |
|---|---|
| `underive.cljc` | 26/26 ✅ — lifted from `knownFailing` |
| `derive.cljc` | 2/3 `are` rows pass; the third row uses `String` via the `:default` reader-conditional branch of the vendored `jank-lang/clojure-test-suite`. Same lift condition as ancestors/parents/descendants. |
| `ancestors.cljc` / `parents.cljc` / `descendants.cljc` | Still file-level fail at top-level `(def AncestorT Object)`. Won't lift without the upstream suite adding a `:lg` reader-conditional branch or let-go growing host-type derivation. Comments updated to reflect that. |
| `isa?.cljc` | not vendored |

## Out of scope

- **Multimethod-dispatch integration** — making `MultiFn.Invoke` walk ancestors via `isa?` and respect `prefer-method` is the natural follow-up.
- **Class-tag derivation and class-isa?** — let-go has no JVM class system. Adding host-type derivation would need its own design and will surface again when the multimethod-dispatch PR lands.

## Verification

- `go test ./...` green
- `go build ./...` clean
- `make generate` clean (AOT regen is the trailing commit)


olivergeorge:
Pulling this back for additional polish based on internal review (lens-pass surfaced PR body framing inaccuracies and an idiom-shape nit). Will re-open as a new PR once tidied — branch `feat-hierarchies-data` retained, no maintainer action needed.

olivergeorge:
Superseded by #24 — tried to reopen this one rather than start fresh, but GitHub wouldn't (the branch was force-pushed during polish and the original head SHA is no longer the branch tip). Same head ref, squashed-and-trimmed diff.

PR #24, opened 2026-05-17
feat(core): data-side ident-tag hierarchies + non-class isa?

olivergeorge:
**PR title:** `feat(core): data-side ident-tag hierarchies + non-class isa?`

**Builds on:** the n-arg `alter-var-root` + root-binding read/write from #18 (merged).

---

Lands the **data-ident subset** of `clojure.core` hierarchies — keyword/symbol tags only — plus the non-class branch of `isa?`. The class-isa? branches (`isAssignableFrom`, `supers` walk) are out of scope — let-go has no Java class model. Multimethod-dispatch integration is also out of scope; that's the natural follow-up.

## Implementation notes

**Storage of the global hierarchy.** Held as a `^:private` non-dynamic var root (`(def ^:private global-hierarchy (make-hierarchy))`), mutated through `alter-var-root` — matches Clojure JVM directly. Uses the n-arg `alter-var-root` with root-binding read/write semantics that landed in #18. Thread-level atomicity is bounded by let-go's single-threaded evaluator today, same caveat #18 carries.

**Validation.** `when-not` + `throw` rather than `assert`, so thrown-ness is predictable independent of let-go's assertion semantics. Error messages match Clojure's wording where they can be exercised from compat tests, but no test asserts exact wording.

**Tag policy.** Keywords and symbols only. The `(class? tag)` branch in Clojure's source is dropped — let-go has no Java classes, no compat test asks for it, and the same rationale applies to `isa?`. The 2-arg `derive`/`underive` forms enforce namespace-qualified idents because the global hierarchy is shared mutable state across libraries; namespacing prevents collisions. The 3-arg forms operate on a local hierarchy map the caller owns, so the choice is theirs (matching Clojure's behaviour — see <https://ask.clojure.org/index.php/14759/>).

**`underive` rebuild.** Mirrors `clojure.core/underive`'s rebuild strategy: remove the direct edge, then replay every remaining edge into a fresh hierarchy via public `derive`. Aligns with upstream for the same perf/simplicity trade Halloway accepted in 2010 ([discussion](https://groups.google.com/g/clojure/c/C3D4vxbeIhY)). Uses a nested reduce over the parents map directly rather than Clojure's `flatten` / `interpose` / `partition` trick — avoids any future hazard if seqable tag types are added. A latent throw on rebuild of legally-constructed redundant edges affects both this implementation and `clojure.core/underive`; see `pkg/rt/core/core.lg` underive comment for context and `backlog/research/260517-1530-jvm-redundant-edge-repro/` for the repro.

**`isa?` semantics.** `(= child parent)` OR `parent ∈ (ancestors child)` in the given (or global) hierarchy, OR — when both arguments are same-length vectors — element-wise tuple `isa?` (empty tuples are `isa?` to each other). The class-related branches in Clojure's `isa?` (direct `isAssignableFrom` and the `supers` walk for child classes) are not implemented; let-go has no host class model. JVM `isa?`'s class branch operates on class *values* (e.g. `(isa? String Object)`), not on instances — multimethods typically reach it via `(class x)` dispatch values. Real Clojure libs that dispatch on classes (e.g. `(defmulti m class)` plus `(derive String ::any)`) won't work here, and would need a host-type story before they could.

## Tests

`test/hierarchies_test.lg` — 31 deftests / 133 assertions covering:

- `make-hierarchy` shape; freshness
- Linear chains, diamonds, multiple parents
- `derive` idempotency, cycle detection, "already has X as ancestor" rejection
- `underive` round-trip; diamond preservation when removing one of two parents
- 1-arg lookup garbage inputs (`nil`, numbers, strings, vectors, maps) for `parents`/`ancestors`/`descendants`
- Global-hierarchy round-trips under both `derive`/`underive`, using namespace-disjoint test tags so a mid-test failure cannot poison the next test
- `isa?`: identity, direct edge, transitive ancestor, diamond, unrelated tags, global form, vector-tuple element-wise (including empty tuples, nested tuples, and asymmetric nested-vs-scalar shapes that fall through to the equality branch), garbage input

## Compat impact

| File | Status |
|---|---|
| `underive.cljc` | 26/26 ✅ — lifted from `knownFailing` |
| `derive.cljc` | 2/3 `are` rows pass; the third row uses `String` via the `:default` reader-conditional branch of the vendored `jank-lang/clojure-test-suite`. Same lift condition as ancestors/parents/descendants. |
| `ancestors.cljc` / `parents.cljc` / `descendants.cljc` | Still file-level fail at top-level `(def AncestorT Object)`. Won't lift without the upstream suite adding a `:lg` reader-conditional branch or let-go growing host-type derivation. Comments updated to reflect that. |
| `isa?.cljc` | not vendored |

## Out of scope

- **Multimethod-dispatch integration** — making `MultiFn.Invoke` walk ancestors via `isa?` and respect `prefer-method` is the natural follow-up.
- **Class-tag derivation and class-isa?** — let-go has no JVM class system. Adding host-type derivation would need its own design and will surface again when the multimethod-dispatch PR lands.

## Verification

- `go test ./...` green
- `go build ./...` clean
- `make generate` clean (AOT regen is the trailing commit)


nooga:
Closing this as superseded by the v2.0.0 compatibility work now on main.

The hierarchy/isa? functionality landed in a native implementation in pkg/rt/hierarchy.go, with type/record/protocol hierarchy support and the Clojure compat suite passing 5522/5522 through the :clj lens.

Thanks for the detailed writeup and test coverage here. Some of the edge-case tests may still be worth folding into the current hierarchy tests separately. Please open a fresh PR based on current main if you want to carry over hierarchy tests; that should target test/hierarchy_test.lg and validate the native implementation we now have, rather than rebasing this older core.lg implementation branch.

This branch itself is now stale/conflicting and would regress the implementation shape.

PR #35, opened 2026-05-19
fix: map/mapcat laziness and Cons/LazySeq hash compliance

nnunley:
Fixes two related laziness regressions:

1. `map*` had an eager fast-path for counted collections of ≤32 elements, returning a fully-realized `PersistentList` instead of a `LazySeq`. Broke rose-tree shrinking, `take` over `(map f infinite)`, and any side-effect-on-realization contract. Removed the fast-path entirely so `map` is always lazy. Use `mapv` for eager-to-vector.

2. `mapLazy1`/`mapLazyN` didn't force their captured `LazySeq` before reading `First()`. `(map f (lazy-seq nil))` produced a non-empty `LazySeq` that invoked `f(nil)` on realization. Added a `forceSeq` helper that resolves the captured seq inside the thunk.

Fixing eagerness exposed a latent bug: `*Cons` and `*LazySeq` had no `Hash()` method and fell back to string hashing, so equivalent seqs hashed differently from `*List`. Broke `(contains? #{'(:b :d) '(:d :b)} (vals m))` because `vals` returns a `LazySeq`. Added `Hash` methods using `hashOrdered` on both.

## Tests
- `test/lazy_map_test.lg` — 10 regression tests (16 assertions) for eagerness, short-circuit via `take`, multi-coll laziness, and empty `lazy-seq` inputs to `map`/`mapcat`.
- `test/clojure_sequences_compat_test.lg` — ports laziness and seq-equality assertions from [`clojure/test/clojure/test_clojure/sequences.clj`](https://github.com/clojure/clojure/blob/master/test/clojure/test_clojure/sequences.clj) (EPL 1.0).

## Verification
Full `TestRunner` green; `TestClojureTestSuite` 232 files / 5621 assertions / 0 fail.

---
This is PR 1 of a 2-PR stack. PR 2 (#36) fixes a separate `lazy-seq` empty-`Sequable` realization bug surfaced by the new compat tests; it sits on top of this branch.


nooga:
This is good but it surfaces a problem with `Unbox` exposing lazy seqs in the Go land. I'm wondering what to do about this - we could make `Ubnox` realize lazy sequences and keep the interop smooth, or add some utility to realize on demand after unboxing, i.e.:

```go
func SeqToSlice(v Value) (out []Value, err error) {
	defer func() {
		recoverThrownPanic(&err)
		if err != nil {
			out = nil
		}
	}()

	if v == nil || v == NIL {
		return nil, nil
	}

	var s Seq
	switch x := v.(type) {
	case Seq:
		s = x
	case Sequable:
		s = x.Seq()
	default:
		return nil, NewTypeError(v, "is not a sequence", SequenceType)
	}

	for s != nil && s != EmptyList {
		out = append(out, s.First())
		s = s.Next()
	}

	return out, nil
}
```

I'm not sure what would be more desirable here. @nnunley do you have any thoughts about this?

nnunley:
Two things, separable.

## Process miss

The `Makefile` `test` target runs `go test ./test` only, so I didn't catch that this PR breaks `pkg/api/interop_test.go::TestSimple` line 24:

```go
{"(map inc [1 2 3])", []vm.Value{vm.Int(2), vm.Int(3), vm.Int(4)}},
...
assert.Equal(t, z.result, v.Unbox(), z.code)
```

After the PR, `(map inc [1 2 3])` returns `*LazySeq` whose `Unbox()` returns `*Cons`, not `[]vm.Value`. `go test ./...` (what CI runs) catches it; the Makefile target doesn't. Independent fix: change `test:` to `go test ./...` so the local loop matches CI. Happy to send that as a small separate PR.

Related break: `pkg/vm/native_func.go:40` uses `Unbox()` + `reflect.Convert` to pass let-go args to Go funcs. Before: `(go-fn (map inc xs))` works because `(map …)` unboxes to `[]vm.Value` → convertible to `[]int`. After this PR: same call fails with `reflect: Call using *vm.Cons as type []int`. Same root cause.

## Design

Your two options solve different things:

- **Eager `Unbox`** restores the existing contract (and the test, and `native_func.go`) but re-introduces the eager-by-default trap at a different layer — pass an infinite seq through `Unbox()` and you hang.
- **`SeqToSlice` helper** keeps `Unbox` lazy and adds an explicit realization escape hatch. Doesn't fix existing callers automatically, including the `native_func.go` bridge.

Given the bidirectional-Go-compat goal, I'd propose layering three pieces.

**1. `vm.Iter(v Value) iter.Seq[Value]` as a free function.** Go 1.23+ idiom (`go.mod` is on 1.26, so available). Mirrors `slices.All` / `maps.Keys` in the stdlib — free function returning an `iter.Seq`, no naked type-assertion at the call site:

```go
for v := range vm.Iter(result) {
    // ...
}
```

Internally it does `if sq, ok := v.(Sequable); ok { ... }` (existing pattern from `pkg/rt/ions.go:267`) and yields nothing for non-seqs. Preserves laziness, composes with `slices.Collect`, `slices.Filter`. Native modern-Go shape for "potentially lazy sequence" — which is what a let-go `Seq` *is*. Roughly:

```go
func Iter(v Value) iter.Seq[Value] {
    return func(yield func(Value) bool) {
        sq, ok := v.(Sequable)
        if !ok { return }
        for s := sq.Seq(); s != nil && s != EmptyList; s = s.Next() {
            if !yield(s.First()) { return }
        }
    }
}
```

**2. Auto-realize at the reflect-Convert bridge** in `native_func.go:40` and the `Unbox()` fallback in `struct_mapping.go:336`. When the target Go param is a slice/array kind, detect `Seq`/`Sequable` and walk it before reflecting. Fixes the `(go-fn (map inc xs))` regression without changing the broader `Unbox()` contract:

```go
if in.Kind() == reflect.Slice || in.Kind() == reflect.Array {
    if sq, ok := args[i].(Sequable); ok {
        rawArgs[i] = reflect.ValueOf(seqToValueSlice(sq.Seq()))
    } else if s, ok := args[i].(Seq); ok {
        rawArgs[i] = reflect.ValueOf(seqToValueSlice(s))
    } else {
        rawArgs[i] = reflect.ValueOf(args[i].Unbox())
    }
}
```

**3. `SeqToSlice(v Value) ([]Value, error)`** roughly per your sketch. One-off realization when a Go caller has a `vm.Value` and explicitly wants the slice. Useful even with (1) and (2) in place.

## Question: should `(*LazySeq).Unbox()` auto-realize?

Two paths:

- **Auto-realize.** `(*LazySeq).Unbox() → []Value`, same shape as `*List.Unbox()`. Restores the existing contract. Existing user code (`v.Unbox().([]vm.Value)`) keeps working. Tradeoff: hostile to infinite seqs via Unbox — users learn to use `vm.Iter` for that case.
- **Keep lazy.** Force migration to `vm.Iter` / `SeqToSlice`. Stronger architectural stance, breaks user code that already exists.

I lean towards keeping it lazy. There's a lot of higher-order Clojure code that generates effectively infinite lists, and silent eager realization on `Unbox()` is a time bomb.

Downstream check before recommending this: `pkg.go.dev` reports 0 known importers, GitHub network-dependents shows 0, code search for the `github.com/nooga/let-go` import path returns only this repo, and all 21 forks are 0 commits ahead of main. The migration cost from "keep lazy" looks bounded.

If you pick auto-realize, the bridge fix in (2) is still needed — `[]vm.Value` doesn't reflect-convert to `[]int` without per-element conversion (would use the existing `unboxInto` machinery from `struct_mapping.go`).

## Practical

Once I have your preference, I'll amend this PR and land anything elsewhere that needs it.

nnunley:
Correction on the "process miss" section — the Makefile is fine. The README at line 298 says `go test ./... -count=1 -timeout 30s` is the canonical test command. `make test` is the lightweight inner-loop target, not the comprehensive one. I used the wrong target; that's a process error on my end, not a Makefile bug.

The two real test breaks (`pkg/api/interop_test.go::TestSimple` and `pkg/vm/native_func.go` reflect bridge) stand. Everything else in the comment above is unchanged.

nnunley:
Pushed a concrete sketch of the "keep Unbox lazy" path as a follow-up commit on top of the original fix. Reads as a proposal, not a decision — happy to revert and rework as "auto-realize" if you'd rather go that direction.

Adds (`47a2bb9`):

- `vm.Iter(v) iter.Seq[Value]` and `vm.SeqToSlice(v)` for Go-side consumption
- `vm.SeqIsEmpty` / `vm.SeqValues` helpers (encapsulate the nil-or-EmptyList sentinel pattern)
- Auto-realize at the reflect-Convert bridge in `pkg/vm/native_func.go` so `(go-fn (map inc xs))` with `go-fn` taking `[]int` works — uses the existing `unboxSliceInto` machinery for per-element conversion
- Fixes a latent EmptyList bug in `unboxSliceInto` (loop missed the EmptyList sentinel; predated PR #35)
- Migrates `pkg/api/interop_test.go::TestSimple` to use the new helpers (so CI goes green)
- `examples/go_interop_lazy/` — 5-demo worked example covering: infinite primes with early break, bounded `SeqToSlice`, Go fn inside a lazy pipeline (with a realization counter proving laziness), `(take 100 infinite)` → Go `[]int`, Go predicate in `(filter …)` consumed from Go

What `Unbox()` returns is unchanged — `*LazySeq.Unbox()` and `*Cons.Unbox()` still expose the seq as-is. The `vm.Iter` / `SeqToSlice` helpers are the new opt-in surface; callers that want eager `[]Value` use `SeqToSlice` explicitly.

Verification: full Go test suite green, `TestClojureTestSuite` 232 / 5621 / 0.

Open to:
- Switching to auto-realize Unbox if you prefer (most of the sketch becomes redundant; `vm.Iter` and `SeqIsEmpty` stay useful)
- Splitting any of these out (the EmptyList fix is independent of the design call)
- Whatever shape works for landing.

nooga:
I'm happy with this solution - I haven't thought about Go iters myself but I think this is the right shape. I'm going to merge this one! Thanks @nnunley

nooga (approved):

PR #37, opened 2026-05-19
feat: add check ns — QuickCheck-style property testing

nnunley:
**Stacked on #35 and #36** — merge those first.

Adds `pkg/rt/core/check.lg`, a small QuickCheck-style property testing library: splittable LCG RNG, rose-tree shrinking, generator combinators (`gen-int`, `gen-bool`, `gen-elements`, `gen-vector`, `gen-tuple`, `gen-frequency`, `bind`, `fmap`), and a property runner with `defprop`/`for-all`/`check`.

```clojure
(require '[check :as c])

(c/defprop addition-commutes
  [a (c/gen-int -100 100)
   b (c/gen-int -100 100)]
  (= (+ a b) (+ b a)))
```

Depends on #35 (lazy `map`/`mapcat`, `Cons`/`LazySeq` `Hash`) for correct shrink-tree composition. Originally developed against let-go in a downstream project; the laziness fixes were extracted as #35/#36 first.

## Tests
`test/check_test.lg` — 39 tests (69 assertions) covering RNG determinism and split, rose-tree map/filter, integer shrinking, all generators, counterexample formatting, and the runner (pass/fail/shrink/exception paths). Includes a smoke-test `defprop` invocation.

## Verification
`TestRunner` green; `TestClojureTestSuite` 232 files / 5621 assertions / 0 fail.

### DIFF
--- a/pkg/resolver/resolver.go
+++ b/pkg/resolver/resolver.go
@@ -158,6 +158,8 @@ func (r *NSResolver) loadEmbedded(name string) *vm.Namespace {
 		src = rt.ZipSrc
 	case "data":
 		src = rt.DataSrc
+	case "check":
+		src = rt.CheckSrc
 	case "term":
 		// term is a pure Go namespace, already registered in init()
 		return rt.NS("term")
--- a/pkg/rt/check.go
+++ b/pkg/rt/check.go
@@ -1,0 +1,6 @@
+package rt
+
+import _ "embed"
+
+//go:embed core/check.lg
+var CheckSrc string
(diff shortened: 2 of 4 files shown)


nnunley:
Restacked. This PR only depends on #35, not on #36 — the `check.lg` library uses `lazy-seq` only in non-empty-Sequable patterns, so it doesn't exercise the empty-Sequable normalization that #36 fixes. Verified by rebasing this branch directly onto #35 and running the full `check_test.lg` suite (39 tests / 69 assertions / 0 fail).

Practical effect: #35 and this PR can land independently of #36's design discussion. The branch is now based on `push-xmxpzzlnozqy` (PR #35) directly.

nooga (approved):

PR #43, opened 2026-05-19
fix: COI bootstrap works on Safari/Firefox, no reload loops

mparrett:
Fixes two compounding bugs in the WASM template's `cross-origin-isolation` bootstrap.

Symptoms: pages flashing between "Interactive Input Unavailable" and "loading..." on Safari and Firefox mobile. See [nooga/xsofy#3](https://github.com/nooga/xsofy/issues/3) for the user-visible report.

## Causes

1. `coi-serviceworker.js` sets `Cross-Origin-Embedder-Policy: credentialless`. That value is Chrome-only — Safari and Firefox ignore it, so `crossOriginIsolated` never becomes true and `SharedArrayBuffer` stays gated.
2. The boot HTML does `register(...).then(() => location.reload())` with no loop guard. When (1) means isolation can never activate, every reload re-enters the same path, leading to  infinite reload.

| Browser | `credentialless` | `require-corp` |
|---------|------------------|----------------|
| Chrome  | accepted         | accepted       |
| Safari  | ignored          | accepted       |
| Firefox | ignored          | accepted       |

Support for `credentialless`:
image

https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/IFrame_credentialless

## Changes

**wasm.go:**
- HTML bootstrap: register the SW at most once per tab (sessionStorage guard), and proactively unregister any stale SW when the page is already cross-origin-isolated from server headers.
- Service worker: only inject COOP/COEP when the response doesn't already carry them (so dev servers that send headers themselves aren't clobbered), and use `require-corp` instead of `credentialless` when injecting.

## Testing

- Chrome (macOS): boots to title screen on first visit; `crossOrigin Isolated === true`; SW auto-unregisters when server provides headers.
- iOS Safari: boots to title screen on first visit; no reload loop.

Cannot test Android Firefox directly (no device); the header-compatibility table above and the loop-guard logic should cover it, but confirmation from the original reporter would be welcome.

## Recovery for clients with the old SW cached

`skipWaiting` + `clients.claim` in the new SW swaps it in on next visit; the HTML's auto-unregister branch then removes it for good once headers isolate the page. Users who want to force the recovery immediately can clear site data once.

## other notes

- `./let-go/wasm.go`: the only file modified in the let-go PR.
- `./let-go/.git/refs/heads/expmt/js-bridge`:  source of the two cherry-picked SHAs (`f3ed218`, `44812e9`).
- `./xsofy/.github/workflows/deploy-pages.yml`: *not modified*, but referenced: confirms `let-go@latest` will pull the fix once merged. Worth considering pinning to a specific SHA instead (out of scope for this plan).
- This PR does not bump the `let-go@latest` pin in xsofy's workflow (no need; `@latest` resolves once the fix is merged and tagged).

## AI Disclosure

Claude Opus 4.7 was used extensively to debug and fix this issue.

### DIFF
--- a/wasm.go
+++ b/wasm.go
@@ -111,8 +111,21 @@ const wasmHTMLTemplate = `<!doctype html>
   <script src="https://cdn.jsdelivr.net/npm/@xterm/xterm@5.5.0/lib/xterm.min.js"></script>
   <script src="https://cdn.jsdelivr.net/npm/@xterm/addon-fit@0.10.0/lib/addon-fit.min.js"></script>
   <script>
-// --- Cross-origin isolation via Service Worker ---
-if (!crossOriginIsolated && window.isSecureContext && 'serviceWorker' in navigator) {
+// --- Cross-origin isolation: prefer server headers, fall back to SW ---
+// When the dev/host server sends the COI headers itself (dev/serve.json
+// does this), crossOriginIsolated is already true and we don't need the
+// SW. Any leftover SW from a prior visit (e.g. earlier GitHub Pages load)
+// would intercept future fetches with stale content — unregister it now
+// so the headers path stays clean.
+if (crossOriginIsolated && 'serviceWorker' in navigator) {
+  navigator.serviceWorker.getRegistrations().then(rs => rs.forEach(r => r.unregister())).catch(()=>{});
+}
+// No isolation? Register the SW shim — but only once per tab. Without a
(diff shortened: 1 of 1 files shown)


nooga (approved):

nooga:
Thanks @mparrett! I had no idea how to fix this one :)

Issue #49, opened 2026-05-20
Running yamlscript atop let-go


VarLad:
Hey, is it possible to run https://github.com/yaml/yamlscript atop let-go?

nnunley:
Short answer: **partially yes, today**, via `ys --compile`. A full embedded port is structurally feasible but not free.

I dug into this because I was curious how close it actually is. Here's what I found, plus three PRs that move the answer from "no" to "yes for a useful subset."

## Three paths, in increasing ambition

### 1. `ys --compile` + a small runtime shim (works today)

`ys --compile foo.ys` emits plain Clojure. With a ~25-line `ys-runtime.lg` providing the YS stdlib symbols let-go doesn't know (`say`, `ARGS`, `%`, `sum`, ...), the compiled output runs unmodified on let-go.

```bash
ys --compile foo.ys > foo.clj
{ echo "(require '[ys-runtime :refer [say ARGS sum %]])"; cat foo.clj; } > foo.lg
lg -source-paths /path/to/runtime foo.lg
```

I ran four progressively-richer YS programs through this pipeline and all matched native `ys` output:

| Program | Result |
|---|---|
| `defn main(): say: "hello"` | ✅ identical |
| FizzBuzz with `cond`/`doseq`/`range` | ✅ identical |
| `defn classify(n)` with nested `cond`, `pos?`/`neg?`, string interpolation | ✅ identical |
| YAML data-mode (`name: Fido\nage: 6\nlikes: [running, treats]`) → `(% "name" "Fido" ...)` | ✅ shimmed `%` as `hash-map`, loads correctly (order not preserved) |

Submitted as **#53** (`examples/ys-on-let-go/` — shim + worked examples + README).

### 2. Same approach for pure data-mode YAML loading

`ys --compile data.yaml` emits `(% "name" "Fido" "age" 6 "likes" ["running" "treats"])`. Aliasing `%` to `hash-map` (or to a real ordered-map type if let-go adds one) gives you a YAML→Clojure-data loader on let-go without any runtime machinery.

### 3. Port SCI to let-go (the "yamlscript actually runs on let-go" path)

This is what I initially assumed was a non-starter, but on closer look it's tractable:

- **SCI is `.cljc`**, not `.clj`. 41 of 45 source files are cross-platform with reader conditionals (444 `:clj` branches, 118 `:cljs` branches). It already runs on Clojure JVM, GraalVM native, ClojureScript, and JS.
[...]

PR #53, opened 2026-05-20
examples: ys-on-let-go — run yamlscript compiled output on let-go

nnunley:
## Summary

Adds `examples/ys-on-let-go/` showing how the output of `ys --compile` runs on let-go with a small `ys-runtime.lg` shim providing the YS stdlib symbols (`say`, `ARGS`, `sum`, `%`, ...) let-go doesn't ship.

Four worked examples: hello, fizzbuzz, conditional classify, and YAML data-mode loading. The `.lg` files are checked in so reviewers can run the examples without `ys` installed; `build.sh` regenerates them from the `.ys` / `.yaml` sources.

```bash
cd examples/ys-on-let-go
lg -source-paths . 01_hello.lg     # → hello from let-go via yamlscript
lg -source-paths . 02_fizzbuzz.lg  # → 1 2 Fizz 4 Buzz Fizz 7 8 Fizz Buzz ...
lg -source-paths . 03_classify.lg  # → -2 is negative ... 2 is positive
lg -source-paths . 04_data.lg      # → {"name" "Fido" ...}
```

All four examples produce identical output to native `ys`.

## Scope

This is a docs/examples-only PR — no core changes. Independent of #51 and #52. The README links #49 and is explicit about what the shim does and doesn't cover (no `babashka.fs`/`process`/`http-client`, no `java-time`, no ordered maps).

## Test plan

- [x] All four `.lg` files run cleanly with `lg -source-paths .`
- [x] Output matches `ys` native run byte-for-byte for examples 01–03
- [x] `build.sh` regenerates the `.lg` files; bash syntax checked with `bash -n`

Context: #49.

### DIFF
--- a/examples/ys-on-let-go/ys-runtime.lg
+++ b/examples/ys-on-let-go/ys-runtime.lg
@@ -1,0 +1,52 @@
+;
+; Copyright (c) 2026 Marcin Gasperowicz <xnooga@gmail.com>
+; SPDX-License-Identifier: MIT
+;
+
+;; ys-runtime.lg — minimal shim namespace that lets the output of
+;; `ys --compile` run on let-go. yamlscript-compiled Clojure refers to
+;; symbols from the YS stdlib (say, sum, %, ARGS, ...) which are not
+;; part of let-go. This namespace provides enough of them to run
+;; small YS programs unmodified.
+;;
+;; Coverage: pure-Clojure subset of YS. No babashka.fs / process /
+;; http-client, no java-time, no ordered maps. Suitable for scripts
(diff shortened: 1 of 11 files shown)


examples/ys-on-let-go/ys-runtime.lg
```diff
@@ -0,0 +1,52 @@
+;
+; Copyright (c) 2026 Marcin Gasperowicz <xnooga@gmail.com>
```
nooga, line 2: @nnunley I think you should put your name here :)

nooga (approved):

PR #60, opened 2026-05-21
Some improvements to Makefile and report generator

ingydotnet:
Makefile now installs a go binary as .cache/local/go-1.26.3/bin/go if user doesn't have go installed.

`make distclean` adds .cache/ to the clean targets.

Some minor bash adjustments. Still passes shellcheck.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,9 @@
 /vendor/
 /Godeps/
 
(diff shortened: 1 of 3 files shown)


nooga:
Thanks @ingydotnet!

nooga (approved):

PR #61, opened 2026-05-21
feat(gogen): switch, composite-lit, func-lit, plus top-level decl walkers

nnunley:
## Summary

Builds on [#57 (gogen-v1)](https://github.com/nooga/let-go/pull/57). Adds:

- New gogen primitives: \`composite-lit\`, \`kv-expr\`, \`case-clause\`, \`func-lit\`, \`switch-stmt\`. (The originals were defined Go-side but not registered with \`installGogenNS\`, so they weren't reachable from Clojure — fixed in this PR.)
- Walker support for **top-level declaration forms**: \`type-decl\` (named types + struct types), \`const-block\` (with iota), \`var-decl\` (file-level with array-of-struct composite literal), \`method\` (receiver functions), and a \`doc\` form for doc comments.
- Position-management infrastructure (\`allocPos\`, \`stripPositions\`) so generated code survives \`go/format.Node\` rendering across multi-line composite literals.

## Why

The hand-written \`pkg/ir/op.go\` (opTable + accessor methods) is exactly the kind of spec-driven code defop is meant to generate. To target it cleanly, gogen needed top-level decl support. The follow-up PR ([ssa-ir](#)) demonstrates the use case: \`pkg/ir/op_generated.go\` is now generated from a single \`examples/go-gen/ir_ops.lg\` spec.

## Stacked

This is part of a stacked set:
- #57 (gogen-v1) — foundation
- **This PR (gogen-v2)** — additional primitives + walkers
- [ssa-ir](#) — IR + optimization passes (depends on gogen-v2)
- [ir-tooling](#) — disassembler + opcode profiler + IR profile fixtures (depends on ssa-ir)

Each PR includes the ancestors' commits; merge them in order.

## Test plan

- [x] \`go test ./pkg/rt/...\` — all gogen tests pass
- [x] \`examples/go-gen/example.lg\` produces byte-identical output to before
- [x] \`examples/go-gen/numeric_specialize.lg\` produces byte-identical output to before

### DIFF
--- a/pkg/rt/gogen.go
+++ b/pkg/rt/gogen.go
@@ -717,6 +717,167 @@ func fieldSlice(v vm.Value) ([]*ast.Field, error) {
 	return out, nil
 }
 
+// func-lit: (gogen/func-lit [params] [results-or-nil] [body-stmts]) -> *ast.FuncLit
+// A function literal — an anonymous function used as an expression
+// (the value, not the declaration). Returns an expression, so it can
+// be passed as a call argument, assigned, etc.
+//
+// Example uses:
+//
+//	cb := func(x int) int { return x * 2 }
+//	vm.NativeFnType.Wrap(func(vs []vm.Value) (vm.Value, error) { ... })
+func cFuncLit(paramsV, resultsV, bodyV vm.Value) (vm.Value, error) {
+	params, err := fieldSlice(paramsV)
+	if err != nil {
+		return vm.NIL, fmt.Errorf("gogen/func-lit: params: %w", err)
+	}
+	results, err := fieldSlice(resultsV)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


nooga:
@nnunley thanks! There is some lint - unused funcs. Is this intended?

nnunley:
@nooga thanks for the catch! Those five constructors (`cFuncLit`, `cCaseClause`, `cSwitchStmt`, `cKVExpr`, `cCompositeLit`) were defined here but their `installGogenNS` registrations had drifted into the downstream PR (#62). Just pushed a fix — the registrations are now in this PR alongside the definitions, so lint is clean and the namespace exports them as soon as gogen-v2 lands.

nooga (approved):

PR #62, opened 2026-05-21
feat(ir): SSA IR + optimization passes (ConstFold, CSE, DCE, block-DCE)

nnunley:
## Summary

Builds an indexed-RPN, block-parameter SSA IR for let-go bytecode. Adds:

- **\`pkg/ir/\`** — the IR types (\`Function\`, \`Block\`, \`Node\`, \`Op\`), \`Build()\` (bytecode → IR via abstract interpretation of stack heights), \`Lower()\` (IR → bytecode with materialize-on-demand operand handling).
- **\`pkg/ir/passes/\`** — \`ConstFold\` (in-place arithmetic folding), \`CSE\` (block-local common-subexpression elimination), \`DCE\` (dead-code elimination), plus tests that exercise full Build→optimize→Lower round-trips.
- **\`pkg/ir/op_generated.go\`** — generated from \`examples/go-gen/ir_ops.lg\` via gogen (depends on PR #61 for the gogen primitives needed). The op catalog is now a single Clojure spec.
- **Source-span preservation** through optimization: \`Node.SourceInfos []vm.SourceInfo\` is a set; ConstFold and CSE union spans across rewrites/merges. Lowered bytecode carries every contributing source location.

## Highlights

- **Loop support**: \`(loop … recur …)\` constructs (e.g., \`(sumto 5) = 10\`) round-trip end-to-end. Stack-shape abstract interpretation pre-pass computes entry-stack-height for each block before translation, eliminating the underflow-on-demand model that previously worked by coincidence.
- **Multi-use lowering**: CSE-merged values (used >1 time) are handled via cheap re-materialization (Const/LoadArg/LoadVar/LoadClosed) or DUP_NTH (everything else). Cross-block multi-use is detected and errors loudly rather than silently miscompile.
- **Auto-prune unreachable blocks**: Build's Phase 4 removes blocks not forward-reachable from Entry (e.g., the dead fall-through after OP_RECUR).
- **Real-code integration**: \`TestIntegration_MandelbrotHelpers\` round-trips the three arithmetic helpers from \`examples/mandelbrot.lg\` (\`cadd\`, \`cmul\`, \`cmagsq\`) through the full pipeline.

## Tests

- All 31 IR/pass tests pass.
- Full repo: all 393 tests pass.

## Stacked

- [PR #57](https://github.com/nooga/let-go/pull/57) (gogen-v1) — foundation
- [PR #61](https://github.com/nooga/let-go/pull/61) (gogen-v2) — top-level decl walkers (this PR depends on it)
- **This PR (ssa-ir)** — IR + passes
- [ir-tooling](#) — disassembler + opcode profiler + IR profile fixtures (depends on this PR)

Merge in order: #57, #61, this, ir-tooling.

## What's deferred (filed as future work)

- **IR performance**: profiling shows the materialize-on-demand DUP_NTH strategy currently increases bytecode dispatches by 19-36% on tested fixtures (more dispatches than the source compiler emits). A future pass (likely a local-slot allocator + smarter operand ordering) is needed to make IR-optimized bytecode no-worse than baseline. The infrastructure is correctness-complete; performance is the next axis.
- **Mandelbrot end-to-end**: blocked on closure support in Build (\`MAKE_CLOSURE\` / \`LOAD_CLOSEDOVER\`), needed for \`lazy-seq\`.

## Test plan

- [x] \`go test ./pkg/ir/...\` (all 31 entries)
- [x] \`go test ./...\` (all 393 entries)
- [x] \`./scripts/generate-ir-ops.sh\` is idempotent (running it twice produces no diff)
- [x] Round-trip correctness via existing test suite


nnunley:
Closing in favor of a sem-ir redesign that uses a different IR shape — Carbon-style InstId-keyed flat array with no separate Block.Params / BranchTarget machinery, and an IR-direct VM that retires Lower entirely.

The data substrate from this PR (Node, NodeID, Function, the build pass, op tables, dominance, the ConstFold/CSE/DCE algorithms) carries over conceptually, but the storage shape and the lowering boundary change enough that landing this PR first then rewriting was likely to cost more than starting from the new design directly.

Will reopen a fresh PR for the redesign once the spec is approved.

PR #63, opened 2026-05-21
feat(rt,vm): disasm primitives + opcode profiler + IR profile fixtures

nnunley:
## Summary

Tooling for measuring IR-optimized bytecode. Three commits:

1. **\`feat(rt)\`** — exposes \`disassemble\` and \`constants\` primitives for bytecode introspection from Clojure.
2. **\`wip(vm,rt)\`** — opcode profiler. \`(profile/enable!)\`, \`(profile/disable!)\`, \`(profile/reset!)\`, \`(profile/snapshot)\`. Per-opcode + per-opcode-pair execution counts. Cost when disabled: one well-predicted branch per dispatch.
3. **\`feat(vm,ir)\`** — make TAIL_CALL terminal for builtin callees + profile fixtures comparing baseline vs IR-optimized bytecode.

## TAIL_CALL change

The VM's OP_TAIL_CALL had two cases: for \`*Func\` callees it replaced the frame (true tail call), but for builtins it pushed the result and advanced past TAIL_CALL — relying on a trailing RETURN. This trailing RETURN was always emitted by the source compiler but is now dead code: when the builtin path runs, the result is returned directly from \`Frame.Run\`. Lower (in the SSA IR PR) treats TailCall as a true terminator without emitting a redundant RETURN.

## Profile fixtures

\`pkg/ir/passes/profile_bench_test.go\` exercises sumto and the three mandelbrot helpers, reporting:
- Bytecode size baseline vs optimized.
- Per-opcode dispatch counts baseline vs optimized.
- Total dispatch delta.

Current findings (actionable signal for future work):

| Fixture | Bytecode size | Opcodes/run | Δ |
|---|---|---|---|
| sumto | 30 → 34 | 109 → 130 | +19.3% |
| cadd | 15 → 18 | 8 → 10 | +25% |
| cmul | 39 → 54 | 22 → 30 | +36.4% |
| cmagsq | 18 → 23 | 10 → 13 | +30% |

The IR optimization currently *regresses* performance on these fixtures because the materialize-on-demand DUP_NTH strategy emits more dispatches than the source compiler's natural stack layout. The infrastructure is correctness-complete; a future pass (likely a local-slot allocator) will close the gap.

## Stacked

- [PR #57](https://github.com/nooga/let-go/pull/57) (gogen-v1)
- [PR #61](https://github.com/nooga/let-go/pull/61) (gogen-v2)
- [PR #62](https://github.com/nooga/let-go/pull/62) (ssa-ir)
- **This PR (ir-tooling)** — depends on #62 for the IR types/passes

Merge in order.

## Test plan

- [x] \`go test ./...\` (all 393 entries)
- [x] Profile tests run in <1s and produce stable opcode counts across runs
- [x] Profiler-disabled overhead: single atomic Load per dispatch (already measured in earlier work)


nnunley:
Closing alongside #62 — see that PR for rationale. The tooling parts of this PR (disassemble / constants primitives, opcode profiler, profile fixtures) are independent of the IR shape and may be revived as a small standalone PR if useful before the sem-ir redesign lands.

PR #64, opened 2026-05-22
feat(core): unchecked-add/subtract/multiply/negate/inc/dec/divide-int — Clojure parity

nnunley:
## Summary

Adds the `unchecked-*` family of arithmetic operations with silent int64 wrap semantics, matching Clojure's `clojure.core` signatures. Required for porting code that needs modular int64 / u64 arithmetic (hash functions, splittable RNGs, etc.).

## What's added

Native (`pkg/rt/lang.go`):
- `unchecked-add` (2-ary)
- `unchecked-subtract` (2-ary)
- `unchecked-multiply` (2-ary)
- `unchecked-negate` (1-ary)
- `unchecked-divide-int` (2-ary)

Clojure-composed (`pkg/rt/core/core.lg`):
- `unchecked-inc` — `(defn unchecked-inc [x] (unchecked-add x 1))`
- `unchecked-dec` — `(defn unchecked-dec [x] (unchecked-subtract x 1))`

Strict-arity to match Clojure's inliner signatures. Float and BigInt inputs coerce to int64 (matching Clojure's long-arithmetic semantics).

## Why

Pairs with the existing `unsigned-bit-shift-right` to enable porting:
- xxh3-64 and other hash families
- splitmix64, SipHash, PCG, xoshiro, and other PRNGs
- Anything assuming modular u64/i64 wraparound

Without `unchecked-*`, every overflow promotes to BigInt or throws — neither matches what hash/RNG code expects.

## Choosing the right family

| Family | Semantics | When to use |
|--------|-----------|-------------|
| `+` `-` `*` | Checked, throws on overflow | Default arithmetic |
| `+'` `-'` `*'` | BigInt-promoting | When you need exact arithmetic past int64 |
| `unchecked-*` | Silent int64 wrap | Hash functions, RNGs, modular u64 code |

## Tests

`test/unchecked_arith_test.lg` covers:
- Basic operations on small integers (parity with Clojure semantics)
- Overflow wrap (e.g. `(unchecked-add Long/MAX_VALUE 1)` → `Long/MIN_VALUE`)
- Float and BigInt coercion paths
- Arity validation

Full clojure-test-suite passes (232 files, 5621 assertions).

## Supersedes

This replaces an earlier `unchecked-arith` local bookmark that only had a partial set (add/subtract/multiply + unsigned-bit-shift-right in `pkg/rt/math.go`). That version is dropped in favor of this fuller implementation.


nooga (approved):

nooga:
No notes!

Issue #42, opened 2026-05-19
Source-fallback in evalInit is broken — lgbgen can't bootstrap from a missing .lgb


nnunley:
`pkg/compiler/eval.go:54-71` has a fallback path: if `CoreCompiledLGB` is empty, fall back to compiling `rt.CoreSrc` (raw `core.lg`) from source. The comment says "Fall through to source compilation on error" — implying graceful degradation.

In practice that branch is dead code. Reproduction:

```bash
: > pkg/rt/core_compiled.lgb
go build -o /tmp/lg .
/tmp/lg -e '(println "hi")'
```

Panics with:

```
panic: core.lg compilation failed: CompileError: compiling do member
  caused by CompileError: compiling def value
  caused by CompileError: compiling fn body
  caused by CompileError: compiling let binding
  caused by CompileError: compiling function position
  caused by CompileError: Can't resolve odd? in this context
```

`odd?` is defined at `core.lg:157` as a regular `defn`. Some later macro body (in the file) references `odd?` and the source-compile path can't resolve it from a macro body context, even though the bytecode path handles the same code fine.

Worse: `lgbgen` itself panics under the same condition. So a fresh clone with an empty/missing `.lgb` has no bootstrap path — `lg` won't run, and `lgbgen` won't run to generate a fresh `.lgb`. The `.lgb` is currently load-bearing build infrastructure committed to git.

Bisect identifies `47e582c` (March 2026, "Add decoding and encoding capabilities for execution units") as where the bytecode path was introduced. The source-compile path was preserved as a fallback but never actively maintained; it has since bit-rotted.

Triage options:
1. **Fix the source path.** Investigate why `odd?` (defined earlier in the file) isn't resolvable from a macro body during forward source compilation. May relate to how top-level def-bound vars are exposed inside macro-body fn compilation.
2. **Delete the dead fallback** at `eval.go:67-71` and document `core_compiled.lgb` as required build infrastructure. Cheaper, honest about current state.

Surfaced while reviewing the `.gitattributes` change in #40 — was trying to confirm whether the `.lgb` could be `.gitignore`d.

nooga:
Linking #71

nooga:
No longer reproducible after merging #65 and #71. Closing.

PR #65, opened 2026-05-22
Bootstrap stdlib bundle generation from source

nooga:
## Summary
- make lgbgen run only under the bootstrap tag so bundled bytecode is regenerated from source stdlib
- keep normal builds on embedded LGB while bootstrap builds force source fallback
- run CI tests against both bundled bytecode stdlib and source bootstrap stdlib, including the Clojure compatibility suite
- regenerate release LGB before GoReleaser tests/builds

Refs #42

## Tests
- go test -v -timeout 60s ./... -skip TestClojureTestSuite
- go test ./test/ -count=1 -run TestClojureTestSuite -v
- go test -tags bootstrap -v -timeout 60s ./... -skip TestClojureTestSuite
- go test -tags bootstrap ./test/ -count=1 -run TestClojureTestSuite -v

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -32,11 +32,20 @@ jobs:
         with:
           go-version-file: go.mod
 
+      - name: Regenerate bundled bytecode
+        run: go generate ./pkg/rt/
+
       - name: Build
         run: go build -v ./...
 
-      - name: Test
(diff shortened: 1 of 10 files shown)


nnunley (approved):
lgtm. A lot simpler than what i was initially thinking about.

PR #77, opened 2026-05-22
feat(ir): SSA IR + Lisp-side pass pipeline

nnunley:
## Summary

Adds an SSA IR layered between the compiler and the bytecode VM, with the optimization passes and data structures written in Lisp on top of a small Go substrate.

**Design:** Go provides only the IR substrate the bridge needs — op kinds, `IROpToBytecode` mapping, chunk-emission primitives, and `vm.Consts` / `vm.SourceInfo` constructors. Everything else — Function/Block/Inst data, build (forms→IR), optimization passes (dce, constfold, cse, licm, thread-block-args, typeinfer), dump, dominance, lower (IR→bytecode) — lives in Lisp under `pkg/rt/core/ir/`.

The pipeline is wired as a Lisp function `(ir.passes.pipeline/compile-form form)` that runs forms→IR→optimize→Lower→bytecode. Direct-compile and pipeline-compile both work; the pipeline path goes through the IR.

Total: ~6000 LOC added, ~1100 of which is Lisp.

## Commits

1. **`feat(ir): SSA IR substrate — op table, bridge primitives, codegen tooling`** — Go-side base (op-table, chunk-emission bridge, generators, disasm, profiler). Self-contained; doesn't activate anything until commit 2.
2. **`feat(ir): Lisp-side IR — data, build, optimize, lower`** — the .lg files (data layer, zipper, passes DSL, build, all passes, lower, dump, dominance) plus bundle integration.
3. **`test(ir): regression coverage for the Lisp pass pipeline`** — two Go-side tests driving the runtime: a constfold behavioral test and an end-to-end pipeline test.

## Test plan

- [x] `go build ./...` succeeds at each commit
- [x] `go test ./...` passes at HEAD
- [ ] `pipeline_bench_test` confirms the Lisp pipeline produces executable chunks across the corpus (no byte-identity requirement; results compared behaviorally)
- [ ] `lisp_constfold_test` confirms each constfold strategy fires on representative inputs

## Followups (out of scope)

- Raw-string `:body` templates in `examples/go-gen/ir_bridge.lg` are ugly (no AST validation, escape gymnastics). Future cleanup: emit via `go/ast`.
- `examples/go-gen/` is a misleading directory name for what's actually production codegen specs. Future cleanup: dedicated `gogen/<target>/` tree.
- `ir_data.lg` (Lisp output) and `ir_bridge.lg` (Go output) parallel each other; a single declarative spec could drive both.
- The remaining `pkg/ir/` is bytecode-adjacent metadata. Could move to `pkg/vm/` and drop `pkg/ir/` entirely.


nooga:
@nnunley We're now getting some new lint that wasn't showing up before. Did you notice that on local?
I'm guessing this is due to the golangci-lint upgrade in CI action.

nnunley:
I Think the best bet (if I don't do it first) is to create a clean baseline that commits the results of go fix, go vet, and gofmt -s -w.

nooga:
Good idea. Pushed #82 - linter migration + mechanical cleanup.

nooga (approved):

PR #80, opened 2026-05-23
feat(bytecode): LGB v2 format with per-tag versioning and batch decode

nnunley:
## Summary

Bumps the `.lgb` bytecode format from v1 to v2 with three goals: (1) give the format room to evolve without breaking old decoders, (2) speed up collection decode, and (3) reserve extension points for upcoming IR-direct VM work.

The v2 header now carries a 2-bit version prefix on every tag byte (`0bVV_TTTTTT`), so future tag semantics can be versioned independently. A capability mask in the header lets future features fail fast on old decoders instead of silently misinterpreting data. Collection tags (map, set, vector) decode via batch constructors (`NewPersistentMap`, `NewPersistentSet`, presized `ArrayVector`) instead of per-element `Assoc` loops, cutting decode time on large collections by ~45–70%.

The decoder keeps a frozen v1 path (`readModuleV1`, `decodeToExecUnitV1`) so old `.lgb` files still load. The encoder writes v2 only. `pkg/rt/core_compiled.lgb` has been regenerated in the new format.

## Key changes

- `pkg/bytecode/tags.go`: tag versioning constants (`TagVer0`–`TagVer3`, `TagID*` 6-bit IDs, `Tag*` wire bytes)
- `pkg/bytecode/module.go`: `Capabilities uint32` field
- `pkg/bytecode/encoder.go`: writes v2 header + optional capability mask
- `pkg/bytecode/decoder.go`: dual-path v1/v2 decoder, batch collection helpers, tag version validation
- `pkg/bytecode/bench_test.go`: `BenchmarkV1vsV2Decode` a-b comparison
- `pkg/rt/core_compiled.lgb`: regenerated in v2 format

## Test plan

- [x] 84 bytecode unit tests pass
- [x] 115 VM tests pass
- [x] 9 compiler tests pass
- [x] Hand-crafted v1 module still decodes (`TestV1DecodeStillWorks`)
- [x] Unknown tag produces error with hex value (`TestUnknownTagError`)
- [x] Capability mask validated (`TestUnsupportedCapabilityMask`, `TestSupportedCapabilityMask`)
- [x] Benchmarks show ~2–3x decode speedup on core, ~45% on large collections


nooga:
I think this is an interesting change. Shaving off core load time sounds amazing.

I decided to benchmark this branch against current `main` using the existing startup workload, `lg -e nil`, with stripped binaries and `hyperfine`.

Results:

| Revision | Startup |
|---|---:|
| `main` (`5f7556c`) | ~7.5 ms |
| PR #80 (`b1d9964`) | ~29.2 ms |

I dug into it with `GODEBUG=inittrace=1`. The regression is mostly not the compiled stdlib load itself:

| Area | `main` | PR #80 |
|---|---:|---:|
| `pkg/compiler` init / compiled core load | ~1.9 ms | ~2.9 ms |
| `pkg/rt` init | ~3.0 ms / 180 KB | ~25 ms / 41.9 MB |

The dominant issue appears to be this package-level init in `pkg/rt/gogen.go`:

```go
func init() {
	for i := 1; i < 1<<20; i++ {
		goSynthFile.AddLine(i)
	}
}
```

That eagerly pre-fills over a million synthetic line offsets on every process startup, even for `lg -e nil`, before the gogen functionality is used. I validated this by building a temp copy of the PR with only that prefill removed:

| Build | Startup |
|---|---:|
| PR #80 as-is | ~29.2 ms |
| PR #80 without synthetic line prefill | ~7.9 ms |

There is a smaller cost from the larger/newer compiled core too: `core_compiled.lgb` grows from ~149 KB to ~274 KB, and the direct LGB init benchmark moves from about ~1.05 ms on `main` to ~2.0 ms here. But that is not the main startup regression.

Worsening boot time is not acceptable, especially for a change whose goal is improving startup/core load behavior.

What do we do about this? My suggestion would be to make the synthetic `token.File` line allocation lazy instead of pre-filling `1<<20` lines at package init. The formatting need makes sense, but the cost should only be paid by the gogen path when it actually generates positioned nodes.

nnunley:
Fair. I'll do some investigation on my side to see what we can get away with before the lazy loading. I don't want to paper over things by hiding. But you're right that the regression in startup times isn't acceptable.

nnunley:
Main doesn't have the init() loop at all — gogen on main lacks allocPos and synthetic line preallocation entirely. The regression was introduced in PR #80 and fixed by our
  lazy-init change before it ever hit main.

  Here's the lg -e nil comparison (20 runs each, in milliseconds):
```
  ┌─────────────────────────────┬─────────┐
  │           Branch            │ Average │
  ├─────────────────────────────┼─────────┤
  │ lgb-v2 (lazy init)          │ 13.4 ms │
  ├─────────────────────────────┼─────────┤
  │ main (no gogen init at all) │ 15.7 ms │
  └─────────────────────────────┴─────────┘
```

nnunley:
@copilot resolve the merge conflicts in this pull request

nnunley:
Accidentally included the IR changes - they shouldn't have been included in this.

nooga (approved):

PR #85, opened 2026-05-25
lginterop: compact positional exports, generics/methods/fields metadata

nnunley:
## Summary

- Replace verbose keyword-map serialization with compact positional vectors: `[:func name [type-params] [params] [results] variadic?]`
- Capture generics signatures (no longer filtered from exports)
- Capture methods and fields for struct types
- Fix multi-result smart wrapper detection (only wrap when second result is `error`)
- Fix float32/intN/uintN unboxing in smart wrappers (two-line assignment)
- Skip unboxable untyped int constants (e.g. `math.MaxUint` overflow)
- Make imports conditional in generated Go (`fmt`, `vm`, package import only when needed)
- Add `lginterop.lg` — Lisp-side macro-driven codegen using `gogen` constructors

## Verification

- `go build ./cmd/lginterop` — clean
- Generated `math` (95 exports, smart wrappers) — compiles
- Generated `bytes` (64 exports, struct registration) — compiles  
- Generated `slices` (40 generic exports, empty-install) — compiles

### DIFF
--- a/cmd/lginterop/main.go
+++ b/cmd/lginterop/main.go
@@ -1,0 +1,620 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"go/build"
+	"go/constant"
+	"go/importer"
+	"go/token"
+	"go/types"
+	"os"
+	"os/exec"
+	"path"
+	"path/filepath"
+	"runtime"
+	"sort"
+	"strconv"
+	"strings"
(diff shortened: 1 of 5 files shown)


nooga:
lgtm, there's just some staticcheck lint

nooga (approved):

PR #104, opened 2026-05-28
feat/perf(ir): substrate refinements + uses-cache + passes (DCE/LICM/CSE/mutability/trace)

nnunley:
## Summary

IR substrate refinements layered on top of #94/#96/#98:

- **Build/lower/data**: ctx-block fix for recur/loop, `(var sym)` and `(set! sym val)` special forms, captures-of walks fn* bodies, refactor `(nth form N)` destructuring, skip unterminated blocks in lower-go, multi-arity closure captures + vector type promotion
- **Passes**: new `mutability.lg`, new `trace.lg` (per-pass instrumentation), incremental uses-cache (Phase 1) with bitvector-of-int64 backing (Phase B) and transient outer vector in compute-uses (Phase A), DCE skips :invalid insts, query uses cache via uses-* API not raw seq ops
- **Validate**: cross-block-ref check + diagnostic cleanup + go-name munge for `!`/`?`/`*`
- **Ops**: `:inc`/`:dec` ops + lowered to Go
- **Tests**: new `lisp_cse_test.go`

## Test plan
- [ ] `go build ./...`
- [ ] `go test ./...`
- [ ] Run `scripts/ir-stress.lg` against xsofy corpus (from PR for tools/scripts)


nnunley:
Retracting — squashing this batch picked up unresolved jj rebase artifacts in pkg/rt/core/ir/build.lg (stray divider lines) that block bundle regeneration via `go run -tags bootstrap ./cmd/lgbgen`. `make build` only compiles the Go binary, not the bundle, so the issue wasn't visible locally either. Restoring to the pre-rebase granular history and will reopen smaller, cleaner PRs from there.

PR #105, opened 2026-05-28
perf(ir/typeinfer): sparse worklist + monotonic store + backward infer-arg-types

nnunley:
## Summary

Stacked on top of #104. Three perf improvements to the typeinfer pass:

- **Monotonic store + RPO sweep + skip redundant normalize** (`typeinfer.lg`)
- **Backward infer-arg-types pass** + `SourceInfo.Symbol` for readable Go identifiers (new `passes/infer_arg_types.lg`)
- **Sparse monotone worklist** replacing the RPO sweep (`typeinfer.lg`)

## Depends on
- #104 (substrate refinements)

## Test plan
- [ ] After #104 lands, rebase will be no-op
- [ ] `go test ./pkg/ir/...`

### DIFF
--- a/pkg/ir/lisp_typeinfer_test.go
+++ b/pkg/ir/lisp_typeinfer_test.go
@@ -11,22 +11,9 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/nooga/let-go/pkg/compiler"
 	"github.com/nooga/let-go/pkg/rt"
 	"github.com/nooga/let-go/pkg/vm"
 )
-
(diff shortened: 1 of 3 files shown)


nnunley:
Retracting — squashing this batch picked up unresolved jj rebase artifacts in pkg/rt/core/ir/build.lg (stray divider lines) that block bundle regeneration via `go run -tags bootstrap ./cmd/lgbgen`. `make build` only compiles the Go binary, not the bundle, so the issue wasn't visible locally either. Restoring to the pre-rebase granular history and will reopen smaller, cleaner PRs from there.

PR #107, opened 2026-05-28
feat(rt+vm): xxh3 hash namespace + numeric-width interop widening

nnunley:
Adds a `hash` namespace exposing `github.com/zeebo/xxh3`:

- Pure fns: `xxh3-64`, `xxh3-64-seed`, `xxh3-64-str`, `xxh3-64-str-seed`.
- Stateful: `(hash/xxh3-hasher)` / `(hash/xxh3-hasher-seed seed)` returns `{:write :write-str :sum :reset}` — methods bound directly from `*xxh3.Hasher`.

Companion change to the boxing layer because xxh3 returns `uint64`:

- `IntType.Box` accepts every signed/unsigned int width (`int8..int64`, `uint..uint64`) instead of `int`-only.
- `BoxValue` in `pkg/vm/value.go` collapses `reflect.Int*` / `reflect.Uint*` and dispatches via `v.Int()` / `v.Uint()` so named int types (e.g. `time.Duration`) round-trip correctly.

The widening is generally useful — any Go library whose signature uses width-typed ints is now usable from let-go without per-call adapters.

## Smoke

```clojure
(hash/xxh3-64-str "hello, world")
;; => 3471384689615110252

(let [h (hash/xxh3-hasher)]
  ((:write-str h) "hello, ")
  ((:write-str h) "world")
  ((:sum h)))
;; => 3471384689615110252
```

## Test plan

- [x] `make test` — full suite passes locally
- [x] Smoke above matches between one-shot and streaming hasher paths

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,10 @@ go 1.26
 require (
 	github.com/chzyer/readline v0.0.0-20180603132655-2972be24d48e
 	github.com/stretchr/testify v1.8.4
+	github.com/zeebo/xxh3 v1.1.0
 )
+
+require github.com/klauspost/cpuid/v2 v2.2.10 // indirect
 
 require (
 	github.com/chzyer/test v1.0.0 // indirect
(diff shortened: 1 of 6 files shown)


nnunley:
Closing — reshaping as lginterop-generated bindings + .lg user-facing wrapper instead of a hand-written Go NS. Will re-open under new branch.

PR #122, opened 2026-05-30
perf(vm): lock-free Var.Deref (atomic root/curr; ~670x parallel deref)

nnunley:
## What

`Var.Deref` — the hottest var operation — took the **global** `bindingsMu`
mutex on *every* read, even for a var with no dynamic binding (just to check
the binding stack was empty). So all var reads across all goroutines serialized
on one lock. That throttles any concurrent Lisp that touches vars — in
particular, parallelizing the IR typeinfer passes, where each worker derefs
`*ns*` / config dynamic vars per function.

## Fix

Make `Deref` lock-free: the `root` and the current dynamic top binding (`curr`)
live in `atomic.Pointer[Value]`. `Deref` is two atomic loads, no mutex. Push /
pop / `RunWithBindings` keep `curr` in sync with the bindings stack under
`bindingsMu` (the cold path). Binding semantics are unchanged.

## Numbers (pkg/vm, 8 cores, before → after)

| benchmark | before | after | speedup |
|---|--:|--:|--:|
| `VarDerefRoot` (single) | 6.4 ns | 0.46 ns | ~14× |
| `VarDerefRootParallel` | 87 ns | 0.13 ns | **~670×** |
| `VarDerefBoundParallel` | 80 ns | 0.11 ns | **~720×** |
| `VarDerefDistinctParallel` | 84 ns | 0.13 ns | ~650× |

`VarDerefDistinctParallel` gives each worker a *distinct* var yet still measured
~84 ns before — proving the contention was the one **global** `bindingsMu`, not
per-var. After the change, reads scale like an atomic load.

The benchmarks ship in `pkg/vm/var_deref_bench_test.go`, so the bench-ratchet
(which sweeps `pkg/vm`) tracks this hot path going forward.

## Correctness

- `pkg/vm` + `pkg/rt` green; the Lisp suite (`binding` / `*ns*` / `future`
  heavy) passes.
- New `pkg/vm/var_binding_test.go`: nested push/pop shadow+restore,
  `RunWithBindings` restore, and a 50k push/pop vs 8 concurrent readers stress —
  all pass under `-race` (no torn reads, no data race).

## Scope / follow-up

This removes the read-side serialization. The binding *stack* itself is still
process-global (not thread-local), so concurrent `(binding …)` / `future` with
*different* values on the *same* var still share state — a separate, deeper
change (true per-goroutine binding frames) if that's ever needed. Not required
for the parallel-read case this unblocks.

### DIFF
--- a/pkg/vm/var.go
+++ b/pkg/vm/var.go
@@ -9,11 +9,20 @@ import (
 	"fmt"
 	"maps"
 	"sync"
+	"sync/atomic"
 )
 
 type Var struct {
-	root      Value
-	bindings  []Value // dynamic binding stack (nil when unused — zero cost)
+	// root and curr are atomic so Deref — by far the hottest var
+	// operation — is lock-free. Previously every Deref (even of a var
+	// with no dynamic binding) took the global bindingsMu just to check
+	// the binding stack, serializing all var reads across goroutines.
+	// curr holds the current dynamic top binding (nil = use root); it is
+	// kept in sync with the bindings stack by push/pop/RunWithBindings
+	// under bindingsMu (the cold path).
+	root      atomic.Pointer[Value] // root binding (lock-free read)
+	curr      atomic.Pointer[Value] // current dynamic top binding; nil = use root
+	bindings  []Value               // full dynamic binding stack (guarded by bindingsMu)
 	nsref     *Namespace
 	ns        string
 	name      string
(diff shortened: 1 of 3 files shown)


nooga:
Wohoo 🚀 Believe it or not I was actually annoyed with var deref implementation but didn't have any idea how to improve it. Thanks @nnunley!

nooga (approved):

PR #157, opened 2026-06-01
Minimal compatibility with `weavejester/medley`

### LINKED ISSUE #155: Minimal compatibility with `weavejester/medley`
**Goal:** make medley 1.10.0 load under let-go (`LG_READ_CLJ=1`) so its core data-manipulation functions run. This covers the map/kv transforms (`map-vals`, `map-keys`, `filter-vals`), the assoc/update helpers (`assoc-some`, `dissoc-in`, `deep-merge`), and the common seq utilities (`index-by`, `distinct-by`, `find-first`, `interleave-all`).

This targets "load and run the data functions", not full medley parity.

medley reaches JVM-only references on its `:clj` reader-conditional branch. Loading a namespace compiles every `defn`, so one unresolved `clojure.lang.*`  or `java.*` symbol aborts the whole load, and even `m/boolean?` becomes unavailable. This issue adds the minimal compat aliases needed to compile the namespace and run the data functions. Functions that need real JVM collections or interop (queues, mutable lists, list methods) stay out of scope and get tracked separately.

abogoyavlensky:
Resolves: #155 

Makes `weavejester/medley` 1.10.0 load under let-go with `LG_READ_CLJ=1`. The `medley.core` namespace now compiles, and 47 of its functions run, covering the everyday map, kv, and seq helpers. I verified this by probing every public  medley function against the built runtime.

Every change is an additive entry in the existing Clojure-compat layer  (`installClojureCompatAliases`), plus small type ancestry and `Receiver`  wiring. The reader, resolver, and compiler core are untouched.

## Supported from Clojure / JVM

| Reference | Kind | Enables in medley |
|---|---|---|
| `clojure.lang.IEditableCollection` marker and ancestry (vectors/maps/sets) | real | `editable?`, transient paths|
| `clojure.lang.MapEntry.` and `->clojure.lang.MapEntry` ctor sugar | real | `map-entry` |
| `Throwable`: `ExInfo` is now a `Throwable`, implements `Receiver` (`.getMessage`/`.getCause`) and `IMeta` (the `^Throwable` hint compiles to `with-meta`) | real | `ex-message`, `ex-cause`|
| `java.util.UUID/fromString` (over `ParseUUID`) and `/randomUUID` (over `random-uuid`) | real | `uuid`, `random-uuid` |
| `java.util.regex.Pattern` aliased to `RegexType` | real | `regexp?` |
| `compare-and-set!` added as a real `Atom` primitive | real | `deref-swap!`, `deref-reset!` |
| `instance?` answers bare marker symbols through type-to-interface ancestry | mechanism | every `instance?` predicate |
| `clojure.lang.PersistentQueue` marker and `/EMPTY` | load-only stub | `queue`/`queue?` resolve; `EMPTY` fails loudly when conj'd instead of returning wrong data |
| `java.util.ArrayList` constructor | load-only stub | `partition-between`/`window` compile |

## Deferred to separate PRs

medley needs more JVM machinery to make the rest work. Each item below gets its own PR:

1. **Real `clojure.lang.PersistentQueue` type.** Makes `queue` and `queue?`  functional. They are load-only stubs today.
2. **Real mutable `java.util.ArrayList` backing.** Unblocks the transducer  arities of `partition-between`, `partition-after`, `partition-before`, and `window`. Their collection-form arities already work.
3. **`find` that returns a real `MapEntry`.** Today `find` returns a plain 2-vector that `key` and `val` reject. This core fix unblocks `update-existing`, `update-existing-in`, and `collate-by` with colliding keys.
4. **`.indexOf` / `^java.util.List` interop, or a native `index-of`.** Unblocks medley's `index-of`.
5. **Stateful-transducer `sequence` support (`transformer-seq*`).** Unblocks `sequence-padded`.

## medley functions not yet working
- `update-existing`, `update-existing-in`: hit `(val (find m k))`. See deferred 3.
- `collate-by`: works for single or distinct keys; fails when keys collide. See deferred 3.
- `index-of`: needs `.indexOf` interop. See deferred 4.
- `sequence-padded`: needs stateful transducers. See deferred 5.
- `queue`, `queue?`: no real queue type. See deferred 1. `(queue coll)` fails loudly, and `queue?` always returns `false`.
- Transducer arities of `partition-between`, `partition-after`, `partition-before`, `window`: need a real `ArrayList`. See deferred 2. Their collection-form arities work.

## Everything else works

The full `map-*`/`filter-*`/`remove-*` kv family, `assoc-some`, `dissoc-in`, `deep-merge`, `find-first`, `index-by`, `interleave-all`, `distinct-by`, `dedupe-by`, `take-upto`, `drop-upto`, `indexed`, `insert-nth`, `remove-nth`, `replace-nth`, `join`, `least`, `greatest` (and `-by`), `abs`, `mapply`, `find-in`, `map-padded`, `boolean?`, `uuid`, `uuid?`, `random-uuid`, `regexp?`, `ex-message`, `ex-cause`, `deref-swap!`, `deref-reset!`, and the collection forms of `partition-*` and `window`.


pkg/rt/hierarchy.go
```diff
@@ -484,6 +507,12 @@ func directTypeParents(tag vm.Value) *vm.PersistentSet {
 		result = setConj(result, vm.AnyType)
 	case vm.PromiseType:
 		result = setConj(result, vm.AnyType)
+	case vm.ExInfoType:
+		// ex-info values are let-go's equivalent of Clojure's ExceptionInfo,
+		// which IS-A Throwable. Reporting the Throwable marker lets medley's
+		// (instance? Throwable ex) guard pass so ex-message/ex-cause work.
+		result = setConj(result, vm.AnyType)
+		result = setConj(result, cljThrowable)
```
nnunley, line 515: This is the idiomatic move done right: modeling `ExInfo` as a `Throwable` *ancestor in let-go's own hierarchy* rather than faking JVM inheritance. `(instance? Throwable ex)` becomes a statement about let-go's ancestry graph, consistent with how `cljAssociative`/`cljSeqable` already work. Clean.

pkg/vm/exinfo.go
```diff
@@ -24,6 +25,42 @@ func (e *ExInfo) Message() string      { return e.message }
...
+	cp.meta = m
+	return &cp
+}
+
+// InvokeMethod implements Receiver so the JVM-interop method forms
+// (.getMessage ex) / (.getCause ex) dispatch to an ex-info value. This lets
+// libraries that branch on (instance? Throwable ex) and then call .getMessage
+// — e.g. medley's ex-message/ex-cause on their :clj branch — work against
+// let-go ex-info values. Mirrors the semantics of core's ex-message/ex-cause.
+func (e *ExInfo) InvokeMethod(name Symbol, args []Value) (Value, error) {
```
nnunley, line 51: Right seam. `Receiver.InvokeMethod` is let-go's designed extension point for `(.method obj)` — the same path `String`, `Record`, and `DTypeInstance` already use. Since `ExInfo` *owns* these methods, implementing `Receiver` directly is correct; no need to route through the pluggable host-method-dispatch registry (that seam is for extending types you don't own). Nicely mirrors core's `ex-message`/`ex-cause` semantics too.

pkg/vm/exinfo.go
```diff
@@ -24,6 +25,42 @@ func (e *ExInfo) Message() string      { return e.message }
...
+// is still *ExInfo, preserving Type()==ExInfoType (so (instance? Throwable ex)
+// stays true) and InvokeMethod (so .getMessage/.getCause still dispatch).
+func (e *ExInfo) Meta() Value {
+	if e.meta == nil {
+		return NIL
+	}
+	return e.meta
+}
+
+func (e *ExInfo) WithMeta(m Value) Value {
```
nnunley, line 40: `IMeta` impl mirrors `List`/`Vector` exactly (shallow copy + `meta` field, hash bookkeeping correctly omitted since `ExInfo` has none). And it's load-bearing, not decorative: type hints compile to `with-meta`, so this is precisely what lets `(.getMessage ^Throwable ex)` reach dispatch. Minor note only: the copy means `(identical? ex (with-meta ex {...}))` is now false — harmless here, just flagging the semantics.

pkg/rt/lang.go
```diff
@@ -7420,6 +7435,81 @@ func installClojureCompatAliases(ns *vm.Namespace) {
...
+	ns.Def("clojure.lang.MapEntry.", create)
+
+	// clojure.lang.PersistentQueue/EMPTY. let-go has no
+	// real PersistentQueue, so this is a load-only stub: it must merely resolve
+	// as a var at compile time (medley derefs it only at runtime inside `queue`).
+	// Bind it to a non-collection marker symbol rather than vm.EmptyList — that
+	// way medley's `(queue coll)` = `(into (queue) coll)` FAILS LOUDLY with
+	// "conj expected Collection" instead of silently returning a reversed list
+	// (a plausible-but-wrong FIFO). queue?/queue stay degraded by design.
+	DefNSBare("clojure.lang.PersistentQueue").Def("EMPTY", vm.Symbol("clojure.lang.PersistentQueue/EMPTY"))
```
nnunley, line 7559: Good call failing loudly here instead of returning `EmptyList` (which would silently hand back a *reversed* queue). One non-blocking thought: the failure is currently *incidental* — it works only because a `Symbol` happens to reject `conj`. If symbols ever gain a `conj`/method fallback, these load-only stubs would quietly start succeeding with garbage. Might be worth making the failure *intentional by construction* — e.g. a dedicated "unsupported" sentinel value, or a stub that throws on use with a `PersistentQueue not implemented` message — so a future runtime change can't turn it silent. Totally fine to defer.

pkg/rt/lang.go
```diff
@@ -7420,6 +7435,81 @@ func installClojureCompatAliases(ns *vm.Namespace) {
...
+		if len(vs) > 1 {
+			return vm.NIL, fmt.Errorf("java.util.ArrayList. expects 0 or 1 args, got %d", len(vs))
+		}
+		return vm.Symbol("java.util.ArrayList"), nil
+	})
+	if err != nil {
+		panic(err)
+	}
+	ns.Def("java.util.ArrayList", vm.Symbol("java.util.ArrayList"))
+	ns.Def("java.util.ArrayList.", arrayListStub)
```
nnunley, line 7578: Same shared fragility as the `PersistentQueue/EMPTY` stub above: it fails loud only because `Symbol` rejects method dispatch, not because the stub itself refuses use. If you do the explicit-sentinel idea, this `ArrayList.` ctor is the other site. Non-blocking.

pkg/rt/lang.go
```diff
@@ -5783,6 +5783,16 @@ func installLangNS() {
 		if t, ok := vs[0].(vm.ValueType); ok {
 			return vm.Boolean(vs[1].Type() == t), nil
 		}
+		// Clojure-compat interface markers (e.g. clojure.lang.IEditableCollection)
+		// are registered as plain symbols by installClojureCompatAliases. Answer
+		// instance? for them via the type→interface hierarchy so libraries that
+		// branch on (instance? clojure.lang.IEditableCollection coll) — like
+		// medley — observe the JVM-equivalent answer.
+		if marker, ok := vs[0].(vm.Symbol); ok {
+			if anc := directTypeAncestors(vs[1].Type()); anc != nil {
```
nnunley, line 5873: Heads-up for the maintainer (not a code issue): PR #146 also adds a branch to this same `instance?` function a few lines up. Whichever of #146/#157 lands second will need a trivial rebase to combine the two cases — no logic conflict. Landing #157 first makes #146 rebase cleanly.

nnunley (commented):
Took a careful pass at this specifically asking: does it stay idiomatic to let-go's own protocols, or does it try too hard to replicate the JVM object model? Really nice work — it leans on let-go's *own* machinery (the type-ancestry hierarchy, `IMeta`, `Receiver`) rather than bolting Java semantics on top. No fake class wrappers, no inheritance, no per-class method tables. `ExInfo`-as-`Throwable` is a claim in let-go's hierarchy (not pretense), the UUID/regex aliases map to real types, and `.getMessage`/`.getCause` go through the designed `Receiver` seam. The comments explaining *why* each stub exists are excellent.

Everything below is either praise or a non-blocking thought — none of it blocks merge. The new `TestMedleyCompat` suite covers the tricky fail-loud paths well. One heads-up inline about a trivial future rebase against #146.

nooga:
I’m comfortable with this as a focused compat step. Thank you for pointing out the stubs - worth tracking separately but not blocking the merge.

nooga (dismissed):

abogoyavlensky:
@nooga Thank you for the review! I've fixed conflicts with main, now should be fine to merge.

nooga (approved):

Issue #132, opened 2026-05-31
Proposal for declarative let-go version requirement / ns metadata


mparrett:
When consumer code calls a let-go feature missing from the running `lg`, the failure surfaces as a "Can't resolve" cascade pointing at symbols that visibly exist in the consumer's own code. The real failure is several requires deep — a transitive dep uses a function/runtime behavior the lg version doesn't ship, that ns fails to compile, every dependent fails to resolve its exports. Hard to diagnose without walking the full require graph by hand.

A declarative way for a `.lg` file (or namespace) to say "I require let-go ≥ X.Y.Z" would let lg fail at parse time with a single clear message.

### Three shapes

**A. Top-of-file form:** `(require-letgo-version! ">=1.9.0")`
- ✅ Trivial; no parser changes; message can name the demanding file + running version.
- ❌ Only catches what loads. A fork-only call deep in a require-graph still cascades if the check file isn't loaded first.

**B. `ns` metadata:** `(ns my.app {:require-letgo ">=1.9.0"} ...)`
- ✅ Every ns is a versioned contract; lg can validate the full require-graph before any symbol resolution. Error: "ns `my.title` needs let-go ≥ 1.9.0, you have 1.7.0".
- ❌ Parser + bootstrap must honor it before other resolution. Larger lift.

**C. Consumer-side scrappy workaround (no lg change):** consumer ships a pin file (`.let-go-version` or similar) and a build wrapper greps `lg --version` against it, bailing with a clear message. Works today; doesn't help anyone running `lg foo.lg` ad-hoc.

The version surface exists: `(System/getProperty "let-go.version")` returns `rt.Version` (goreleaser ldflags at release, `"dev"` locally).

Filing as discussion since the choice between A/B (or some third design) is yours. Happy to share xsofy's recent cascade as a concrete example.

nnunley:
This seems like a really good idea. Do you want to support github commit hashes, too?

nooga:
Good idea! One note, here @mparrett. let-go v2.x is older than v1.8.0 and dead. We had to go back to v1.x versioning because otherwise Go would force us to omve the source packages into v2/ directory. The latest released version now is v1.9.0 and it should run xsofy.

mparrett:
Thanks both. Thanks for confirming the v1.9.0 / v2.x correction. Fixed the example.

Building on @nnunley's commit-hash question, here's a concrete shape prototyped on a local branch off `origin/main` at d57d0cd:

[...]

PR #164, opened 2026-06-01
feat(semver): require-letgo (Layer 3) + npm/cargo range sugar (Layer 4) (#132)

nnunley:
Builds on #160 (merged) to add the next two layers from the #132 plan, plus the `:lg` portability guard. **Design credit: @mparrett** — Layer 3 follows the `require-letgo` shape he specced in #132. His prototype lived on a *local, unpushed* branch ("local branch off origin/main at d57d0cd"), so this implements his documented design rather than merging his branch.

## Layer 3 — `enforce` + `require-letgo`
A file asserts its host `lg` build in one line; on a known mismatch it throws a single clear message at load time instead of a "can't resolve" cascade.

```clojure
#?(:lg (require-letgo ">=1.9.0"))   ; => throws: "require-letgo: this file requires >=1.9.0 but running v1.8.0"
```

- **Auto-detect (Matt's disjoint classes):** a 7–40 hex string is a commit pin (prefix-matched against `let-go.commit`); anything with `. < > = ^ ~` is a semver range.
- **Per-property skip-when-unknown (Matt + nnunley in #132):** a range warns-and-passes when `version="dev"`; a SHA pin warns-and-passes when `commit="none"`. Known mismatch = hard fail.
- `enforce` is a **pure** `(enforce running-version running-commit spec) → :ok | [:skip …] | [:fail msg]`, fully unit-tested; `require-letgo` is the thin throwing/​warning wrapper over `(current)`/`(current-commit)`.

## Layer 4 — npm/cargo range sugar
`satisfies-range?` now understands caret `^1.2.3`, tilde `~1.2`, x-ranges (`1.2.x`, `1.*`, `*`), and `||` OR-composition, desugared into the existing comparator clauses. Existing comparator/bare behavior from #160 is a strict subset (regression-tested). Hyphen ranges (`1.2.3 - 2.3.4`) intentionally excluded (ambiguous with the `-pre` separator).

## Safety notes (`:lg`)
`require-letgo` is let-go-specific, so the canonical form guards both the `:require` and the call: `#?(:lg (:require [let-go.semver :refer [require-letgo]]))` + `#?(:lg (require-letgo …))`. New `.cljc` fixture + driver prove it loads clean under let-go (dev-skip) and is invisible to a non-`:lg` reader. README documents the operators + the guard.

## Makefile
One-line `-X main.commit=$(COMMIT)` (Matt's), so SHA pins actually fire on `make` builds (previously `commit="none"` → always skipped). `version` stays `"dev"`. Verified end-to-end: a matching SHA pin passes, a mismatch throws.

## Scope / structure
Kept entirely inside `let-go.semver` (no core changes), so the merged public API stays stable and this reviews as a clean layer on #160. Bare-form `(require-letgo …)` and Matt's `#semver` reader tags remain forward-compatible follow-ups (bootstrap-load + lazy `*data-readers*`), out of scope here.

## Verification
- `let-go.semver` tests (Layer 3 + Layer 4) and portability tests green on **both** the bytecode VM and `-tags gogen_ir`.
- Bundle regenerated (`make generate`); `core_go_lowered/**` restored from main since `let-go.semver` is `;; lgbgen:skip-go` (no lowered output) — avoids the known regeneration drift.

cc @nooga @mparrett — review welcome, especially the skip-when-unknown policy and the Layer 4 caret/tilde edge cases.


mparrett:
@nnunley Nice PR. Thanks for the collaboration!

Reviewed end-to-end and probed locally against the PR branch. Overall, clean and faithful to the #132 consensus: `enforce` as a pure `(running-version running-commit spec) → :ok | [:skip msg] | [:fail msg]` is a nice refinement, skip-when-unknown matches the agreed defaults, and the `.cljc` fixture closes the portability story. The Makefile `-X main.commit` change works as intended: SHA pins fire correctly on `make` builds (`(require-letgo "deadbee")` against a running `6e26558` throws the documented one-line message).

**Three things worth tightening before merge:**

**1. Caret/tilde/x-range tokens silently misbehave on a leading `v`.** Concrete repro on this branch:

```clojure
(satisfies-range? "1.5.0" "^v1.0.0")  ;; => false (silent, no diagnostic)
(satisfies-range? "1.5.0" "^1.0.0")   ;; => true
```

`parse-clause` (`>=`, `<=`, etc.) routes its suffix through `version` → `parse`, which strips `v`. But caret/tilde/x-range tokens go through `parse-partial`, which calls `parse-int` directly on each component — the `v` is never stripped and the result is a silent `false` rather than a parse error. Two-line fix: tolerate `v` at the top of `parse-partial` (mirror what `parse` already does). Keeps the v-tolerance uniform across all token kinds.

**2. The throwing path of `require-letgo` has no automated test.** `enforce` is covered for `[:fail …]` returns, but the wrapper's `(throw …)` step is unreachable on a dev/none test binary. Verified by hand that the throw works (produces a `let-go.lang.String` exception carrying the expected one-line message), but a regression wouldn't bite. `with-redefs` over `current` / `current-commit` closes it in a few lines.

**3. Optional: `ex-info` instead of raw string throw.** Not blocking — `(throw "string")` works in let-go (catchable as `let-go.lang.String`). The original sketch in #132 mentioned `ex-info`; the implementation drifted to a raw string. `ex-info` would let catchers pull `{:required spec :found running-version :check-type :semver|:sha}` for programmatic handling while keeping the same human message. Small ergonomics upgrade, easy to defer.

**Smaller stuff, probably skip:** `^0`/`^0.0`/`^0.0.0` edge cases are correctly handled by the code (verified in probing) but lack tests — cheap insurance, not a correctness fix. Empty-string spec produces `[:fail "require-letgo: this file requires  but running v…"]` (double space) — below the bar to fix in this PR.

Skip-when-unknown and Layer 4 core semantics both look right under probing. Appreciate the design-vs-implementation framing in the PR body. Looks good!

nooga:
Agree with @mparrett's points here. I think we can merge it. Thanks @nnunley @mparrett :)

nooga (approved):

nnunley:
Thanks @mparrett, @nooga — addressed all three points, and rebased onto current `main` while here.

**1. Leading-`v` in caret/tilde/x-range (your repro).** `parse-partial` now strips a leading `v` (mirroring `parse`), so `^v1.0.0` / `~v1.2.3` / `v1.x` match instead of silently returning `false`. v-tolerance is now uniform across the comparator and the caret/tilde/x-range paths.

**2. Throw-path coverage.** `with-redefs` didn't actually exist in core yet — it was only a placeholder in `ir.passes.mutability`'s `known-var-mutating-builtins`. So I implemented it (`with-redefs` + `with-redefs-fn`: temporary root rebind of any var, restored even on throw). The throw path now has a direct test via the new `enforce!` plus an end-to-end `require-letgo` test driven by `with-redefs`.

**3. `ex-info`.** `require-letgo` now throws an `ex-info` carrying `{:required :found :check-type}`, with the same one-line human message via `ex-message`. The decision is split into the pure `enforce` and an effectful `enforce!` — that split is what makes the throw testable without a release build.

Two notes worth flagging:
- let-go's `try`/`finally` *without* a `catch` swallows the exception rather than propagating it, so `with-redefs-fn` restores roots explicitly and re-throws (it does not rely on `finally`). Happy to file that separately if it's news.
- The large `core_go_lowered/` diff is mechanical: adding a `core` def shifts the generator's global temp-id counter, renumbering `arg__NNN` across every lowered module. No semantic change; `make check-generated` is green.

Full Go + `.lg` suite and `check-generated` both green.

mparrett:
A separate PR to address the swallowed exception seems appropriate, unless that's by design.

Rationale: compat and visibility. Java/python/Clojure follow the same pattern (propagate if no catch).

PR #165, opened 2026-06-02
feat(rt/term): wake read-key on SIGWINCH via BEL self-pipe

mparrett:
## Summary

Enables immediate* resizing during terminal rows/cols changes.

`term/read-key` blocks on `os.Stdin.Read` until input arrives, so terminal resizes are invisible to a game loop sitting at the prompt — the loop's `term/size` check only runs after a keypress, even when the OS has already delivered SIGWINCH. This is the native analog of the WASM-side `window._lgWake` pattern: an external poker needs to interrupt the blocking read so the next iteration can sample the new dimensions.

## Approach

A SIGWINCH handler writes one BEL byte (`0x07`) into a self-pipe. `read-key` uses `unix.Poll` to wait on stdin OR the pipe; when the pipe fires it returns BEL as if the user typed it. Caller-side `parse-key`-equivalents see `:unknown`, the game loop's `(term/size)` check fires on the next iteration and triggers redraws at the new dimensions. Zero caller-side wiring needed beyond what was already in place for keypress-driven resize detection.

The lazy `sync.Once` setup means processes that never call `read-key` don't install the signal handler. The plain `os.Stdin.Read` fallback when `os.Pipe()` fails preserves prior behavior on systems where the self-pipe trick can't be set up.

`golang.org/x/sys/unix` is already a dependency — no new imports beyond `os/signal`, `sync`, `unix.Poll`, `unix.SIGWINCH`.

## Why a self-pipe?

(rather than `(go (term/read-key))` + `alts!`)

The Clojure-idiomatic alternative — one-shot reader goroutine writing into a channel, parent doing `alts!` over reader + signal — was the first thing I tried. It works fine on native: goroutines blocked on stdin park independently while the scheduler runs others.

It cannot work on WASM: `term/read-key` blocks via synchronous `syscall/js Atomics.wait`, which holds the worker thread. Go's WASM runtime is single-threaded; while the worker is parked on Atomics.wait, no other goroutine can run — including the sibling doing `alts!`. The decoupling is illusory on that target.

The self-pipe + `unix.Poll` design is a wake-byte pattern: an external trigger (SIGWINCH here) writes a non-printable byte into the input path the worker is parked on. Same shape would generalize cleanly to a hypothetical WASM-side equivalent that needs to interrupt a blocking read from outside the worker — both targets end up parametrized by the signaling source, with the read side identical.

## Test plan

- [x] `go test ./pkg/rt ./test` — green
- [x] `GOOS=js GOARCH=wasm go vet ./pkg/rt` — clean (change is `//go:build !js && !plan9` only; no WASM impact)
- [x] Manual: drove a native TUI built on this runtime in a terminal, resized the window mid-game without pressing a key, verified the screen redraws at the new dimensions immediately rather than after the next keystroke

### DIFF
--- a/pkg/rt/term.go
+++ b/pkg/rt/term.go
@@ -10,13 +10,57 @@ package rt
 import (
 	"fmt"
 	"os"
+	"os/signal"
+	"sync"
 	"syscall"
 	"unicode/utf8"
 	"unsafe"
 
 	"github.com/nooga/let-go/pkg/vm"
+	"golang.org/x/sys/unix"
 	"golang.org/x/term"
 )
+
+// SIGWINCH wake-byte plumbing for term/read-key.
+//
+// Without this, term/read-key sits in a blocking os.Stdin.Read until input
+// arrives, so terminal resizes are invisible to a game loop sitting at the
+// prompt — the loop's (term/size) check only runs after the next keypress
+// even though the OS has already delivered SIGWINCH.
+//
+// Design: a SIGWINCH handler writes one BEL byte (0x07) into a self-pipe.
+// read-key uses unix.Poll to wait on stdin OR the pipe. When the pipe fires,
+// read-key returns BEL as if the user typed it — caller-side parse-key
+// equivalents see :unknown, the game loop's (term/size) check fires on the
+// next iteration and triggers a redraw at the new dimensions. No caller-side
+// wiring needed beyond what's already in place for keypress-driven resize
+// detection.
+//
+// The wake-byte pattern generalizes to any blocking read where an external
+// event needs to be surfaced — only the signaling source changes.
+var (
+	winchPipeR     *os.File
+	winchSetupOnce sync.Once
+)
+
+func setupWinch() {
+	winchSetupOnce.Do(func() {
+		r, w, err := os.Pipe()
+		if err != nil {
+			return // best-effort; read-key falls back to plain stdin
+		}
+		winchPipeR = r
+		sigCh := make(chan os.Signal, 1)
+		signal.Notify(sigCh, unix.SIGWINCH)
+		go func() {
+			for range sigCh {
+				// One BEL per SIGWINCH. Storms collapse — the read side
+				// drains whatever's pending in a single Read regardless.
+				_, _ = w.Write([]byte{0x07})
+			}
+		}()
+	})
+}
 
 type winsize struct {
 	rows, cols, xpix, ypix uint16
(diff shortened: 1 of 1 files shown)


nnunley:
The code looks good to me, but the AI minions have feedback:

### POLLHUP/POLLERR on stdin is unhandled → spurious BEL / busy-loop on disconnect, EOF never surfaced

pkg/rt/term.go — new read-key poll path (after unix.Poll(fds, -1), the fds[0].Revents&unix.POLLIN check and the trailing return vm.String("\x07"))

Bug: POLLHUP, POLLERR, and POLLNVAL are always reported in Revents regardless of requested Events. The code only tests Revents & POLLIN. When stdin reports hangup/error without POLLIN (e.g. controlling pty master closes — terminal window closed, SSH session dropped), neither branch is taken and the function falls through to return vm.String("\x07") (BEL). A caller looping on read-key re-enters, Poll returns the same POLLHUP immediately, and the loop spins at ~100% CPU emitting BEL forever. The end-of-input condition is never reported.

Impact: Regression. Pre-PR, stdin EOF produced os.Stdin.Read → n==0 → vm.NIL, letting the caller detect end-of-input. The new path can convert disconnect into an infinite BEL/CPU spin. The target use case (interactive TTY game) makes terminal disconnect a realistic event. Triggering requires POLLHUP to arrive without POLLIN, which is kernel/fd-dependent (some pty closes set POLLHUP|POLLIN, in which case the existing read 0 → NIL path is taken correctly) — so this fires on a subset of disconnects, not all. The defect is the complete absence of hangup/error handling, not the probability.

Suggested fix: After Poll, before falling through to BEL, handle error revents explicitly, e.g.:

if fds[0].Revents&(unix.POLLHUP|unix.POLLERR|unix.POLLNVAL) != 0 {
    // surface EOF/disconnect as the prior code did
    buf := make([]byte, 16)
    n, err := os.Stdin.Read(buf)
    if err != nil || n == 0 {
        return vm.NIL, nil
    }
    return vm.String(buf[:n]), nil
}
Only fall through to BEL when the wake pipe (fds[1]) actually fired.

Confidence: Confirmed (empirical). Reproduces 19/20 on Linux 6.8 via a closed-writer pipe (see Empirical Validation below). Not triggered on darwin/FreeBSD (they set POLLIN at pipe EOF).

### unix.Poll not retried on EINTR → spurious nil from read-key

pkg/rt/term.go — if _, err := unix.Poll(fds, -1); err != nil { return vm.NIL, nil }

Bug: Raw golang.org/x/sys/unix.Poll returns EINTR when the blocking poll(2) is interrupted by a signal and does not auto-retry. Go’s runtime uses SIGURG for async goroutine preemption (Go 1.14+), and this PR additionally installs a SIGWINCH handler — both can interrupt the blocked Poll. The code treats any Poll error as return vm.NIL, so read-key returns nil spuriously.

Impact: New robustness regression. The pre-PR os.Stdin.Read went through Go’s os package, which retries EINTR internally (ignoringEINTR); the new raw unix.Poll does not. Effect is occasional spurious nil from read-key. Benign in examples/term_demo.lg (it re-loops), but a game that branches on (read-key) returning nil will see phantom “no input” events.

Suggested fix: Wrap the poll in a retry loop (combine with C1’s fix):

for {
    _, err := unix.Poll(fds, -1)
    if err == unix.EINTR {
        continue
    }
    if err != nil {
        return vm.NIL, nil
    }
    break
}
Confidence: Medium-High

### Doc comment overclaims WASM parity — references window._lgWake, which does not exist in this repo

pkg/rt/term.go — new setupWinch block comment (“native analog of WASM’s window._lgWake”, “Mirror the WASM design: a SIGWINCH handler writes one BEL byte…”) and the read-key comment (“see … the WASM-side window._lgWake design”)

Bug: Repo-wide grep finds no _lgWake/lgWake in any .go/.js/.html/.lg file. The actual WASM read-key (pkg/rt/term_wasm.go:59) blocks on Atomics.wait(keyInt32, 0, 0) and has no resize-wake byte path. The comment describes parity with code that isn’t present here.

Impact: Maintenance/accuracy. A future maintainer who reads “see the WASM-side window._lgWake design” will search and find nothing, or assume an in-repo invariant that doesn’t hold. Per review standard, doc claims that overstate what the codebase guarantees are a maintenance issue, not a nit. (If _lgWake lives in an external JS harness, the comment should say so explicitly.)

### key-pending? does not report the resize wake — feature silently no-ops for gated callers

pkg/rt/term.go — relationship between new read-key and existing key-pending? (FIONREAD on stdin)

Bug: The wake only unblocks a blocking read-key. key-pending? checks stdin via FIONREAD and knows nothing about the winch pipe, so a caller written as (when (key-pending?) (read-key)) will never observe a resize wake — key-pending? stays false, read-key is never entered. The design works only for callers that call read-key as a pure blocking wait (as examples/term_demo.lg:45 does).

Impact: Contract gap. In-repo callers are unaffected (the demo doesn’t gate). But the WASM side does expose key-pending? for exactly the “poll loop breaks out on input” pattern, so a portable game using that idiom would get resize wakes on WASM-by-design but not here, or neither — the asymmetry is a latent surprise.

Suggested fix: Document on read-key/key-pending? that resize wakes are delivered only to blocking read-key, not signaled through key-pending?. (Optionally, have key-pending? also poll the winch pipe.)

mparrett:
Thanks for the careful pass — all four are valid, addressed in `383b1c1`. By section:

**POLLHUP/POLLERR on stdin unhandled** — confirmed regression, fixed exactly as suggested. Added an explicit `POLLHUP|POLLERR|POLLNVAL` branch before the BEL fall-through; on hangup/error it routes through `os.Stdin.Read` so EOF surfaces as `vm.NIL`, matching the pre-PR contract.

**unix.Poll not retried on EINTR** — confirmed, fixed with the standard retry loop. Folded together with the POLLHUP handling so the poll-then-classify-revents sequence is one cohesive block.

**Doc comment overclaims WASM parity** — right, `_lgWake` and the `c218882` SHA reference were fork-only context that shouldn't have leaked into upstream. Rewrote both the setupWinch and read-key doc comments to describe the wake-byte pattern abstractly without claiming a cross-target invariant or referencing artifacts outside this repo.

**key-pending? doesn't see resize wakes** — went with the "also poll the winch pipe" option rather than the docs-only path. `key-pending?` now FIONREADs both stdin and the winch pipe (and arms `setupWinch` lazily, so a caller hitting `key-pending?` before any `read-key` still gets resize wakes). The `(when (key-pending?) (read-key))` idiom now sees resize wakes here too, matching the property the WASM side naturally has — BEL would appear in the SAB queue and surface to its `key-pending?` check.

Native `go test ./pkg/rt ./test` green, `GOOS=js GOARCH=wasm go vet ./pkg/rt` clean. Tested the happy path against a native TUI on darwin; the empirical Linux validation in your review covers the case I can't repro locally.

nnunley (approved):

PR #172, opened 2026-06-02
feat(examples): revive the malli-on-let-go shim example

nnunley:
## Summary
- recover the historical examples/malli-on-let-go example and shim bundle
- restore the small host-class registration seam needed for preload-style compatibility shims
- make with-meta tolerate non-IMeta values again so protocol-tagged compatibility code can load

## Verification
- go test ./pkg/compiler -run 'TestHost(Class|Method)Seam|TestWithMetaOnProtocolDoesNotError' -count=1
- go test ./... -skip TestClojureTestSuite (push hook)
- go test ./test/ -count=1 -run TestClojureTestSuite -v (push hook)
- built let-go and loaded a local malli smoke script via the recovered shim; observed:
  - true
  - #Tags{:values {:x 1 :y "hi"}}

## Notes
- the example still assumes a local malli checkout at /Users/ndn/development/malli, as documented in examples/malli-on-let-go/lgx.edn


nnunley:
Follow-up context, corrected: recovering the historical malli example onto current upstream/main was not quite example-only. The example still needed two small compatibility restores to load cleanly:

- restore register-host-class! plus compiler-side host-class resolution so preload-style shims can register class symbols for later compilation units
- make with-meta return the original value for non-IMeta inputs again, which lets protocol-tagged compatibility code load without tripping a runtime error

Everything else the old branch needed appears to have already landed. I verified the recovered example with a local malli smoke script after these two restores.

nooga (approved):

nooga:
It looks like there's quite a lot of shim code here - shouldn't we think how to incorporate it into stock let-go so that it can just run malli as is?

nnunley:
We should. I just wanted to be principled. There's a lot of the JDK surface being replicated there. Things like Java's timeunits, etc.

Would it be better to have a java compat layer? Use Ingy's gojava as a bridge? Use my new work to allow defining constructors?

nnunley:
## Shim reduction plan — what moves into core vs. stays a Java bridge

Following up on the "shouldn't this just run on stock let-go?" question. Principle: a thing moves into core only if it has a real `clojure.core`/`clojure.string` name **and** a *correct* let-go implementation (never a stub that silently lies). The rest stays in the example shim, which becomes a pure Java-bridge layer — no recreation of the JDK object model.

While scoping this I verified the shim was overstating what's missing. Stock let-go already has more than the shim assumed.

### Moves into core
- **`unchecked-{inc,dec,add,multiply}-int`** — compose on the existing native `unchecked-*` (let-go is int64; `-int` ≡ base op), mirroring the current `unchecked-inc`/`-dec`.
- **`every-pred`** — pure HOF; core already ships its complement `some-fn`.
- **`eduction`** — over the real, tested transducer machinery (`transduce`/`sequence`/`into`). Seqable + reducible; docstring notes it isn't the JVM's reusable `Eduction` recipe (single-pass only — all malli needs).
- **`hash-combine`** — public clojure.core; the shim's bit-mix is standard.
- **`ns-resolve`** — **real**, not a stub: `resolve`/`find-ns`/`the-ns` already exist and `Namespace.Lookup` resolves qualified syms (nil for absent ns), which is exactly the answer malli's one call site needs.
- **`inst-ms`** — **real**, backed by let-go's existing time (`(now)` → `time.Time`, `System/currentTimeMillis`), not a `0` stub.
- **`*unchecked-math*`** — honest no-op dynamic var (no unchecked compiler mode here).
- **`->MapEntry`** — **real**, over let-go's own first-class `MapEntry` type (already mapped to arity-2 `[k v]`); also makes `(MapEntry. k v)` work via the unified ctor dispatch.
- **`bytes?`** — **real** native predicate. Correction: let-go *does* have byte arrays (`TypedArray`/`ArrayByte`, `byte-array` builtin), so `false` would be wrong. Needs to be native (same layer as `array?`, which doesn't distinguish element kind) — the one Go change.

### Already in core — drop from shim
- **`clojure.string/replace-first`** — already present (`string.lg`), shim entry is stale.

### Stays in the example shim (Java object model — *not* core)
Host classes (`java.util.Map`, `Pattern`, …), constructors/methods/deftypes (`HashMap`, `ArrayDeque`, `FutureTask`, `Thread`, `URI`, `BigDecimal`, …), statics (`TimeUnit`, `System/arraycopy`, `Long/parseLong`, `Float/parseFloat`, the `java.time`/`java.util` namespaces), the `-vmap` iterator override, and the non-`clojure.core` predicates `indexed?` / `uri?`.

### Deferred to its own PR
- **`print-method`** — a faithful version is the whole print-dispatch multimethod subsystem (pr/pr-str routing through it); too big to fold in here. Shim stub stays until then.

Net effect: the shim shrinks to a documented Java-interop bridge; everything Clojure-standard-and-correct lands in core. Implementation will be TDD `.lg` tests per fn (+ native `bytes?` test), then `make generate` / `make check-generated`.

nooga:
I'm ok with merging this as is. We can defer the rest :)

PR #173, opened 2026-06-02
perf: add timeline page and main snapshots

nooga:
## What

This adds a static "Are we fast yet?" perf page and a main-only CI job that records full perf snapshots after merges to main. Quick and dirty!

## How It Works

- `bench-ratchet` gets a new `snapshot` mode. It captures and aggregates benchmark output like the existing modes, but writes the current run as an immutable JSON snapshot instead of ratcheting `docs/perf/baseline.json`.
- Snapshots are written under `docs/perf/timeline/<commit-time>-<sha>.json`.
- `.github/workflows/perf-timeline.yml` runs on pushes to `main` except perf-artifact-only commits. It measures the post-merge tip once, regenerates the static page, and commits the snapshot plus `docs/perf/index.html` back to `main`.
- `cmd/perf-page` renders `docs/perf/baseline.json`, `docs/perf/historical/*.json`, and `docs/perf/timeline/*.json` into a self-contained static HTML page with SVG trend charts.
- The existing Pages workflow already uploads `wasm/static`; `wasm/Makefile` now also generates `wasm/static/perf/index.html`, so the page deploys at `/perf/`.

## Open Questions / Risks

- The current ratchet baseline was captured on an Apple M3. The timeline workflow will run on GitHub-hosted Linux runners, so the displayed machine metadata and absolute `ns/op` values will change after the first CI snapshot.
- GitHub-hosted runners are noisy and may vary by CPU model. The page graphs `ratio_to_anchor` for the main timing charts, but we should merge this after review and observe how stable the data looks after the first main merge.
- If the runner noise is too high, the next step is likely a self-hosted stable runner or more grouping/filtering by machine fingerprint.

## Verification

- `GOCACHE=/private/tmp/let-go-gocache go test ./... -count=1 -skip TestClojureTestSuite`
- `GOCACHE=/private/tmp/let-go-gocache make perf-page`
- Validated `bench-ratchet snapshot` from an existing raw `.jsonl` capture.
- Render-checked the generated page in local headless Chrome at desktop and mobile widths.

## Proposal

Merge after review and see how the first post-merge main snapshot behaves in CI before tuning runner selection, benchmark duration, or chart grouping.

### DIFF
--- a/.github/workflows/perf-timeline.yml
+++ b/.github/workflows/perf-timeline.yml
@@ -1,0 +1,77 @@
+name: Perf Timeline
+
+on:
+  push:
+    branches: [main]
+    paths-ignore:
+      - 'docs/perf/index.html'
+      - 'docs/perf/timeline/**'
+
+permissions:
+  contents: write
+
+concurrency:
+  group: perf-timeline-main
+  cancel-in-progress: false
+
+jobs:
+  profile:
+    if: github.actor != 'github-actions[bot]'
+    runs-on: ubuntu-latest
+    timeout-minutes: 90
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+          fetch-depth: 0
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
+
+      - name: Configure git author
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+
+      - name: Generate lowered runtime
+        run: make lowered
+
+      - name: Capture full perf snapshot
+        id: perf
+        run: |
+          short_sha="$(git rev-parse --short=12 HEAD)"
+          commit_epoch="$(git show -s --format=%ct HEAD)"
+          stamp="$(date -u -d "@${commit_epoch}" +%Y%m%dT%H%M%SZ)"
+          snapshot_name="${stamp}-${short_sha}.json"
+          snapshot_tmp="${RUNNER_TEMP}/${snapshot_name}"
+          raw_tmp="${RUNNER_TEMP}/${snapshot_name%.json}.jsonl"
+
+          go run ./cmd/bench-ratchet \
+            -full \
+            -count 3 \
+            -benchtime 2s \
+            -timeout 30m \
+            -out "${raw_tmp}" \
+            -baseline "${snapshot_tmp}" \
+            snapshot
+
+          echo "snapshot_name=${snapshot_name}" >> "${GITHUB_OUTPUT}"
+          echo "snapshot_tmp=${snapshot_tmp}" >> "${GITHUB_OUTPUT}"
+
+      - name: Commit perf snapshot and page
+        run: |
+          git checkout -B main origin/main
+          git pull --rebase origin main
+          mkdir -p docs/perf/timeline
+          cp "${{ steps.perf.outputs.snapshot_tmp }}" "docs/perf/timeline/${{ steps.perf.outputs.snapshot_name }}"
+          go run ./cmd/perf-page -out docs/perf/index.html
+
+          git add docs/perf/timeline docs/perf/index.html
+          if git diff --cached --quiet; then
+            echo "No perf changes to commit."
+            exit 0
+          fi
+          git commit -m "perf: record timeline snapshot for ${{ github.sha }}"
+          git push origin HEAD:main
--- a/.github/workflows/perf-timeline.yml
+++ b/.github/workflows/perf-timeline.yml
@@ -56,6 +56,7 @@ jobs:
             -out "${raw_tmp}" \
             -baseline "${snapshot_tmp}" \
             snapshot
+          # The committed snapshot embeds the per-run samples from raw_tmp.
 
           echo "snapshot_name=${snapshot_name}" >> "${GITHUB_OUTPUT}"
           echo "snapshot_tmp=${snapshot_tmp}" >> "${GITHUB_OUTPUT}"
--- a/cmd/perf-page/main.go
+++ b/cmd/perf-page/main.go
@@ -1,0 +1,1313 @@
+package main
+
+import (
+	"bytes"
+	"encoding/base64"
+	"encoding/json"
+	"flag"
+	"fmt"
+	"html/template"
+	"math"
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+	"time"
+)
+
+const modulePrefix = "github.com/nooga/let-go/"
+
+type Baseline struct {
+	Version       int                       `json:"version"`
+	CapturedAt    string                    `json:"captured_at"`
+	CapturedAtSHA string                    `json:"captured_at_sha"`
+	Machine       Machine                   `json:"machine"`
+	Anchor        Anchor                    `json:"anchor"`
+	Benchmarks    map[string]BenchmarkEntry `json:"benchmarks"`
+}
+
+type Machine struct {
+	OS        string `json:"os"`
+	Arch      string `json:"arch"`
+	NumCPU    int    `json:"num_cpu"`
+	CPUModel  string `json:"cpu_model"`
+	GoVersion string `json:"go_version"`
+}
+
+type Anchor struct {
+	Name       string  `json:"name"`
+	Package    string  `json:"package"`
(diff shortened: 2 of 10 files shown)


nnunley (commented):
## Review summary (assistant-assisted)

Solid, well-structured addition. `html/template` is used correctly (no HTML/SVG injection), and the workflow's `main`-only + bot-actor guard is a sound design.

**One high-signal call-out** (matches the worry about not using the bench library fully for reporting): the new `perf-page` uses `golang.org/x/perf` for essentially nothing. Only `benchfmt` is used, and only in the pre-existing `bench-ratchet` parser. All of the *reporting* is hand-rolled:
- unit formatting → `benchunit`
- `median()` with no confidence interval, and a hard-coded ±5% faster/slower threshold → `benchmath`
- a timeline of bare best-value points with no error bars → `benchseries`

The structural root: the per-run distribution (the `-out` jsonl) is captured in CI but never committed — only the collapsed single-scalar snapshot is. So the statistical packages have nothing to operate on downstream. See inline notes.

**Verified false alarms (please ignore these — they came up in automated passes and don't hold):** no `int64→float64` data loss (`encoding/json` coerces numbers into float64 fields); GitHub Actions runs `run:` steps with `bash -eo pipefail`, so "missing `set -e`" is a non-issue; no template injection (html/template auto-escapes).

### Fast-follow? Yes — with one caveat.
Nothing here is a correctness/data bug on the realistic path, it's purely additive, and the security posture is fine — so it's reasonable to merge and treat the reporting depth as a fast-follow. The **one thing worth doing before the timeline starts accumulating committed history** is persisting the per-run samples into the snapshot schema (even if rendering stays scalar for now). That's the only piece that's expensive to retrofit: you can't add intervals/error bars to history you didn't keep. Everything else — wiring up benchunit/benchmath/benchseries, atomic snapshot writes, skip-bad-file on load, and the divide-by-zero guard in the pre-existing compare code — is genuinely fast-follow.

Full report lives at `paad/code-reviews/perf-timeline-page-2026-06-02-19-03-06-ab1b054c.md`.

cmd/perf-page/main.go
```diff
@@ -0,0 +1,1313 @@
...
+func splitBenchmarkName(fullName string) (string, string) {
+	name := strings.TrimPrefix(fullName, modulePrefix)
+	idx := strings.Index(name, ".Benchmark")
+	if idx < 0 {
+		return "", name
+	}
+	return name[:idx], name[idx+1:]
+}
+
+func compare(current, reference Baseline) (map[string]float64, Summary) {
```
nnunley, line 541: **x/perf is used only for parsing — reporting is hand-rolled.** `compare` classifies faster/slower on a hard-coded ±5% (`delta <= -0.05` / `>= 0.05`), and `median()` below returns a bare point estimate. `golang.org/x/perf/benchmath` gives medians *with* confidence intervals plus a real significance test (`MedianAssumption.Summary`, `Assumption.Compare`), so noise around the threshold stops flipping the faster/slower counts. Prerequisite: the per-run samples, which are currently discarded (see my note on the workflow's `git add`).

cmd/perf-page/main.go
```diff
@@ -0,0 +1,1313 @@
...
+}
+
+func shortSHA(value string) string {
+	if len(value) <= 12 {
+		return value
+	}
+	return value[:12]
+}
+
+func formatNS(value float64) string {
```
nnunley, line 687: `formatNS` / `formatBytes` / `formatCount` reimplement unit scaling that `golang.org/x/perf/benchunit` already does: `benchunit.Time.Scale(v).Format(v)`, plus `Binary` for bytes (yours caps at MiB; benchunit goes to TiB) and `Decimal` for counts. Roughly a drop-in replacement.

cmd/perf-page/main.go
```diff
@@ -0,0 +1,1313 @@
...
+		Timeline:          timeline,
+		Charts:            buildCharts(timeline),
+		Rows:              rows,
+		TopImprovements:   improvements,
+		TopSlowdowns:      slowdowns,
+		RecentlyTightened: recent,
+	}
+}
+
+func buildCharts(timeline []Snapshot) []Chart {
```
nnunley, line 325: The timeline plots a single `Y` per snapshot with no variance — a regression seen in 1/3 runs looks identical to one seen in 3/3. `golang.org/x/perf/benchseries` is purpose-built for an ordered-axis (commit/time) comparison with summary bounds, i.e. error bars. Same prerequisite as above: keep the per-run samples.

cmd/perf-page/main.go
```diff
@@ -0,0 +1,1313 @@
...
+	sort.Slice(all, func(i, j int) bool {
+		if !all[i].captured.Equal(all[j].captured) {
+			return all[i].captured.After(all[j].captured)
+		}
+		return all[i].path > all[j].path
+	})
+	return all[0].baseline, all[0].name, nil
+}
+
+func loadTimeline(timelineDir, historicalDir string, current Baseline) ([]Snapshot, error) {
```
nnunley, line 214: One corrupt/truncated snapshot aborts the entire page: this returns on the first `loadBaseline` error, so a single bad file — in a directory that grows by one-per-commit from CI — blanks the whole timeline. Prefer skip-with-warning + `continue`. (The writer `writeBaseline` is also non-atomic `os.WriteFile`; pre-existing, but a temp-file + rename would stop a crash mid-write from producing the truncated file this then chokes on.)

cmd/perf-page/main.go
```diff
@@ -0,0 +1,1313 @@
...
+}
+
+type Anchor struct {
+	Name       string  `json:"name"`
+	Package    string  `json:"package"`
+	NSPerOp    float64 `json:"ns_per_op"`
+	Iterations int64   `json:"iterations"`
+}
+
+type BenchmarkEntry struct {
```
nnunley: Heads-up, **not a bug**: this `BenchmarkEntry` is a hand-copy of the one in `cmd/bench-ratchet` and has drifted — `allocs_per_op`/`bytes_per_op` are `float64` here vs `int64` there, and the ratio field is `Ratio` here vs `RatioToAnchor` there. JSON still round-trips (tags match, numbers coerce — no data loss), but the two copies will diverge silently over time. Worth extracting a shared `pkg/perfdata` types package consumed by both binaries.

.github/workflows/perf-timeline.yml
```diff
@@ -0,0 +1,77 @@
...
+
+      - name: Commit perf snapshot and page
+        run: |
+          git checkout -B main origin/main
+          git pull --rebase origin main
+          mkdir -p docs/perf/timeline
+          cp "${{ steps.perf.outputs.snapshot_tmp }}" "docs/perf/timeline/${{ steps.perf.outputs.snapshot_name }}"
+          go run ./cmd/perf-page -out docs/perf/index.html
+
+          git add docs/perf/timeline docs/perf/index.html
```
nnunley, line 72: This is the root of the reporting limitation: `-out "${raw_tmp}"` (the jsonl containing every individual run) is captured above but never committed — only the collapsed single-scalar snapshot is `git add`ed here. Once the timeline accumulates history, those distributions are gone for good and benchmath/benchseries can't add intervals/error bars retroactively. If only one thing lands before merge, it's persisting the per-run samples into the committed snapshot schema; the rendering can follow.

nnunley (commented):
_(Duplicate — this review was accidentally submitted twice due to a shell error. Please see the other review with the same title for the full write-up and inline comments. Ignore this entry.)_

nooga:
Rebased on latest main, resolved the Makefile conflict by keeping both the perf targets and the new fanout targets, and pushed the updated branch.

I also addressed the review comments in commit 9733c57b:
- moved the perf JSON schema into shared pkg/perfdata so bench-ratchet and perf-page no longer drift
- embedded per-run benchmark samples in committed timeline snapshots, so future benchmath/benchseries work has the distributions available
- made baseline writes atomic via temp file + fsync + rename
- made perf-page skip corrupt/truncated timeline snapshots with a warning instead of failing the whole page
- switched perf-page unit scaling to golang.org/x/perf/benchunit
- added tests for sample retention, atomic readable writes, and corrupt snapshot skipping

Verified after rebase with:

```
GOCACHE=/private/tmp/let-go-gocache go test ./... -count=1
```

I did not wire full confidence intervals/error bars yet; this preserves the raw samples first, which was the non-retrofit-able part. The richer benchmath/benchseries rendering can now be a follow-up without losing history.

nnunley (approved):

Issue #79, opened 2026-05-23
Proposal: improved browser interop


mparrett:
Hi there. Following up on nooga/xsofy#22, where we discussed let-go should "provide a more decent browser interop surface" for clients to build their own web shells. Here's a proposal for a minimal first step here for discussion before any PR.

## Idea for a first cut: extract `lg-host.js` + add registration API

Two changes, behavior-preserving from the user's perspective:

1. **Extract `lg-host.js` from `wasm.go`** — pull the JS body from the Go string template into `pkg/rt/wasm/lg-host.js`, embedded via `//go:embed`. Build still emits the same single self-contained HTML; output bytes unchanged.

2. **Add a registration API** — introduce `window.LetGoHost.{onOutput, onEmit, sendInput, wake, setSize}` in `lg-host.js`. Existing `_lg*` globals stay as internal aliases so xsofy keeps working. Default `index.html` registers a `console.log` fallback sink.

**Rationale for starting here:** Step 1 is invisible (verified by byte-identical build output). Step 2 is the first observable improvement: custom shells can subscribe to VM stdout and `js/emit` without hardcoded `term.write` or `window.dispatchEvent`. Unblocks moving xterm.js out of let-go (future work) without committing to anything beyond this.

**Motivation disclosure:** Working through a longer dependency tree in the fork (`mparrett/let-go` / `experiment/js-bridge`): xterm into `examples/`, a `host/*` namespace for event-framed input with a wake protocol that doesn't leak `BEL` into user code, slimming `term/*` to ANSI helpers only. Each piece independent and none is required for this slice. Happy to propose them one at a time as needed.

I've taken an initial stab at deeper design and exploring tradeoffs — JS API shape (namespaced object vs. flat globals vs. ES modules), event-framed input vs. byte stream, where the `term/*` ↔ host I/O line lives, etc. I'm happy to share the analysis as it relates to this issue or larger design. Nothing in this proposal locks any of it.

Working implementations exist in my fork if a concrete reference for any piece would help. Looking forward to hearing your ideas.

## Open questions

- General alignment on this approach?
- Namespace the JS API under `window.LetGoHost`, or keep flat `_lg*` globals with a documented contract?
- One PR or two (source-split separate from API addition)?
- Anything in the current `wasm.go` template you'd want preserved that I should call out explicitly?

nooga:
Hi @mparrett I have not forgotten about this but I'm still thinking about what would be the right approach to this. Please bear with me :)

[...]

PR #174, opened 2026-06-02
feat: browser interop (js/emit + LetGoHost API)

mparrett:
## Summary

Concrete proposal for #79, opened as draft to make the design tangible. Three commits in conceptual order.

### 1. `feat: add js namespace for browser event dispatch`

The `(js/emit event-name data)` primitive. Fire-and-forget bridge from .lg code to the host. WASM dispatches via a single `_lgEmit(name, jsonString)` hand-off; native validates args and no-ops so .lg code stays portable. ~140 LOC, three new files in `pkg/rt/`, self-registering namespace.

### 2. `feat(wasm): extract lg-host.js + host.html from wasm.go template`

Step 1 from the issue body. Pulls the inline `<script>` body and HTML template body out of `wasm.go` into `pkg/rt/wasm/{lg-host.js, host.html}`, embedded via `//go:embed` and assembled back via `AssembleHTML`. A golden test in `testdata/` pins the assembled bundle so any future drift surfaces clearly.

### 3. `feat(wasm): add window.LetGoHost registration API`

Step 2 from the issue body. Adds:

```js
window.LetGoHost = {
  onOutput(cb),    // register sink for VM stdout; cb(string)
  onEmit(cb),      // register sink for js/emit; cb(name, parsedData)
  sendInput(s),    // inject UTF-8 keystrokes/bytes toward the VM
  setSize(c, r),   // advertise terminal size to the VM
};
```

Defaults: `onOutput` is `console.log` (overridden by xterm registration on boot), `onEmit` dispatches a `CustomEvent` on window. Internal `_lg*` globals stay in place as the existing wiring `LetGoHost` is built on. The default template registers xterm via `LetGoHost.onOutput((s) => term.write(s))`.

## Behavioral changes (vs prior inline template)

The `LetGoHost` layer in commit 3 introduces two intentional behavioral differences — they're the point of the new API, not regressions:

- Worker stdout now flows through `LetGoHost._output(cb)` (default: `console.log`; xterm registers its own callback on boot) instead of calling `term.write` directly.
- `_lgKey("")` returns `false` rather than falling through to the worker.

Framing the whole PR as a "pure refactor + new API" rather than byte-identical-vs-prior — h/t @nnunley for catching the overclaim.

## Verified locally

- `go test ./...` passes across the whole codebase.
- `lg -w` produces a working 7.1MB bundle.
- Headless playwright run: `(js/emit "test-event" {:answer 42 :source "lg"})` dispatches `CustomEvent("test-event", {detail: {answer: 42, source: "lg"}})` on window. Map and vector payloads both flow through the bridge.
- Existing browser REPL still works (terminal renders, REPL accepts input, output flows through `LetGoHost._output`).

## What's NOT in this PR

- **`wake()`**. Intentionally absent from the `LetGoHost` surface. Unblocking a parked `read-key` without sending real input requires a SAB-level protocol change (wake-epoch cell, tri-state ready flag, etc.) so the Go-side `read-key` can tell a real keystroke from an unblock. Deserves its own PR with a concrete contract.
- **xterm to `examples/` + `--shell` flag.** The most opinionated change in the original. Saved for follow-up once this API surface settles.

## Open questions

- **`js/emit` namespace name.** Kept as `js`; the broader interop-surface design (CLJS-familiar syntax, externs-vs-typed-boundary, marshalling) feels like its own conversation, not something to pre-resolve via a rename now.
- **`LetGoHost` shape: single-sink vs event-bus.** Current `onOutput(cb)` replaces; multi-subscriber is a consumer-side compose. Happy to switch to `addEventListener`-style if you'd prefer.
- **`_lg*` compat hooks.** Left in place for existing callers; they predate this PR (`_lgKeyInt32`, `_lgFlush` go back to the first WASM commit). PR doesn't promise them as stable API or deprecate them; `LetGoHost` is a documented layer on top. Eventual removal would need a setup-time handshake giving the runtime stored references rather than name lookups; I have a shape in mind, but out of scope here.

Opened as draft to make the design concrete; happy to reshape or close it.

### DIFF
--- a/pkg/rt/js.go
+++ b/pkg/rt/js.go
@@ -1,0 +1,56 @@
+/*
+ * Copyright (c) 2026 Matt Parrett
+ * SPDX-License-Identifier: MIT
+ *
+ * Platform-independent half of the js namespace. Argument validation and
+ * marshaling live here so that bugs (wrong arity, bad event-name type) are
+ * caught the same way whether .lg code is running native or in WASM. The
+ * platform-specific files (js_wasm.go, js_other.go) only provide the
+ * dispatch primitive.
+ */
+
+package rt
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/nooga/let-go/pkg/vm"
+)
+
+// prepareEmit validates args for (js/emit event-name data) and returns the
+// event name and the JSON-marshaled data ready to hand to the platform
+// dispatcher. Same contract on every platform.
+func prepareEmit(vs []vm.Value) (string, string, error) {
+	if len(vs) != 2 {
+		return "", "", fmt.Errorf("js/emit expects 2 args (event-name data), got %d", len(vs))
+	}
+	name, err := eventName(vs[0])
+	if err != nil {
+		return "", "", err
+	}
+	data, err := fromValue(vs[1])
+	if err != nil {
+		return "", "", err
+	}
+	buf, err := json.Marshal(data)
+	if err != nil {
+		return "", "", fmt.Errorf("js/emit: %w", err)
+	}
+	return name, string(buf), nil
+}
+
+// eventName coerces a let-go value into the string event name passed to
+// CustomEvent. Accepts keyword (:stats), symbol (stats), or string ("stats").
+func eventName(v vm.Value) (string, error) {
+	switch v.Type() {
+	case vm.KeywordType:
+		return string(v.(vm.Keyword)), nil
+	case vm.SymbolType:
+		return v.(vm.Symbol).String(), nil
+	case vm.StringType:
+		return string(v.(vm.String)), nil
+	default:
+		return "", fmt.Errorf("js/emit event-name must be keyword, symbol, or string; got %s", v.Type().Name())
+	}
+}
--- a/pkg/rt/js_other.go
+++ b/pkg/rt/js_other.go
@@ -1,0 +1,35 @@
+//go:build !(js && wasm)
+
+/*
+ * Copyright (c) 2026 Matt Parrett
+ * SPDX-License-Identifier: MIT
+ *
+ * Native (non-WASM) stub for the `js` namespace. The namespace IS installed
+ * so .lg code that uses (js/emit ...) parses and runs identically on native
+ * and in the browser — emits just no-op on native, mirroring the way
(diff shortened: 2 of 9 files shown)


nnunley:
Nice extraction — the `js/emit` platform split is clean and the data path checked out end-to-end (Go `prepareEmit` → `_lgEmit` → worker `postMessage` → main-thread `JSON.parse` → `CustomEvent`), build tags are mutually exclusive/exhaustive, and the JSON escaping in `AssembleHTML` is correct. No correctness bugs.

One callout, on framing rather than code: this isn't quite a "pure refactor," and the golden test doesn't guard what its comment claims it does.

- `assemble_test.go` says it "guards the byte-identity claim that backs the 'pure refactor' framing." It actually only proves `AssembleHTML` is *self-consistent* (same inputs → committed golden bytes). Nothing compares the new output against the **old inline template**, so it can't catch divergence from prior behavior — and there is some: worker stdout now flows through `LetGoHost._output` instead of `term.write` directly, and `_lgKey("")` now returns `false` where it used to fall through. Those are good changes (the host API is the point), but they're behavioral, so I'd reframe the PR/comment as "refactor + new host API" rather than byte-identical.

Minor robustness item worth folding in while you're here: `AssembleHTML`'s `strings.Replace(..., 1)` silently no-ops if a marker ever goes missing from the template, and `TestMarkersGone` can't catch that (markers absent == markers "gone"). A one-line assert that each `__MARKER__` is present exactly once before substitution would turn a future silent-broken-build into a clear test failure.

mparrett:
Thanks for the careful read! Pushed `1e32de4`:

- `mustReplaceOnce` asserts each marker appears exactly once before substitution — `AssembleHTML` panics with a clear message if lg-host.js or host.html ever loses a marker.
- Retuned the `TestAssembleHTMLGolden` docstring — it pins against drift, which is what it actually proves; dropped the byte-identity / "pure refactor" framing.

Also reframed the PR description: `LetGoHost` is intentional new behavior (worker stdout via `_output`, `_lgKey("")` → `false`), not a regression riding on a refactor. Agreed those are the point of the layer.

nnunley (approved):

nooga (approved):

PR #178, opened 2026-06-03
feat: Add io/resource with -resource-paths roots and -b binary embedding

### LINKED ISSUE #177: Support resource directories: `io/resource`, with embedding into `-b` binaries

abogoyavlensky:
Resolves: #177 

Add resource directories: read non-source files (templates, static assets, data) at runtime with `io/resource`, and embed them into `-b` standalone binaries.

## Usage

`io/resource "path"` returns a reader-coercible handle, or `nil` if the file is absent. It composes with `io/slurp`, `io/reader`, and `io/line-seq`:

```clojure
(when-let [r (io/resource "templates/index.html")]
  (io/slurp r))
```

Pass resource roots with `-resource-paths` (separate multiple roots with the OS path-list separator, or set `LG_RESOURCE_PATHS`). Names are relative to a root; the first root wins on a conflict.

```bash
lg -resource-paths resources app.lg         # dev: read from ./resources
lg -b app -resource-paths resources app.lg  # embed resources into the binary
```

A bundled binary reads only its embedded resources, so it runs anywhere without the source tree.

## Design notes

- Resource roots are separate from source paths, so the always-present `.` source root can't drag the whole repo into a bundle.
- The flag is explicit: no default directory. Project tooling (e.g. lgx) owns conventions like a default `resources/` dir and passes the flag.
- Bundles with resources use a new 20-byte LGB2 trailer holding the gzipped archive. Resource-less bundles keep the 12-byte LGBX trailer, so binaries built before this change still load.
- A symlink is followed only when it resolves to a path inside its resource root (in dev and when bundling), which blocks path traversal.

## Out of scope

WASM and `.lgb` embedding, a bytes API, and resource listing.


nnunley:
Reviewed this carefully — it's in good shape. The security-sensitive guarantees in the description all hold up against the code and the tests back them: `..`-escape rejection, symlink containment via `resolveReal` + `WithinRoot`, legacy-LGBX backward compat (and non-bundle binaries are safely ignored by the `default` case), and first-root-wins. A couple of scary-sounding angles I chased down turned out to be non-issues: a `C:/passwd`-style Windows path and the symlink TOCTOU are both caught by the `WithinRoot`/`resolveReal` backstop, and the embedded-archive trust boundary is build-time (anyone who can corrupt those bytes already owns the binary).

One callout: the binary-trailer reading has two parsers that should be one. `readBundledLGB` (lg.go:155) carefully error-checks every `Seek`/`ReadFull`, but `getBaseBinarySize` (lg.go:444) re-implements the same LGB2/LGBX discrimination and **drops the Seek error checks** (lg.go:466, 475) — exactly the kind of drift duplicated format-parsing invites. Extracting a single `parseBundleTrailer(f) (lgbSize, resSize uint64, kind, err)` used by both would fix that and give one place to add a size sanity-check: a crafted/truncated trailer can currently reach `make([]byte, lgbSize)` (lg.go:187) and panic rather than failing with a clean "corrupt bundle" error. Robustness, not security — but cheap to get right now.

nnunley (changes requested):
See the comments on the PR for the request. But TL;DR - collapse the two trailer parsers into one (getBaseBinarySize dropped its Seek error checks) and add a size guard so a corrupt binary fails cleanly instead of panicking at make()

abogoyavlensky:
@nnunley Hi, thanks for the review! It's fixed: one `parseBundleTrailer` for both call sites (Seek checks restored), plus a size guard so a corrupt trailer errors cleanly instead of panicking at `make()`. The fit check subtracts rather than sums to avoid `uint64` overflow. Tests added.

nnunley (approved):

PR #181, opened 2026-06-03
feat(core): upstream Clojure-standard fns (every-pred, eduction, ns-resolve, inst-ms, bytes?, …)

nnunley:
## What

Upstreams the genuinely Clojure-standard functions that the `malli-on-let-go` example previously shimmed, limited to those whose let-go implementation is **correct** (not a misleading stub). The JVM object-model surface (host classes/ctors/statics, `java.time`/`java.util` types, `TimeUnit`, `System/arraycopy`) stays in the example shim.

These landed on the original #172 branch but were not part of the merge; this re-PRs them, rebased onto current `main`.

| Added | Notes |
|-------|-------|
| `unchecked-{inc,dec,add,multiply}-int` | compose on the native `unchecked-*` ops, like the existing `unchecked-inc`/`-dec` (let-go is int64) |
| `every-pred` | complement of the existing `some-fn` |
| `eduction` | over `sequence`/`comp`; seqable + reducible (single-pass) |
| `hash-combine` | faithful `clojure.core` bit-mix of two hash codes |
| `*unchecked-math*` | honest no-op `^:dynamic` var (no unchecked compile mode here) |
| `->MapEntry` | constructor for let-go's first-class `MapEntry` type |
| `ns-resolve` | real, via native `resolve`/`ns-name`; nil for an absent ns/var |
| `inst-ms` | real, epoch millis of a let-go time value (e.g. `(now)`) via Go time |
| `bytes?` (native) | true for a byte-kind `TypedArray` |

## Verification

- `go test ./...` green; `test/malli_core_additions_test.lg` covers each fn (36 assertions).
- `make check-generated` green (bundle lockstep + `core_go_lowered/` native dispatch under `-tags gogen_ir`).

## Notes

- `string/replace-first` was found already in core, so it is **not** included (the shim entry was stale).
- Independent of #179 (faithful print-method); both touch `core.lg`/`lang.go`/`core_compiled.lgb`, so whichever merges second will need a trivial rebase + bundle regen.


nnunley:
Follow-up (not in this PR): `examples/malli-on-let-go/shim-manifest.edn` still documents the pre-trim surface counts (it lists e.g. `unchecked-*-int` under `:missing-core-fns`, and counts `print-method` as a stub). Once **both** this PR and #179 land, those entries are obsolete — the manifest should be refreshed to reflect the trimmed shim:

- drop from `:missing-core-fns`: `unchecked-inc-int`, `unchecked-dec-int`, `unchecked-add-int`, `unchecked-multiply-int` (now in core); keep `indexed?`
- drop `print-method` from `:stdlib-stubs` (now a real core multimethod)
- `->MapEntry` no longer a shim host-constructor (real core ctor)
- note that `hash-combine` and `inst-ms` are *intentionally retained* overrides (malli's `:bb` `hash-combine` hashes its 2nd arg; the example's identity-stubbed `java.time` types feed `inst-ms` non-time values), so they stay even though core now provides faithful versions

Deferred deliberately: splitting the manifest edit across this PR and #179 would make it inconsistent with each PR's partial shim state, so it's cleanest to refresh in a small doc-only change after both merge.

nooga (approved):

PR #182, opened 2026-06-03
fix(lower-go): deterministic gensym naming so lowered Go is reproducible

nnunley:
## Problem

The gogen_ir lowered tree (`pkg/rt/core_go_lowered/`) was **non-reproducible across regenerations**. Emitted Go identifiers embedded the process-global `gensym` counter, so a function's locals drifted depending on how many gensyms had already run in the process:

```
arg__8368   ->   arg__52670     (same source, different regen)
```

Fresh regenerations therefore diverged from the committed tree, which made `make check-generated` / `TestGeneratedArtifactsAreFresh` unreliable as freshness gates.

## Fix

Two parts, both **pure naming** — no control-flow, value, or type changes:

**1. `build.lg` — ctx-scoped build counter.** `new-context` seeds `:gensym 0`; a new `ctx-gensym!` replaces `(gensym "arg__")` / `(gensym "head__")` for IR-build slot keys. Generated names become a pure function of build order *within a single function build*, not of global process state.

**2. `lower_go.lg` — per-function dense renumber at emission.** `gensym-renumber-map` assigns dense ordinals (`<prefix>__1`, `__2`, …) by deterministic first-occurrence order (`oos-all-nids` traverses blocks→params→insts structurally). `canonical-name` applies it at final identifier emission (`value-name` + `coalesce-map`).

A **collision guard** in `coalesce-map` drops any name that would munge to the same Go identifier as a *distinct* source name — `go-name` maps `-`→`_`, so a real `x--1` could collide with a canonicalized `x__1`. Conflicting names fall back to `value-name`'s versioned `_<nid>` path: strictly conservative, never aliases live values.

## Verification

- **Reproducibility:** two fresh-process regenerations of the full lowered tree are **byte-identical** (`diff -r` clean).
- **Fix active:** emitted gensym names are dense per-function ordinals; **zero** global-counter remnants (no `arg__`/`head__` with 4+-digit suffixes).
- **Compiles:** `go build -tags gogen_ir ./...`.
- **Behavioral parity:** the full `.lg` test suite passes on **both** the bytecode path and `-tags gogen_ir` (the native renamed tree behaves identically).
- Going forward this makes `make check-generated` a **reliable** determinism guard.

## Scope

Source-only: `build.lg` + `lower_go.lg` + the regenerated `core_compiled.lgb`. The lowered tree (`pkg/rt/core_go_lowered/`, a gitignored build artifact) is **not** included — it regenerates deterministically from these sources.

> Note: a pre-existing `WARNING: lowering failed for lower: ir/build: unresolved symbol e` on main is unrelated to this change and is addressed separately in the direct-call stack.


nooga (changes requested):
Requesting changes for the `with-meta` behavior change. The other items are smaller cleanup points, but the first one widens runtime semantics enough that I would not merge this as-is.

1. `pkg/rt/lang.go:5194-5198`

   `with-meta` now returns the original value for every non-`IMeta` value. That fixes `(with-meta P ...)` for protocols, but it also makes invalid calls like `(with-meta 1 {})` and `(with-meta "x" {})` silently succeed instead of reporting unsupported metadata.

   Please keep the unsupported-value error and make `Protocol` metadata-aware, or explicitly special-case only protocol values if the intended compat behavior is a no-op there.

2. `docs/contribution-policy.md:89` and `docs/contribution-policy.md:217-219`

   This policy is now marked authoritative, but it still says `core_go_lowered/` is a committed tree checked as committed bytes. The PR/Makefile now treats that tree as a gitignored regenerated artifact, so the doc should describe the new checkpoint accurately.

3. `pkg/compiler/programs_bench_test.go:43`

   The benchmark ignores `fr.Run()` errors, so a benchmark program can fail at runtime while the benchmark still reports success. Please check the error and fail the benchmark; this should also release the frame after each run so allocation results are not skewed by bypassing the frame pool.

nooga (approved):

nnunley:
### Context on the `with-meta` change + what "doing it properly" requires

Investigation while reconciling this against current `main`: the `with-meta` non-`IMeta` passthrough isn't a malli-only nicety — it's **load-bearing for type hints**. The reader turns a value-position `^long x` into a runtime `(with-meta x {:tag (quote long)})` (`reader.go:readMeta`). So making `with-meta` strict (erroring on non-`IMeta`) breaks **every `^long`/`^Long`-hinted scalar**, language-wide — minimal repro:

```clojure
((fn [x] (< 1 ^long x)) 5)   ; strict with-meta => "with-meta not supported on let-go.lang.Int"
```

Scalars can't simply be made `IMeta`: `Int`/`Float`/`Char`/`Keyword`/`Symbol` are value types (`type Int int`), so carrying a `meta` field means boxing every number (perf/identity/`=`), and Clojure doesn't do it either.

**What I'm landing separately (clean, independent):** `Protocol` and `DTypeInstance` (reify) now implement `IMeta`, so reference types carry *real* metadata via the normal path instead of the no-op. That tightens the actual gap behind `(with-meta P …)` / reify-schema metadata.

**What the *deeper* integration genuinely requires** (so it's scoped honestly, not hand-waved) — threading source hints to where they belong (the compiler/IR) instead of onto runtime values:

1. **Reader** already preserves the hint (as the `with-meta` form) — nothing to change there.
2. **Bytecode compiler**: the binding-strip for fn params (`compiler.go:199`) and let-names (`compiler.go:1103`) currently *unwraps and discards* the `{:tag …}` ("we don't yet attach meta to locals, just drop it"). The `locals` table is `map[Symbol]int` (symbol→slot) — **no slot for a tag**. Preserving hints means a parallel per-binding tag map.
3. **IR build → typeinfer**: typeinfer's `:tag` (`ir/passes/typeinfer.lg`) is the IR's *own* type lattice (`:int`/`:float` on typed nodes) — **there is no pipeline from source `^long` hints into it today**. Real integration = emit the binding tag onto IR local nodes and have typeinfer consume it (unboxing / type narrowing).

That's a multi-layer reader→compiler→IR feature, not a one-liner — and it's the prerequisite for ever making `with-meta` strict without regressing hinted scalars.

**Minimal step that unblocks "hints aren't discarded" without the full feature:** stop the binding-strip from dropping the tag — record it on the binding (a parallel `map[Symbol]Value`) so it's *available* to later stages at zero runtime cost, no scalar boxing, and no change to `with-meta` semantics. Deeper typeinfer consumption can follow when there's a consumer.

Recommendation: drop the strict-`with-meta` change from this PR (it can't ship without step 2–3 above); keep the gensym-determinism fix on its own; take `Protocol`/`DTypeInstance` `IMeta` as the separate win; and track the hint-propagation feature on its own.

PR #199, opened 2026-06-06
feat(source-paths): explicit -source-paths/LG_SOURCE_PATHS fully define the search path

### LINKED ISSUE #198: source-paths: Drop the implicit `"."` so an explicit `-source-paths` fully controls the search path
## Problem

When you pass `-source-paths` (or set `LG_SOURCE_PATHS`), `lg` still prepends the current directory `"."` to the namespace search path. The explicit roots never stand alone:

- `-source-paths a:b` -> `[".", "a", "b"]`
- `LG_SOURCE_PATHS=a:b` -> `[".", "a", "b"]`

So a stray `foo.lg` in the working directory can satisfy a `(require 'foo)` that was meant to resolve from one of the declared roots. The caller asked for a specific set of roots and silently got an extra, invisible one.

## Motivation

The trigger is lgx (the project manager for let-go, in `../lgx`). lgx resolves dependencies to cached directories and passes them through `-source-paths`. When it does, it expects those roots to be the complete, authoritative list - nothing else loadable. The implicit `"."` undermines that: namespaces in the invocation directory leak in and can shadow the resolved ones, so a build depends on where `lg` happens to run.

The principle: providing source-paths, by any channel, means "I am taking control of the search path."

## Target behavior

When the search path is supplied explicitly - the `-source-paths` flag is present, or `LG_SOURCE_PATHS` is set  being
non-empty) - it fully defines the search path. `"."` is not added; list it to search the current directory.

| Invocation | Before | After |
|---|---|---|
| `-source-paths a:b` |  `[".","a","b"]` | `["a","b"]` |
| `-source-paths ""` | `["."]` | `[]` (no paths) |
| `LG_SOURCE_PATHS=a:b` (no flag) | `[".","a","b"]` | `["a","b"]` |
| `LG_SOURCE_PATHS=` (present, empty) | `["."]` | `[]` (no paths) |
| `LG_SOURCE_PATHS` unset, no flag | `deps.edn` / `["."]` | unchanged |
| bare `lg foo.lg` | `["."]` | unchanged |

- To include the working directory, list it: `-source-paths .:lib`.
- A present-but-empty value on either channel means "no paths" (only embedded namespaces resolve). The flag and the env var behave the same.
- The script passed on the command line is always loaded by its path, independent of the search path, so `lg path/to/app.lg` is unaffected.

abogoyavlensky:
Resolves: #198


nnunley (changes requested):
Per our discussion, it would good to have a warning for a release or two in order to indicate that "." isn't implicitly added.

abogoyavlensky:
@nnunley Thank you for the comment, I've added warning for this, and also added a temp env var to suppress it `LG_SUPPRESS_SOURCE_PATHS_WARNING`, because the warn message can be quite noisy to see every time.

nnunley:
Can you add an issue to remove it after the next release? I don't think we need to spam people beyond that.

abogoyavlensky:
@nnunley Yes, sure, I've created a cleanup issue: #200

nnunley (approved):

PR #205, opened 2026-06-08
feat(args): add core/*command-line-args* var

### LINKED ISSUE #204: Add *command-line-args* for reliable access to a program's arguments
### Problem

A let-go program can handle its own arguments only with `os/args`, the full process argv, and its shape depends on how the program was launched:

| Invocation | `os/args` |
|---|---|
| `lg app.lg a b` | `["lg" "app.lg" "a" "b"]` |
| `lg -d app.lg a b` | `["lg" "-d" "app.lg" "a" "b"]` |
| `./bundled a b` | `["./bundled" "a" "b"]` |

Because the layout shifts, each program has to reconstruct where the user's args begin. Every reconstruction is a guess, and each one breaks somewhere:

- `(drop 2 os/args)` - wrong when a flag precedes the script, or for a bundle.
- `.lg`-suffix sniffing to find the script - wrong if a real arg ends in `.lg`.
- a `--` marker plus an env flag to detect run mode - breaks on inherited env, and collides with a literal `--` in the user's command (`app git checkout -- file`).

These heuristics exist only because `os/args` has already discarded the structure needed to answer the question.

### Proposal

Add `clojure.core/*command-line-args*`, matching Clojure and Babashka: the program's arguments (the positionals after the script) as a seq of strings, or `nil` when there are none. `os/args` stays the full process argv, so the change is additive and breaks nothing.

`lg` is the only layer that knows the boundary with certainty: it ran the flag parser. It should compute the args once and publish them, rather than have every script re-derive them.

```clojure
;; greet.lg - run as `lg greet.lg Alice Bob` or `./greet Alice Bob`
(doseq [name *command-line-args*]
  (println "Hello," name))
```

### Behavior

| Invocation | `*command-line-args*` |
|---|---|
| `lg app.lg a b` | `("a" "b")` |
| `lg -source-paths . app.lg a b` | `("a" "b")` - a flag before the script doesn't shift it |
| `lg app.lg git checkout -- file` | `("git" "checkout" "--" "file")` - literal `--` preserved |
| `lg app.lg` / `./bundle` / REPL | `nil` |
| `./bundle a b` | `("a" "b")` |

abogoyavlensky:
Resolves: #204


nooga (approved):

nooga:
Thanks @abogoyavlensky! Good addition :)

PR #206, opened 2026-06-09
feat(rt): route print/pr/prn/println through *out* and error sites through *err*

mparrett:
## Summary

let-go already has the right abstraction for I/O: `*in*`/`*out*`/`*err*` are dynamic vars holding IOHandles (`pkg/rt/iort.go`), and `(write! *err* "x")` and `(binding [*out* h] (write! *out* ...))` both work today. What's missing is that the four top-level print fns (`print`/`pr`/`prn`/`println`) and the runtime's error-output sites **don't consult those vars** — they write straight to `os.Stdout`/`os.Stderr`. So `(binding [*out* *err*] (println "x"))` silently writes to stdout, where Clojure writes to stderr.

This PR finishes the abstraction: the print fns resolve the current dynamic binding of `*out*` at each call, error sites resolve `*err*`, and the two internal output-capture mechanisms that depended on swapping `os.Stdout` are rewritten to use the binding stack instead.

Default behavior is unchanged — `*out*`'s root binding is still the IOHandle wrapping `os.Stdout`, so bare `(println ...)` is identical to today.

## Context — where this fits

This is the runtime-semantics layer under the browser-interop direction in #79. #174 (LetGoHost: `onOutput`/`onEmit`/…) landed the **JS-side** host surface; this PR is the **Go-side** plumbing that makes it clean — once `print`/`pr`/`prn`/`println` honor `*out*`, the runtime stops reaching for `os.Stdout` directly, and `pkg/api.WithStdout` (a follow-up PR) becomes the exact dual of `LetGoHost.onOutput`, eventually retiring the `fs.writeSync` shim. Scope here is deliberately just the runtime semantics; the embedder API and the WASM host writer are separate, sequenced follow-ups.

## Why this matters (it's a real compat bug, not just ergonomics)

`(binding [*out* *err*] ...)` is the idiomatic Clojure way to send output to stderr, and downstream let-go code already relies on it. Surveyed instances that are silently broken today:

- `nooga/lgcr` ships three copies of `(defn die [& m] (binding [*out* *err*] (println ...)) (os/exit 1))` (cli.lg, imgstore.lg, container.lg) — every error message has been going to **stdout**.
- `lgcr`'s `stream-file` does `(binding [*out* sink] (print chunk))` to redirect a stream — the redirect is a no-op today, and it's on the container log-streaming path (`stdout.log`/`stderr.log`).
- *(Both lgcr cases are filed and repro'd at nooga/lgcr#1.)*
- `examples/ys-on-let-go/ys-runtime.lg`'s `warn` has the same `die`-shaped bug.

After this PR, each starts behaving as written once the consumer bumps its let-go pin to a release that includes it. That pin bump is the one required step — there are no code changes on the consumer side beyond it.

## The 6 commits (reviewable independently)

1. **`refactor(rt): IOHandle wraps io.Writer/io.Reader`** — IOHandle was `*os.File`-locked; now it wraps an `io.Writer`/`io.Reader` pair, keeping the `*os.File` for handles that need `Fd()`/`Sync()` (raw-mode term, fsync). One-way handles are nil-guarded so a write-only handle passed to a reader (or vice-versa) returns a clear error instead of panicking.
2. **`feat(rt): print/pr/prn/println consult dynamic *out*`** — the 5 stdout sites (4 print fns + the pod stdout router) go through a small `WriteToOut` helper that resolves `*out*` via the core ns + `Var.Deref` (lock-free).
3. **`feat(rt): error sites consult dynamic *err* (10 sites)`** — error output in `pkg/rt/*` goes through `*err*`. Six of these were *double-wrong*: naked `fmt.Println` targeting **stdout** for semantically-error output (async-fn errors, HTTP error logs).
4. **`feat(rt): with-out-str* uses binding-based capture`** — see "the interlock" below.
5. **`feat(nrepl): eval-capture uses binding-based *out* rebind`** — the same rewrite for nREPL's per-eval capture; exports `rt.LookupCoreVar` for it.
6. **`test(io): pin the *out* binding semantics`** — 8 `.lg` tests covering each print fn, the `die` pattern, the arbitrary-handle pattern, nesting, and the default-unchanged case.

## Breaking changes

Two, both worth a reviewer's attention:

1. **`IOHandle.File` (exported field) → `IOHandle.File()` (method).** The struct now wraps an `io.Writer`/`io.Reader` pair and keeps the `*os.File` privately, so the field became an accessor. Any Go embedder reading `handle.File` must switch to `handle.File()`. All in-repo callers are updated; this only affects external code holding the old field.
2. **`print`/`pr`/`prn`/`println` no longer follow a runtime reassignment of `os.Stdout`.** They resolve `*out*`, whose default root binding wraps the `os.Stdout` present at runtime init. Go code that does `os.Stdout = w` at runtime to capture output — the pattern this PR removes from `with-out-str`/nREPL — won't be picked up; rebind `*out*` instead. The WASM path is unaffected: the default handle still wraps fd 1, so writes still reach the `wasm_exec.js` fd hook (browser-verified, see Testing).

## Critical: the print and `with-out-str` changes are coupled

**The print refactor and `with-out-str` are non-separable.** `with-out-str*` (and the structurally identical nREPL eval-capture) captured output by swapping the process-global `os.Stdout`. Once `println` stops writing to `os.Stdout` and goes through the IOHandle instead, that swap captures nothing — `(with-out-str (println "x"))` would return `""` and leak to the real stdout. So commits 4 and 5 rewrite both to capture by pushing a `bytes.Buffer`-backed binding on `*out*` for the body's scope. That's why they're in the same PR as the print change; they can't ship separately.

Two side effects of that rewrite: it removes a latent **64 KB truncation** in the old pipe-read path (it read the capture pipe into a fixed `64*1024` buffer), and a panic in the body no longer leaks process-global state (the binding is popped via `defer`).

## Concurrency

The binding-based capture isn't an isolation guarantee. `vm.Var` holds a single process-global binding stack (not goroutine-local, as on the Clojure JVM), so concurrent `with-out-str` calls — or a `with-out-str` racing an nREPL eval — can interleave push/pop and capture each other's output. In practice the surface is a multi-client nREPL server, or output produced from `(go …)` blocks during a capture. It removes the package-global side effect of the old `os.Stdout` swap and otherwise matches its safety: the global-state hazard is identical, not new. Real isolation would need goroutine-local binding stacks in `vm.Var` or serialization at the capture layer — out of scope here. The docstrings say this explicitly.

## Deliberately out of scope

- **3 `pkg/vm/*` error sites** stay on `os.Stderr` (panic recover, core-shadow warning, GLS drain timeout): `pkg/vm` can't import `pkg/rt` without a cyclic dep, and they're last-resort diagnostics.
- **44 `term/*` ANSI control sites** — separable Phase 1.5; open question whether `(term/clear)` inside a `*out*` rebind should follow it.
- **`pkg/api` embedder options** (`WithStdout`/`WithStderr`) — the follow-up PR, so the instance-isolation question gets its own review.

## Testing

`go test ./...` green across all packages. The 8 new `.lg` tests are in `test/io_binding_test.lg`, named in the descriptive-phrase `deftest` style already used across `test/` (e.g. `assoc-preserves-metadata`, `async-go-returns-channel`).

Beyond the unit suite:
- **WASM, in a real browser:** built the hosted REPL and confirmed `(println …)` still reaches fd 1 → the `wasm_exec.js` console hook, `(with-out-str …)` captures correctly, captured output does not leak, and `(binding [*out* *err*] …)` routes to fd 2. All pass.
- **Cross-platform build:** `GOOS=js GOARCH=wasm go build ./pkg/...` and the Plan9 `pkg/rt` build are both clean.
- **Throughput:** a 1M-`println` loop shows no reliable regression vs base — the added per-call `*out*` resolution is within measurement noise.

### DIFF
--- a/pkg/rt/lang.go
+++ b/pkg/rt/lang.go
@@ -3468,7 +3468,10 @@ func installLangNS() {
 		vm.CurrentScope().Go(func(ctx context.Context) {
 			v, err := at.Invoke(nil)
 			if err != nil {
-				fmt.Println(err)
+				// Async (go ...) error — route to *err*. Previously
+				// fmt.Println, which targets stdout despite being error
+				// output: double-wrong.
+				_ = WriteToErr(fmt.Sprintln(err))
 			}
 			// The result send is cancellable via the registry. (Channel
 			// ops <!/>! INSIDE the block are still synchronous and not yet
--- a/pkg/rt/lang.go
+++ b/pkg/rt/lang.go
@@ -6,6 +6,7 @@
 package rt
 
 import (
+	"bytes"
 	"context"
 	crand "crypto/rand"
 	_ "embed"
@@ -7452,7 +7453,19 @@ func installLangNS() {
 	})
 	ns.Def("pop!", popBang)
 
-	// with-out-str — capture stdout as string (macro helper)
+	// with-out-str* — capture *out* output as string. Binding-based capture
+	// (no process-global os.Stdout swap). Rebinds *out* to a bytes.Buffer-
+	// backed IOHandle for the thunk's scope; restores via defer so a
+	// panicking thunk doesn't leak the binding.
+	//
+	// Concurrency caveat: vm.Var holds a single process-global binding
+	// stack guarded by bindingsMu (pkg/vm/var.go:38). Concurrent
+	// with-out-str calls on the same *out* DO NOT isolate captures from
+	// each other — their push/pop interleavings can cause one goroutine's
+	// println to land in another's buffer. This is no worse than the
+	// previous os.Stdout swap (which had the same global-state race),
+	// but it's also not better. Real isolation would require either
+	// goroutine-local binding stacks in vm.Var or external serialization.
 	withOutStrf, _ := vm.NativeFnType.Wrap(func(vs []vm.Value) (vm.Value, error) {
 		if len(vs) != 1 {
 			return vm.NIL, fmt.Errorf("with-out-str* expects 1 arg (a thunk)")
@@ -7461,23 +7474,19 @@ func installLangNS() {
 		if !ok {
 			return vm.NIL, fmt.Errorf("with-out-str* expects a function")
 		}
-		// Capture stdout
-		old := os.Stdout
-		r, w, err := os.Pipe()
-		if err != nil {
-			return vm.NIL, err
+		outVar := lookupNSCached(NameCoreNS).LookupLocal(vm.Symbol("*out*"))
+		if outVar == nil {
+			return vm.NIL, fmt.Errorf("with-out-str*: *out* not installed")
 		}
-		os.Stdout = w
+		buf := &bytes.Buffer{}
+		handle := vm.NewBoxed(NewWriterHandle("with-out-str*", buf))
+		outVar.PushBinding(handle)
+		defer outVar.PopBinding()
 		_, callErr := fn.Invoke(nil)
-		w.Close()
-		os.Stdout = old
-		buf := make([]byte, 64*1024)
-		n, _ := r.Read(buf)
-		r.Close()
 		if callErr != nil {
 			return vm.NIL, callErr
 		}
-		return vm.String(buf[:n]), nil
+		return vm.String(buf.String()), nil
 	})
 	ns.Def("with-out-str*", withOutStrf)
 
--- a/pkg/rt/lang.go
+++ b/pkg/rt/lang.go
@@ -4828,14 +4828,13 @@ func installLangNS() {
 		return prThroughToString(vs, true)
 	})
 
-	// prn: print readably + newline to stdout
+	// prn: print readably + newline through *out*
 	prn, _ := vm.NativeFnType.Wrap(func(vs []vm.Value) (vm.Value, error) {
 		s, err := prThroughToString(vs, true)
 		if err != nil {
 			return vm.NIL, err
 		}
-		fmt.Fprintln(os.Stdout, string(s.(vm.String)))
-		return vm.NIL, nil
+		return vm.NIL, WriteToOut(string(s.(vm.String)) + "\n")
 	})
 
 	// prn-str: print readably + newline to string
@@ -5454,34 +5453,31 @@ func installLangNS() {
 		return vm.MakeInt(c), nil
 	})
 
-	// print — print human-readably to stdout, no newline
+	// print — print human-readably through *out*, no newline
 	printf, _ := vm.NativeFnType.Wrap(func(vs []vm.Value) (vm.Value, error) {
 		s, err := prThroughToString(vs, false)
 		if err != nil {
 			return vm.NIL, err
 		}
-		fmt.Fprint(os.Stdout, string(s.(vm.String)))
-		return vm.NIL, nil
+		return vm.NIL, WriteToOut(string(s.(vm.String)))
 	})
 
-	// pr — print readably to stdout, no newline
+	// pr — print readably through *out*, no newline
 	prf, _ := vm.NativeFnType.Wrap(func(vs []vm.Value) (vm.Value, error) {
 		s, err := prThroughToString(vs, true)
 		if err != nil {
 			return vm.NIL, err
 		}
-		fmt.Fprint(os.Stdout, string(s.(vm.String)))
-		return vm.NIL, nil
+		return vm.NIL, WriteToOut(string(s.(vm.String)))
 	})
 
-	// println — print human-readably to stdout with newline
+	// println — print human-readably through *out* with newline
 	printlnf, _ := vm.NativeFnType.Wrap(func(vs []vm.Value) (vm.Value, error) {
 		s, err := prThroughToString(vs, false)
 		if err != nil {
 			return vm.NIL, err
 		}
-		fmt.Fprintln(os.Stdout, string(s.(vm.String)))
-		return vm.NIL, nil
+		return vm.NIL, WriteToOut(string(s.(vm.String)) + "\n")
 	})
 
 	// --- Bitwise ops ---
--- a/pkg/rt/pods.go
+++ b/pkg/rt/pods.go
@@ -141,12 +141,15 @@ func (p *Pod) startRouter() {
 				return
 			}
 
-			// Handle out/err on any message
+			// Handle out/err on any message — route through the runtime's
+			// dynamic *out*/*err*, not raw stdio. Pod stdout/stderr is
+			// pod content; it should follow whatever capture/redirect the
+			// caller has set up via (binding ...) or pkg/api options.
 			if out, ok := msg["out"].(string); ok {
-				fmt.Print(out)
+				_ = WriteToOut(out)
 			}
 			if errStr, ok := msg["err"].(string); ok {
-				fmt.Fprint(os.Stderr, errStr)
+				_ = WriteToErr(errStr)
 			}
 
 			// Route by id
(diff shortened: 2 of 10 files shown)


nooga (approved):

pkg/rt/lang.go
```diff
@@ -7453,7 +7453,19 @@ func installLangNS() {
...
 	ns.Def("pop!", popBang)
 
-	// with-out-str — capture stdout as string (macro helper)
+	// with-out-str* — capture *out* output as string. Binding-based capture
+	// (no process-global os.Stdout swap). Rebinds *out* to a bytes.Buffer-
+	// backed IOHandle for the thunk's scope; restores via defer so a
+	// panicking thunk doesn't leak the binding.
+	//
+	// Concurrency caveat: vm.Var holds a single process-global binding
+	// stack guarded by bindingsMu (pkg/vm/var.go:38). Concurrent
```
nnunley, line 7462: **The concurrency model described here is inverted, and the cited symbol doesn't exist.** This comment — and the matching block in `pkg/nrepl/server.go:245` — state bindings are *"process-global, not per-goroutine"* guarded by `bindingsMu (pkg/vm/var.go:38)`. Verified against the tree:

- `bindingsMu` exists nowhere (`grep -r bindingsMu pkg/` → 0 hits), and `var.go:38` is unrelated code inside `Var.Invoke`.
- Bindings are actually **goroutine-local** — `pkg/vm/binding_frames.go:15,61` (`goroutineLocalBindingStore`, "a goroutine sees only its own frame unless bindings are explicitly propagated"); `Var.Deref()` reads the current goroutine's frame.

So the real behavior is *safer* than documented: concurrent nREPL evals / `with-out-str` calls on **different goroutines have isolated captures**, not interleaved ones. Not a runtime bug, but worth correcting so a future reader doesn't add unnecessary serialization or misdiagnose a capture issue.

nnunley (commented):
Three inline notes on the Important findings from an agentic review of this branch. None are Critical — the refactor is compile-safe on all platforms and a net improvement; these are accuracy/robustness call-outs.

pkg/rt/unix_linux.go
```diff
@@ -112,7 +112,7 @@ func installUnixNS() {
 			return int(fv), nil
 		case *vm.Boxed:
 			if h, ok := fv.Unbox().(*IOHandle); ok {
-				return int(h.File.Fd()), nil
+				return int(h.File().Fd()), nil
```
nnunley: **Missing nil-guard — potential panic.** After the refactor `File()` returns `nil` for any non-file-backed handle, so `h.File().Fd()` panics if a writer/reader-only `IOHandle` (e.g. a `bytes.Buffer`-backed `*out*`) reaches `extractFd`. The PR added exactly this guard to the sibling site `term.go:81`, so this reads as an oversight rather than intent. Suggest mirroring it:

```go
f := h.File()
if f == nil {
    return 0, fmt.Errorf("IOHandle is not file-backed; fd ops require a real file descriptor")
}
return int(f.Fd()), nil
```

The same applies to `syscall_linux.go:763`, which returns `h.File()` unchecked (a non-file handle as spawn stdio → nil `*os.File` downstream).

pkg/rt/pods.go
```diff
@@ -141,12 +141,15 @@ func (p *Pod) startRouter() {
 				return
 			}
 
-			// Handle out/err on any message
+			// Handle out/err on any message — route through the runtime's
+			// dynamic *out*/*err*, not raw stdio. Pod stdout/stderr is
+			// pod content; it should follow whatever capture/redirect the
```
nnunley: **This routing can't actually follow a caller's `(binding [*out*/*err*] …)`.** `startRouter` runs in its own goroutine, and neither `vm.Scope.Go` nor a bare `go` propagates dynamic bindings (`pkg/vm/scope.go:97`; binding frames are goroutine-local). So `WriteToOut`/`WriteToErr` here — plus the stderr-stream `resolveIOHandleVar("*err*")` at ~L523 and the async `(go …)` error path in `lang.go` — always resolve to the **root** `*out*`/`*err*` (process stdio), never the caller's rebinding.

**Not a regression** (the prior `fmt.Print` / `io.Copy(os.Stderr, …)` also hit process stdio, and the async case is an improvement since errors now go to `*err*`→stderr instead of stdout). The note is just that the comment promises capture-following the goroutine-local model doesn't deliver — either soften the comment, or snapshot + propagate bindings before spawning if caller-scoped capture is genuinely wanted.

pkg/rt/lang.go
nnunley: Following up after reading the #206 and #207 descriptions: both describe `vm.Var`'s binding stack as *process-global, not goroutine-local*, and treat "goroutine-local binding stacks in vm" as future / out-of-scope work (in #207 it's framed as "the open question — instance isolation").

That backend already appears to be wired up on this branch, though: `pkg/vm/binding_frames.go:33` builds `newGoroutineLocalBindingStore()` as the *only* store, keyed by a real per-goroutine `goID()` (`pkg/vm/scope_gls.go:34`, parsed from the runtime stack). If that's the live path, the per-goroutine isolation #207 raises as the open question would already hold for `Run`s on different goroutines, and this caveat overstates the hazard rather than understating it.

Worth reconciling the comments + both PR descriptions with the actual backend — or confirming there's a reason it still behaves process-globally in practice.

mparrett:
Thanks for the careful pass, @nnunley — appreciate it. Replies inline:

**`unix_linux.go` / `syscall_linux.go` nil-guard — agreed, fixed.** Good catch: after the field→method change `File()` returns `nil` for a non-file-backed handle, and both sites called through it unchecked. Mirrored the `term.go:81` guard at both in `aae8b51`.

**On the concurrency model — I think the agentic pass may have been reading a different tree.** On this PR's base (`origin/main` `9aa5354`) the `*out*`/`*err*` dynamic bindings are process-global, not goroutine-local:

- `bindingsMu` is at `pkg/vm/var.go:38`, guarding the per-`Var` `bindings []Value` stack (`var.go:25`, locked in `PushBinding`/`PopBinding`).
- `Var.Deref` reads the per-`Var` atomic `curr`/`root` (`var.go:95`) — lock-free since your own #122 (`a898a39`), but still process-global: `curr` is a single pointer shared across goroutines, not a per-goroutine frame.
- `grep -rn 'goroutineLocalBindingStore\|binding_frames' pkg/` → 0 hits; `binding_frames.go` isn't in the worktree, and I couldn't find it in any commit's history either.

The goroutine-local machinery that *does* exist is `Scope` (`scope_gls.go`, keyed by `goID()`, from #123) — the supervision-tree scope store, which is separate from the `Var` binding stack. My guess is the review conflated those two, or ran against a branch with a goroutine-local binding refactor that isn't on this base. If such a tree exists, I'd genuinely like a pointer — it would change the picture.

As written, then, the "process-global binding stack" caveat in the docstrings and in both PR descriptions matches the code, and the instance-isolation question in #207 is genuinely open: concurrent `with-out-str` / multi-client nREPL evals on different goroutines can interleave a shared binding. Happy to be corrected if I'm misreading something.

Where this gets interesting: that open question in #207 is precisely "should `*out*`/`*err*` bindings become goroutine-local?", and its stated resolution is goroutine-local binding stacks in `vm.Var` — for which your `Scope` GLS (#123) looks like the natural substrate. So the behavior your review describes may be less "what this branch does" than "where this should land" — and if you think goroutine-local bindings are the right call, that's a great direction to pin down on #207. It would close the isolation question outright and make the per-goroutine capture isolation real.

**`pods.go` routing comment — fair, reworded in `67594aa`.** You're right that `startRouter` resolves to the root `*out*`/`*err*` in practice. The reason on this tree is binding *lifetime* rather than goroutine-locality: the router is a long-lived async goroutine that typically outlives the caller's `(binding …)` scope, so the binding is popped before pod messages arrive — process-global visibility doesn't help once the binding is gone. The comment now promises root-by-default instead of caller-following.

Thanks again — the nil-guard was a genuine bug, and the prompt to reconcile the docs against the actual backend was worth doing.

PR #207, opened 2026-06-10
feat(api): WithStdout/WithStderr embedder options for *out*/*err*

mparrett:
## Summary

Adds two `pkg/api` constructor options that let a Go embedder redirect a runtime's output without touching globals:

```go
api.New(api.WithStdout(w), api.WithStderr(w2))
```

Each option pushes a dynamic binding on `*out*`/`*err*` **for the scope of `Run`** and pops it on return (per-`Run` scope, leaving the root vars untouched), so two runtimes configured with different writers stay independent. This is the Go-side dual of the JS `LetGoHost.onOutput` that #174 already shipped; once it's in, the `fs.writeSync` shim can eventually retire.

`WithStdin`/`WithEmit` are deliberately **not** exposed here (input is tied to the SAB/`wake()` conversation #174 deferred; a public no-op `WithEmit` is worse than none).

## Follows #206 (merged)

Builds on #206 (`d6d0128`, now merged), which routes print/error through `*out*`/`*err*`; this consumes that `*out*` resolution. Single commit on `main` — `pkg/api/{api,options_test}.go`, +239/-1.

## The open question — instance isolation

`vm.Var`'s binding stack is process-global, not goroutine-local (unlike Clojure-on-JVM). So per-`Run` scoping holds for the common embed-and-run case, but two runtimes whose `Run`s overlap *on different goroutines* can interleave each other's bindings. The tests pin the sequential-isolation guarantee (`TestTwoRuntimesIsolation`, `TestNoOptionRuntimeUnaffectedByConfiguredOne`); true concurrent isolation would need goroutine-local binding stacks in `vm`, which is a bigger change I kept out of scope. Flagging it because it's the part most worth a maintainer's eye.

## Tests

6 Go tests in `pkg/api/options_test.go`: `TestWithStdout`, `TestWithStderr`, `TestPrintFnsConsultOut`, `TestWithOutStrUnaffected`, `TestTwoRuntimesIsolation`, `TestNoOptionRuntimeUnaffectedByConfiguredOne`. `go test ./...` green.

---

Totally open to prototyping other approaches if you had something in mind.

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"io"
 	"reflect"
 
 	"github.com/nooga/let-go/pkg/compiler"
@@ -9,13 +10,86 @@ import (
 	"github.com/nooga/let-go/pkg/vm"
 )
 
+// Option configures a LetGo runtime instance at construction. Options
+// are applied in order; later options override earlier ones for the
+// same configuration key.
+type Option func(*config)
+
+// config collects the resolved option values. Pure data, no behavior.
+type config struct {
+	stdout io.Writer
+	stderr io.Writer
+}
+
+// WithStdout configures the runtime to route output written via *out*
+// (i.e. (println ...), (print ...), (pr ...), (prn ...), and any user
+// code that consults *out*) through w for this runtime instance.
+//
+// Implementation: each Run call pushes an IOHandle wrapping w as a
+// dynamic binding on *out*, then pops it on return. Per-Run scope
+// means two LetGo instances with different WithStdout values DO NOT
+// interfere with each other's sequential output capture, and a
+// LetGo constructed without WithStdout sees the unaltered os.Stdout
+// default regardless of what other instances have done.
+//
+// Concurrency caveat: vm.Var's binding stack is process-global. If
+// two goroutines call Run on different LetGo instances concurrently,
+// their bindings push onto the same *out* stack — push/pop
+// interleavings can scramble captures. For deterministic isolation,
+// serialize Run calls or run different instances in different
+// processes.
+//
+// Default: os.Stdout (no binding pushed; runtime root unchanged).
+func WithStdout(w io.Writer) Option {
+	return func(c *config) { c.stdout = w }
+}
+
+// WithStderr configures the runtime to route output written via *err*
+// (HTTP server error logs, pod stderr, panic reports, etc.) through w
+// for this runtime instance. Same per-Run binding semantics as
+// WithStdout; same concurrency caveat.
+//
+// Default: os.Stderr.
+func WithStderr(w io.Writer) Option {
+	return func(c *config) { c.stderr = w }
+}
+
+// (Other options deliberately NOT exposed:
+//
+//   - WithStdin: stdin substitution is tied to the wake() / SAB protocol
+//     deferred from nooga/let-go#174 and the readline-driven REPL path.
+//     *in*'s root binding remains os.Stdin; embedders that need stdin
+//     substitution today can rebind *in* manually before calling Run.
+//
+//   - WithEmit: a callback for (js/emit ...) would require either
+//     a *emit-fn* var in the core ns that js/emit consults, or a
+//     refactor of pkg/rt/js_wasm.go's _lgEmit lookup path. Neither is
+//     done. Until the wiring is real, a public option that silently
+//     does nothing would be misleading.)
+
 type LetGo struct {
 	cp     *vm.Consts
 	c      *compiler.Context
 	loader *resolver.NSResolver
+
+	// stdoutHandle/stderrHandle are pre-constructed Boxed IOHandles, or
+	// nil when no option was supplied. Run pushes them as dynamic
+	// bindings on *out*/*err* and pops on return.
+	stdoutHandle vm.Value
+	stderrHandle vm.Value
 }
 
-func NewLetGo(ns string) (*LetGo, error) {
+// NewLetGo constructs a runtime. With no options, behavior is exactly
+// as it was pre-option: *out* and *err* keep their global default root
+// bindings (os.Stdout / os.Stderr). Each I/O option installed gets
+// pushed as a dynamic binding around each Run call, so different
+// LetGo instances don't share or overwrite each other's I/O routing.
+func NewLetGo(ns string, opts ...Option) (*LetGo, error) {
+	cfg := &config{}
+	for _, o := range opts {
+		o(cfg)
+	}
+
 	cp := vm.NewConsts()
 	nso := rt.NS(ns)
 	c := compiler.NewCompiler(cp, nso)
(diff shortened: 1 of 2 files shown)


pkg/api/api.go
```diff
@@ -9,13 +10,86 @@ import (
...
+// code that consults *out*) through w for this runtime instance.
+//
+// Implementation: each Run call pushes an IOHandle wrapping w as a
+// dynamic binding on *out*, then pops it on return. Per-Run scope
+// means two LetGo instances with different WithStdout values DO NOT
+// interfere with each other's sequential output capture, and a
+// LetGo constructed without WithStdout sees the unaltered os.Stdout
+// default regardless of what other instances have done.
+//
+// Concurrency caveat: vm.Var's binding stack is process-global. If
```
nnunley, line 35: **This concurrency caveat is inverted.** It says the binding stack is *process-global* and that two goroutines calling `Run` on different instances will interleave on the same `*out*` stack. But `vm.Var` bindings are **goroutine-local** — `pkg/vm/binding_frames.go:33` wires `newGoroutineLocalBindingStore()` as the only backend, keyed by a real per-goroutine `goID()` (`scope_gls.go:34`), and the frame map is `framesMu`-guarded. So two `Run`s on **different goroutines get isolated** binding frames; they don't scramble each other's captures.

The genuine caveats are different: (a) the unguarded `Var.isDynamic` write flagged below, and (b) nested/recursive `Run` on the *same* goroutine sharing one frame. Worth rewording so an embedder doesn't serialize Runs (or reach for separate processes) to avoid a hazard that isn't there.

pkg/api/api.go
```diff
@@ -9,13 +10,86 @@ import (
...
+// serialize Run calls or run different instances in different
+// processes.
+//
+// Default: os.Stdout (no binding pushed; runtime root unchanged).
+func WithStdout(w io.Writer) Option {
+	return func(c *config) { c.stdout = w }
+}
+
+// WithStderr configures the runtime to route output written via *err*
+// (HTTP server error logs, pod stderr, panic reports, etc.) through w
```
nnunley, line 48: **These enumerated sinks are exactly the ones `WithStderr` can't capture.** The `*err*` binding is pushed on the `Run` goroutine, but HTTP error logs, pod stderr, and async/panic reports are emitted from *other* goroutines that never inherit it:

- `pkg/rt/http.go` `WriteToErr` → net/http handler goroutines
- `pkg/rt/pods.go` router + stderr-stream → their own `go func()`s
- `pkg/rt/lang.go` `(go …)` error handler → `vm.Scope.Go` (no binding propagation)

Each resolves to the **root** `*err*` (`os.Stderr`), not `w`. Only synchronous `*err*` writes on the `Run` goroutine (`(println …)`, `(write! *err* …)`, etc.) are captured. An embedder wiring `WithStderr(auditBuf)` to capture diagnostics would silently miss precisely these categories — worth narrowing the doc to what's actually routed and calling out the cross-goroutine + Go-panic exclusions.

pkg/api/api.go
```diff
@@ -44,7 +126,29 @@ func (l *LetGo) Def(name string, value any) error {
...
+// return (including on error / panic).
+//
+// Two LetGo instances calling Run sequentially each get their own
+// configured streams: A's Run pushes A's handle, evaluates, pops; then
+// B's Run pushes B's handle. The root binding of *out*/*err* is
+// never mutated.
 func (l *LetGo) Run(expr string) (vm.Value, error) {
+	if l.stdoutHandle != nil {
+		if v := rt.LookupCoreVar("*out*"); v != nil {
+			v.PushBinding(l.stdoutHandle)
```
nnunley, line 141: **Real concurrency hazard for concurrent multi-instance `Run` lives here.** Both instances `PushBinding` on the *same* shared core `*out*`/`*err*` `Var`. `Var.PushBinding` does `v.isDynamic = true` **unguarded** (`pkg/vm/var.go:109`), while `Deref` reads `isDynamic` (`var.go:73,101`). Two `Run`s on different goroutines → concurrent write/write (+ write/read vs a third goroutine's `Deref`) on `isDynamic` = a data race that `go test -race` will flag. The written value is always `true`, so it's practically benign, but it's UB and detector-tripping.

Note the binding *store* is safe (goroutine-local + `framesMu`), so this flag — not the interleaving described in the `WithStdout` doc above — is the actual thing between this API and safe concurrent use. Guard the write with `Var.mu` or make `isDynamic` an `atomic.Bool` (it's outside this diff, but it's what the PR's "open question — instance isolation" really hinges on).

nnunley (commented):
Three inline notes on the unique `pkg/api` commit (the rest of the diff is #206, reviewed separately). The options mechanism itself is clean — per-Run push/defer-pop, root never mutated, signature backward-compatible. All three notes trace to the same fact as the #206 review: `vm.Var` bindings are goroutine-local, not process-global.

mparrett:
Thanks @nnunley. I've been trying to line up your three notes with the code and I genuinely can't find what they point at — wanted to flag that honestly rather than guess at a revision.

On this PR's base (`main` @ `d6d0128`) I'm seeing process-global `Var` bindings: `PushBinding` (`pkg/vm/var.go:125`) appends to a shared `bindings` slice under the package-global `bindingsMu`; `Deref` (`var.go:95`) reads only the `curr`/`root` atomics and never consults `isDynamic`; and `isDynamic` is written only by `SetDynamic()` (`var.go:321`). I also can't find a `binding_frames.go` / `newGoroutineLocalBindingStore` anywhere on the tree — the only goroutine-local store I see is `Scope` (`scope_gls.go`, #123), which is a separate mechanism from `Var` dynamic bindings.

The `binding_frames.go:33` and `var.go` lines your notes cite don't resolve for me on `d6d0128`, so I think we may be on different trees. Could you surface where you're seeing the goroutine-local `Var` bindings? If there's a refactor in flight I'm missing, I'd genuinely like to see it — it's exactly what would make `WithStdout`/`WithStderr` concurrency-safe, and I'd happily rebuild the caveat on top of it.

nnunley:
My apologies. Juggling too many things at once. I'm about to push the relevant code. I'll make sure to @ you so you can review it. Alternatively, I can just update my commit to fix your commit once it's ready.

mparrett:
@nnunley I'm open to whatever path creates the least churn for you — happy to hold here and wait on your code. Take your time.

I was pushing forward a little faster than I normally would on this one, mostly because of the number of slices that need to land together-ish to form a reasonable API design when all the dust settles. One more chunk after this PR. But making sure this and subsequent changes are aligned with yours is important, too. Excited to see where all of this ultimately lands... the direction is looking pretty clean so far.

nnunley:
Cool. I'm pushing to get my better per goroutine/thread isolation code in review shape now. I'd prefer to wait until that lands, but feel free to do stacked PRs to build towards your vision if I have you blocked for more than a few more hours today (10 Jun 2026)

mparrett:
Sounds good — we'll wait for your isolation code. It's the right base for this PR anyway: once goroutine-local `Var` bindings are in, the concurrency caveat here flips from "serialize Runs" to genuinely thread-safe, so I'd rather rebase onto yours and rewrite it than land it as-is. Ping me when it's up and I'll review. Thanks for the heads-up.

nooga (approved):

PR #210, opened 2026-06-11
feat(vm,ir): ExecContext threading — explicit dynamic-binding + scope context, delete goid

nnunley:
## What

Replaces the goroutine-id (goid) machinery for dynamic-var bindings and structured-concurrency scope with a single **`ExecContext`** that is threaded explicitly through every execution path — the bytecode VM **and** the `-tags gogen_ir` native-lowered path.

One `ExecContext` carries both the dynamic-binding stack and the `Scope`. `Frame` gains an `ec`; context-needing builtins become `CtxNativeFn`s (`WriteToOut(ec)`, print family, `in-ns`, channel ops, `future`/`go`/`bound-fn`); spawn sites capture a binding snapshot and run the worker under `ec.Child()`. The goid lookup tables (`scope_gls.go`: `scopeByGID`/`scopedLive`/`goID`) are deleted.

## Clojure parity

Brought thread/binding semantics in line with Clojure and locked them with **differential gold tests** (same `.cljc` script run under `clojure -M`, `bb`, and `lg`, byte-identical stdout — `gold_dynvar_e2e_test.go`):

- `test/gold/dynvar_threads.cljc` — future binding conveyance, isolation from later rebinds, nested-future inheritance, concurrent non-interference, `bound-fn` capture, thread-local `set!` mutation, **protocol + multimethod dispatch under a future binding**.
- `test/gold/ns_threads.cljc` — `*ns*` conveyance / isolation / `in-ns` thread-local mutation across nested threads.

Fixes where let-go diverged from Clojure: `set!`/`in-ns` now mutate the **top binding frame** of the current `ec` (not the var root) when a binding is active, falling back to root only when unbound; protocol-method and multimethod dispatch now route the selected impl through `ec.Invoke`, so dynamic vars read inside a method body resolve against the caller's context. Spots where let-go is intentionally **more permissive** than Clojure (e.g. `set!` with no active binding) are documented at the call sites.

## gogen_ir dangling-call fix

The two-pass native lowering built its direct-call registry from pass 1, but a fn registered there could take a different (registry-visible) path in pass 2 and bail without emitting a decl — leaving callers with a direct Go call to an undefined function. This surfaced as an intermittent `undefined: lower_terminator` when compiling the regenerated `-tags gogen_ir` tree (nondeterministic, per-pass typeinfer/gensym state). Fixed in `pipeline.lg` `lower-ns-to-go` by iterating pass 2 to a fixpoint where the registry holds only fns that actually emit a decl (`reg' = reg ∩ emitted ⊆ reg`, monotone → terminates; worst case empties → trampoline, always compiles).

## Verification

- Full bytecode suite green incl. `-race` on `pkg/vm` + `pkg/rt`; gold tests match live Clojure.
- `goid` deletion verified complete (no `scopeByGID`/`scopedLive`/`goID` remnants in `.go`/`.lg`).
- gogen_ir: 5/5 fresh `make generate`-style regen + `go build -tags gogen_ir ./...` clean (was intermittently failing before the dangling-call fix); `make check-generated` bundle lockstep OK.
- Two prior multi-agent reviews; the second found **0 Critical / 0 Important** after empirical verification (the loudest raw findings — "panic in spawned goroutine crashes the process" — were falsified: workers recover at the `Frame.Run` boundary).

## Known residual

Compile-time `*ns*` conflation: the compiler still switches `*ns*` at compile time via the shared var root (`tryDetectInNS`), so `in-ns` as the **first** form of a do-body leaks past the binding/thread boundary. Runtime `*ns*` isolation is correct; the compile-time case is tracked separately. `*in*` / `*compiling-aot*` ambient readers are deferred (design-note "Step 3 remainder").

Design note: `docs/design/exec-context-threading.md`.


mparrett (commented):
Read this through and pulled it into a worktree: `go build ./...` is clean, `go test ./...` is green, and `-race` on the binding/exec tests is green too. One caveat on the parity check: locally only the **pinned golden-constant** assertions run. The **live-Clojure differential** needs `clojure` on PATH, which this box doesn't have, and the CI workflow I checked doesn't appear to install it. Worth confirming that live comparison runs somewhere, since that's the load-bearing parity check.

The design direction makes sense to me, and the design note made the tradeoffs easy to follow: two goid-keyed mechanisms, the reuse footgun, and the load-bearing `pmap`/lgbgen isolation that rules out a naive global stack.

**Strengths**

- **Deleting goid earns its keep** — it removes a footgun (a goroutine exiting with a live binding + Go reusing goids → a later goroutine inheriting dead state, plus unbounded map growth under leaks), not just the `runtime.Stack` parse cost.
- **`CtxFn` is additive** — pure builtins untouched; `RootExecContext` keeps un-isolated/host/lowered code behaving as before.
- **`isDynamic` as `atomic.Bool`** closes the push/deref race cleanly.
- **The gold tests cover the hard spawn-boundary cases** — future conveyance, isolation from later rebinds, `bound-fn`, `set!`, protocol/multimethod dispatch under a future binding, `*ns*` conveyance.

---

**1. (P1) Dynamic bindings don't convey through eager callback-invoking builtins.**

The eval loop threads `ec` correctly through *direct* evaluation, but Go-level builtins that re-enter user code via the context-free `Fn.Invoke([]Value)` run the callback under `RootExecContext`. So a binding established in a non-root context is invisible to the callback. Repro on the PR head, all inside `@(future (binding [*scale* 10] …))`:

| expression | result |
|---|---|
| `*scale*`, `(str *scale*)`, `(let [a *scale*] a)`, `((fn [] *scale*))` | `10` ✓ |
| `(mapv (fn [x] *scale*) [0])` | `[1]` ✗ |
| `(reduce (fn [a x] (+ a (* x *scale*))) 0 [1 2 3])` | `6` (vs `60` at root) ✗ |
| `(with-out-str (print *scale*))` | `1` ✗ |

(All the same expressions at root → `10`.) This is a Clojure-parity divergence — Clojure conveys to eager callbacks because they run in the binding's thread — and the gold tests miss it (they cover the spawn boundary, not *binding-inside-a-child-context, read through a callback*). It also likely affects embedder `Run`s: if `Run` executes the program under a per-Run `ec` (§6), then `(binding [*x* v] (mapv …))` inside a `Run` loses `*x*` in the callback.

This is broader than the design note's "~17 ambient-reader builtins" framing: the conveyance surface is every builtin that re-enters user code. I count ~**28 sites** (`lang.go` 23, `async.go` 4, `pods.go` 1), plus the lowered-path HOF emissions. The async callback sites — `reduce`/`map`/`split`/`pub` — are the same root cause. `mult`/`into` don't take user fns, so their concern is spawn-scope/cancellation, not callback bindings.

The behavior split seems straightforward:

- **Eager** callback sites (`reduce`, `mapv`, `filterv`, `run!`, `every?`, `sort-by`, `swap!`, transducer application, …) — the callback runs synchronously, so thread `ec`:

  ```go
  // before — callback runs under RootExecContext:
  acc, err = mfn.Invoke([]vm.Value{acc, x})
  // after — eager HOF becomes a CtxFn and threads the caller's ec (your own prn/alts! pattern):
  reduce := vm.NewCtxNativeFn("reduce", func(ec *vm.ExecContext, vs []vm.Value) (vm.Value, error) {
      ...
      acc, err = ec.Invoke(mfn, []vm.Value{acc, x})
      ...
  })
  ```

- **Lazy** callback sites (`map`, `filter`, …) — the callback runs at realization. `(binding [*scale* 10] (map f coll))` realized *after* the extent returns `[1 2 3]`, which matches Clojure's lazy-seq-across-dynamic-binding caveat. I'd leave lazy as-is; force-fixing it would make let-go stricter than Clojure.

A goroutine-ambient "current ec" would paper over these sites, but it would re-add the kind of ambient lookup this PR deletes. Explicit threading seems like the right tradeoff; it means the callback surface has to be included in the migration. I’d address this or scope it explicitly before merge, because today the conveyance contract is "works at root and through direct evaluation, but is lost in eager HOFs in any child context."

**2. Deref hot-path takes a lock where `main` is lock-free.**

`ec.deref` → `bindingStack.current(v)` takes `b.mu` for any var whose `isDynamic` is set (`exec_context.go:133`, `binding_stack.go:45`), and `isDynamic` is sticky — set on first `push`, never cleared. #206 deliberately made `Var.Deref` lock-free (atomic `curr`/`root`) because it's "by far the hottest var operation." So any var that's *ever* been dynamically bound pays a mutex on every subsequent deref, for the life of the process — and `*out*` is that kind of var (`with-out-str`, an embedder's `WithStdout`) deref'd in tight loops. Rough local numbers: root ~2.8ns, bound ~25ns, bound-parallel ~107ns. **But** the existing `var_deref` bench looks unreliable (`_ = v.Deref()` can be dead-code-eliminated; the base produced impossible sub-ns numbers), so it needs a package-level sink before those numbers should drive a merge decision. If the bound-var cost holds up, an atomic top-of-stack cache per var (mirroring #206's `curr`), or clearing `isDynamic` when the stack empties, would keep bound-var deref lock-free.

**Minor.** The design note's §8 top risk — fast-path opcodes that bypass `NativeFn.Invoke` — I poked at `OP_LOAD_VAR`/`OP_SET_VAR`/invoke in `vm.go` and they're using `f.ec`; no concrete miss found, so just "worth a second pass."

---

Finding 1 is the one I’d address or explicitly scope before merge. I can send a focused follow-up PR for eager-HOF conveyance, with lazy HOFs left at Clojure parity and the repro above promoted to a gold test. Wanted to get your thoughts before dropping a patch into #210 or a follow-up branch.

This is also a clean base for the `WithStdout` PR (#207): once `Run` builds its own `ExecContext` and pushes `*out*`/`*err*` there (§6), the concurrency caveat in that PR flips from "serialize Runs" to isolated. I'll rebase onto this rather than land it as-is. Thanks for doing this properly instead of bolting goroutine-local onto the global stack.

nnunley:
Finding 1 confirmed and fixed in 060b7e0, following the eager/lazy split you proposed — so no follow-up PR needed for conveyance; if you want to take the bound-deref lock optimization as a follow-up instead, it's open.

**Eager-HOF conveyance.** New `ExecContext.Bind` pins a callback to its creation context for callbacks that escape into context-free machinery (`Atom.Swap` retry loops, `sort.SliceStable` comparators); it's the identity at root, so root callers pay nothing. Converted: `reduce`, `some`, `apply*`, `update`, `swap!`, `swap-vals!`, `vswap!`, `sort`, `mapv`, `complement`, `alter-meta!`, `alter-var-root`, `macroexpand`, `pmapv`, and `with-out-str*` — which had a second bug: it pushed the `*out*` binding on the var root rather than the caller's context. The async loop sites (`reduce`/`pub`/`split`/`map`) now run under `ec.Child()`, same conveyance as `future`. Two of your repro rows had a different root cause: core.lg implemented `mapv`/`filterv` as `(vec (map …))` — lazy thunks losing the context even when realized inside the extent — they're now eager via `reduce`, like Clojure's. Your repro is promoted to `test/gold/dynvar_callbacks.cljc` (mapv, filterv, reduce, run!, every?, sort-by, swap!, into+xform, with-out-str inside a child-context binding), byte-identical under clojure 1.12 and bb.

Scoped out, intentionally: lazy realization after the extent (Clojure's own caveat), pod async callbacks (no defined conveyance contract), and comparators stored in sorted collections (invoked at later op sites, not in this extent).

**Deref lock.** Bench sink added; honest numbers on an M3: root 3.9ns, bound 13.2ns, bound-parallel 142.7ns — so the cost is real and your top-of-stack-cache / clear-`isDynamic`-on-empty suggestion is worth doing, but as a follow-up with numbers driving it rather than in this PR.

**CI.** The live differential now runs in CI: the workflow provisions the Clojure CLI on demand via makeplus (`make -f <(curl -sL clojure.cc/cmd.mk) clj`) before the test steps.

On #207: agreed, rebasing onto this once merged is the right order — `Run` building its own `ExecContext` and pushing `*out*`/`*err*` there flips the concurrency story to isolated.

Verification: full suite + `-race` on root/vm/rt green, `make check-generated` lockstep OK, `-tags gogen_ir` build clean.

mparrett (approved):
Pulled `060b7e0` into a worktree and verified the eager/lazy split against the original repros.

Inside `@(future (binding [*scale* 10] …))`, the eager-HOF callbacks now convey the bound value: `mapv` → `[10 20 30]`, `reduce` → `60`, `with-out-str` → `10`, all matching their root-context results. The boundary matrix is uniform — direct read, `let` body, direct call, and HOF callback all see `10`. The lazy-seq-realized-after-extent case still returns the unbound value (`[1 2 3]`), so the fix is scoped to the eager path and doesn't reach past Clojure's own caveat. The `with-out-str*` var-root push was a second bug — that one wasn't in our repro set.

One caveat on my end: this box has neither `clojure` nor `bb`, so `TestGoldDynvarCallbacksMatchesClojure` ran against the pinned golden only, not the live differential. The CI provisioning you added closes that gap.

LGTM. I'll take the bound-deref lock optimization as a follow-up, numbers-driven against the sink you added.

mparrett:
Follow-up on my earlier caveat: I've since put `clojure` and `bb` on my box, so the live differential runs. `TestGoldDynvarCallbacksMatchesClojure` passes against Clojure 1.12.5 (no longer the pinned-golden fallback), and running all three gold scripts under babashka 1.12.218 by hand gives byte-identical output too.

mparrett:
Re-checked at `ffe1268` (my approval above was on `060b7e0`, before the upstream merge and the gold-test restructure). Gating the live Clojure differential behind `LETGO_GOLD_REDERIVE` and pinning the `.out` goldens is a better shape than what I endorsed — it takes the slow-JVM-shutdown flake out of every run while keeping the differential for CI. Verified both modes locally: default is clojure-free and fast, `LETGO_GOLD_REDERIVE=1` re-derives byte-identical to the committed goldens, and the merge left conveyance intact (the future-binding repros still match). Approval stands.

mparrett:
@nooga — #210 (ExecContext threading) is approved and `mergeable`. Two things are based on it: the dynamic-var deref work in #211 (nnunley's leaning C, validation's posted) and the IO threading in #207. Both rebase onto whatever #210 merges as, so they're parked until it lands.

No rush if you want another pass first — otherwise it's ready to merge whenever you are.

nooga (approved):

nooga:
Thanks @mparrett and @nnunley. I just finished reviewing this one. I'm comfortable with this as is. Approved #207 and commented in #211 - I agree, C is the most sensible in my opinion.

PR #218, opened 2026-06-13
Publish Homebrew formula to shared tap

nooga:
## Summary
- publish the let-go Homebrew formula to nooga/homebrew-tap
- merge the generated formula PR in the shared tap repo
- update install docs to use nooga/tap

## Checks
- ruby YAML parse for .goreleaser.yml and release workflow
- git diff --check

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,7 +30,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GORELEASER_PAT }}
           GORELEASER_PAT: ${{ secrets.GORELEASER_PAT }}
 
-      - name: Merge Homebrew cask PR
+      - name: Merge Homebrew formula PR
         env:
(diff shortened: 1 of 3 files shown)


mparrett:
Drive-by from the sidelines, this looks good to me. The cask→formula move is the right call for a CLI release binary, and dropping the quarantine `xattr` hook falls out naturally since formulae don't get the cask quarantine treatment. The `unix`/`plan9` build ids line up with the new `archives` and `brews` references, and the README shorthand (`brew install nooga/tap/let-go`) auto-taps `nooga/homebrew-tap`.

The auto-merge step and the goreleaser `brews.repository` push both target `nooga/homebrew-tap`, a separate repo, so `GORELEASER_PAT` needs write plus `--admin` merge rights there, not just on `let-go`. That won't cause pain until you cut a release tag — worth confirming the token scope first.

nnunley (approved):

PR #219, opened 2026-06-13
Makefile fixes

ingydotnet:
Add missing `$(GO)` dep to Makefile targets using go.

This makes sure everyone is using the same go including CI. 
Also lets people do everything here without a go installed.

Also remove PHONY targets likely added by claude.


nnunley (approved):
Awesome! Thanks Ingy!

PR #220, opened 2026-06-13
perf(vm): lock-free dynamic-var deref via persistent frame-list binding stack

### LINKED ISSUE #211: Follow-up to #210: lock-free dynamic-var deref — a design fork (A: copy-on-write vs B: active-binding counter)
Picking up the deref-lock optimization you flagged as a follow-up in #210 ("your top-of-stack-cache / clear-`isDynamic`-on-empty suggestion is worth doing, but as a follow-up with numbers driving it"). I built one version, measured it, and found the decision forks two ways, and which way is right depends on let-go's real read/write ratio for dynamic vars, which you'll have a better feel for than I do.

## Cost

Since `isDynamic` is the declaration flag, it stays set for the var's whole life: `SetDynamic()` is called for every `^:dynamic` var at compile time (compiler.go:1700), and any push sets it too. So `ExecContext.deref` consults the per-context binding stack on *every* read of a dynamic var, taking `bindingStack.mu` even when the var has no active binding in this context. For `*out*`/`*ns*`/etc. that's a lock on a hot path.

That also rules out a literal "clear-`isDynamic`-on-empty": clearing it would break the declaration semantic (a `^:dynamic` var with no current binding must still report dynamic and must still be checked, because it can be bound at any time). So the lever is the lookup, not the flag.

## Option A — copy-on-write binding map

Built and measured this. Hold the binding map behind an `atomic.Pointer`, immutable once published. Reads load it lock-free; writers (push/pop/setCurrent/installSnapshot) serialize on the retained mutex, copy the map, and atomic-swap. Per-context isolation is unchanged. Full suite + `-race` green.

Reads (benchstat, n=8, local M-series; treat the deltas as the signal, absolutes will differ on your box):

```
                                  before        after      delta
VarDerefPreviouslyBound          18.2 ns       6.7 ns      -63%   (declared dynamic, no active binding — the common *out* read)
VarDerefPreviouslyBoundParallel  98.2 ns      11.7 ns      -88%
VarDerefBound                    22.8 ns      14.5 ns      -36%   (read inside an active binding)
VarDerefBoundParallel             118 ns      17.6 ns      -85%   (the contention you raised)
VarDerefRoot / RootParallel / DistinctParallel      unchanged (already lock-free)
geomean                                                    -57%
```

The cost lands on the write side: every binding establishment now allocates fresh maps:

```
BindingPushPop    84 ns -> 364 ns   (+335%)    16 B -> 704 B   (+43x)    1 -> 7 allocs
```

So A fixes every read path, including the parallel-bound contention you called out, but makes `(binding [...] ...)` establishment more expensive. Whether that's a good trade depends on how binding-heavy real workloads are.

## Option B — per-context active-binding counter

Described, not yet built. Keep `isDynamic` as the declaration flag, but add an atomic count of active bindings to each `bindingStack`, and gate the lock: `if v.isDynamic.Load() && ec.bindings.count > 0`. A context with no active bindings reads every dynamic var lock-free; the write path is unchanged but for one atomic add. This is closer to the "clear-on-empty" framing: it returns a context to the fast path when its stack drains.

The trade is narrower: B fixes the common unbound read (`VarDerefPreviouslyBound`, roughly to root speed) at ~zero write cost, but leaves reads *inside* an active binding (including `VarDerefBoundParallel`) on the lock, since the counter is non-zero there. I can build and measure it if the trade looks right to you.

## The decision point

Which tradeoff fits let-go? A buys lock-free reads everywhere (and kills the bound-parallel contention) at a real write-path regression; B is cheaper and lower-risk but only covers the unbound read. My lean is B as the conservative default: don't regress the write path without evidence the reads it buys are hot. But you raised the bound-parallel contention specifically, which only A addresses, so I didn't want to pick for you.

A is on `perf/bound-deref-lock` on my fork if you want to check it out and run the benches. Happy to build B for a side-by-side, or to take this whichever direction you prefer once #210 lands (this stacks on it).

mparrett:
Implements option C from #211, which @nnunley and @nooga settled on. Built on #210 (ExecContext threading), now merged. Closes #211.

## What changes

Dynamic-var binding state moves from a per-`ExecContext` mutex-guarded map to a persistent immutable singly-linked list of `(var, val)` frames behind an atomic head pointer.

- **Read**: atomically load the head, walk to the first matching var. No lock.
- **Push**: prepend one frame — one allocation, O(1).
- **Pop / set!**: rebuild only the frames above the target; none when the target is on top, which is the LIFO common case.
- **Child context**: shares the parent's chain by pointer, so spawning a future or go-block forks bindings without copying.

The diff is `pkg/vm/binding_stack.go` plus benchmarks in `pkg/vm/var_deref_bench_test.go`.

## Correctness

Immutable frames plus atomic head publication mean a reader observes a complete old-or-new chain, never a torn one; `set!` semantics are preserved because a set rebuilds the chain above the target and republishes the head atomically. Verification:

- `go test ./...` green (13 packages); `pkg/vm` binding/deref/`ExecContext` tests green under `-race`.
- The `.lg` suite (`go test ./test`) green; dynamic-binding conveyance repros intact.
- An adversarial second review found no correctness refutation.

## Performance

Nested bindings at depth 8 / 16 / 32, three read:write ratios. ns/op, M2, benchstat n=8, vs current main (the mutex-map baseline):

```
                  baseline      C (frame-list)
d8  read-only      1046 ns      308  (-71%)
d8  balanced        129 ns       61  (-53%)
d8  write-heavy      83 ns       40  (-52%)
d16 read-only      1047 ns      383  (-63%)
d16 balanced        128 ns       71  (-44%)
d16 write-heavy      80 ns       45  (-44%)
d32 read-only      1063 ns      637  (-40%)
d32 balanced        127 ns      108  (-15%)
d32 write-heavy      81 ns       65   (~, within noise)
```

C wins every cell from -15% to -71%, except d32 write-heavy, which is within noise (p=0.105). Push in isolation is -34% (`BenchmarkBindingPushPop`), at the cost of one frame allocation per push: still one alloc/op, but 32 B/op vs the baseline's 16. So the write path trades a larger allocation for lock-free reads, and still comes out ahead on wall time across the workloads.

### Caveat: O(depth) read walk

A read of a var buried under many unrelated active bindings walks the chain, so it is O(depth). Head hits stay ~2 ns at any depth; tail and miss reads grow with depth:

```
              baseline    C
d1  head        16.0     2.1
d16 head        15.1     2.1
d16 tail        15.0     8.2   (-45%)
d16 miss        15.0     8.4   (-44%)
d64 head        15.0     2.3
d64 tail        14.9    49.0  (+229%)
d64 miss        15.0    49.8  (+233%)
```

C wins tail/miss through depth 16 (-44% to -45%); the crossover to worse-than-baseline falls between depth 16 and 64, where deep tail/miss reads cost ~49 ns. That isn't a pattern expected from `*out*`/`*ns*`-style usage. If pathologically deep binding stacks turn out to matter, a hybrid is possible; the workload numbers say it isn't worth the complexity now.

## Concurrency validation

`test/dynvar_concurrency_test.lg` is a multi-tenant request pipeline (thread-per-request fan-out, per-request bound context, tenant-routed data access) that exercises dynamic-var deref under concurrency. It runs under `go test ./test` and asserts binding isolation, so a bleed fails the suite. Two findings from it:

- Binding isolation holds across futures: concurrent results equal a serial golden run elementwise, no cross-request bleed.
- Because #210 gives each goroutine its own `ExecContext`, the deref lock is uncontended in idiomatic concurrent code, so the win here is the uncontended single-thread read cost rather than contention relief. The parallel microbenchmarks measure a shared-context case the per-goroutine design makes rare.


nooga:
I applied the benchmark fix locally and reran the affected benchmarks.

The issue was that `BenchmarkVarDerefDepth` called `bindDepth(b, depth)` on the parent benchmark `b`, so the cleanup did not run until the whole top-level benchmark finished. That meant later `miss` cases measured `current depth + previous depths`, not the labeled depth.

The local fix is to scope setup/cleanup to a per-depth sub-benchmark while preserving the same names:

```go
for _, depth := range []int{1, 4, 16, 64} {
	depth := depth
	b.Run(fmt.Sprintf("d%d", depth), func(b *testing.B) {
		vars, miss := bindDepth(b, depth)
		// ...
		for _, pos := range []string{"head", "tail", "miss"} {
			t := targets[pos]
			b.Run(pos, func(b *testing.B) {
				for i := 0; i < b.N; i++ {
					derefSink = t.Deref()
				}
			})
		}
	})
}
```

Command:

```sh
go test ./pkg/vm -run '^$' -bench 'BenchmarkVarDerefDepth|BenchmarkNsWorkload' -benchmem -count=5
```

Representative medians on my M1 Pro:

| case | before | after |
| --- | ---: | ---: |
| `BenchmarkVarDerefDepth/d4/miss` | 4.48 ns/op | 4.14 ns/op |
| `BenchmarkVarDerefDepth/d16/miss` | 13.74 ns/op | 10.07 ns/op |
| `BenchmarkVarDerefDepth/d64/miss` | 75.72 ns/op | 53.26 ns/op |
| `BenchmarkVarDerefDepth/d64/tail` | 52.52 ns/op | 52.66 ns/op |

So this does not change runtime performance; it fixes the benchmark so the miss cases report the intended active depth. Tail/head results are essentially unchanged, which is what I would expect because the extra leaked frames sat below the target frames.

Also ran:

```sh
go test ./pkg/vm -count=1
```

That passed.

nooga (approved):

mparrett:
Thanks for pushing the fix (`980eac8`) — confirmed the diagnosis: `bindDepth`'s `b.Cleanup` on the parent `b` left earlier depths' frames on the stack, so the later miss cases walked the accumulated depth.

Reran on M2 (n=8); the corrected numbers track yours. The bug had inflated C's miss column, so the caveat is milder than the body had it:

| case | was (buggy bench) | now |
| --- | ---: | ---: |
| d16 miss | 10.9 ns (-29%) | 8.4 ns (-44%) |
| d64 miss | 68.7 ns (+354%) | 49.8 ns (+233%) |

tail/head are unchanged, as you found — the leaked frames sat below the target frames. Updated the PR description's caveat table to match.

examples/dynvar-concurrency-validation.lg
```diff
@@ -0,0 +1,117 @@
...
+      golden (mapv handle reqs)
+      w1     (mapv (fn [_] (run-once reqs golden)) (range reps))
+      ;; --- W2: deref-dominated concurrent reads (allocation-free) ---
+      hotk   2000
+      hreqs  (mapv (fn [i] (assoc i :shard (mod (count (:id i)) 7))) reqs)
+      hgold  (mapv (fn [r] (handle-hot r hotk)) hreqs)
+      w2     (mapv (fn [_] (run-hot-once hreqs hotk hgold)) (range reps))]
+  (println "=== W1: realistic mixed handler ===")
+  (println "  " n-req "concurrent reqs, work=" work "; ~" (+ (* work 8) 6) "derefs/req")
+  (println "   binding isolation:" (if (every? second w1) "OK" "FAIL — bleed!"))
```
nnunley: These FAIL - bleed conditions should probably cause the script to fail with a non-0 exit code so it can be used as a sentinel/ci/cd case.

examples/dynvar-concurrency-validation.lg
nnunley, line 1: This file probably should be moved to the tests directory, or something like a benchmarks directory. The measurements are good, but we lose the ability to detect drift/breakage with it being pinned in the examples directory.

pkg/vm/binding_stack.go
```diff
@@ -6,89 +6,161 @@
...
+func replaceTopBinding(head *bindingFrame, v *Var, val Value) (*bindingFrame, bool) {
+	var above []*bindingFrame
+	for f := head; f != nil; f = f.next {
+		if f.v == v {
+			return rebuildAbove(above, &bindingFrame{v: v, val: val, next: f.next}), true
+		}
+		above = append(above, f)
+	}
+	return head, false // not bound
+}
```
nnunley, line 139: These are near duplicates. It might be good to refacto the first to be in terms of the  second.

nnunley (changes requested):

mparrett:
Thanks @nnunley — addressed the first two; the third I pushed back on, with numbers.

**Exit non-zero on bleed + move out of `examples/`** (`561d241`): folded both into one move — the validation is now `test/dynvar_concurrency_test.lg`, a deftest. It runs under `go test ./test`, and a bleed (concurrent result != serial golden) fails the suite with a non-zero exit, so it's a CI sentinel now. Perf stays covered by the `pkg/vm` benchmarks.

**The two near-duplicates** (`removeTopBinding`/`replaceTopBinding`): I tried the shared higher-order helper (`rebuildTopBinding(head, v, newTail func)`) and backed it out. `removeTopBinding` is the hot pop path (every binding exit), and the helper can't inline (its `append` walk blocks it), so the `newTail` func arg doesn't devirtualize and the dedup adds a non-inlined plus indirect call per pop (`-gcflags=-m` confirms; allocs/op stays 1, so it's call overhead, not GC). On a ~32 ns op that's a few percent on the path the PR exists to make fast. I left them separate with a comment explaining why (`561d241`), so the next reader doesn't re-flag it. If you'd still rather take the dedup and eat the overhead, happy to — your call on the tradeoff.

nooga:
My 2 cents on the dedup: keep the fast path. It's Go, DRY never really applied ;)

pkg/vm/binding_stack.go
mparrett: Thanks for calling this out... these are near-identical and it looks temping to factor them together. left them separate because removeTopBinding is on the hot path this PR is trying to speed up, and folding them adds an indirect call right there. sounds like that's the idiomatic call anyway. Apparently there's a Go proverb for it, "a little copying is better than a little dependency." 🤷🏻

mparrett:
> My 2 cents on the dedup: keep the fast path. It's Go, DRY never really applied ;)

Thanks for adding your perspective. Haven't worked in golang very much, and especially not recently. Apparently there's a [go proverb ](https://go-proverbs.github.io/) about this, "A little copying is better than a little dependency." TIL.

nnunley (approved):
LGTM

PR #222, opened 2026-06-14
feat(ir): lower try/catch/finally to native Go (gogen_ir self-AOT)

nnunley:
> **📚 Stacked PR (rebased 2026-06-15).** This now descends from the collection-protocol fix rather than sitting alone on `main`:
> 1. #227 — lgbgen CPU profiling
> 2. #228 — Indexed collection protocol (nth/chunked/conj correctness)
> 3. **this PR** — lower try/catch/finally to native Go
>
> Why the restack: this PR's gogen self-AOT lowering regenerated an **empty** tree until #228 fixed `nth` dispatch on transient/chunk types. With #228 beneath it, `make check-generated` passes (bundle in lockstep, `core_go_lowered/` compiles + dispatches natively under `-tags gogen_ir`). Diff includes #227+#228 until they merge; review/merge bottom-up.

---

Completes Task 1 of the **complete-gogen-self-aot-lowering** change: `try`/`catch`/`finally` now lower to native Go under `-tags gogen_ir` instead of falling back to the bytecode VM.

## What changed

A new `:try` IR op and its gogen_ir lowering. Body, handler, and finally are built as inner `fn*` closures and invoked via `rt.InvokeValueEC`; a caught Go error is converted to the handler-bound value via `vm.ErrorToValue`, mirroring the bytecode VM's handler path. The emitted statements are wrapped in a bare block so the `:=` temps stay block-scoped — a `goto` in the surrounding emission must not jump over their declarations.

| File | Change |
|---|---|
| `examples/go-gen/ir_ops.lg` | append `Try` op (iota-safe, last) → `op_generated.go` (+2) |
| `pkg/rt/core/ir/build.lg` | `parse-try` + `build-try`; `try` cases in `free-vars`/`captures-of` |
| `pkg/rt/core/ir/lower_go.lg` | `try-assign-stmts` (catch+finally / catch / finally / bare); `:try` wired into `local-carrying-op?`, `lower-inst-stmts`, `function-needs-error?`, `function-needs-rt?` |
| `pkg/rt/gogen.go` | `block-stmt` builder (`cBlockStmt`) for goto-safe temp scoping |
| `pkg/vm/errors.go` | export `ErrorToValue` for lowered handler value conversion |
| `pkg/rt/core_compiled.lgb` | regenerated (in lockstep with sources) |

## Verification

- `go build -tags gogen_ir ./...` — clean
- `go test ./pkg/ir/` (interpreted) — pass
- **Parity-identical across bytecode and gogen_ir** (`make parity-check`):
  - jank Clojure compat `./test`: **5631 / 0** on both engines
  - ir-stress lower-go: identical pass/fail buckets on both engines
- `make check-generated` — `core_compiled.lgb` in lockstep; `core_go_lowered/` compiles + dispatches natively

try/catch defns now lower to native Go instead of bytecode fallback, eliminating the try-related lower-go failure buckets.

## Second commit — parity gate hardening

`test(parity)`: `scripts/gogen-parity.sh` could report `PARITY` (exit 0) when a suite never ran — `time_run` swallowed the inner exit code, and an empty summary compared `empty == empty`. Now `time_run` records the real inner exit, and a new `require_ran` guard asserts each suite emitted its `TOTALS`/`Passed` line (>0 tests ran) before any parity comparison, exiting 2 on a run/setup failure.


nooga:
One correction/clarification on the `try/finally` discrepancy I found while reviewing:

```clojure
(try
  (throw :x)
  (finally :f))
```

Current let-go bytecode evaluates this to `:x` because the no-catch `try/finally` path converts the thrown value into the expression result and then discards the `finally` value.

This PR's lowered Go path instead propagates the body error after running `finally`, which is what Clojure does: `finally` runs, its value is ignored, and the exception continues unless a `catch` handles it.

So the discrepancy is real, but I think the bytecode implementation is most likely the bug, not this PR's lowered Go behavior. The remaining concern is parity: after this lands, bytecode and `gogen_ir` disagree for no-catch `try/finally` until the bytecode compiler/VM semantics are fixed, and core comments such as the one in `with-redefs-fn` that mention no-catch `finally` swallowing throws should be updated as part of that fix.

nooga (approved):

PR #223, opened 2026-06-14
feat(rt): route term/* ANSI output through *out*

mparrett:
## Summary

#206 made `print`/`pr`/`prn`/`println` consult the current dynamic binding of `*out*`. The `term/*` ANSI ops never got the same treatment: `move-cursor`, `clear`, `set-fg`, `write`, and the rest wrote their escape sequences straight to `os.Stdout` via `fmt.Print`. So `(binding [*out* x] (term/clear))` silently ignored `x`, the same compat bug #206 fixed for the print fns, still live for the terminal-control ops.

This routes all 15 ANSI output ops through the existing `WriteToOut` helper, so they honor `*out*` the way `println` does. This is the "Phase 1.5" the I/O proposal called out.

## Context: where this fits

This is a small, separable follow-up in the #79 browser-interop line, after the two merged PRs:

- **#206** (PR1) — print fns honor `*out*`, error sites honor `*err*`.
- **#207** (PR2) — `pkg/api.WithStdout`/`WithStderr` embedder options.
- **this PR** — the `term/*` ANSI ops, the remaining set that still bypassed `*out*`.
- **PR3** (next) — the WASM host writer that lets a browser shell receive output directly, retiring `fs.writeSync`.

It depends only on what's already merged. It does not touch `pkg/vm`, so it's independent of the in-flight deref work (#220).

## Why this matters

`(binding [*out* h] ...)` is how you redirect output in let-go, and it already works for the print fns and `write!`. For `term/*` it was a no-op: the escape sequence went to `os.Stdout` regardless of the binding. That blocks the obvious uses: capturing a TUI frame to a string, rendering into a buffer, or sending control and text to the same redirected sink a browser shell provides.

It also means ANSI and text now ride the same stream. `(with-out-str (term/move-cursor 1 1) (term/write "X") (term/reset-style))` captures `"\033[1;1HX\033[0m"` as one in-band string: control and text interleaved, exactly as on a real TTY, instead of the control half leaking past the capture.

## What changed

All 15 ANSI output ops across the three build targets — `term.go` (native), `term_wasm.go`, `term_plan9.go` — become context-aware builtins. Each is now a `vm.NewCtxNativeFn` that receives the `ExecContext` and threads it into `WriteToOut(ec, …)`, which resolves `*out*`'s current binding through the explicit context #210 introduced. This is the same shape `println`/`print`/`pr` use post-#210.

The ops covered: `move-cursor`, `clear`, `clear-line`, `hide-cursor`, `show-cursor`, `set-fg`, `set-bg`, `reset-style`, `bold`, `underline`, `inverse`, `write`, `write-at`, `alternate-screen`, `main-screen`.

The PTY/termios ops — `raw-mode!`, `restore-mode!`, `read-key`, `key-pending?`, `size`, `set-size`, `tty?`, `open-pty` — stay plain native fns on real file descriptors. They're fd/ioctl operations, not byte output, and have no business on `*out*`. `flush` and `char-width` likewise stay as-is (`flush` syncs the real fd; `char-width` computes, emits nothing).

## Transparency / compatibility

Byte-identical default output on every target. `*out*`'s root binding is the IOHandle wrapping `os.Stdout`, so a bare `(term/clear)` writes the same bytes to the same place it always did: native terminal, WASM via the existing `fs.writeSync` path until the host writer lands, plan9. The only new behavior is that `term/*` now honors a `*out*` rebind.

## Tests

`test/term_out_binding_test.lg` (7 deftests) pins the fixed behavior by capturing escapes through `with-out-str`: each op routes through the current `*out*` binding; control and text capture as one stream; and `(binding [*out* *err*] (term/clear))` hits `*err*`, not the outer `*out*`.

## Pre-merge checks

- `go test ./test` (full `.lg` suite) and `go test ./pkg/...`, both green.
- Cross-target builds: `GOOS=linux`, `GOOS=js/wasm` (`./pkg/rt`), `GOOS=plan9` (`./pkg/rt`), all build.
- `gofmt` clean; `go vet ./pkg/rt` clean.
- Browser smoke: a WASM bundle built with this branch boots headless (`crossOriginIsolated`, no console errors or exceptions) and renders its `term/*`-drawn title screen, confirming the wasm `term/*` output still reaches xterm.js through `*out*` (the path `go test` doesn't cover).

### DIFF
--- a/pkg/rt/term.go
+++ b/pkg/rt/term.go
@@ -536,10 +536,16 @@ func installTermNS() {
 	})
 	ns.Def("main-screen", mainScreen)
 
-	// flush — flush stdout
-	flushFn, _ := vm.NativeFnType.Wrap(func(vs []vm.Value) (vm.Value, error) {
-		os.Stdout.Sync()
-		return vm.NIL, nil
+	// flush — sync the active *out* binding so flush hits the same sink the
+	// term/* bytes did. File-backed handles fsync; buffered embedder writers
+	// flush; otherwise no-op. Falls back to os.Stdout only at early boot
+	// (before *out* is installed). Pre-#223 this synced os.Stdout directly,
+	// which diverged once the term/* ops started honoring *out*.
+	flushFn := vm.NewCtxNativeFn("flush", func(ec *vm.ExecContext, vs []vm.Value) (vm.Value, error) {
+		if h := resolveIOHandleVar(ec, "*out*"); h != nil {
+			return vm.NIL, h.Sync()
+		}
+		return vm.NIL, os.Stdout.Sync()
 	})
 	ns.Def("flush", flushFn)
 
--- a/pkg/rt/term.go
+++ b/pkg/rt/term.go
@@ -361,7 +361,7 @@ func installTermNS() {
 	ns.Def("open-pty", openPty)
 
 	// move-cursor — (move-cursor col row) — 1-based ANSI positioning
-	moveCursor, _ := vm.NativeFnType.Wrap(func(vs []vm.Value) (vm.Value, error) {
+	moveCursor := vm.NewCtxNativeFn("move-cursor", func(ec *vm.ExecContext, vs []vm.Value) (vm.Value, error) {
 		if len(vs) != 2 {
 			return vm.NIL, fmt.Errorf("move-cursor expects 2 args (col row)")
 		}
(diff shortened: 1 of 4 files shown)


nooga:
I would treat the `term/flush` behavior as a blocker before merging.

This PR makes the ANSI/text-producing `term/*` ops respect the current `*out*` binding, but `term/flush` still flushes the old backend (`os.Stdout` on native/Plan 9, `_lgFlush` on WASM). That means a redirected terminal render can write bytes to the rebound handle while flushing somewhere else:

```clojure
(binding [*out* h]
  (term/write "...")
  (term/flush))
```

If `h` is buffered or embedder-provided, the caller expects the flush to apply to the same sink that received the terminal bytes. `IOHandle.Sync()` already exists, so I think `term/flush` should become context-aware too and resolve/sync the active `*out*` handle, with the existing platform fallback behavior preserved where appropriate.

Separate non-blocking cleanup: `test/term_out_binding_test.lg` currently proves rebinding by sending `term/clear` to `*err*`, which emits a real `ESC[2J` during verbose test runs. I saw that escape sequence in the test output. It would be better to bind to a temporary file or another non-terminal writable handle so tests do not clear/pollute terminal or CI logs.

mparrett:
Agreed it's a blocker, and you're right that this PR introduced the divergence: before #223 both `term/write` and `term/flush` went to `os.Stdout`; once `write` started honoring `*out*`, `flush` kept syncing the old backend. It bites the embedder case directly too: xsofy's render loop calls `term/flush` every frame, so a buffered or embedder-provided `*out*` would never see the flush.

Fixed in `407be02`. `term/flush` is now a context-aware builtin that resolves the active `*out*` handle and calls `IOHandle.Sync()` (`File.Sync()` for file-backed handles, `Flush()` for a buffered writer, no-op otherwise), mirroring `WriteToOut`. Native and Plan 9 fall back to `os.Stdout.Sync()` only at early boot, before `*out*` is installed.

One spot I'd like your read on: wasm. There `flush` drove `_lgFlush` (the JS-side postMessage flush to xterm.js), which isn't an fd sync, and `os.Stdout.Sync()` is a no-op in wasm. So I kept `_lgFlush` and sync the active `*out*` ahead of it: a rebound buffered `*out*` is flushed by `Sync()`, and `_lgFlush` then no-ops on the empty screen buffer; the default-root `*out*` wraps `os.Stdout`, so `_lgFlush` stays the real screen flush. That's a "do both" rather than identity-checking the root handle. Correct in effect, but it does call `_lgFlush` even when output is redirected away from the screen (harmless, since nothing's queued there). If you'd rather gate `_lgFlush` to the default root, easy to switch; your call.

On the test cleanup: retargeted off `*err*` to a temp file, so no real `ESC[2J` reaches the terminal or CI log. It now asserts both halves: the outer `*out*` captured nothing, and the rebound handle received the escape. Added a `flush` test in the same shape.

One caveat on that flush test: `open :write` handles are unbuffered, so it pins that flush resolves `*out*` and targets it without error, but it can't catch a byte-level missed flush. The buffered case, where the flush is load-bearing, is exercised by the wasm render path every frame. If you'd want a unit-level guarantee, I can add a Go test against a `bufio`-backed handle; happy to.

nooga (approved):

PR #224, opened 2026-06-15
perf(vm): cut redundant value hashing on the lowering hot path

nnunley:
## Summary

Profiling self-AOT lowering (`lgbgen --target=go`) showed it was dominated by **hashing**. Three fixes on the value-hash hot path:

### 1. Cache `Boxed.Hash()`
`*Boxed` implemented no `Hash()`, so `hashValue` fell back to `computeHash` → `hashBytes(String())`, formatting the wrapped value via `fmt.Sprintf("<%s %v>")` on **every** hash. Boxed values (e.g. `SourceInfo`) embedded in IR vectors were re-stringified on every `=`, map/set key, and CSE hash. `*Boxed` is immutable; cache the hash (same value as the old fallback).

### 2. Hash strings by UTF-8 bytes
`hashUnencodedChars` decoded to runes and re-encoded to UTF-16 code units (Java/Clojure hash parity), boxing into an interface (a `convTstring` alloc) + virtual `At()`/`Len()` in the Murmur3 loop. The hash is only for internal bucketing — no normalization/case-folding — so hash the UTF-8 bytes directly in one pass. **ASCII (every keyword/identifier) is byte-identical**; non-ASCII changes but stays consistent.

### 3. Skip the hash fast-negative in `valueEquiv` for collections
`valueEquiv` (map/set key comparison) computed both operands' hashes as a "fast negative" — but for `Sequable` types `Hash()` is **O(size)**, exactly as costly as the structural `Equals` it precedes, which short-circuits on length then the first differing element. And `valueEquiv` is reached only **after** the hash trie already matched, so the operands' hashes usually agree (pure overhead). Skipping it for `Sequable` operands eliminated the dominant remaining cost — `PersistentVector`/`ArrayVector` re-hashing during map `assoc`. `Boxed` (not `Sequable`, no `Equals`) keeps the check. It's only a negative shortcut, so correctness is unaffected.

## Impact (CPU profile composition, lowering)

| | before | after |
|---|---|---|
| `hashValue` cum | 49.7% | **~1%** |
| `fmt.Sprintf` | 31% | eliminated |
| `PersistentVector.Hash` | dominant | off the profile |

## Verification

- Lowered Go output **byte-identical** (0 files differ) — no codegen change
- `go test ./pkg/...` green
- **jank clojure-test-suite: 5631 assertions pass, 0 fail** under both VM variants (bytecode + gogen_ir), parity-identical

### DIFF
--- a/pkg/vm/boxed.go
+++ b/pkg/vm/boxed.go
@@ -24,12 +24,14 @@ func (t *aBoxedType) Box(value any) (Value, error) {
 	if !reflect.TypeOf(value).ConvertibleTo(t.typ) {
 		return NIL, NewTypeError(value, "can't be boxed as", t)
 	}
-	return &Boxed{value, t}, nil
+	return &Boxed{value: value, typ: t}, nil
 }
 
 type Boxed struct {
-	value any
-	typ   *aBoxedType
+	value  any
+	typ    *aBoxedType
+	hash   uint32
+	hashed bool
 }
 
 // Type implements Value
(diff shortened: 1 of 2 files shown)


nooga (approved):

mparrett:
Measured this to confirm the win — it holds up. Same-machine A/B (`d4eb364` → `f6e4fd9`) over the vector/value/collection benches the change touches, `benchstat -count 12`:

**~36% geomean speedup on those ops, allocations unchanged** — pure compute, consistent with cutting the redundant hashing.

Biggest movers (all p<0.05, n=12):

| Benchmark | Δ time |
|---|---:|
| `LargeVectorAppend/ArrayVector` | −73% |
| `Conj/Map` | −65% |
| `VectorAssoc/ArrayVector/1000` | −64% |
| `MapAssoc/HAMT-Assoc/10` | −64% |
| `VectorSeq/PersistentVector/1000` | −56% |
| `VectorVersioning/PersistentVector` | −53% |

Two small-case spots came back as significant regressions, worth a glance:
- `VectorConj/PersistentVector/1000` +63% (p=0.000)
- `VectorCreation/ArrayVector/10` +62% (p=0.005)

Could be real small-N regressions or artifacts of the wide parent-side variance — flagging rather than hiding.

**Caveat for anyone re-checking via the perf timeline:** don't diff two snapshots pairwise for this. I tried, and the two `ubuntu-latest` dispatches landed on different CPUs (EPYC 9V74 vs 7763, anchor 1.41 vs 1.25 ns) — the anchor doesn't bridge memory bandwidth across different silicon, so the vector benches show bogus +50–96% "regressions." `var_deref_bench_test.go` was also rewritten here, so those bench names aren't comparable across the two commits. The timeline is for trends over many commits; for a head-to-head, controlled local hardware + benchstat is the one to trust.

PR #232, opened 2026-06-15
feat(ir): native Go lowering for deftype and defprotocol

nnunley:
## What

Native Go lowering for `deftype`/`defprotocol`, now carried through to an
**idiomatic, `ec`/`error`-minimized Go interop surface** — the original intent
of this PR.

Two layers, additive:

1. **Internal dynamic-dispatch seam** (unchanged): every protocol method is
   lowerable to the uniform IFn ABI `func (recv *T) M(ec *vm.ExecContext, …)
   (vm.Value, error)`. This remains the devirtualized fast path and the shape
   the generated interface + IFn dispatch agree on.
2. **ec-free typed surface** (new): for each protocol method we compute the
   *tightest unified signature* the method's impls support and emit that, so
   **total methods drop the spurious `error`** and read as ordinary Go:
   `func (recv *T) M(ec *vm.ExecContext) vm.Value`.

## How

- `lower-ns-to-go` (pipeline.lg) computes a per-method **unified signature**
  by building each impl's IR fn and joining across impls:
  `needs-error?` = OR over impls (MVP keeps params/results at `vm.Value`, since
  deftype fields are stored as `vm.Value`). It binds this as the dynamic
  `*protocol-method-sigs*`.
- The four emission sites read it: `lower-defprotocol` (interface),
  `lower-method` (concrete receiver), `devirtual-call-stmts` (call site), and
  `call-error-free?`. A method gets the clean signature only when all impls are
  total; otherwise it falls back to the uniform `(ec)(vm.Value, error)`.
- **Additive by construction:** when `*protocol-method-sigs*` is unbound
  (per-fn lowering / pass 1), every site reverts to the prior forced-error
  behavior — output is byte-identical to the uniform-ABI baseline.

## Example (generated Go)

```go
type Measurable interface {
    Measure(ec *vm.ExecContext) vm.Value            // total → error-free
}
func (arg0 *Box) Measure(ec *vm.ExecContext) vm.Value { … }                        // total
func (arg0 *Box) Describe(ec *vm.ExecContext, arg1 vm.Value) (vm.Value, error) { … } // effectful → keeps error

v := b.Measure(ec)                     // external caller: single value, no error to unpack
_, err := b.Describe(ec, vm.Int(0))    // effectful keeps the 2-value form
```

## Verification

- `make check-generated` green (compiles the whole `-tags gogen_ir` corpus).
- `go test ./...` green.
- Trampoline parity (bytecode == native) on all 5 `test/gogen/deftype_*`
  fixtures, including the new `deftype_total_method.lg` which exercises **both**
  a total method (error-free) and an effectful method (keeps error).

## Scope / non-goals (tracked separately)

- C-ABI / `c-shared` codegen (`//export`, value↔C/JSON marshaling).
- External Go types *implementing* let-go protocols (impl registration).
- Typed `deftype` fields (to unlock typed method results).
- Dropping the `core_compiled.lgb` bootstrap.

Spec: `openspec/changes/add-cabi-interop-layer/`. The db-efpw prerequisite
(c-shared loader eager-loads core) lives on `spike/libyamlstar-cshared`.


mparrett:
Codex findings (5.5-high) on [nooga/let-go#232](https://github.com/nooga/let-go/pull/232):

1. **Receiver methods with calls generate invalid Go**  
   [lower_go.lg](/private/tmp/let-go-pr232/pkg/rt/core/ir/lower_go.lg:2330) emits receiver methods without an `ec` parameter, but normal call/load-var lowering inside the method still emits `rt.InvokeValueEC(ec, ...)`. Also [devirtual-call-stmts](/private/tmp/let-go-pr232/pkg/rt/core/ir/lower_go.lg:1330) always assigns a single return from `recv.Method(...)`, while `lower-method` may emit `(vm.Value, error)`.  
   Repro fixture: method body `(plus1 n)` generated `undefined: ec` and `assignment mismatch: 1 variable but arg__3.Calc returns 2 values`.

2. **Bare field rewrite miscompiles shadowed locals/params**  
   [dt-rewrite-field](/private/tmp/let-go-pr232/pkg/rt/core/ir/passes/pipeline.lg:53) rewrites every matching symbol in the body to `(. this field)` without lexical binding awareness. A method `(pick [this side] side)` on a type with field `side` should return the parameter, but native lowering returned the field.  
   Trampoline repro: `bytecode=9 native=3`.

3. **Deftype field identifiers are not Go-munged**  
   [lower-deftype-struct](/private/tmp/let-go-pr232/pkg/rt/core/ir/lower_go.lg:2388) passes raw `(str fld)` into struct fields, constructor params, composite keys, and idents. `gogen/field-decl` rejects invalid Go identifiers at [gogen.go](/private/tmp/let-go-pr232/pkg/rt/gogen.go:1122), so normal Lisp names like `the-side` break native lowering.

4. **Generated artifact manifest is stale**  
   `go test ./...` fails in `pkg/genmanifest`: recorded hash `65f36aff...` vs computed `bc573031...`. Focused package tests passed, but the full suite is currently not merge-clean.

Lower severity: [RecordBase.Unbox](/private/tmp/let-go-pr232/pkg/vm/record_base.go:46) claims it returns the embedding struct, but a promoted embedded method can only return `*RecordBase`. That may be fine if nobody depends on unboxing native records, but the comment/test overstate the contract.

mparrett:
Worth another pass for correctness. The native deftype lowering is currently safe only for the narrow skeleton case: field-only arithmetic method, no shadowing, Go-valid field names. The first two findings are the blockers. Minimal `.lg` fixtures that demonstrate the main issues:

**1. Method Calls Break Generated Go**

```clojure
(ns examples.deftype-method-call)

(defn plus1 [x]
  (+ x 1))

(defprotocol MethodCall
  (calc [this]))

(deftype BoxCall [n]
  MethodCall
  (calc [this] (plus1 n)))

(defn run []
  (calc (->BoxCall 4)))
```

Expected result: `5`.

Native lowering emits a receiver method with no `ec` parameter, but the method body still lowers the helper call through `ec`:

```go
func (arg0 *Boxcall) Calc() (vm.Value, error) {
    v4, callErr = rt.InvokeValueEC(ec, ...)
    //                         ^^ undefined
}
```

It also devirtualizes the call as if the method returns one value:

```go
v9 = arg__3.Calc()
```

but `Calc()` returns `(vm.Value, error)`, so the generated Go fails to compile.

**2. Shadowed Field Names Miscompile**

```clojure
(ns examples.deftype-shadow-param)

(defprotocol ShadowParam
  (pick [this side]))

(deftype SquareShadow [side]
  ShadowParam
  (pick [this side] side))

(defn run []
  (pick (->SquareShadow 3) 9))
```

Expected result: `9`, because `side` in the method body refers to the method parameter.

Observed native result: `3`.

The field rewrite treats every bare symbol matching a field name as a field access, so it effectively changes the method body to:

```clojure
(pick [this side] (. this side))
```

This is silent wrong-code.

**3. Hyphenated Field Names Fail Native Lowering**

```clojure
(ns examples.deftype-hyphen-field)

(defprotocol HyphenField
  (read-field [this]))

(deftype HyphenBox [the-side]
  HyphenField
  (read-field [this] the-side))

(defn run []
  (read-field (->HyphenBox 7)))
```

Expected result: `7`.

Native lowering fails before or during Go source construction because raw Lisp field names are passed to Go AST helpers without munging. `the-side` is valid in Lisp, but not a valid Go identifier. The lowerer needs to consistently munge it, for example to `the_side`, everywhere it appears in generated Go: struct field, constructor parameter, composite literal key, and field selector.

nnunley:
Thanks for the careful review — all five findings are addressed in `86b0b2df`.

1. **Receiver methods with calls / `undefined: ec` + return arity** — protocol methods now lower to a uniform IFn ABI: `func (recv *T) M(ec *vm.ExecContext, …) (vm.Value, error)`. The generated interface, the devirtualized direct call (`recv.M(ec, …)`), and `call-error-free?` all agree on that one signature, so a method body containing a call compiles and returns/handles the error. (`(calc [this] (plus1 n))` repro is green.)

2. **Shadowed field names** — `dt-rewrite-field` is now lexical-binding-aware: it threads a `bound` set and handles `let`/`loop`/`fn` binding forms, mirroring `core.lg`'s `-dt-rewrite`, seeded with the method params. `(pick [this side] side)` now returns the parameter, not the field.

3. **Hyphenated field names** — field identifiers are munged via `go-name` everywhere they become Go identifiers: struct field, constructor param, composite-literal key, **and** the `(. obj field)` selector. `the-side` → `the_side` consistently.

4. **Stale generated manifest** — regenerated `core_compiled.lgb` + `generated.sums` (`make generate`); `pkg/genmanifest` (`TestGeneratedArtifactsAreFresh`) is green.

5. **`RecordBase.Unbox` doc** — corrected to state it returns the embedded `*RecordBase` (the promoted-method receiver), not the outer embedding struct.

Verification: three new trampoline fixtures (`test/gogen/deftype_{method_call,shadow_param,hyphen_field}.lg`) plus the existing skeleton are all at bytecode↔native parity; `make check-generated` is green (including the `-tags gogen_ir` corpus compile); full `go test ./...` passes.

nnunley:
Note: I don't think this is fully stabilized yet. But I think it might be good for incremental progress.

mparrett:
> Note: I don't think this is fully stabilized yet. But I think it might be good for incremental progress.

Agreed it's good for incremental progress. I was curious to understand this better. Asked the swarm to probe the post-fix head (`86b0b2d`) along the axes outside the skeleton, using the trampoline harness. Found a few things for your consideration before integrating this:

## A structural issue: dispatch is correct only when the receiver type is statically known

Native protocol dispatch works only when the call site can infer the receiver as `[:dtype T]` and devirtualize to a direct `recv.Method(ec)`. When the type is erased, the lowerer emits a runtime protocol-var dispatch instead:

```go
ec.Invoke(rt.CachedVarFn(&__v_..._val5, "...", "val5"), []vm.Value{c})
```

and that path has no impl registered for the native record — the lowered package only registers top-level defns as overrides, so the protocol methods are reachable only through devirtualization. So any erasure point fails at runtime rather than falling back to bytecode.

Two minimal repros (both return the right answer on the bytecode VM, both fail natively):

```clojure
;; a method calling a sibling protocol method on `this` — `this` isn't seeded as [:dtype T]
(defprotocol PSelf (base-val [this]) (doubled [this]))
(deftype SelfBox [v] PSelf
  (base-val [this] v)
  (doubled [this] (* 2 (base-val this))))
(doubled (->SelfBox 7))   ; bytecode 14, native: runtime dispatch failure
```

```clojure
;; a receiver whose type is erased through an ordinary fn
(defprotocol PFlow (val5 [this]))
(deftype FlowBox [v] PFlow (val5 [this] v))
(defn pick-it [b] b)
(let [c (pick-it (->FlowBox 8))] (val5 c))   ; bytecode 8, native: runtime dispatch failure
```

The skeleton sidesteps this because every call site there has a statically known concrete receiver. The fix I'd suggest is a guard: when inference can't prove `[:dtype T]` at a protocol call site, fall back to bytecode for that site instead of emitting a dispatch that fails at runtime. That keeps the native path "narrow but safe" rather than "fails on ordinary code."

## Smaller gaps

- **Zero-field deftype** emits no Go — `lower-ns-to-go` returns nil for `(deftype Marker [] ...)`.
- **Munge collision**: fields `a-b` and `a_b` both munge to `a_b`, so the struct fails to compile (`a_b redeclared`). Loud today, but it's one step from silently selecting the wrong field — `go-name` does no collision check.
- **Multi-arity methods** emit no Go — consistent with the scope you noted as follow-on.

## Note about the harness

The trampoline asserts `native == bytecode`, not `native == correct`. A `for`/`doseq` binding that shadows a field name fails identically on both paths (the `dt-rewrite-field` docstring already flags `for`/`doseq`/`letfn` as a shared limitation with `core.lg`), so a pure parity gate stays green on it. Not a blocker for this PR — it's a pre-existing bytecode-path gap — but worth knowing the parity check can't catch bugs the two paths share.

None of this changes the incremental-progress call. The devirtualization-or-fallback point is important to resolve before deftype lowering gets pointed at anything beyond hand-picked fixtures.

nooga (approved):

PR #233, opened 2026-06-15
docs(design): runtime I/O host-decoupling proposal

mparrett:
A `status: planning` design doc capturing where the runtime-I/O work is heading and why, so the direction is reviewable in one place.

It follows the framing from nooga/xsofy#22 — get host specifics out of let-go, expose an interop surface, let the client build its own shell — and works it out for I/O: `*in*`/`*out*`/`*err*` as seams, each host (native, embedder, WASM) binding the concrete stream. The output half is done (#206, #207, #223, with #231 in review); the doc lays out the open input decisions (the `HostReader` shape and the `read-key` wake model) before that code lands.

No code change — `docs/design/runtime-io-host-decoupling.md` plus an index row. Posting it as a proposal to get direction-level feedback; §5 (the design decisions, each with an alternative and tradeoff) is where I'd most value your read.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -38,6 +38,7 @@ Design plans, execution roadmaps, and policy for the let-go implementation. Each
 | Babashka pod support | `pods.md` |
 | IR fixup / link pass | `els2023-ir-fixup-audit.md` |
 | Parallel IR lowering + determinism | `parallel-lowering-and-type-cache.md` |
+| Runtime I/O, host decoupling | `design/runtime-io-host-decoupling.md` |
 
 ## Reading order if starting cold
 
--- a/docs/design/runtime-io-host-decoupling.md
+++ b/docs/design/runtime-io-host-decoupling.md
@@ -1,0 +1,127 @@
+---
+status: planning
+last-verified: 2026-06-15
+human-verified: 2026-06-15
+---
+
+# Decoupling runtime I/O from the host
+
+**Status:** design proposal — where the runtime-I/O work is going and why
+**Decision leaning:** the runtime exposes a small set of I/O *seams*; each host (native, embedder, WASM) binds the concrete stream. Output is done; input is the live frontier.
+
+## 1. Where this came from
+
+This direction comes from nooga/xsofy#22:
+
+> The xterm.js bundle was sort of a temporary hack to get xsofy going. My thinking now is that we should get it out of let-go itself and provide a more decent browser interop surface instead. This should allow the client code, xsofy in this case, to build its own web shell without disturbing let-go.
+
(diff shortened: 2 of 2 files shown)


mparrett:
@nooga — this picks up the direction you sketched in nooga/xsofy#22 (get host specifics out of let-go, expose an interop surface, let the client build its own shell) and works it out for I/O as a planning doc, so the whole direction is reviewable in one place.

The output half is already in: #206/#207/#223, with #231 in review. What I'd most value your read on is §5, the open decisions: especially the input shape (5.2) and the `read-key` wake model (5.3), since those gate the next PR. No rush.

nooga (approved):

mparrett:
**TL;DR — where the browser/host interop arc landed**

A few weeks back you floated two things: `js/emit`, and wanting a better browser/host interop story (the xsofy#22 framing — get host specifics out of let-go, expose an interop surface, let the client build its own shell). I took that as the thesis and built it out:

**let-go stops owning the host. It exposes typed seams; each host — native, embedder, WASM/browser — binds them. The client owns its shell.**

That's now real from the print path up to a custom browser shell, proven by a real client (xsofy, a TUI game on let-go) running interactively on an iPhone in its own shell.

## The shape

This doc is the spine: `*in*`/`*out*`/`*err*`, emit, and input as seams the runtime exposes and hosts bind. Everything else implements one seam at a time:

- **Output** — merged: print/term routed through `*out*`/`*err*` (#206, #207, #223), embedder options, and a WASM `HostWriter` retiring the `fs.writeSync` shim (#231).
- **Emit** — #241 (open): typed `HostEmitter`, promoting `js/emit` off the raw `_lgEmit` lookup. Your `js/emit` thread, made a first-class seam.
- **Input** — #244 (open): typed `KeySource` for `term/read-key` + `key-pending?`.
- **Shell** — #245 (draft): `lg -w -w-shell none` splits the host JS so a client binds its own shell against `window.LetGoHost`, with the default xterm kept. xsofy#78 is the companion client that proves it.

## Where it stands for you

- Merged and live: the output half plus this doc.
- Open, ready to land: #241 (emit) and #244 (input) — both green and mergeable, independent code.
- Draft, for direction: #245 + xsofy#78 — the "client owns the shell" endgame. Draft because it sits behind the seams in the sequencing here. #245 lays out four delivery shapes — A (inline single-file, what it ships) through D (let-go emits glue + a separate, client-hosted `main.wasm`) — and we lean **ship A now, add D when the need arises**. That A/D call is the main thing worth your read before merge.

Nothing's blocked on you. #241/#244 are the ready pieces; #245 is where a direction-level nod (or redirect) helps most.

PR #234, opened 2026-06-15
refactor(test): move e2e tests under test/e2e + pkg/bundle extraction + root-test guard

nnunley:
**Stacked on #232** (`feat/gogen-interface-recordbase`). Until #232 merges, the
diff here includes its commits; afterward it collapses to just the two commits
below. Review/merge after #232.

## What

Gets all test code out of the repository root and adds a guard so it stays out.

Six `*_e2e_test.go` files lived at the repo root as `package main`. This moves
them out and removes the only reason one of them couldn't move cleanly.

### `refactor(bundle): extract lg.go standalone-binary subsystem into pkg/bundle`

`resources_e2e_test.go` was white-box — it tested `lg.go`'s unexported bundle
trailer/resource internals, so it was pinned to `package main` at the root. This
extracts that subsystem (the LGBX/LGB2 trailer format, payload read/write,
resource collection) into a new **`pkg/bundle`** package with a small exported
API (`ReadBundled`, `BaseBinarySize`, `CollectResources`, `AppendTrailer`); the
format internals stay unexported. `lg.go` imports it at the existing call sites —
behavior unchanged. The bundle-internals tests move to `pkg/bundle/bundle_test.go`.

### `refactor(test): move e2e tests under test/e2e; guard the repo root`

The five black-box e2e tests (build the `lg` binary + exec it) move to a new
**`test/e2e`** package, plus the black-box half of the resource tests. Since
`go test` runs a package's tests with cwd = the package dir, `test/e2e/helpers_test.go`
adds `repoRoot()` (walks up to `go.mod`); `buildLG` builds from there, and the
trampoline / gold tests set the subprocess cwd to the repo root so their
repo-relative paths (`scripts/`, `test/gold/`) resolve.

**Guard:** `scripts/check_test_location.py` fails if any `*_test.go` / `*_test.lg`
is added at the repo root, wired as a `check-test-location` pre-commit hook **and**
a CI job (jj doesn't run git hooks, so CI is the real gate).

Follow-through: the `gold-differential` CI job now targets `./test/e2e/`
(`TestGold*` moved there); the Makefile `test` target runs `./test/...`.

## Verification

- `go build ./...` and `-tags gogen_ir` / `-tags bootstrap` — clean.
- `go test ./pkg/bundle/` and `go test ./test/e2e/` — pass (repo-relative paths
  resolve via `repoRoot` despite the non-root cwd).
- Full untagged suite (15 packages) green; `make check-generated` green.
- `python3 scripts/check_test_location.py` exits 0 on the now-clean root and 1
  when a root test file is planted.


mparrett:
**Codex Findings (5.5-high)**

- **[P1] `go test ./...` fails because generated artifacts are stale.**  
  [pkg/rt/generated.sums](/private/tmp/letgo-pr234/pkg/rt/generated.sums:5) records `65f36...`, but `pkg/genmanifest` computed `bc573...`. The test output says to run `make generate`, and the suite fails at `github.com/nooga/let-go/pkg/genmanifest`.

- **[P1] Native `deftype` method rewriting ignores lexical shadowing.**  
  [pipeline.lg](/private/tmp/letgo-pr234/pkg/rt/core/ir/passes/pipeline.lg:53) rewrites every matching field symbol recursively. For `(let [side 4] side)` inside a method, the native lowering rendered `return arg0.side`, dropping the local binding and changing semantics.

- **[P2] `RecordBase.Unbox()` returns the embedded base, not the generated record.**  
  [record_base.go](/private/tmp/letgo-pr234/pkg/vm/record_base.go:48) cannot return the outer `*Square` through a promoted embedded method; it returns `*vm.RecordBase`. That breaks callers expecting `Value.Unbox()` to expose the underlying Go value.

- **[P2] Native `deftype` field lowering uses raw Lisp field forms as Go identifiers.**  
  [lower_go.lg](/private/tmp/letgo-pr234/pkg/rt/core/ir/lower_go.lg:2389) passes `(str fld)` into `gogen/field-decl`, constructor params, and struct literals. Legal fields like `side-len` fail with `not a valid identifier`, and metadata fields like `^:mutable items` stringify as `(with-meta items {:mutable true})`.

**Notes**

GitHub reports PR #234 as `CONFLICTING` with `main`.

Checks run: `go test ./...` failed only on `pkg/genmanifest`; `python3 scripts/check_test_location.py` passed; `git diff --check origin/main...HEAD` passed; `go build -tags gogen_ir ./...` passed; untagged `go build` passed.

nooga (approved):

PR #235, opened 2026-06-15
perf(ir): thread typeinfer state + preserve [:dtype T]; complete numeric typing (contagion, :ratio, quot/div ops)

nnunley:
> **Scope note:** this PR was approved at its original scope (typeinfer
> threading + `[:dtype T]` preservation). Addressing a review finding about
> a stale `:number` crashing strict `lower-go` expanded it into a fuller
> **numeric typing** pass — including two new first-class IR ops (`quot`,
> `div`) and a new `OP_DIV` bytecode op. Flagging for re-review. The
> op-definition scattering this surfaced is tracked for consolidation in
> `tk lg-6r45`.

### Typeinfer threading + exact `[:dtype T]` (original)

Threads typeinfer state through a shared epoch instead of writing
`Inst.Type` on the hot path, and adds an exact-`[:dtype T]` preservation
path so concrete record types survive to the lowering boundary.

- **`ir.lattice` (new, `pkg/rt/core/ir/lattice.lg`)** now owns fact/state
  handling. Exact `[:dtype T]` facts are kept *out* of the bitmask
  universe and flushed back unchanged, so a concrete type is never
  widened away by a join with the bitset lattice.
- **`infer-arg-types`, `typeinfer`, `pipeline`** run through the shared
  epoch state rather than mutating `Inst.Type` through the hot path.
- **`lower-go`** treats a preserved exact dtype fact as a concrete
  pointer type at the lowering boundary, so a preserved `[:dtype Square]`
  lowers as `*Square` instead of falling back to `vm.Value`.
- **`cmd/lgbgen`** registers the new `ir.lattice` module so it is bundled
  and lowered alongside the rest of the IR namespaces.

### Numeric typing (added in review)

The reviewer found that a stale `:number` survived the second typeinfer
epoch and crashed strict `lower-go` with `unsupported result type`. The
root causes were that mixed-numeric arithmetic was mis-typed and that the
lowering target table wasn't total. Fixed properly:

- **Numeric contagion** for `+ - *` (and `inc`/`dec`): operands all int →
  `:int`; any concrete float → `:float`. Matches the runtime
  (`(+ 1 2.0)` ⇒ `3.0`), so mixed int/float lowers to a native `float64`
  instead of the ambiguous `:number`.
- **Total `go-type-spec`:** `:number` and `:ratio` lower to `vm.Value`
  (neither has a native Go scalar type), so an honest int|float / ratio
  result never crashes strict `lower-go`.
- **`:ratio` as a first-class type:** a non-maskable lattice kind (like
  `:dtype`) so it survives the threaded state; ratio literals (`1/2`) are
  typed `:ratio`. No bitset bit, so `ti-bitset-proptest` stays green.
- **`quot` + `div` as first-class IR ops** (previously `quot` was an
  unregistered op that threw `unknown op quot` on the lowering path, and
  `div` was an untyped generic call):
  - `:quot` registered in the op catalog (`OP_QUOT`). Typed `:int` for
    int/int (Go `/` truncates toward zero like `quot`); any other mix
    boxes via `rt.QuotValue`.
  - `:div` is a **new** op: `OP_DIV` opcode + `rt.DivValue` (both via
    `vm.NumDiv`, identical to `clojure.core//`). Typed `:float` for
    float/float (native Go `/`); int/int (`Ratio`) and mixed box via
    `rt.DivValue`. Unary `(/ x)` lowers to the reciprocal `(/ 1 x)`;
    n-ary folds left. `lower-go` dispatches `quot` vs `div` by op-kw
    (both render Go `/`).

## Tests

- `lisp_typeinfer_test.go` — threaded typeinfer state.
- `lisp_lower_go_test.go` — dtype-preservation (`[:dtype T]` → `*T`),
  plus the numeric paths: mixed int/float → native `float64`, ambiguous
  `:number` → `vm.Value`, `quot` int/int native vs mixed `QuotValue`,
  `div` float/float native vs int/int `DivValue`.

## Benchmarks (`scripts/ti-*.lg`)

- **`ti-bitset-proptest.lg`** — exhaustively checks `mask-join == type-join`
  over the closed type universe (26,244 non-const pairs PASS), proving the
  bitset fast path is equivalent to the reference join.
- **`ti-microbench.lg`** — mask vs. keyword-join microbench: ~3× on common
  (early-cond) joins, ~247× on the wide-union normalize path.
- **`ti-baseline.lg`** — corpus profile over the IR sources with a capped
  per-fn budget, reporting join/normalize counts and wall-time.

## Verification

- `go test ./...` → pass (incl. `pkg/vm` for `OP_DIV`, `pkg/ir`, full
  `./test` suite)
- Runtime Clojure-parity confirmed for `quot`/`div`/`rem`/`mod`/`+`/`*`
  (e.g. `(/ 1 2)` ⇒ `1/2` Ratio, `(/ 4 2)` ⇒ `2` Int, `(quot 7.0 2)` ⇒
  `3.0`, `(/ 2)` ⇒ `1/2`)
- `make check-generated` → bundle in lockstep with `.lg` sources; lowered
  tree compiles + dispatches natively under `-tags gogen_ir`
- `ti-bitset-proptest.lg` → PASS (lattice change is sound)

Generated artifacts (`core_compiled.lgb`, `generated.sums`,
`op_generated.go`) were regenerated with `make generate` after the source
changes.


mparrett:
Codex (5.5-high) found one blocking issue:

- **[P1] Stale inferred types survive the second typeinfra epoch** in [lattice.lg](/private/tmp/let-go-pr235/pkg/rt/core/ir/lattice.lg:312) / [typeinfer.lg](/private/tmp/let-go-pr235/pkg/rt/core/ir/passes/typeinfer.lg:349). The new path seeds every existing `Inst.Type` instead of clearing derived slots. `optimize-fn` runs inference, then `constfold`, then inference again; `constfold` rewrites ops/refs/aux without clearing the old type. Repro on PR head: optimizing `(defn mixed-zero [] (+ 1 2.0))` leaves `v2 = Const ; 3.0 : number`, and strict `lower-go` fails with `ir.lower-go: unsupported result type` because `:number` has no Go type spec. The old pass avoided this by resetting non-`load-arg` types to `:bottom` before recomputing. Fix by seeding only true input pins for a fresh epoch, or by clearing/updating type slots when mutating passes replace an instruction.

Verification: `go test ./pkg/ir` passed. A temporary focused regression test for the repro above failed as described, then I removed the scratch test from the throwaway checkout.

nooga (approved):

PR #236, opened 2026-06-15
ci(perf-timeline): add workflow_dispatch for historical backfill

mparrett:
## What

Add a `workflow_dispatch` trigger to the perf-timeline workflow so it can
snapshot an arbitrary commit, then dispatch it once per historical commit
to backfill the timeline.

## Why

The "Are we fast yet?" timeline only records a snapshot on push to `main`,
so it has just the seed points: no trend line, and branch work produces
nothing to look at. Filling the v1.8.0→main gap meant either a one-off
local sweep (different hardware, manual perf-data push) or this — dispatch
the existing job per commit, on the same `ubuntu-latest` hardware as every
other point, auto-published to `perf-data`.

## How

The job was already commit-agnostic: snapshot name, sha, and commit time
all derive from `HEAD`. The only thing tied to the push event was which
commit `checkout` grabbed.

- `checkout` takes `ref: ${{ github.event.inputs.ref || github.sha }}`
  (`fetch-depth: 0` was already set, so arbitrary shas resolve).
- The `push` trigger and every downstream step are unchanged, so per-merge
  behavior is identical.
- The perf-data commit message no longer uses `github.sha`, which under
  dispatch is the branch tip rather than the backfilled commit; it names
  the snapshot's own commit via `snapshot_name`.

`make lowered` regenerates the gitignored IR tree per run, so no generated
code is committed.

## Backfill

After this merges, the sweep is a sampled dispatch loop:

    for sha in $(git rev-list --reverse v1.8.0..main | awk 'NR % 10 == 1'); do
      gh workflow run perf-timeline.yml -f ref="$sha"
    done

Runs queue on the `perf-timeline-main` concurrency group and drain onto
`perf-data` one at a time. Start coarse, fill in around interesting steps
later.

### DIFF
--- a/.github/workflows/perf-timeline.yml
+++ b/.github/workflows/perf-timeline.yml
@@ -6,6 +6,15 @@ on:
     paths-ignore:
       - 'docs/perf/index.html'
       - 'docs/perf/timeline/**'
+  # Manual backfill: dispatch once per historical commit to fill the timeline
+  # gap. Snapshot identity is derived from the checked-out HEAD (name, sha,
+  # commit time), so the rest of the job is already commit-agnostic.
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: 'Commit SHA/ref to snapshot (backfill)'
+        required: true
+        default: 'main'
 
 permissions:
   contents: write
@@ -13,6 +22,11 @@ permissions:
 concurrency:
   group: perf-timeline-main
   cancel-in-progress: false
+  # Retain every queued run (FIFO, up to 100) instead of the default
+  # queue: single, which keeps only the latest pending run. The backfill
+  # dispatches one run per historical commit; under queue: single all but
+  # the last would supersede each other and be canceled.
+  queue: max
 
 jobs:
   profile:
@@ -22,6 +36,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
+          ref: ${{ github.event.inputs.ref || github.sha }}
           submodules: recursive
           fetch-depth: 0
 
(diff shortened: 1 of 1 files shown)


mparrett:
**Codex Findings (5.5-high)**

- **P1** [`.github/workflows/perf-timeline.yml:22`](/Users/matt/projects-new/3p/joint-xsofy/let-go/.github/workflows/perf-timeline.yml:22): the backfill dispatch loop will drop most requested commits. The workflow keeps the existing fixed concurrency group with `cancel-in-progress: false`, but GitHub Actions’ default concurrency queue allows only one pending run; each newly queued run cancels/replaces the previous pending run. The PR body’s loop (`gh workflow run ...` per SHA) would therefore keep the active run plus the latest pending run, not drain every historical commit. Existing `Perf Timeline` run history already shows pending runs being canceled under this group. Fix by adding `queue: max` under `concurrency`, or throttle the dispatch script to wait for each run to finish before submitting the next.

Checks on PR #236 are green: `build`, `lint`, `gold-differential`, `generated-artifacts`, and `docs-frontmatter`.

Reference used: [GitHub Actions concurrency docs](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency).

mparrett:
Good catch — confirmed and fixed. Amended into `1925669`: added `queue: max` to the concurrency block. Under the default `queue: single` the per-commit dispatch loop would've kept only the latest pending run and canceled the rest; FIFO retention drains every commit, and it also stops rapid pushes to `main` from superseding intermediate timeline points.

One caveat worth noting: the FIFO cap is 100 pending, so a full every-commit sweep of a long range needs batching. The sampled loop in the description stays well under that.

nooga (approved):

mparrett:
Heads-up for anyone dispatching this by hand: `actions/checkout@v4` only treats a **full 40-char SHA** as a commit. A short SHA goes through as a branch/tag name and fails at checkout — `A branch or tag with the name '<sha>' could not be found`. The documented `git rev-list` backfill loop is fine since it emits full SHAs; this only bites manual dispatch from the Actions UI with a short SHA.

#238 hardens it: drop the `ref` input and resolve the target in a script step (`git checkout "$ref"`) so any committish — short SHA, full SHA, branch, or tag — works.

PR #238, opened 2026-06-16
ci(perf-timeline): resolve dispatch ref in a script step (support short SHAs)

mparrett:
## What

Resolve the dispatch `ref` in a script step so the timeline job accepts any committish — short SHA, full SHA, branch, or tag — not just full 40-char SHAs.

## Why

`actions/checkout@v4` only treats a **full 40-char SHA** as a commit. A short SHA passed to the `ref` input is interpreted as a branch/tag name and fails:

```
##[error]A branch or tag with the name 'd4eb364' could not be found
```

The documented backfill loop (`git rev-list …`) is unaffected, since it emits full SHAs. But hand-dispatching from the Actions UI with a short SHA — the natural thing to do — fails at checkout. Hit while validating the #236 backfill on a fork.

## How

Drop the `ref` input; check out at `fetch-depth: 0` (which already fetches every branch and tag), then `git checkout "$ref"` in a script step. Git resolves any committish locally against the full clone. Submodules are re-synced to the resolved commit, and snapshot identity still derives from `HEAD` (now the resolved target).

Follow-up to #236.

### DIFF
--- a/.github/workflows/perf-timeline.yml
+++ b/.github/workflows/perf-timeline.yml
@@ -36,9 +36,19 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          ref: ${{ github.event.inputs.ref || github.sha }}
           submodules: recursive
           fetch-depth: 0
+
+      - name: Check out target ref
+        run: |
+          ref="${{ github.event.inputs.ref || github.sha }}"
+          # actions/checkout only treats a *full* 40-char SHA as a commit; a
+          # short SHA is read as a branch/tag name and fails ("a branch or tag
+          # with the name '<sha>' could not be found"). Resolve here instead —
+          # fetch-depth:0 already pulled every branch and tag, so any committish
+          # (short SHA, full SHA, branch, tag) resolves locally.
+          git checkout --force "$ref"
+          git submodule update --init --recursive
 
       - name: Set up Go
         uses: actions/setup-go@v5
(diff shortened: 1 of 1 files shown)


mparrett:
**Codex Findings (5.5-high)**

- **[P1] Shell-injection risk in `workflow_dispatch` input**: [`.github/workflows/perf-timeline.yml#L43`](https://github.com/nooga/let-go/pull/238/files#diff-8d4f9f5f2c0c1e14d7de0c68e59c2d99db4f8b02ad8a7cc0611cb3181813b68fR43) interpolates the manual `ref` input directly into shell source:
  ```sh
  ref="${{ github.event.inputs.ref || github.sha }}"
  ```
  A dispatch input is free-form text, so a value like `$(...)` would be evaluated by the shell before `git checkout` ever validates it. This workflow has `contents: write`, so this should not land as-is. Pass the expression through `env` and expand only the env var inside the script:
  ```yaml
  - name: Check out target ref
    env:
      TARGET_REF: ${{ github.event.inputs.ref || github.sha }}
    run: |
      git checkout --force "$TARGET_REF"
      git submodule update --init --recursive
  ```

**Notes**
- PR CI is currently green: `lint`, `build`, `gold-differential`, `generated-artifacts`, and `docs-frontmatter`.
- I did not run the full perf workflow locally.

nnunley (approved):

mparrett:
Addressed — the input no longer lands in a `run:` block. It's passed via `env: TARGET_REF` and the script references `"$TARGET_REF"`, so the value can't break out into the shell. Pushed in the current head.

PR #239, opened 2026-06-16
fix(ir): close four gogen self-AOT lowering gaps (nth-non-int, free-vars, identifier munging, method interop)

nnunley:
Closes four `:gogen/*` self-AOT lowering gaps surfaced by `scripts/ir-stress.lg lower-go`. On the corpus subset the harness covers, this moves **673→700 passed (32→8 failures)** with no new buckets. Each fix is a focused source change in the `.lg` lowering pipeline and ships a red→green regression test under `test/`.

### Commits

1. **`fix(ir): guard dominance/successors against dead-block nil terminator`** — `:gogen/nth-non-int` (13→0, +14)
   `structurize/post-idoms` walks every block via `(range n)` and calls `successors` on each, including dead/unterminated blocks `build` can leave behind (the goto-fallback path in `lower_go` already skips these). For such a block the terminator is `nil`, so `successors` evaluated `(ir/op nil f)` → `(nth insts nil)` → "nth index must be an integer", caught by `lower` as a `:fallback`. Guard `successors` to return `[]` on a nil terminator. (+14 = the 13 plus one dependent multi-arity fallback.)

2. **`fix(ir): detect free vars inside map and seq literals in build`** — unresolved-symbol fallbacks (+3)
   `ir.build/free-vars` recursed into list/vector forms but not map or seq literals, so a closure capturing an outer local only inside a map (e.g. `(fn [] {:k x})`) never registered `x` as free → `x` reported unresolved during lowering. Adds `(map? form)`/`(seq? form)` branches.

3. **`fix(ir): match Clojure CHAR_MAP for Go identifier munging + _U fallback`** — `:gogen/bad-identifier` (4→0, +5)
   `go-name` munged ~10 chars and passed any **unmapped** char through verbatim, so any special char not in the map produced an invalid Go identifier (e.g. a `%1` var-decl from a reused `#()` param). Replaces the map with the canonical Clojure `CHAR_MAP` (`clojure.lang.Compiler/CHAR_MAP`) and adds a catch-all to both munge sites: ASCII alphanumerics and `_` pass through; any other char becomes `_U<hex>_`. Makes `:gogen/bad-identifier` structurally impossible rather than whack-a-mole.

4. **`fix(ir): lower Go method-call interop (.Method recv args)`** — `.WriteString`/`.UnixMilli` (3→0, +5)
   `build-list` had no dispatch for a call head that is a symbol starting with `.`, so `(.WriteString w s)` fell through to symbol resolution and threw "unresolved symbol .WriteString". Mirrors the bytecode VM's rewrite (`compiler.go`): `(.Method recv arg...)` → `(. recv (quote Method) arg...)`, which builds as a generic call to the `.` builtin and lowers through the existing trampoline — semantically identical to the VM. (A literal `recv.Method(args)` isn't pursued: the lowered receiver is a boxed `vm.Value` with no static concrete Go type, so the trampolined dispatch is the correct equivalent.)

### Verification
- Each commit: `make generate` + `make check-generated` clean (lowered tree compiles + dispatches under `-tags gogen_ir`), self-consistent bundle so history bisects.
- Combined tip: `ir-stress` 700/8 (all four target buckets at 0), full `TestRunner` + `pkg/ir` suites green.
- Each fix has a focused regression test that is red before the fix and green after: `test/gogen_nth_non_int_test.lg`, `test/gogen_freevars_capture_test.lg`, `test/gogen_percent_identifier_test.lg`, `test/gogen_method_interop_test.lg`.


mparrett:
**Codex Findings (5.5-high)**

- [P1] [pkg/rt/generated.sums](/private/tmp/let-go-pr239/pkg/rt/generated.sums:4) is stale. From a fresh PR worktree, `make check-generated` rewrote the digest from `26bf3313...` to `a33b220d...`, leaving the worktree dirty. The PR should commit the regenerated `generated.sums`; otherwise the genmanifest/pre-commit freshness checks can fail or disagree with the committed sources.

- [P2] [pkg/rt/core/ir/build.lg](/private/tmp/let-go-pr239/pkg/rt/core/ir/build.lg:492) detects method-call heads with `(name head)`, which strips namespaces. That means a qualified var call like `(foo/.bar x)` would be rewritten as method interop, while the bytecode compiler only rewrites symbols whose raw text starts with `.`. Use the raw symbol string or require `(nil? (namespace head))`, and add a regression for a qualified symbol whose name starts with `.`.

**Verification**

- `go test -timeout 60s -count=1 ./test` passed.
- `make check-generated` completed, but left `pkg/rt/generated.sums` modified, which is the stale-artifact finding above.

mparrett:
**Codex Findings (5.5-high)**

- [P1] `pkg/rt/generated.sums:5` is still stale. On a fresh worktree at PR head `f9091ed`, `make check-generated` completed successfully but rewrote the committed digest from `1ce230b...` to `61a3f04...`, leaving `pkg/rt/generated.sums` dirty. The regenerated digest needs to be committed.

**Conflict Notes**

GitHub reports the PR as `CONFLICTING` / `DIRTY`. A merge attempt against current `origin/main` (`ad24d9b`) conflicts in:

- `pkg/rt/core/ir/build.lg`
- `pkg/rt/core_compiled.lgb`
- `pkg/rt/generated.sums`

PR #239 cannot merge until it is rebased/merged onto current `main`, keeping both `build-dot` from `main` and the PR’s `.Method` rewrite, then regenerating the bundle/digest artifacts.

The real source conflict is in `build.lg`: current `main` added `(= head '.) (build-dot form ctx)`, while this PR adds the `(.Method recv ...)` rewrite. The reconciliation should keep both, with the method-call rewrite before the direct `.` field-selector branch. Then regenerate `core_compiled.lgb` and `generated.sums`.

Verification run:

- `go test -timeout 60s -count=1 ./test` passed.
- `make check-generated` passed, but left `pkg/rt/generated.sums` modified as noted above.

nnunley:
Closing this in favor of the consolidated gogen self-AOT lowering work.

When rebased onto current `main`, the four gap fixes here are necessary but not sufficient: the lowered `ir.lower-go` tree no longer compiles under `-tags gogen_ir` (`lower_method`: dead locals + missing return), so `make check-generated` fails. The fix is a lowering-pipeline change (closure-local arg prefixing to stop captured-name shadowing, plus `:const`-Var recovery) that doesn't live in this PR.

All of this PR's content — `build.lg` method-interop rewrite (byte-identical), the nth-non-int / free-vars / identifier-munging fixes, and all four `gogen_*_test.lg` regression tests — is already carried forward in the consolidated self-AOT branch, which also includes that lowering fix. So nothing is lost by closing.

To keep giving consumers incremental slices, the lowering fix is being extracted as its own minimal PR (the foundation), with the gap-fixes layered on top in the correct dependency order.

PR #240, opened 2026-06-16
fix(generate): clean lowered-tree target dir before regen (hermetic generation)

nnunley:
## What

Make `make generate` **hermetic**: it now clears the lowered-tree target
directory (`pkg/rt/core_go_lowered/`) before regenerating, so stray files from a
previous run can't survive into the new one.

## Why

`pkg/rt/core_go_lowered/` is gitignored and fully regenerated, but nothing
cleared it first. Leftover packages from another branch's generation (e.g. a
renamed/removed namespace, or an experiment's output sitting on disk) would
linger. Those strays are invisible to `jj`/`git status` (gitignored) — **yet
they are digested into `generated.sums`**. The result: `make generate` produces
a different digest than what's committed, and the content-freshness check
(`make check-generated` / `TestGeneratedArtifactsAreFresh`, which `go test ./...`
runs) fails with no visible cause. This cost real debugging time.

## Fix

In `scripts/generate.lg`, immediately before `lgbgen --target=both`:

```clojure
(let [lowered "pkg/rt/core_go_lowered"]
  (println ">> clean" lowered)
  (os/sh "rm" "-rf" lowered))
```

The dir is recreated in full by `lgbgen --target=both`, so removing it wholesale
is safe and makes generation reproducible regardless of prior on-disk state.

## Verification

- `make generate` → `>> clean pkg/rt/core_go_lowered`, exit 0, all artifacts regenerated.
- **Stray-injection test:** created `core_go_lowered/STRAY_PKG/x.go`, ran `make generate` → stray removed, `generated.sums` unchanged (`a4a6854…`).
- `make check-generated` → 0. Touches no committed artifact (only `scripts/generate.lg` changes).


mparrett:
@nnunley —

**Codex Findings (5.5-high)**

1. **P1: `make check-generated` still has the same stale-tree problem.**  
   The clean was added only in `scripts/generate.lg` at `scripts/generate.lg:125`, so it only affects `make generate`. But the PR says the failure mode includes `make check-generated`, and that target bypasses `generate.lg` entirely: `Makefile:225` runs `go run -tags bootstrap ./cmd/lgbgen --target=both` directly. `make lowered` and the lowered-tree Makefile rule do the same at `Makefile:69` and `Makefile:140`. Since `runGoTarget` only `MkdirAll`s the output dir and writes current packages, stale package directories can still survive those paths and be picked up by `go build -tags gogen_ir ./...`. The cleanup probably belongs inside `lgbgen`’s Go target path, before `runGoTarget` writes into `pkg/rt/core_go_lowered`.

2. **P2: the cleanup ignores `rm -rf` failure.**  
   `scripts/generate.lg:127` calls `(os/sh "rm" "-rf" lowered)` but does not inspect `:exit`. If removal fails, generation continues and the exact non-hermetic state this change is meant to prevent remains in place. This should abort via `die`, or use a helper that checks shell command status.

3. **P3: the new comment says strays are digested into `generated.sums`, but they are not.**  
   `generated.sums` is computed from `pkg/rt/core/**/*.lg` and `cmd/lgbgen/**/*.go`, not from `pkg/rt/core_go_lowered/`. The stale lowered tree can still break `go build ./...`, but the in-code explanation at `scripts/generate.lg:121` is misleading.

This was a diff/code-path review against `origin/pr-240` and the current `Makefile`/`lgbgen` implementation.

---

I smoked it in a disposable copy of `origin/pr-240`. The current PR does not cover the direct `lgbgen` path used by `make check-generated`; moving the clean into `runGoTarget` does.

```md
I think the cleanup needs to live in `lgbgen`, not only in `scripts/generate.lg`.

Repro against this PR:

```bash
stray_dir="pkg/rt/core_go_lowered/zz_stray_bad"
mkdir -p "$stray_dir"
cat > "$stray_dir/zz_stray_bad.go" <<'GO'
package zz_stray_bad

func Broken = this is intentionally invalid Go
GO

go run -tags bootstrap ./cmd/lgbgen --target=both >/tmp/lgbgen.log

test -e "$stray_dir/zz_stray_bad.go" && echo "STILL PRESENT" || echo "REMOVED"

go build -tags gogen_ir ./...
```

Observed on PR 240:

```text
STILL PRESENT
# github.com/nooga/let-go/pkg/rt/core_go_lowered/zz_stray_bad
pkg/rt/core_go_lowered/zz_stray_bad/zz_stray_bad.go:3:13: syntax error: unexpected =, expected (
```

This matters because `make check-generated` calls `lgbgen` directly:

```make
go run -tags bootstrap ./cmd/lgbgen --target=both >/dev/null
go build -tags gogen_ir ./...
```

So `make generate` gets the new clean, but `make check-generated`, `make lowered`, and the lowered-tree Makefile rule still allow stale ignored packages to survive.

A fix that passed the same smoke test for me was to clean in `runGoTarget`, where both `--target=go` and `--target=both` converge:

```go
func runGoTarget(outDir string) {
    var failed []string
    var generated []string

    pipelineVar := rt.LookupVar("ir.passes.pipeline", "lower-ns-to-go")
    if pipelineVar == nil {
        fmt.Fprintf(os.Stderr, "fatal: ir.passes.pipeline/lower-ns-to-go not found\n")
        os.Exit(1)
    }

    if err := os.RemoveAll(outDir); err != nil {
        fmt.Fprintf(os.Stderr, "clean output dir %s: %v\n", outDir, err)
        os.Exit(1)
    }
    if err := os.MkdirAll(outDir, 0755); err != nil {
        fmt.Fprintf(os.Stderr, "create output dir %s: %v\n", outDir, err)
        os.Exit(1)
    }

    // existing lowering loop...
}
```

With that placement, the same repro prints:

```text
REMOVED
go build status: 0
```

One small wording nit: the new `generate.lg` comment says stale `core_go_lowered` packages are digested into `generated.sums`, but `genmanifest.Compute` only hashes `pkg/rt/core/**/*.lg` and `cmd/lgbgen/**/*.go`. The stale package breaks the `gogen_ir` build path, but not the manifest digest directly.

nnunley:
Addressed in 7e0ad91:

- **P1**: moved the clean into `lgbgen`'s `runGoTarget` (`os.RemoveAll(outDir)` before `MkdirAll`), the single point where `--target=go` and `--target=both` converge. `make generate`, `make lowered`, and `make check-generated` now all get the clean. Your repro now prints `REMOVED` + `go build status: 0`.
- **P2**: the unchecked `os/sh rm -rf` is gone — removal now happens in Go and aborts (`os.Exit(1)`) if `RemoveAll` fails.
- **P3**: dropped the incorrect claim that strays are digested into `generated.sums`; the new comment states they break the `gogen_ir` build path instead.

`make check-generated` is clean and `go test ./test ./pkg/genmanifest/...` passes.

mparrett:
**Codex Findings (5.5-high)**

1. **P1: `--target=go <out-dir>` can now delete arbitrary directories.**  
   The updated fix is in the right place for the Makefile paths, but `runGoTarget` now does `os.RemoveAll(outDir)` at `cmd/lgbgen/main.go:532`. `outDir` is user-controlled for documented CLI usage: `go run -tags bootstrap ./cmd/lgbgen --target=go <out-dir>` parsed at `cmd/lgbgen/main.go:326`. That means `--target=go .`, `--target=go pkg/rt`, or a mistyped existing directory becomes a recursive delete before generation. I’d add a guard before `RemoveAll`, at least refusing `.`, repo root, `/`, and maybe anything outside the known generated tree unless explicitly intended.

Suggested shape:

```go
func cleanGoOutputDir(outDir string) {
    clean := filepath.Clean(outDir)
    if clean == "." || clean == string(filepath.Separator) {
        fmt.Fprintf(os.Stderr, "refusing to clean unsafe output dir %q\n", outDir)
        os.Exit(2)
    }

    if err := os.RemoveAll(clean); err != nil {
        fmt.Fprintf(os.Stderr, "clean output dir %s: %v\n", clean, err)
        os.Exit(1)
    }
}
```

I’d probably also compare against `genmanifest.FindRepoRoot(".")` and refuse the repo root explicitly.

**Smoke Result**

The original stale-package repro now passes on updated head `7e0ad91f014451c2a299c783200cfceb85139d68`:

```text
stray after lgbgen:
REMOVED
checking gogen_ir build...
go build status: 0
```

So the previous functional issue is addressed.

<removed note about stale description- our comment was stale>

mparrett:
Main risk is a documented CLI invocation such as `go run -tags bootstrap ./cmd/lgbgen --target=go .` would recursively delete the current repo directory before recreating it as the output dir; simple guard should take care of it.

nnunley:
Addressed in 7f53f42c.

`runGoTarget` now calls a `cleanGoOutputDir` guard before `os.RemoveAll`. It refuses `.`, the filesystem root, and the repo root (resolved via `genmanifest.FindRepoRoot` + abs-path comparison), exiting 2 without deleting. The default and every Makefile caller pass the in-repo `pkg/rt/core_go_lowered` subdir, which is unaffected.

Verified with the built binary:
- `--target=go .` → `refusing to clean unsafe output dir "."`, exit 2; a sentinel file in the cwd survived (no delete).
- `--target=go <repo-root>` → `refusing to clean repo root …`, exit 2.
- `make check-generated` green (bundle lockstep + native dispatch); `go test ./test ./pkg/genmanifest/...` passes; your stale-package repro still prints REMOVED + build status 0.

The branch was also rebased onto current `main` (it had drifted onto an older base, conflicting on `generated.sums`).

mparrett:
Looks good. Btw opened #248 to fix/stabilize the failing gold-differential job.

mparrett:
One more Codex pass. Non-blocking but if it's easy, nice to do now.

Safer than the previous version; one remaining edge: `outDir` is still arbitrary user input, and the guard only rejects `.`, `/`, and the repo root.

⚠️ Examples like these below still pass the guard and recursively delete substantial non-generated directories before writing output:

```bash
go run -tags bootstrap ./cmd/lgbgen --target=go ..
go run -tags bootstrap ./cmd/lgbgen --target=go pkg
go run -tags bootstrap ./cmd/lgbgen --target=go pkg/rt
```

Given the Makefile only needs the default `pkg/rt/core_go_lowered`, the safest version may be to only auto-clean the canonical generated dir, or require a generated-dir marker before deleting a custom output directory.

I attempted to rerun the full stale-package smoke test against the new head, but this run stopped advancing during `lgbgen` and I killed the disposable temp process after several minutes. Static inspection shows the functional fix is still in the right place (`runGoTarget`), and the new remaining question is the breadth of the delete guard.

nnunley:
Reworked the clean to sanitize properly (9fc25cea).

Instead of `os.RemoveAll(outDir)`, the clean now deletes **only files lgbgen generated**:
- Every generated file (per-package lowered files + the wireup files) carries the `// Code generated by lgbgen --target=go. DO NOT EDIT.` banner.
- `cleanGoOutputDir` walks outDir and removes only `*.go` files bearing that banner, then prunes directories left empty. User-authored files have no banner and are never touched. The `.`/fs-root/repo-root checks stay as a backstop.

So a mistargeted `--target=go pkg/rt` can no longer destroy hand-written files — only lgbgen's own output goes. Real orphans (a renamed/removed namespace whose previously-generated `<pkg>/<pkg>.go` lingers) carry the banner and are still removed; that's the actual stale-package case the PR targets.

One behavior note on your earlier repro: a hand-planted `zz_stray_bad/zz_stray_bad.go` with no banner is now **intentionally preserved** — by design we only delete files we generated, not arbitrary content dropped into the tree. A faithful orphan repro is to generate, rename/remove a namespace, regenerate, and confirm the old package is gone (covered by the new unit test `TestRemoveGeneratedFilesDeletesOnlyBanneredFiles`, run under `-tags bootstrap`).

Verification: `go test -tags bootstrap ./cmd/lgbgen` (incl. the new test) passes; `make check-generated` green (lockstep + native dispatch with banners); `go test ./test ./pkg/genmanifest/...` passes.

(Follow-up tracked separately: evolving `generated.sums` into a per-file hash manifest, which would drive incremental re-lowering and double as the authoritative generated-file list.)

mparrett:
@nnunley Thanks for addressing the feedback. This looks much safer now (no nasty surprises with mistargeted custom dirs, as a zealous bot or unfocused human might be prone to doing). And the per-file hash manifest is an interesting idea. Probably outside the scope of #252, but adjacent and I can give that some thought if you like.

One small finding I considered and rejected is below; bold emphasis is mine. The failure scenario is probably rare and isolated, but you'd know better. A good example of scope creep in LLM review — overly defensive analysis and contingency handling for an assumed larger-than-actual developer pool or user base.

> the latest version of this PR protects _future generated files_, but it may not clean legacy orphan packages already on a developer/CI runner disk because old lowered files lack the new banner. So the original failure can still happen **once for anyone carrying an old pkg/rt/core_go_lowered tree with removed/renamed packages**. We can leave it alone if the team is comfortable with a manual escape hatch. It is not a correctness issue in fresh clones, and CI runners are usually clean. The risk is mostly local developer confusion, especially across branches.

The escape hatch mentioned above:
```
rm -rf pkg/rt/core_go_lowered
make generate
# or
make check-generated
```

I'm good with shipping this or open to making it more robust. Your call. 🚀

nnunley:
So, ready to merge if you give me approval @mparrett

mparrett (approved):
Sorry for the delay. Ship it!

Issue #41, opened 2026-05-19
CI doesn't detect stale core_compiled.lgb


nnunley:
`pkg/rt/core_compiled.lgb` is a checked-in build artifact regenerated by `go run ./cmd/lgbgen` (or `make generate`). It must be in sync with `pkg/rt/core/core.lg`, or runtime behavior silently diverges from the source.

Today, a contributor who edits `core.lg` without running `lgbgen` can commit a PR where the source change is invisible to the running binary (precompiled bytecode loads first per `pkg/compiler/eval.go:58`). CI doesn't catch this.

Suggested CI step in `.github/workflows/go.yml`:

```yaml
- name: Verify core_compiled.lgb is in sync with core.lg
  run: |
    go run ./cmd/lgbgen
    git diff --exit-code pkg/rt/core_compiled.lgb || {
      echo "::error::core_compiled.lgb is stale. Run: go run ./cmd/lgbgen"
      exit 1
    }
```

Bytecode regeneration appears deterministic. Raised during review of #40.

Related: #42 (the source-fallback path is broken so this check assumes the current `.lgb` is at least a valid starting point — which it has to be for `lgbgen` to run at all today).

nooga:
I'll take this one and #42 - I think LGB regen should go into a pre-commit hook AND we should also validate this in CI as you suggested.

nnunley:
Heads up before you wire the drift check: `lgbgen` is currently nondeterministic. Three back-to-back runs against an unchanged tree:

```
$ go run ./cmd/lgbgen >/dev/null && md5sum pkg/rt/core_compiled.lgb
fa4d85cb1f2856df65ec8d5c31315dda
603e207f61855df39be77cab80978442
19b1b9b9010cea0b300604dab8859c56
```

Same byte length, different content. Likely map-iteration-order somewhere in `pkg/bytecode/encoder.go` or `pkg/vm/constpool.go`. `go generate && git diff --exit-code` will false-positive every time until that's fixed. (Closed #44; folding here.)

nnunley:
Took a stab at the determinism part in #45 since it blocks the drift check — happy to discard if you'd rather own the whole thing. Two map-iteration sources (cmd/lgbgen + writeNSTable); after the fix six back-to-back runs hash identical. Doesn't touch the pre-commit hook / CI piece — that's still yours.

nooga:
[...]

PR #241, opened 2026-06-16
feat(rt,wasm,api): typed HostEmitter seam — promote js/emit off the raw _lgEmit lookup

mparrett:
Promotes `(js/emit ...)` from the untyped `_lgEmit` bridge to a typed host capability bound at an `*emit*` var — the emit dual of the `*out*` / `HostWriter` seam from #231. This is item 2 of the §6 sequence in the runtime-I/O decoupling doc (#233): "promote the existing `_lgEmit` bridge to a typed host capability, the same shape as the writer."

### Why

`js/emit` is split across `js_wasm.go` (calls `js.Global().Get("_lgEmit")`) and `js_other.go` (no-op), with the host wiring hardcoded in the WASM path. A Go embedder has no way to receive the events its guest dispatches — the `pkg/api` note has carried a `WithEmit` "deferred" comment since #174. Output already solved this with `HostWriter` + `*out*` + `api.WithStdout`; emit hadn't been given the same treatment.

### What

The change mirrors the writer arc:

- **`pkg/rt/emitter.go`** — `Emitter` interface (`Emit(name, dataJSON string)`), `FuncEmitter` adapter, a default `nopEmitter`, and `resolveEmitterVar` / `EmitVia` (duals of `resolveIOHandleVar` / `WriteToOut`).
- **`*emit*` core var** (`iort.go`), root-bound to `nopEmitter` so emit is harmless when nothing is listening, the same fallback shape as `*out*`.
- **`pkg/rt/hostemitter_js_wasm.go`** — `HostEmitter`, forwarding to `_lgEmit` (dual of `HostWriter` → `_lgOutput`). Installed at the `*emit*` root by the generated bundle main (`wasmMainTmpl` in `wasm.go`).
- **`js/emit`** is now a single ctx-aware native in `js.go` that validates args (unchanged) and dispatches through `EmitVia`. The per-platform `js_wasm.go` / `js_other.go` pair is deleted: the only platform difference left is which `Emitter` sits at the root, so the split collapses — the same simplification #231 produced for output.
- **`api.WithEmit(fn func(name, dataJSON string))`** — per-Run binding on `*emit*`, the emit dual of `WithStdout`. Same per-instance isolation and the same process-global-binding-stack concurrency caveat.

### Behavior

- WASM bundle: unchanged. `(js/emit :stats {...})` still reaches `LetGoHost.onEmit` via `_lgEmit`.
- Native, no option: unchanged. `(js/emit ...)` validates its args and no-ops.
- Native with `api.WithEmit`: the embedder's callback receives `(name, dataJSON)`, the same pair the bundle hands to `onEmit`.
- `(binding [*emit* ...] ...)` now works for guest-side redirection, like `*out*`.

No `.lg` sources change, so `core_compiled.lgb` is untouched.

### Testing

`pkg/api/emit_test.go` covers the `WithEmit` capture path (`(js/emit :stats {:hp 7})` → `("stats", {"hp":7})`) and the no-op/arg-validation path with no emitter. Existing `WithStdout` / `WithStderr` tests still pass. Native `go build ./...` and `GOOS=js GOARCH=wasm go build ./pkg/rt/` are both clean, as is `go vet`.

### Out of scope

`pkg/genmanifest`'s `TestGeneratedArtifactsAreFresh` already fails on a clean `main` (recorded `a4a6854…` vs computed `5daa727…`): the recorded manifest is stale relative to current `.lg` / lgbgen sources, independent of this change. A proper fix means regenerating the bundle + lowered tree, so it stays separate (#41 territory).


mparrett:
Codex (5.5-high): No blocking findings in either PR. 

A few non-blocking review notes:

- #241 fits the larger interop direction cleanly: `js/emit` now resolves through `*emit*` and `api.WithEmit`, matching the HostWriter pattern.
- #244 is also consistent with the design, but it intentionally does not solve `wake()` / parked `read-key` interruption for custom hosts yet. That should stay explicit in review discussion.
- The two PRs conflict mechanically in [pkg/api/api.go](/Users/matt/projects-new/3p/worktrees/let-go-keysource/pkg/api/api.go:58) and [pkg/rt/iort.go](/Users/matt/projects-new/3p/worktrees/let-go-keysource/pkg/rt/iort.go:330). `merge-tree` reports conflicts there, as expected. The second merge should preserve both `emit` and `keys` config/handles/options, and both `*emit*` and `*keys*` root vars.
- If #241 lands first, #244’s stale “WithEmit not exposed” comment in [pkg/api/api.go](/Users/matt/projects-new/3p/worktrees/let-go-keysource/pkg/api/api.go:82) needs to disappear during the rebase.

Verification run:

- #241: `go test ./pkg/api ./pkg/rt` passed; `GOOS=js GOARCH=wasm go build ./pkg/rt` passed.
- #244: `go test ./pkg/api ./pkg/rt` passed; `GOOS=js GOARCH=wasm go build ./pkg/rt` passed.

I also read #79, #233, xsofy#22, and #233’s design doc. The only larger-picture cleanup I’d suggest is a doc/status refresh: #233 still describes #231 as open and its §6 ordering differs from the current issue status sequence.

nooga (approved):

PR #90, opened 2026-05-26
repl: migrate from alimpfard/line to chzyer/readline for FreeBSD support

nnunley:
The alimpfard/line library uses Linux-specific termios/ioctl calls that don't compile on FreeBSD:

  ../go/pkg/mod/github.com/alimpfard/line@v0.0.0-20230131232016-03b4e7dee324/impl.go:284:6: undefined: setTermios
  ../go/pkg/mod/github.com/alimpfard/line@v0.0.0-20230131232016-03b4e7dee324/impl.go:536:10: undefined: getTermios

Replace with github.com/chzyer/readline which has broad platform support (Linux, macOS, FreeBSD, Windows) while maintaining feature parity:

- Tab completion via AutoCompleter interface
- Syntax highlighting via Painter interface (ANSI escape codes)
- Interrupt handling (Ctrl+C)
- History file support

Fixes #84

PR #243, opened 2026-06-16
build: keep the readline REPL out of GOOS=js builds

### LINKED ISSUE #242: build: GOOS=js go build ./... fails — REPL drags in chzyer/readline (no wasm support)
### What

`GOOS=js GOARCH=wasm go build ./...` fails to compile — the root `main` package pulls in the REPL, which imports `github.com/chzyer/readline`, and readline has no `js`/`wasm` support.

```
$ GOOS=js GOARCH=wasm go build ./...
# github.com/chzyer/readline
.../readline/utils.go:81:29: undefined: State
.../readline/operation.go:234:4: undefined: ClearScreen
.../readline/readline.go:129:20: undefined: GetScreenWidth
.../readline/readline.go:132:22: undefined: DefaultIsTerminal
.../readline/terminal.go:51:2: undefined: SuspendMe
... (12 errors)
```

### Root cause

`lg_repl.go` is tagged `//go:build !plan9`, so it compiles under `GOOS=js`. It imports `chzyer/readline`, which only builds for native terminals. `lg.go` (the root `main`) has no build constraint, so a wasm build of the root package drags the REPL — and readline — in.

### Impact

Build/CI hygiene, not a runtime bug. The shipped `lg -w` bundle is unaffected: the bundler compiles a generated `main` (the `wasmMainTmpl` temp tree) plus `pkg/rt`, not the root package or the REPL. So `pkg/rt` and the bundle build clean for wasm; only `go build ./...` under `GOOS=js` breaks. It surfaces as noise when cross-checking wasm builds locally.

### Suggested fix

Mirror the existing plan9 split. `repl(ctx *compiler.Context)` already has a plan9 stub in `lg_repl_plan9.go`; the readline-backed implementation in `lg_repl.go` can take the same treatment:

- tag `lg_repl.go` `//go:build !plan9 && !js`, and
- provide a `js`-tagged stub `repl()` (or fold js into the existing stub's constraint).

That keeps the REPL and readline out of any `GOOS=js` build while leaving native untouched.

Will send a PR. Let me know if anything looks off with the approach.

mparrett:
`GOOS=js GOARCH=wasm go build ./...` fails to compile: the root `main` package builds the REPL, which imports `github.com/chzyer/readline` — a termios/ioctl library with no js/wasm support. `lg_repl.go` was tagged `!plan9` only, so it compiled under `GOOS=js` and dragged readline in.

This has never built for `GOOS=js` — the gap predates the `chzyer/readline` migration (#90); `alimpfard/line` before it was equally termios-bound. It went unnoticed because the shipped `lg -w` bundle builds a generated `main` (the `wasm/` tree), not the root package, and CI builds only the bundle (`pages.yml`), never `go build ./...`.

Fix: route js to the readline-free stub REPL the same way plan9 already does.

- `lg_repl.go`: `!plan9` → `!plan9 && !js`
- `lg_repl_plan9.go` → `lg_repl_stub.go`, tag `plan9` → `plan9 || js` (the `bufio.Scanner` REPL needs no termios, so it serves both)

`go build ./...` now passes for `GOOS=js`; native is unchanged.

Closes #242.

### DIFF
--- a/lg_repl.go
+++ b/lg_repl.go
@@ -1,4 +1,4 @@
-//go:build !plan9
+//go:build !plan9 && !js
 
 /*
  * Copyright (c) 2021 Marcin Gasperowicz <xnooga@gmail.com>
--- a/lg_repl_stub.go
+++ b/lg_repl_stub.go
@@ -1,4 +1,4 @@
-//go:build plan9
+//go:build plan9 || js
 
 /*
  * Copyright (c) 2021 Marcin Gasperowicz <xnooga@gmail.com>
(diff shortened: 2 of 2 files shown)


mparrett:
**Codex (5.5-high) : No findings.**

I verified the exact PR head `571caa0` in a temp checkout. `go build ./...` and `GOOS=js GOARCH=wasm go build ./...` both exited successfully. `go list` also confirmed the intended file split: native selects `lg_repl.go`, wasm selects `lg_repl_stub.go`, and the wasm root package no longer imports `github.com/chzyer/readline`.

nnunley (approved):

PR #244, opened 2026-06-16
feat(rt,api): typed KeySource seam — route term/read-key + key-pending? through a host capability

mparrett:
Promotes blocking key input to a host-bound capability, the input peer of the `*out*` HostWriter (#231) and `*emit*` HostEmitter (#241). `term/read-key` and `key-pending?` consult a `KeySource` at the `*keys*` var instead of reaching for `os.Stdin` / the SharedArrayBuffer globals inline — the same decoupling move output and emit made. This is the input item of the §6 sequence in the host-decoupling design doc (#233).

### What

- **`pkg/rt/keysource.go`** — `KeySource` interface (`ReadKey() (string, error)` + `KeyPending() bool`), a default `nopKeySource`, and `resolveKeySourceVar` / `boundKeySource` (duals of `resolveIOHandleVar` / the emit resolver).
- **`*keys*` core var** (`iort.go`), root-bound per platform: a stdin+SIGWINCH `nativeKeySource` on native, a SharedArrayBuffer `HostKeySource` in the WASM bundle (`keysource_js_wasm.go`).
- **`term/read-key` + `key-pending?`** are now ctx-aware natives dispatching through the bound source, so `api.WithKeySource` and `(binding [*keys* …] …)` are honored.
- **`api.WithKeySource(ks)`** — per-Run binding on `*keys*`, the input dual of `WithStdout`. An embedder or test feeds the keys the guest reads, with no real terminal.

The native `read-key` / `key-pending?` logic moves verbatim behind the interface — the SIGWINCH wake-pipe, the `unix.Poll` loop, the EINTR retry, and the POLLHUP/EOF handling are unchanged, just relocated into `nativeKeySource`.

### Scope

`ReadKey` + `KeyPending` only. Terminal geometry (`size` / `set-size`) stays a separate `term` op — native `size` takes a handle arg and queries real winsize, which doesn't belong in a key source.

**Wake stays deferred.** Unblocking a parked `read-key` without a real key is the one open design question (a `wake()` method vs the native `BEL`-on-resize sentinel vs a SAB wake-epoch cell). It's already called out as out of scope in `lg-host.js`; this seam leaves room for it but doesn't introduce the SAB-level protocol. Holding it back keeps the input seam a clean mirror of the writer; happy to take it up next as its own change.

No `.lg` or `lg-host.js` change, so `core_compiled.lgb` and the assemble golden are untouched.

### Testing

`pkg/api/keysource_test.go` binds a scripted `KeySource` via `WithKeySource` and proves keys flow out of `(term/read-key)` in order, exhaustion surfaces as nil, and `(term/key-pending?)` tracks the queue — all without a TTY. Native `go build ./...` and `GOOS=js GOARCH=wasm go build ./pkg/rt/` are clean, as is `go vet`.

### Diffstat

```
 pkg/api/api.go              |  32 ++++-
 pkg/api/keysource_test.go   |  77 ++++++++++
 pkg/rt/iort.go              |   6 +
 pkg/rt/keysource.go         |  73 +++++++++
 pkg/rt/keysource_js_wasm.go |  69 +++++++++
 pkg/rt/term.go              | 225 ++++++++++++++++++-------------
 pkg/rt/term_wasm.go         |  61 +++-----
 7 files changed, 397 insertions(+), 146 deletions(-)
```

### Relationship to #241

Independent — both branch off `main` and neither needs the other. They do touch the same two files in different spots (`iort.go` adds `*keys*` vs `*emit*`; `api.go` adds an option), so whichever merges second will want a trivial rebase there.

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -20,6 +20,7 @@ type config struct {
 	stdout io.Writer
 	stderr io.Writer
 	emit   func(name, dataJSON string)
+	keys   rt.KeySource
 }
 
 // WithStdout configures the runtime to route output written via *out*
@@ -70,10 +71,27 @@ func WithEmit(fn func(name, dataJSON string)) Option {
 	return func(c *config) { c.emit = fn }
 }
 
+// WithKeySource routes term/read-key and key-pending? through ks for this
+// instance, the input dual of WithStdout. An embedder or test supplies the
+// keystrokes the guest reads — driving a TUI without a real terminal, or
+// feeding a deterministic script. ks is an rt.KeySource (ReadKey +
+// KeyPending); see pkg/rt/keysource.go.
+//
+// Implementation: each Run pushes ks as a dynamic binding on *keys*, popped
+// on return. Same per-Run isolation and process-global-binding-stack
+// concurrency caveat as WithStdout.
+//
+// Default: the platform key source (stdin+SIGWINCH on native, the SAB in
+// WASM). Wake — unblocking a parked read-key without a real key — is not part
+// of this seam yet; see the note in pkg/rt/wasm/lg-host.js.
+func WithKeySource(ks rt.KeySource) Option {
+	return func(c *config) { c.keys = ks }
+}
+
 // (Other options deliberately NOT exposed:
 //
-//   - WithStdin: stdin substitution is tied to the wake() / SAB protocol
-//     deferred from nooga/let-go#174 and the readline-driven REPL path.
+//   - WithStdin: line-stream stdin substitution is the *in* io.Reader dual
+//     of WithStdout, separate from WithKeySource's interactive key input.
 //     *in*'s root binding remains os.Stdin; embedders that need stdin
 //     substitution today can rebind *in* manually before calling Run.)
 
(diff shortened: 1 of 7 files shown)


nooga (approved):

mparrett:
Check failing (no `clojure` on PATH for LETGO_GOLD_REDERIVE=1):

```
Run LETGO_GOLD_REDERIVE=1 go test -run '^TestGold' -count=1 -v -timeout 120s ./test/e2e/
=== RUN   TestGoldDynvarThreadsMatchesClojure
    gold_dynvar_test.go:63: LETGO_GOLD_REDERIVE=1 but no `clojure` on PATH: exec: "clojure": executable file not found in $PATH
--- FAIL: TestGoldDynvarThreadsMatchesClojure (0.00s)
=== RUN   TestGoldNsThreadsMatchesClojure
    gold_dynvar_test.go:64: LETGO_GOLD_REDERIVE=1 but no `clojure` on PATH: exec: "clojure": executable file not found in $PATH
--- FAIL: TestGoldNsThreadsMatchesClojure (0.00s)
=== RUN   TestGoldDynvarCallbacksMatchesClojure
    gold_dynvar_test.go:65: LETGO_GOLD_REDERIVE=1 but no `clojure` on PATH: exec: "clojure": executable file not found in $PATH
--- FAIL: TestGoldDynvarCallbacksMatchesClojure (0.00s)
FAIL
FAIL	github.com/nooga/let-go/test/e2e	0.002s
FAIL
Error: Process completed with exit code 1.
```

PR #245, opened 2026-06-17
feat(wasm): client-owned shell — split lg-host.js, add -w-shell none

mparrett:
Make "the client owns the shell" a supported `lg -w` path. Today `lg -w` emits one `index.html` with let-go's xterm shell baked in; a client can only post-patch that artifact. This splits the host JS so a client can supply its own shell and bind the published `window.LetGoHost` surface, with the default xterm shell kept as the convenience. This is the endgame the decoupling doc (#233 §6) names: the generator wiring the seams, not the client patching the output.

Companion: [nooga/xsofy#78](https://github.com/nooga/xsofy/pull/78) — the real client that proves the contract. Conceptually stacks on the I/O seams (#241 emit, #244 input) but is independent code.

### The split

The host JS becomes two assets around one contract:

- **`lg-host-core.js`** — host-agnostic glue: COI bootstrap, wasm decode, the `window.LetGoHost` surface, the worker/main-thread boot. Never names xterm. Fires `LetGoHost.onReady(mode)` once the glue is wired (`mode` is `'worker'` | `'main'`), before the VM runs.
- **`lg-shell-xterm.js`** — the default xterm shell, binding output/input/size purely through `LetGoHost` in its `onReady` handler. The only asset that touches xterm.

`onReady` is the one new surface. A shell — the default, or a client's — opens its UI and binds in that callback, instead of the boot path hardcoding `term.*` calls mid-flight. Output emitted before a shell binds is buffered and replayed on the first `onOutput`, so a shell that binds late (async CDN load under `-w-shell none`) loses nothing.

### The flag

- **`-w-shell xterm`** (default) — today's bundle, byte-for-byte behavior; the default golden is regenerated because the JS is reorganized through `onReady`, but the shipped behavior is unchanged.
- **`-w-shell none`** — emits core only, no xterm shell, no CDN tags. The client provides its own `index.html` shell and binds `LetGoHost`. `coi-serviceworker.js` still ships (COI is glue, not shell).

`embed.go`'s `AssembleHTML` gains a `shell bool`; `wasm.go` threads the flag; `lg.go` validates it. A second golden pins the shell-less output, and `TestMarkersGone` covers both modes.

### Validated end-to-end

xsofy (a TUI game on let-go), built with `-w-shell none` and wearing its own shell bound only through `LetGoHost.onReady`, runs **interactively on a real iPhone** (worker mode, input live) over a header-capable HTTPS serve. The client owns presentation outright: its own embedded fonts, an audio hook, graphics rendered alongside the terminal, whatever it needs — none of it something let-go has to know about. What used to mean patching the generated shell is now just the client's own code against `LetGoHost`.

### Scope and honest edges

- **Shell ownership is presentation ownership.** A client owns the terminal, font, theme, DOM. Two adjacent concerns are *not* covered by the `LetGoHost` contract and stay where they are: passing env/config into the worker VM (xsofy reads `os/getenv` in the worker) and the COI bounded-retry for header-less hosts (GitHub Pages). Those touch the worker boot and COI lifecycle, not the shell — candidates for follow-up seams (a `setEnv`/config surface, a COI option), noted but out of scope here.
- **Wake** stays deferred (the input-seam follow-up, #244), unrelated to this.

### Delivery shapes

`-w-shell none` decouples *who owns the shell*. A second, related axis is *how the glue and the compiled wasm reach the client*. Four shapes, in increasing order of decoupling:

- **A — inline, single file (what this PR ships).** One self-contained `index.html`: the host glue plus the gzipped-base64 wasm payload baked in, default xterm dropped. The client attaches its shell by appending. Zero-config, works on any dumb static host, validated on-device. Cost: the payload parses as a JS string with no streaming compile, and it can't be cached separately — so every load re-ships and re-decodes it. (The base64's ~33% inflation washes out under gzip/brotli, so it's not a transfer cost — see the perf follow-up below.)
- **B — glue+payload as one emitted `.js`.** The client authors `index.html` from scratch and `<script src>`s it — truly owns the page, no version skew (glue and payload ship together). The payload is still embedded, so A's size and streaming costs remain.
- **C — client supplies the glue, let-go emits only the payload.** Most independent, but the client's vendored glue can drift from the payload's boot/`LetGoHost` contract — a skew risk, and it needs a stable, published glue API that doesn't exist yet.
- **D — let-go emits glue/loader and a separate `main.wasm`; the client hosts the payload wherever.** The glue stays versioned with let-go (no skew, unlike C); the wasm becomes a real artifact — separately servable, cacheable, `instantiateStreaming`-able, HTTP-compressible (no base64). Costs one new surface: a configurable payload URL (default `./main.wasm`, client-overridable). Caveat: a cross-origin payload must satisfy COEP (`Cross-Origin-Resource-Policy` or CORS), since the page is cross-origin-isolated for SharedArrayBuffer.

**Leaning: ship A as the default, treat D as the production target.** A is the right default — self-contained and zero-config is what most `-w` users want, and it's what's validated here. I benchmarked the two (perf follow-up comment below): on a server that streams (HTTP/1.1 chunked + `application/wasm`), D is faster both cold (+14–29%) and on warm/repeat load (+80–90%, roughly network-flat), with transfer size a wash. So D isn't a speed tradeoff — it's the better-performing shape for a client that owns its page and serves from real infrastructure; A's remaining edge is purely zero-config simplicity. B and C are waypoints worth skipping: B is D without pulling the payload out, and C is D with a skew risk. (A separate surfacing question — a `-w` template-path flag versus the glue-only mode this builds — is orthogonal; glue-only is the smaller change and what's here.)

---

Opening as a **draft** to demonstrate the shell-extraction path end-to-end alongside the companion client. Marked draft because it sits behind the I/O seams (#241, #244) in the #233 sequencing; the delivery-shape choice above (A now, D later) is the main thing I'd want a read on before merge.

### DIFF
--- a/lg.go
+++ b/lg.go
@@ -587,6 +587,7 @@ var compileOutput string
 var bundleOutput string
 var bundleBase string
 var wasmOutput string
+var wasmShell string
 var sourcePaths string
 var resourcePaths string
 
@@ -602,6 +603,7 @@ func init() {
 	flag.StringVar(&bundleOutput, "b", "", "bundle .lg file into a standalone executable (specify output path)")
 	flag.StringVar(&bundleBase, "bundle-base", "", "path to target-platform lg binary for cross-OS bundling (defaults to current executable)")
 	flag.StringVar(&wasmOutput, "w", "", "build .lg file into a WASM web app (specify output directory)")
+	flag.StringVar(&wasmShell, "w-shell", "xterm", "shell for -w: 'xterm' (default) or 'none' (emit core only; client supplies its own shell via window.LetGoHost)")
 	flag.StringVar(&sourcePaths, "source-paths", "",
 		"namespace search paths separated by the OS path-list separator "+
 			"(':' on Unix, ';' on Windows). When given, fully defines the search "+
@@ -833,7 +835,11 @@ func main() {
 			fmt.Fprintln(os.Stderr, "error: -w requires exactly one input file")
 			os.Exit(1)
 		}
-		if err := buildWasm(context, nsResolver, files[0], wasmOutput); err != nil {
+		if wasmShell != "xterm" && wasmShell != "none" {
+			fmt.Fprintf(os.Stderr, "error: -w-shell must be 'xterm' or 'none', got %q\n", wasmShell)
+			os.Exit(1)
+		}
+		if err := buildWasm(context, nsResolver, files[0], wasmOutput, wasmShell == "xterm"); err != nil {
 			fmt.Fprintf(os.Stderr, "error: %v\n", err)
 			os.Exit(1)
 		}
--- a/pkg/rt/wasm/assemble_test.go
+++ b/pkg/rt/wasm/assemble_test.go
@@ -25,30 +25,39 @@ func TestAssembleHTMLGolden(t *testing.T) {
 	const wasmExecJS = "// stub wasm_exec.js for the golden test\nconsole.log('exec stub');\n"
 	const wasmGzB64 = "STUBWASMBLOBB64=="
 
-	got := AssembleHTML(wasmExecJS, wasmGzB64)
-	goldenPath := filepath.Join("testdata", "assemble_golden.html")
+	// Pin both bundle shapes: the default xterm shell and the shell-less
+	// core-only build (-w-shell none).
+	for _, tc := range []struct {
+		name   string
+		shell  bool
+		golden string
+	}{
+		{"default xterm shell", true, "assemble_golden.html"},
+		{"shell-less core", false, "assemble_golden_shellless.html"},
+	} {
+		got := AssembleHTML(wasmExecJS, wasmGzB64, tc.shell)
+		goldenPath := filepath.Join("testdata", tc.golden)
 
-	if *updateGolden {
-		if err := os.MkdirAll("testdata", 0755); err != nil {
-			t.Fatalf("mkdir testdata: %v", err)
+		if *updateGolden {
+			if err := os.MkdirAll("testdata", 0755); err != nil {
+				t.Fatalf("mkdir testdata: %v", err)
+			}
+			if err := os.WriteFile(goldenPath, []byte(got), 0644); err != nil {
+				t.Fatalf("writing golden: %v", err)
+			}
+			t.Logf("golden updated: %s (%d bytes)", goldenPath, len(got))
+			continue
 		}
-		if err := os.WriteFile(goldenPath, []byte(got), 0644); err != nil {
-			t.Fatalf("writing golden: %v", err)
-		}
-		t.Logf("golden updated: %s (%d bytes)", goldenPath, len(got))
-		return
-	}
 
-	golden, err := os.ReadFile(goldenPath)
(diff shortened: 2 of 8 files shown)


mparrett:
**Perf follow-up: measured inline (A) vs separate-streaming (D) delivery**

I benchmarked the two delivery shapes from the description — metric is boot time (navigation → first VM output), headless Chromium, same program and `lg`, `-w-shell none`, both served compressed with COOP/COEP, realistic caching (immutable `main.wasm`, `no-store` HTML). Median of 5–6 runs.

On a real server (`npx serve` v14, HTTP/1.1 chunked + brotli):

| ms (median) | A cold | D cold | A warm | D warm |
|---|--:|--:|--:|--:|
| unthrottled | 1135 | 810 | 950 | 194 |
| 50 Mbps | 1889 | 1626 | 1766 | 335 |
| 20 Mbps | 3233 | 2765 | 3087 | 294 |

D wins both: cold **+14–29%**, warm **+80–90%**. Warm-D is roughly network-flat (~190–335 ms) — the immutable `main.wasm` is a cache hit, while A re-ships and re-decodes its inline payload on every load.

Two corrections to the "Delivery shapes" framing above, from the measurements:

1. **Cold-load is server-dependent.** On a naive server (Python `http.server`, HTTP/1.0, fixed `Content-Length`) D *lost* cold by 6–21% — no chunked streaming, so `instantiateStreaming` couldn't overlap compile with download. With a server that streams (chunked + correct `application/wasm`), the overlap lands and D wins. The serving posture matters as much as the shape.
2. **Transfer size is a wash** (~5.7 MB either way once compressed) — base64 inflation is undone by the outer gzip/brotli. So "fewer bytes" isn't the argument; streaming + caching is.

So the call in the description holds, sharpened: A is the zero-config default; D is faster on both axes once served correctly, which a deployed app already is. Streaming was verified to actually stream (1 fetch, no `arrayBuffer` fallback), not silently buffer.

One orthogonal note: the real cold-start lever is payload size (~24 MB), not delivery shape. A TinyGo build is ~10× smaller but currently breaks runtime on a non-trivial app, so it's gated on VM work rather than a quick win.

Happy to share the full writeup (method, the Python-vs-serve contrast, and IndexedDB compiled-`Module` caching as the next warm-path step) if it's useful.

mparrett:
> A TinyGo build is ~10× smaller but currently breaks runtime on a non-trivial app

Side note... this was a neat partial success experiment; I am eager to get back to it, but I fear it's riddled with deep challenge.

mparrett:
**Codex Review (5.5-high):**

`#245`: I didn’t find a blocking issue in the shell split itself. The stack boundary matters: two findings in #246 are introduced by the incremental external-payload commit in `#246`, not by `#245`; see findings there for more info.

I reviewed `#245` as `origin/main...794034a329f2eb393791f62618cdd9c1297a5391` and `#246` incrementally as `794034a329f2eb393791f62618cdd9c1297a5391..9eac3e627e262f75b998a745ec0e17cdc4e3969a`.

nooga (approved):

PR #246, opened 2026-06-17
feat(wasm): -w-wasm external — ship the payload as a separate streaming main.wasm

mparrett:
> Stacks on #245 (builds on its `lg-host-core` core/shell split). Opened against `main`, so until #245 merges the diff below includes #245's changes too — the payload-delivery work is the top commit (`feat(wasm): -w-wasm external …`); it rebases to a clean diff once #245 lands.

Make wasm delivery a choice, orthogonal to the shell. #245 splits *who owns the shell*; this adds the second axis from its "Delivery shapes" note — *how the compiled wasm reaches the client*:

- **`-w-wasm inline`** (default) — today's behavior: the gzip-base64 payload is baked into `index.html` and decoded in JS. Byte-for-behavior unchanged.
- **`-w-wasm external`** — emit a separate `main.wasm`; the loader fetches and stream-compiles it (`WebAssembly.compileStreaming`), with an `arrayBuffer` fallback if the host doesn't serve `application/wasm`.

The two axes are independent — `xterm`/`none` × `inline`/`external` are all valid. `xterm` + `external` (default shell, separate payload) is the combination I validated on-device.

### Why

A deployed app spends its life on repeat visits, and there the inline payload is re-shipped and re-decoded every load. A separate, immutable `main.wasm` is a cache hit. Measured (headless Chromium, boot = navigation → first VM output, served compressed under COOP/COEP, medians):

| `npx serve` v14 | A cold | external cold | A warm | external warm |
|---|--:|--:|--:|--:|
| unthrottled | 1135 | 810 | 950 | 194 |
| 50 Mbps | 1889 | 1626 | 1766 | 335 |
| 20 Mbps | 3233 | 2765 | 3087 | 294 |

External wins both — cold +14–29%, warm +80–90% (warm is roughly network-flat; the payload never re-crosses the wire). Two caveats: cold-load only wins on a server that *streams* (HTTP/1.1 chunked + correct MIME) — on a server that buffers, inline is faster; and **transfer size is a wash** once both are compressed (base64 inflation is undone by gzip/brotli). The win is streaming overlap plus caching, not bytes. Validated interactively on a real iPhone over an HTTPS serve.

### How

- `lg-host-core.js` carries both load paths behind an injected `WASM_MODE`. Inline keeps the `atob`+`DecompressionStream` decode; external fetches, and — to overlap download+compile with worker spin-up — `compileStreaming`s on the main thread and posts the `WebAssembly.Module` to the worker (structured-cloneable), which only instantiates. The worker is a Blob worker, so the payload URL is resolved absolute (a relative fetch would resolve against `blob:`).
- `AssembleHTML` gains an `externalWasm` param; `wasm.go` writes `main.wasm` only in external mode; `lg.go` adds + validates `-w-wasm`.
- Goldens pin all four shapes (shell × payload); `TestMarkersGone` covers the new `__WASM_MODE__` marker. Full suite green except the pre-existing `genmanifest` staleness (nooga#41), which this doesn't touch.

### Scope / honest edges

- The payload loads from a sibling `./main.wasm`. A client-overridable URL (to host the wasm on a CDN) is the obvious next surface — left out here because the naming (`LetGoHost.wasmUrl` vs a `data-` attr) is worth settling first.
- A cross-origin payload must satisfy COEP — `Cross-Origin-Resource-Policy: cross-origin` or CORS — since the page is cross-origin-isolated for SharedArrayBuffer. Same-origin needs nothing.
- Draft because it stacks on #245 and the seams (#241, #244); the override surface above is the main open question.

### DIFF
--- a/lg.go
+++ b/lg.go
@@ -294,6 +294,7 @@ var bundleOutput string
 var bundleBase string
 var wasmOutput string
 var wasmShell string
+var wasmPayload string
 var sourcePaths string
 var resourcePaths string
 
@@ -310,6 +311,7 @@ func init() {
 	flag.StringVar(&bundleBase, "bundle-base", "", "path to target-platform lg binary for cross-OS bundling (defaults to current executable)")
 	flag.StringVar(&wasmOutput, "w", "", "build .lg file into a WASM web app (specify output directory)")
 	flag.StringVar(&wasmShell, "w-shell", "xterm", "shell for -w: 'xterm' (default) or 'none' (emit core only; client supplies its own shell via window.LetGoHost)")
+	flag.StringVar(&wasmPayload, "w-wasm", "inline", "wasm delivery for -w: 'inline' (default; gzip-base64 baked into index.html) or 'external' (emit a separate main.wasm the loader fetches + streams)")
 	flag.StringVar(&sourcePaths, "source-paths", "",
 		"namespace search paths separated by the OS path-list separator "+
 			"(':' on Unix, ';' on Windows). When given, fully defines the search "+
@@ -545,7 +547,11 @@ func main() {
 			fmt.Fprintf(os.Stderr, "error: -w-shell must be 'xterm' or 'none', got %q\n", wasmShell)
 			os.Exit(1)
 		}
-		if err := buildWasm(context, nsResolver, files[0], wasmOutput, wasmShell == "xterm"); err != nil {
+		if wasmPayload != "inline" && wasmPayload != "external" {
+			fmt.Fprintf(os.Stderr, "error: -w-wasm must be 'inline' or 'external', got %q\n", wasmPayload)
+			os.Exit(1)
+		}
+		if err := buildWasm(context, nsResolver, files[0], wasmOutput, wasmShell == "xterm", wasmPayload == "external"); err != nil {
 			fmt.Fprintf(os.Stderr, "error: %v\n", err)
 			os.Exit(1)
 		}
--- a/pkg/rt/wasm/assemble_test.go
+++ b/pkg/rt/wasm/assemble_test.go
@@ -25,17 +25,20 @@ func TestAssembleHTMLGolden(t *testing.T) {
 	const wasmExecJS = "// stub wasm_exec.js for the golden test\nconsole.log('exec stub');\n"
 	const wasmGzB64 = "STUBWASMBLOBB64=="
 
-	// Pin both bundle shapes: the default xterm shell and the shell-less
-	// core-only build (-w-shell none).
+	// Pin every bundle shape across both axes: shell (xterm | none) and
+	// payload delivery (inline | external).
 	for _, tc := range []struct {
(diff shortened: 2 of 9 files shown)


mparrett:
**Codex Findings (5.5-high)**

`#246`:

1. [pkg/rt/wasm/lg-host-core.js#L137](https://github.com/nooga/let-go/blob/9eac3e627e262f75b998a745ec0e17cdc4e3969a/pkg/rt/wasm/lg-host-core.js#L137) can expose input before the worker is actually started. In external mode `_ready('worker')` opens/focuses the shell, then the function waits for `main.wasm` to fetch/compile before posting `init` to the worker at [L240-L244](https://github.com/nooga/let-go/blob/9eac3e627e262f75b998a745ec0e17cdc4e3969a/pkg/rt/wasm/lg-host-core.js#L240-L244). If the user types while that fetch is in flight, the first key fills the SAB slot and a second key hits the busy-wait at [L121](https://github.com/nooga/let-go/blob/9eac3e627e262f75b998a745ec0e17cdc4e3969a/pkg/rt/wasm/lg-host-core.js#L121), but no worker is running yet to clear it. That can lock the main thread. I’d either delay `onReady`/input hookup until after `worker.postMessage(initMsg)`, or make `_lgKey` nonblocking when the slot is occupied.

2. [pkg/rt/wasm/lg-host-core.js#L331](https://github.com/nooga/let-go/blob/9eac3e627e262f75b998a745ec0e17cdc4e3969a/pkg/rt/wasm/lg-host-core.js#L331) drops the promise returned by `startWorkerMode()`. Since `#246` made `startWorkerMode` async, external fetch/compile failures after the first `await` bypass the surrounding `try/catch`; the page stays at “fetching wasm...” with an unhandled rejection instead of surfacing `error: ...`. `await startWorkerMode()` would at least route loader failures through the existing status/error path.

mparrett:
Thanks — both findings addressed in 5271ebb.

1. **Input before the worker exists (L137):** gated `_lgKey` behind a `workerReady` flag set right after `postMessage(initMsg)`. Keys typed during the external fetch/compile are now dropped rather than busy-waiting the main thread with no consumer to drain the slot. The shell still renders early via `onReady` (output buffers and replays), so external mode keeps its fast first paint.
2. **Dropped loader promise (L331):** `await startWorkerMode()`, so external fetch/compile rejections surface through the entry `try/catch` as `error: …` instead of stranding the page at "fetching wasm...".

Follow-up, not this PR: the producer busy-wait in `_lgKey` (L121) can still spin the main thread in the narrow window after the worker starts but before the VM first reads a key — pre-existing in inline mode, just widened by external's load. The real fix is a non-blocking producer (or the SPSC ring buffer); that's input-seam territory (#244), so I left it as a noted follow-up rather than fold it in here.

Re-verified both modes boot and stay interactive (headless, worker mode) after the change.

nooga (approved):

PR #247, opened 2026-06-17
fix(ir): prefix closure-local arg names to prevent captured-name shadowing miscompile

nnunley:
## Problem

An inline closure (`lower-fn-lit`) lowers its parameters as `arg0/arg1/…`, tail-call mutable locals as `a0/a1/…`, and its rest slot as `args0` — the **same** naming scheme as any other lowered fn. But a closure's captured values (`:load-closed`) are spliced into the body as the **enclosing** scope's identifiers (e.g. `arg0`). So a closure parameter `arg0` *shadows* a captured `arg0`, and the closure reads its own parameter where it meant the capture.

Concretely, `(map (fn [bid] (write-block f bid)) …)` lowered to `write_block(ec, arg0, arg0)` — passing `bid` for `f`. This is the "captured name shadowed by block param" miscompile.

## Fix

Lower each **capturing** closure body under a per-closure-site prefix derived from its make-closure nid (`c<nid>_`), threaded through `lower-fn-lit` / `lower-template-closure-expr` and applied via two helpers (`go-arg-name`, `go-args-name`) at every param/local/rest naming site. This keeps the closure-local and captured namespaces disjoint; unique per site so nested closures don't collide either.

Capture-free closures and top-level fns leave the prefix `nil`, so their emitted Go is **byte-for-byte unchanged** — the change only touches closures that actually capture.

Pure `.lg` lowering change; no Go-runtime change.

## Context

This is the foundational slice of the larger gogen self-AOT lowering work. It is a prerequisite for the per-gap lowering fixes (the gaps from the now-closed #239): those reshape `ir.lower-go`'s own IR in a way that triggers this latent miscompile, so the self-AOT lowered tree won't compile under `-tags gogen_ir` until this lands first.

## Verification

- `make check-generated` green — `core_compiled.lgb` in lockstep with sources, and `core_go_lowered/` compiles + dispatches natively under `-tags gogen_ir`.
- `go test ./test` passes.


mparrett:
**Codex (5.5-high) found one correctness issue.**

- **[P1] Nested capturing closures can still collide** in [pkg/rt/core/ir/lower_go.lg](/Users/matt/projects-new/3p/let-go/pkg/rt/core/ir/lower_go.lg:2390). The new prefix is only `c<cid>_`, but `cid` is an IR nid from the current function, not globally unique across nested function templates. If an outer capturing closure is lowered with `c7_arg0`, and an inner capturing closure inside it also has closure-site nid `7`, the inner closure’s param becomes `c7_arg0` too and can shadow the captured outer `c7_arg0`, recreating the same miscompile for nested closures. The prefix should include the current lexical prefix, e.g. derive nested prefixes like `c_outer_c_inner_`, rather than replacing `*closure-arg-prefix*`.

Full verification suite was not run; this was a patch/code-path review against `origin/main...origin/pr/247`.

---

Focused `pkg/ir/lisp_lower_go_test.go` regression test, not a broad suite case:

```go
func TestLowerGoNestedCapturedClosurePrefixesAreLexical(t *testing.T) {
	ensureLoader()

	fn := buildLispIR(t, `(defn nested-captures [x]
		(fn* [y]
			(fn* [z]
				[x y z])))`)
	optimizeLispIR(t, fn)
	result := lowerGo(t, fn, ":strict")

	if got := result.ValueAt(vm.Keyword("status")); got != vm.Keyword("lowered") {
		t.Fatalf("expected :lowered status, got %v (reason: %v)", got, result.ValueAt(vm.Keyword("reason")))
	}

	rendered := bindAndRenderGoDecl(t, result)

	re := regexp.MustCompile(`func\((c[0-9_]+arg0) vm\.Value\) vm\.Value \{\s*return rt\.BoxNativeFn\(func\((c[0-9_]+arg0) vm\.Value\) vm\.Value`)
	m := re.FindStringSubmatch(rendered)
	if m == nil {
		t.Fatalf("expected nested captured closures with prefixed params\n--- go ---\n%s", rendered)
	}
	if m[1] == m[2] {
		t.Fatalf("nested capturing closure reused outer closure arg prefix %q; inner param can shadow captured outer param\n--- go ---\n%s", m[1], rendered)
	}
}
```

Run it with:

```sh
go test ./pkg/ir -run TestLowerGoNestedCapturedClosurePrefixesAreLexical -count=1 -v
```

The smaller source repro behind it is:

```clojure
(defn nested-captures [x]
  (fn* [y]
    (fn* [z]
      [x y z])))
```

This targets the exact failure mode: a capturing closure inside another capturing closure. With the current PR shape, both closure sites can derive the same `c<nid>_` prefix because `nid` is only unique within a function IR, so the inner param can still shadow the captured outer param.

nnunley:
Fixed in c6d5b084.

`lower-fn-lit` now extends the enclosing prefix instead of replacing it, so `*closure-arg-prefix*` is lexically cumulative: a capturing closure nested in a capturing closure gets `c<outer>_c<inner>_…`. That's unique by construction across nesting depth, so it no longer matters that `cid` (an IR nid) is only unique within one function template.

One note on the repro: your exact fixture happens not to collide on current head — the two templates get nids 3 and 5, lowering to `c3_arg0` / `c5_arg0` (distinct by luck of nid assignment), and the supplied regex also assumed both inner/outer closures share a `vm.Value) vm.Value` signature (the inner here returns `(vm.Value, error)`). The underlying concern is still right, though: flat per-nid prefixes make nesting-safety accidental, not guaranteed. So I added a regression that asserts the structural invariant directly — the inner closure's prefix must lexically extend the outer's — which fails on the old flat scheme (`c3_` vs `c5_`: not a prefix relationship) and passes on the cumulative one (`c3_` -> `c3_c5_`):

`go test ./pkg/ir -run TestLowerGoNestedCapturedClosurePrefixesAreLexical -count=1 -v`

`make check-generated` green (lockstep + native dispatch); `go test ./test ./pkg/ir` passes.

mparrett:
Ran the regeneration to completion on a clean checkout of `b170ad1`, which is the piece that settles the open question. The staleness is real and deterministic, but the fix is one line.

**The staleness is genuine, not an environment artifact.** `generated.sums` records a SHA256 over source files only: every `.lg` under `pkg/rt/core` plus every `.go` under `cmd/lgbgen` (`genmanifest.Compute`). No codegen output, Go version, or mtime feeds into it, and `TestComputeIsDeterministic` guards that. On a clean worktree at `b170ad1` with no local modifications:

```
recorded: fb55e8aa8fcf6ccacae0449fb8af0dc6696b5601b95cee1477287290199a2f5d
computed: 6d9cfb5e6091a411bab8bfb884a875aa8f8210635dd01585af8f03276aac088f
```

Because the digest is pure source hashing, every machine computes the same value, so it reproduces regardless of sandbox.

👍🏻 **Only the digest is stale; the artifacts are correct.** I ran `go run -tags bootstrap ./cmd/lgbgen --target=both` to completion (exit 0). The only file it changed:

```
 M pkg/rt/generated.sums   # fb55e8... -> 6d9cfb...
```

`core_compiled.lgb` and the whole `core_go_lowered/` tree were already in sync with the sources; regen left them byte-identical. So the recorded digest is just an intermediate value, as if a source (`lower_go.lg`) was edited after the last regen and the digest wasn't refreshed before commit.

**lgbgen isn't hung, it's slow.** `--target=both` took about 8 minutes here, almost all of it in the silent `--target=go` lowering phase. The `WARNING: not lowered (:fallback) ... gogen: expected seq` lines are the pre-existing fallback gaps, not new and not fatal.

**Fix:** rerun `make generate`, let it finish (~8 min), and commit; only `generated.sums` moves. Or set the digest directly to `6d9cfb5e6091a411bab8bfb884a875aa8f8210635dd01585af8f03276aac088f`.

mparrett (approved):
Changes look good. Manifest refresh would unblock a green merge.

mparrett:
Correction to my earlier comment: I called this a CI blocker, and it isn't. CI on the merge (`e2aeace`) is green, `generated-artifacts` included.

Two reasons the stale digest doesn't surface in CI:

1. The `test` job runs `go generate ./pkg/rt/` before `go test ./...`, and any lgbgen run recomputes and rewrites `generated.sums` (`cmd/lgbgen/main.go`). So the freshness test sees a fresh digest no matter what's committed.
2. `make check-generated` doesn't read `generated.sums` at all — it checks `core_compiled.lgb` byte-lockstep and `-tags gogen_ir` dispatch. Both pass, because the actual codegen isn't drifted.

So the committed digest is cosmetic to CI, and `main` was already carrying a stale one before this PR (`0a87097`: recorded `82525d…`, sources hash to `6e4e04…`). The only real symptom is a false-positive `pkg/genmanifest` failure for anyone running `go test ./...` locally without regenerating first.

Opening a one-line follow-up to refresh the digest so local runs are clean. Separately worth a look: the freshness test can't gate drift in CI given the regen-first ordering — `make check-generated`'s lockstep check is the thing actually catching drift.

mparrett:
@nnunley correction to my staleness note above. I said `generated.sums` is "pure source hashing ... every machine computes the same value." That holds only on a clean checkout.

`genmanifest.Compute` hashes every `.go` under `cmd/lgbgen`, which includes `cmd/lgbgen/main_gogen_ir.go` — the gitignored gogen_ir wireup `lgbgen` emits. That's a build artifact, not a source; the Makefile already classifies the gogen_ir wireup as "NOT committed ... a build artifact." It's absent on a clean checkout and present only after a generation run, so the digest depends on local build state:

- post-generate (wireup on disk): `c5e0c94f...`
- clean / pre-generation: `f82ff7e4...`

#252 caught this: it adds a non-mutating gate that checks the manifest *before* regeneration. The committed sum was recorded post-generate, so the pre-regen gate computed the clean digest, disagreed, and went red.

Fix in #252: skip files carrying the standard `// Code generated ... DO NOT EDIT.` marker so the digest is sources-only and invariant to whether the wireup is on disk, then refresh the sum to the clean value. The gate passes either way now.

PR #249, opened 2026-06-18
feat(ir): native-Go lowering for set!/binding/var/quot + self-AOT diff harness; fix cluster-A engine divergence

### LINKED ISSUE #263: gogen_ir: close ir.structurize fallbacks for the 18 func-decl-bail defns
**Bucket: 18/749** — the largest remaining gogen_ir lowering failure. Sub-issue of #262.

## Root cause: `ir.structurize` fallbacks, not try/catch

`lower-body` is **structured-first** (`lower_go.lg:2241`): it runs `ir.structurize/structurize`, and when `:fallbacks` is 0 emits a control tree via `lower-tree` (if/for/break/continue); otherwise it drops to the goto+label emitter. These 18 defns are where **both paths bottom out** — `structurize` reports `:fallbacks > 0` (a CFG it can't reduce) **and** the goto emitter then hits a block it can't lower, so `body-stmts` is `nil` and the fn bails.

In `lower-go` (full func-decl emission) that bail surfaces as the opaque `#error {:message "gogen/func-decl: body: gogen: expected seq, got let-go.lang.ExceptionInfo", :data ...}` — `:data` truncated by the harness bucketer. In `trace` mode the same defns report `:lower-status :fallback` gracefully (same root, different surface). So this is the `ir.structurize` coverage frontier — its docstring goal is "`:fallbacks` ratchets toward zero on the real corpus" — **not** the try/catch-value gap (that is the single `:set-var` case under #260).

## The work

Drive `ir.structurize` (+ the residual goto block-lowering) so these 18 reach `:status :lowered`:

- `core.lg::replace`
- `ir/build.lg::expand-fn-args`
- `ir/build.lg::expand-vector-pattern`
- `ir/dominance.lg::dfs-postorder`
- `ir/dominance.lg::dominators`
- `ir/lattice.lg::normalize-type`
- `ir/lower_go.lg::distinct-imports`
- `ir/lower_go.lg::emit-assignments-for-target`
- `ir/passes/cse.lg::cse`
- `ir/passes/dce.lg::one-pass`
- `ir/passes/licm.lg::reachable-without`
- `ir/passes/licm.lg::thread-hoisted-through-header!`
- `ir/passes/typeinfer.lg::div-op-type`
- `ir/passes/typeinfer.lg::numeric-op-type`
- `ir/passes/typeinfer.lg::quot-op-type`
- `ir/structurize.lg::g-idoms`
- `ir/structurize.lg::g-postorder`
- `ir/validate.lg::check-no-cross-block-refs!`

**Subtask 0 (diagnostics):** de-truncate the inner `:data` (or fall through to the trace-mode `:lower-status :fallback` surface) so each defn names *which* structurize fallback shape it hits, instead of one opaque signature. This splits the 18 into the actual CFG patterns to teach `structurize`.

## Verification

Per-defn, already a standing count — no new harness wiring:
```
LG_STRESS_PASSES=1 LG_STRESS_TIMEOUT_MS=8000 LG_STRESS_LOG=/tmp/s.log \
/tmp/lg-stress scripts/ir-stress.lg lower-go pkg/rt/core <full corpus incl. ir/lattice.lg>
```
(`ir-stress.lg` records `:lower-status :lowered|:fallback|:throw` per defn at L296/L400.) A defn is **done** when it leaves this bucket and `Passed` rises; the bucket is closed at **0/749**. Baseline at `main` `d4c36cf2`: 18.

> Note: counts are vs merged main; #249 (in-flight) already lowers part of this surface.

### LINKED ISSUE #264: gogen_ir: 6 defns lower to 'unsupported function body shape' (invisible-nil)
**Bucket: 6/749** — `:other | unsupported function body shape`. Sub-issue of #262. (local `lg-fa99`)

`lower_go.lg` hits an "invisible-nil" body shape it can't lower. Defns: `ir/lower_go.lg::lower-fn-lit`, `ir/passes/pipeline.lg::{compile-form,protocol-method-sigs}`, `test.lg::{clear-registered-tests!,register-test!,use-fixtures}`.

Done = bucket → 0 in the #262 repro.

### LINKED ISSUE #265: gogen_ir: multi-arity partial fallback (4 defns)
**Bucket: 4/749** — `:multi-arity-partial-fallback`. Sub-issue of #262. (local `lg-q2cf`)

Defns: `core.lg::dorun`, `ir/lower_go.lg::lower-body`, `ir/passes/infer_arg_types.lg::infer-arg-types`, `ir/passes/typeinfer.lg::typeinfer`.

Done = bucket → 0 in the #262 repro.

### LINKED ISSUE #267: gogen_ir: resolve Go-method interop in lowering (.WriteString, .UnixMilli)
**Bucket: 3/749** — Go-method interop in lowering. Sub-issue of #262. (local `lg-izk9`)

`ir/build: unresolved symbol` for host-method calls: `.UnixMilli` @ `core.lg::inst-ms`; `.WriteString` @ `core.lg::pr-sequential` and `hash.lg::hasher-map`. Lowering needs to resolve/emit Go method-interop call sites.

Done = bucket → 0 in the #262 repro.

nnunley:
## Summary

Extends native-Go lowering (`-tags gogen_ir`) to `set!`/`binding`/`var`/`quot`, adds a differential self-AOT harness, and fixes a class of bytecode-vs-AOT engine-divergence bugs (`pkg/ir` "conj expected Collection", 13 → 0).

### Native lowering + infrastructure
- Native-Go lowering for `set!`, `binding`, `var`, `quot`
- Closure-arg naming shadow fix (per-closure `c<nid>_` prefix)
- Go method-interop rewrite
- Full Clojure `CHAR_MAP` munge table
- Dominance dead-block guard; `EqValue` for `=`; `for`-loop gensym fix
- Differential self-AOT harness (`test/e2e/gogen_aot_diff_test.go`): runs each `test/gold-aot/*.lg` fixture under both engines (bytecode VM vs `gogen_ir`) and diffs output, gated by a shrink-only xfail allowlist

### Lowering-correctness fix (cluster-A: `conj expected Collection` 13 → 0)
- `build.lg`: reference defined-once "cheap-load" captures by InstId instead of threading them as block-args; genuine merges still use block-args
- `validate.lg` / `lower.lg`: permit function-scoped InstId refs
- `lower_go.lg`: box re-emitted cheap-load values to `vm.Value`
- `native_func.go`: `theNativeFnType.Box` passes `vm.NIL` (not a Go nil interface) when `NIL` binds to an interface param — fixes `IsTruthy` treating Go-nil as truthy, which had broken `(or s [])` and made `conj` reject the value

### Verification
- `pkg/ir -tags gogen_ir`: cluster-A `conj expected Collection` → **0**
- `make check-generated`: green (bundle lockstep + native dispatch)
- untagged `go test ./...`: no failures
- `test/e2e`: green

---

## Issues addressed (Epic #258 — Lowering & cross-target codegen)

Part of **#258**. The self-AOT diff harness + native lowering + the cluster-A fix take the `ir-stress lower-go` corpus from **40 failures on `main`** to **5 `:stress/timeout`** on this branch (per-defn evidence below).

**Advances #262** (Close gogen_ir self-AOT lowering gaps). Closes its bucket sub-issues:
- Closes #263 — `ir.structurize` func-decl-bail fallbacks (18 → 0; `dominators`, `cse`, `numeric-op-type`, `g-idoms`, `normalize-type`, … all now `:ok`)
- Closes #264 — unsupported function body shape (6 → 0)
- Closes #265 — multi-arity partial fallback (4 → 0)
- Closes #267 — Go-method interop in lowering (3 → 0; `inst-ms`/`.UnixMilli`, `pr-sequential` + `hasher-map`/`.WriteString` now `:ok`)

Partially addresses #266 — unresolved-dep symbols cleared; the 5 remaining failures are `:stress/timeout` (`for-emit`, `mask->type`, `inst-rhs`, `try-assign-stmts`, `thread-hoisted-through-header!`).

**Related:**
- #260 — this PR's `set!`-target LoadVar snapshot is the var-specific materialisation; #260 tracks the *unconditional* version (gated on the DCE liveness fix).
- #274 — the cluster-A change references defined-once captures by `InstId` instead of threading them as block-args; that is the Level-1 work the flat index-RPN bytecode (Level 2) builds on.

> Bucket counts are from the branch `ir-stress lower-go` run; worth a fresh confirmation after merge before the sub-issues are auto-closed.


mparrett:
**Codex (5.5-high), found two issues.**

`[P1] LoadVar is not safe to treat as a function-scoped cheap value`  
[pkg/rt/core/ir/build.lg](/private/tmp/let-go-pr249/pkg/rt/core/ir/build.lg:722) excludes every `ir/op-cheap-load?` capture from block-arg threading, and `op-cheap-load?` includes `LoadVar`, even though [pkg/ir/op_generated.go](/private/tmp/let-go-pr249/pkg/ir/op_generated.go:33) documents Var roots as mutable. In `lower_go`, a later use of that local is re-emitted as a fresh `ec.Deref(rt.LookupVar(...))` at [pkg/rt/core/ir/lower_go.lg](/private/tmp/let-go-pr249/pkg/rt/core/ir/lower_go.lg:598). That changes semantics for forms like:

```clojure
(let [x *v*]
  (if p (set! *v* 2) nil)
  x)
```

Correct behavior returns the pre-branch local `x`; generated Go can re-read `*v*` after the `set!`. My temporary render repro produced a function that returned `ec.Deref(rt.LookupVar(...))` after the `SetBinding` block, with no local preserving the original value. `LoadVar` needs to stay threaded unless a stability analysis proves that specific Var cannot be mutated across the region.

`[P2] gogen-diff misses multi-line output divergences`  
[test/e2e/gogen_aot_diff_test.go](/private/tmp/let-go-pr249/test/e2e/gogen_aot_diff_test.go:72) compares only `lastNonEmptyLine(stdout)`. A fixture can print incorrect earlier output and still pass if the final line matches. Since this is intended as a cross-engine execution gate, it should compare normalized full stdout, or explicitly require single-output-line fixtures and fail otherwise.

**Verification**

`go test ./pkg/ir -run 'TestLowerGo' -count=1` passed.

I also ran a temporary lower-go render repro confirming the `LoadVar` re-read shape above. `make gogen-diff` did not complete: it stayed in `lgbgen --target=go` during `lowered` with no output, so I killed the stuck process rather than leaving it running.

---

The most useful extra snippet is a focused regression test for the `LoadVar` re-read issue. This can go into `pkg/ir/lisp_lower_go_test.go`:

```go
func TestLowerGoDoesNotReemitLoadVarAcrossSetVar(t *testing.T) {
	ensureLoader()
	runLispExpr(t, `(def ^:dynamic *lv-review* 1)`)

	fn := buildLispIR(t, `(defn lv-review []
	  (let [x *lv-review*]
	    (if true
	      (set! *lv-review* 2)
	      nil)
	    x))`)
	optimizeLispIR(t, fn)

	result := lowerGo(t, fn, ":bridge")
	if got := result.ValueAt(vm.Keyword("status")); got != vm.Keyword("lowered") {
		t.Fatalf("expected :lowered, got %v reason=%v", got, result.ValueAt(vm.Keyword("reason")))
	}

	rendered := bindAndRenderGoDecl(t, result)
	if strings.Contains(rendered, "return ec.Deref(rt.LookupVar") {
		t.Fatalf("LoadVar local was re-read after set!, losing the pre-branch value:\n%s", rendered)
	}
}
```

The rendered bad shape I saw was essentially:

```go
func lv_review(ec *vm.ExecContext) vm.Value {
	if true {
		if !ec.SetBinding(rt.LookupVar("core", "*lv-review*"), vm.Int(2)) {
			rt.LookupVar("core", "*lv-review*").SetRoot(vm.Int(2))
		}
	} else {
	}
	return ec.Deref(rt.LookupVar("clojure.core", "*lv-review*"))
}
```

For the diff harness issue, a small unit test around the comparison helper would be ideal, but the current code has no helper for “compare fixture outputs.” I’d first extract normalization/comparison, then pin this case:

```go
func outputsEqual(a, b string) bool {
	return normalizeOutput(a) == normalizeOutput(b)
}

func normalizeOutput(s string) string {
	lines := strings.Split(strings.ReplaceAll(s, "\r\n", "\n"), "\n")
	for len(lines) > 0 && strings.TrimSpace(lines[len(lines)-1]) == "" {
		lines = lines[:len(lines)-1]
	}
	return strings.Join(lines, "\n")
}

func TestOutputsEqualChecksAllLines(t *testing.T) {
	if outputsEqual("wrong\nsame\n", "right\nsame\n") {
		t.Fatalf("must not accept outputs that only share the final line")
	}
}
```

mparrett:
I ran an independent review pass over this and landed on the same two issues Codex flagged (the `LoadVar` re-read and the diff harness comparing only the last line), so consider those corroborated. Four more below, ordered must-fix first. Line refs are against `1483ef0`.

## Confirms Codex's P1: `LoadVar` re-emit across `set!`/`binding`

Reproduced the same bad codegen independently. `OpLoadVar` is in `isCheapMaterializeOp` (`pkg/ir/op_generated.go:134`), so `validate.lg:102` / `build.lg:722` exempt it from block-arg threading and `lower_go.lg:598` re-emits `ec.Deref(rt.LookupVar(...))` at each use site. With `set!`/`binding` now lowering natively, a read consumed after a mutation reads the new value in Go but the old value in the VM:

```clojure
(defn f [] (let [old *flag*] (set! *flag* true) old))   ; VM → old, gogen_ir → true
```

Var roots are mutable (your own comment at `op_generated.go:33` says so), so `LoadVar` doesn't belong in the cheap-load set next to `const`/`load-arg`/`load-closed`: those are immutable for the function's extent, `LoadVar` isn't. Keep it threaded unless a stability analysis proves the specific var can't be mutated across the region. Codex's `TestLowerGoDoesNotReemitLoadVarAcrossSetVar` is a good gate for this.

While confirming it I noticed the namespace spelling is asymmetric — the re-read resolves to `clojure.core` but the `set!` resolves to `core` for the same var:

```go
if !ec.SetBinding(rt.LookupVar("core", "*flag*"), vm.Boolean(true)) { ... }
return ec.Deref(rt.LookupVar("clojure.core", "*flag*"))
```

They land on the same var so it doesn't change the above, but the asymmetric spelling looks like a latent aliasing bug in the lowerer worth a glance.

## `NIL` to a non-`vm.Value` interface param fails the call

`pkg/vm/native_func.go:63` — the new NIL branch assigns `reflect.ValueOf(NIL)` for any interface kind with no `AssignableTo`/`CanConvert` guard, where the non-NIL branch right above it has one. `NIL` is a typed `*vm.Nil`, which satisfies only the `vm.Value`-shaped interface, not `error` / `io.Reader` / etc. Boxed Go-method reflection (`boxed.go:113`, reachable from interop dispatch at `lang.go:3330`) routes arbitrary method signatures through this loop, so calling a method that takes one of those interfaces with `nil` hits `reflect: Call using *vm.Nil as type error`. It's recovered into a `GoPanicError` rather than crashing, but the call fails where it used to succeed.

The fix is to gate the NIL branch the same way the non-NIL one is gated:

```go
if args[i] == NIL && in.Kind() == reflect.Interface {
    nilVal := reflect.ValueOf(NIL)
    if nilVal.Type().AssignableTo(in) {
        rawArgs[i] = nilVal          // in is vm.Value (or a super-interface)
    } else {
        rawArgs[i] = reflect.Zero(in) // genuine nil interface for error/io.Reader/…
    }
}
```

A regression test would box a Go func taking `error` (or `io.Reader`) and invoke it through the VM with `nil` — it should return cleanly, not surface a `GoPanicError`.

## `captures-of` wasn't generalized alongside `free-vars`

This PR taught `free-vars` to recurse into map and seq literals (`build.lg:361`) but left `captures-of` (`build.lg:163`) falling through to `:else #{}` for map literals. `captures-of` is what threads loop-header block-args (`build-loop`, `build.lg:838`), so a value captured only inside a loop-body map literal isn't threaded, and the loop header ends up with a cross-block direct ref to an InstId that isn't in scope:

```clojure
(loop [acc 0 i 0]
  (let [acc2 (+ acc (get {:k captured} :k))]   ; `captured` from outside the loop
    (if (< i 2) (recur acc2 (inc i)) acc2)))
```

It's latent today because the cases I could build resolved to cheap/dominating values the lowerer re-materializes, but a non-cheap value in that position would miscompile the same way the cluster-A bugs did. The two walkers need the same container coverage: give `captures-of` the map/seq arms `free-vars` got. (The `(. obj method arg…)` trailing-arg variant is unreachable, since `build-dot` rejects multi-element dot forms, so only the literal arms matter.)

## Confirms Codex's P2, plus: typeinfer worklist drain went O(1) → O(N²)

On the harness comparison — agree with normalizing full stdout or rejecting multi-line fixtures; `quot.lg` and `var.lg` pack several asserts into one `println`, so they under-cover the paths they look like they test.

Separately, the typeinfer rewrite swapped the worklist's `peek`/`pop` for `(first queue)` + `(subvec queue 1)` (`typeinfer.lg:442`). let-go's `subvec` is a full make+copy (`lang.go:6049`/`6064`), not an O(1) view, so draining N entries one at a time is O(N²) per pass, and typeinfer is the dominant per-compile cost, run twice per `optimize-fn`. The lattice converges order-independently, so the FIFO buys nothing over the old LIFO. Same loop also reads `System/nanoTime` every iteration; the pre-PR code amortized that 128:1 with a tick counter. Restoring `peek`/`pop` (or a head index) and the tick throttle gets both back.

## Smaller stuff

- `pipeline.lg:143` / `:182` — deftype methods are built and run through the full `optimize-fn` pipeline twice (once for `:needs-error?`, once to lower). Build once, cache, feed both.
- `pkg/vm/record_base.go:59` — `recordType` duplicates the existing `vm.RecordType` (`record.go:21`); a native record's `Type()` also won't be `==`-identical to the defrecord `RecordType` of the same name. `NewRecordType` already does this.
- `resolveEmitterVar` / `resolveKeySourceVar` (`emitter.go:40`, `keysource.go:47`) are verbatim copies of `resolveIOHandleVar` (`iort.go:352`) bar the final assertion — one generic helper covers all three.
- `lower_go.lg:78` `go-type-spec` evaluates `dtype-go-type-spec` twice and leaves a dead `[:dtype]` clause at line 98.
- `lower_go.lg:310` — the CHAR_MAP comment claims verbatim glojure parity but omits `.`→`_DOT_` and a few others; safe via the `_U%04X_` fallback, but the comment overstates it.

The two I'd want fixed before merge are the `LoadVar` threading and the `subvec` drain; the NIL guard and `captures-of` gap are real but narrower. The empty xfail ratchet and the op-table-driven cheap/merge split are the right shape for this.

nnunley:
Addressed the review — thanks, these were good catches. Pushed:

- **P1 (LoadVar threading):** `:load-var` excluded from the cheap-load direct-ref at all sites — LoadVar stays threaded (var roots are mutable). Added differential regression fixture `test/gold-aot/setvar_load_stability.lg`.
- **P2 (harness):** the gogen-diff harness now compares **full normalized stdout**, not just the last line.
- **NIL→interface guard:** `theNativeFnType.Box` now gates the NIL branch with `AssignableTo` — `vm.NIL` only for `vm.Value`-assignable interfaces, `reflect.Zero` (genuine nil) for `error`/`io.Reader`/etc.
- **`captures-of` gap:** now walks map literals (mirrors `free-vars`), so loop-header threading doesn't miss a value captured only inside a map literal.
- **`subvec` drain:** restored `peek`/`pop` (O(1)) + the 128:1 `nanoTime` tick throttle — worklist drain back to O(N).
- **pipeline double-optimize:** deftype methods now built+optimized once via a per-compile cache, fed to both `:needs-error?` and lowering.
- Rebased onto **#248**, which clears the `gold-differential` red (that was the runner missing `clojure`, not this PR).

Two small corrections:
- **`go-type-spec` `[:dtype]` clause isn't dead** — `dtype-go-type-spec` only matches `(= 2 (count t))`, so the later clause is the live fallback for `[:dtype …]` forms with count ≠ 2.
- **`core` vs `clojure.core` is semantic, not cosmetic** — `rt.LookupVar` resolves the exact ns; native core vars register under `clojure.core`, lowered-core siblings under `core` (runtime tries both). Forcing one canonical breaks resolution (an attempt did, caught by `TestLowerGoCrossNsCallLiftsToIFn`), so leaving the asymmetry as-is.

**Fast-follow (separate PR):**
1. **recordType identity** — give native records a shared `RecordType` via a name→`*RecordType` registry, so `Type()` is `==`-identical to the `defrecord` of the same name (real bug; deeper #232 change).
2. **Namespace home-ns resolution** — emit each var's actual home ns at all sites, the correct fix for the `core`/`clojure.core` asymmetry.

mparrett:
Thanks!

Codex (5.5-high) re-fetched PR #249 at `82f8c78` and re-reviewed the pushed fixes. Findings:

`[P1] LoadVar is still materialized too late when the if condition mutates the Var`  
The previous branch-body case is fixed, but `:load-var` still has no defining Go statement, so when it is threaded as a branch arg the transfer assignment happens after the `branch-if` condition has executed. That still changes `(let [x *v*] (if (do (set! *v* 2) true) ...) x)`: `x` should capture the value before the condition, but generated Go reads after the condition’s `set!`.

Relevant spots: [build.lg](/private/tmp/let-go-pr249-new/pkg/rt/core/ir/build.lg:757) builds the condition before branch target args are transferred; [lower_go.lg](/private/tmp/let-go-pr249-new/pkg/rt/core/ir/lower_go.lg:1072) emits condition evaluation before `transfer-stmts`; [lower_go.lg](/private/tmp/let-go-pr249-new/pkg/rt/core/ir/lower_go.lg:1018) materializes the incoming `LoadVar` during the edge copy.

Temporary repro rendered:

```go
func lv_cond_review(ec *vm.ExecContext) vm.Value {
    var x vm.Value
    if !ec.SetBinding(rt.LookupVar("core", "*lv-cond-review*"), vm.Int(2)) {
        rt.LookupVar("core", "*lv-cond-review*").SetRoot(vm.Int(2))
    }
    if true {
        x = ec.Deref(rt.LookupVar("clojure.core", "*lv-cond-review*"))
    } else {
        x = ec.Deref(rt.LookupVar("clojure.core", "*lv-cond-review*"))
    }
    return x
}
```

Expected shape is to materialize `x` before the condition side effect, or otherwise treat `LoadVar` as a local-carrying op once it must be threaded.

Minor: [gogen_aot_diff_test.go](/private/tmp/let-go-pr249-new/test/e2e/gogen_aot_diff_test.go:26) still says the harness compares the last non-empty stdout line. The implementation now compares full normalized stdout, so this is just stale commentary.

Verification run:
`go test ./pkg/ir -run 'TestLowerGo|TestLisp|TestPipeline' -count=1` passed.  
`go test ./pkg/vm -run 'Test.*Native|Test.*Reflect|Test.*Box|TestExecContextSetBinding' -count=1` passed.  
`make gogen-diff` passed.

mparrett:
@nnunley Finishing up the review now.

mparrett (approved):
Second-pass review against 6b483c9. Two findings, both P3 / **non-blocking** — detail below.

## P3 (LoadVar re-read across dynamic binding) — real but latent, already tracked

The mechanism is correct. `set-var-target-keys` only scans `:set-var` ops, so push-binding! targets never enter the snapshot hazard set, and `load-var-snapshot?` returns false for them. The render confirms it: in the repro, the pre-push snapshot `x` is emitted as `ec.Deref(rt.LookupVar(...))` inline at the `identity` call, i.e. re-derefed while the pushed binding is active. Bytecode returns 1, the generated Go returns 2. Same class as the set! bug.

Two reasons this isn't blocking:

- It's a defect in *rendered Go*, not an observed runtime divergence. The `gogen_ir` binary only swaps the pre-lowered stdlib; user script code still runs on the bytecode VM. Running the repro through both binaries prints `1`/`1`. It only bites code that actually gets lowered — the stdlib (I don't see a function that hits this shape) or future user-code AOT. The self-AOT diff harness doesn't lower fixture code, so it wouldn't catch a regression of this class either.
- The PR already discloses the snapshot is set!-target-only and tracks the unconditional version in #260. The dynamic-binding case is a subset of that gap.

So this is an extension of the #260 limitation, not a hidden blocker — P3. The follow-up: extend the #260 note to name binding-target vars, and add a shrink-only regression fixture. When the fix lands, the hazard set needs the push-binding! arg's var, which arrives as `(var *x*)` — a `:const` var value or load-var to dig out — and both keying paths must canonicalize ns (`core` vs `clojure.core`) to the same key.

## P3 (stale harness comment) — confirmed

gogen_aot_diff_test.go:26-27 says "last non-empty stdout line"; `runFixture`/`normalizeOutput` compare normalized full stdout. Doc drift, not a blocker.

Issue #194, opened 2026-06-05
async: add non-blocking channel take `poll!` (and `alts!` with :default)


nnunley:
## Motivation

Building a poll-based UI event loop (xsofy [#56](https://github.com/nooga/xsofy/issues/56)) that multiplexes a keyboard source and a cosmetic clock source and **must never block while idle**. This is mandatory under single-threaded Go-wasm, where a parked `<!` (backed by `Atomics.wait`) freezes the *entire* runtime, not just the goroutine. The loop needs to drain a producer-fed channel **once per frame if and only if a value is immediately ready**.

## What's missing

`async` exposes only **blocking** takes (`<!`/`<!!`). There is no non-blocking variant and no `alts!`. Registered channel fns today are just `chan`, `>!`, `<!`, `>!!`, `<!!` (`pkg/rt/lang.go:5969-5973`).

Everything *else* the use case needs already exists, which is why only this one primitive is requested:
- **Cancellable channel ops** — `<!`/`>!` already `select` on `vm.CurrentContext().Done()` (`pkg/rt/lang.go:3446`, `:3472`), so a producer goroutine parked on a put/take is released by `scope-close!`'s cancel+drain.
- **Scoped lifecycle** — `with-scope` (`pkg/rt/core/core.lg:2091`), `scope-open`/`scope-close!`/`scope-live` (`pkg/rt/lang.go:5974-5976`), `*scope-drain-timeout-ms*` (`core.lg:2089`).

So a ctx-aware, drainable producer is writable today; the only gap is the **consumer's non-blocking peek**.

## Requested

1. **`(poll! ch)`** — non-blocking take. Returns a value if one is immediately available, else a distinct empty signal (so a `nil` value can be distinguished from "nothing ready" — e.g. return a sentinel, or `[val ok]`). Go semantics:
   ```go
   select {
   case v, ok := <-ch:
       // ok==false => closed
   default:
       // nothing ready
   }
   ```
2. *(nice-to-have)* **`(alts! [ch…] :default v)`** — select over several channels with a non-blocking default; generalizes `poll!` to multiple upstreams.
3. *(symmetry, optional)* **`(offer! ch v)`** — non-blocking put.

## Notes

- `poll!` alone unblocks the use case; `alts!` is the general form.
- Keep ctx-awareness for consistency, though a `:default` branch means the non-blocking path never parks anyway.

nooga:
[...]

PR #250, opened 2026-06-18
feat(async): alts! :default for non-blocking select

mparrett:
## What

Add a `:default` option to `alts!`: `(alts! ports :default v)` returns `[v :default]` when no port is immediately ready, instead of parking. Matches `clojure.core.async`.

## Why

Closes the remaining ask in #194. `poll!` and `offer!` already shipped in #120; `alts!` was registered but only in its blocking form. The non-blocking variant is what a single-threaded Go-wasm event loop needs — a parked take freezes the whole runtime, not just the goroutine, so a frame loop multiplexing several channels has to drain "whatever's ready, else move on" without ever blocking.

## How

When `:default` is present, append a `reflect.SelectDefault` case so `reflect.Select` returns immediately if no port is ready. The ctx-`Done` case is dropped on that path: a non-blocking select can't park, so there's nothing for a cancel to release. Blocking `alts!` is unchanged.

## Test plan

- `go test ./pkg/rt/... ./test` green.
- New `.lg` tests: `:default` returns the fallback when no port is ready, and still prefers a ready port when one is.

### DIFF
--- a/pkg/rt/async.go
+++ b/pkg/rt/async.go
@@ -420,15 +420,28 @@ func installAsyncNS() {
 	})
 
 	// alts! — select on multiple channel operations
-	// (alts! [ch1 ch2 [ch3 val]]) → [val port]
+	// (alts! [ch1 ch2 [ch3 val]] :default v) → [val port]
 	// Each entry is either a channel (take) or [channel value] (put).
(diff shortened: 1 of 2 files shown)


nnunley (approved):
Should the README be updated now? Are we closer to the right semantics?

PR #251, opened 2026-06-18
fix(lang): def/declare forward-declaration promise semantics

nnunley:
## Summary

Aligns `(def x)` (no-init) and `declare` with Clojure's documented semantics: a no-init `def` interns the var but leaves its **root binding unaffected** — a forward declaration is an unbound *promise* that a later `(def x v)` fulfils, never a clobber.

Today let-go does `(def name nil)` for both the no-init `def` form and `declare`, which eagerly binds the root to `nil`. That:
- destroys an existing or later definition (a real clobber), and
- makes a declared var look *bound-to-nil*, which breaks the `defonce`-after-`declare` distinction (`defonce` gates on `bound?`).

Clojure reference: *"If init is not supplied, the root binding of the var is unaffected"*; `declare` *"defs the supplied var names with no bindings."*

## Changes

- **`pkg/compiler/compiler.go`** — 1-arg `(def x)` interns the var but no longer emits `OP_SET_VAR`; the root binding is left untouched. The var is still the expression result.
- **`pkg/rt/core/core.lg`** — `declare` emits `(def name)` instead of `(def name nil)`, so forward declarations create genuinely unbound vars.
- **`test/def_promise_test.lg`** — covers no-clobber, unbound-after-declare, and fulfilment.
- Regenerated `core_compiled.lgb` + `generated.sums`.

The VM already modeled unbound vars correctly (`NewVar` interns without a root; `IsBound()` tracks the root pointer) — only the compiler was prematurely fulfilling the promise.

Note: deref of an unbound var still returns `NIL` rather than throwing `Unbound` — a pre-existing, deliberate deviation (the deref hot path has no error channel); unchanged here.

## Verification

- `make check-generated` — bundle in lockstep; lowered tree compiles + dispatches under `-tags gogen_ir`.
- Full untagged `go test ./...` green, including the new `def_promise_test.lg`.


mparrett:
**Codex Findings (5.5-high)**

- [pkg/compiler/eval.go](/Users/matt/projects-new/3p/let-go/pkg/compiler/eval.go:92) still decodes missing bundle var references with `DefStub`, which installs a `nil` root. With this PR, `(def x)` now emits only `OP_LOAD_CONST` and returns before `OP_SET_VAR` at `pkg/compiler/compiler.go:1709`, so precompiled/LGB-loaded code can still make no-init defs and declares appear bound to `nil` before any bytecode runs. That preserves the bug for `core_compiled.lgb` and any LGB decode path using this resolver. I’d change the decode resolver to intern an unbound var, for example via `LookupOrAdd` or a new unbound stub helper, and add an LGB regression for `(def x) -> (bound? (var x)) == false`.

**Verification**

- Passed: `go test ./test -run 'TestRunner/def_promise_test.lg' -count=1`
- Failed focused temp check on PR: decoding a chunk with only a var constant through the runtime-style `DefStub` resolver produced a bound var before execution.

---

```
package bytecode_test

import (
	"bytes"
	"testing"

	"github.com/nooga/let-go/pkg/bytecode"
	"github.com/nooga/let-go/pkg/rt"
)

func TestNoInitDefLGBParityTemp(t *testing.T) {
	const nsName = "temp-no-init-lgb-decode-pr251"
	const varName = "x"

	consts := vm.NewConsts()
	chunk := vm.NewCodeChunk(consts)
	idx := consts.Intern(vm.NewVar(nil, nsName, varName))
	chunk.Append(vm.OP_LOAD_CONST)
	chunk.Append32(idx)
	chunk.Append(vm.OP_RETURN)

	var buf bytes.Buffer
	if err := bytecode.EncodeCompilation(&buf, consts, chunk); err != nil {
		t.Fatalf("encode failed: %v", err)
	}

	resolve := func(ns, name string) *vm.Var {
		n := rt.DefNSBare(ns)
		if v := n.LookupLocal(vm.Symbol(name)); v != nil {
			return v
		}
		return n.DefStub(name)
	}
	if _, err := bytecode.DecodeToExecUnit(&buf, resolve); err != nil {
		t.Fatalf("decode failed: %v", err)
	}

	resolved := rt.NS(nsName).LookupLocal(vm.Symbol(varName))
	if resolved.IsBound() {
		t.Fatalf("decoded var constant is bound before bytecode execution")
	}
}
```

nnunley:
Fixed in the latest push (this branch was force-updated).

Root cause was exactly as described: `Namespace.DefStub` — the stub the LGB decode resolver in `pkg/compiler/eval.go` interns for missing bundle var refs — force-set a NIL root, so `IsBound()` returned true. Since a no-init `(def x)` emits only `OP_LOAD_CONST`, that stub is the var that survives into the namespace, so bundle/LGB-loaded code saw `(bound? (var x))` as true while freshly-compiled code was unbound.

Changes:
- `pkg/vm/namespace.go`: `DefStub` now interns an **unbound** var (drops `SetRoot(NIL)`), matching `LookupOrAdd`. Its only callers are the two decode resolvers (`pkg/compiler/eval.go`, `pkg/rt/disasm.go`), so both paths are covered.
- `pkg/bytecode/lgb_noinit_def_test.go`: added your encode→decode parity test as a permanent regression (decoded var const must be unbound before execution). Red before the change, green after.

`go test ./pkg/vm ./pkg/bytecode ./pkg/compiler ./pkg/rt ./test` all pass and `make check-generated` is in lockstep.

On the deeper point raised in review — the permissive `set!` (root fallthrough instead of throwing when there's no thread-local binding) is a genuine Clojure deviation, but making it strict breaks core compilation itself: `defmethod` (`core.lg:2601`) uses root-level `set!` as the multimethod mutation primitive (e.g. `print-method`), and the whole `test.lg` harness `set!`s at root too. It needs `defmethod`→`addMethod` and the harness moved onto `binding`/refs first, so I've split it out as #256 rather than expand this PR.

nooga (approved):

PR #252, opened 2026-06-18
fix(genmanifest): exclude generated Go from digest; gate before regen

mparrett:
This PR now does two things; the second is the real fix.

`ci(genmanifest): fail manifest gate before regeneration` — the original intent: a non-mutating front gate (`make check-generated-manifest`) so CI fails before `lgbgen` can rewrite the manifest and mask drift. Wired into `check-generated` and the build job.

`fix(genmanifest): exclude generated Go from the source digest` — what the gate exposed. `generated.sums` hashes every `.go` under `cmd/lgbgen`, which sweeps in `cmd/lgbgen/main_gogen_ir.go`, the gitignored gogen_ir wireup `lgbgen` emits. That's a build artifact, not a source; the Makefile already calls the gogen_ir wireup "NOT committed ... a build artifact." It's absent on a clean checkout, present only after a generation run. So the digest wasn't checkout-independent: a sum recorded post-generate (`c5e0c94f`, wireup hashed in) can't match a pre-generation check (`f82ff7e4`, without it). The fix skips files with the standard `// Code generated ... DO NOT EDIT.` marker and refreshes the sum to the clean value.

Correction to my note on #247: I said the digest was "pure source hashing ... every machine computes the same value." That held only on a clean checkout where the wireup is absent. The second commit makes it true in general.

Verified locally (rebased on `main`): the gate passes with and without the wireup on disk; `go test ./pkg/genmanifest` (including a new exclusion test); `go vet`; and `linux/amd64`, `js/wasm`, and `plan9/amd64` cross-builds.


mparrett:
**Codex Findings (5.5-high)**

[P1] CI currently masks stale `pkg/rt/generated.sums` rather than proving it was committed fresh. `.github/workflows/go.yml:37` runs `go generate ./pkg/rt/` before `go test`, so `pkg/genmanifest/staleness_test.go:20` sees the regenerated manifest, not the checkout state. The dedicated check has the same shape: `.github/workflows/go.yml:134` runs `make check-generated`, and `Makefile:223-225` regenerates with `lgbgen --target=both` before checking bundle bytes. `cmd/lgbgen/main.go:476` / `:483-499` refreshes `generated.sums`, so stale manifest drift can pass CI. PR #252 demonstrates that gap.

No issue with the PR’s actual one-line manifest refresh: `fb55e8...` to `6d9cf...` is exactly the kind of deterministic digest-only change I’d expect. All PR checks are now green and the PR is mergeable.

On CI automation: I would not make CI auto-commit this by default. Better fix is to make CI fail before mutation. Add an early non-mutating check, for example `go run ./cmd/check-generated` or `go test ./pkg/genmanifest` before any `go generate` / `make check-generated`, or make `make check-generated` start with `go run ./cmd/check-generated` and only then run the heavier regeneration/behavioral checks.

Auto-update can be useful as a bot convenience on trusted branches, but it should be secondary, maybe a workflow that comments/uploads the diff or opens a small bot PR. Directly mutating contributor branches tends to hide the discipline problem and gets messy with forks and permissions.

nnunley (approved):
LGTM. I think there's a broader story for allowing some variation of regeneration for the sums and the .lgb artifacts as part of CI, but there's definitely blocks on that, due to current configuration decisions owned by @nooga

nnunley (approved):

PR #253, opened 2026-06-18
fix(reader): skip leading no-value forms in Read()

nnunley:
## Summary

`Read()` dispatched a reader macro and returned its result directly. Line comments (`readLineComment`) and `#_` discard (`readFormComment`) return the `VOID` no-value sentinel, so single-form callers like `read-string` received `VOID` instead of the next real form — e.g. `(read-string ";; c\n(+ 1 2)")` returned `VOID`.

Collection readers and `CompileMultiple` each independently work around `VOID`; `read-string` did not.

## Change

Loop `Read()` past `VOID`-returning macros so it yields the next real form, or surfaces EOF when the input has no form (comment/whitespace only). Comments produce no value (not a comment atom), per Clojure reader semantics.

## Tests

`TestReaderSkipsLeadingNoValueForms` (leading comments, `#_`, indented, multi-line) and `TestReaderEOFAfterCommentOnly` (comment/whitespace-only → EOF). Full `pkg/compiler` suite green.

### DIFF
--- a/pkg/compiler/eval.go
+++ b/pkg/compiler/eval.go
@@ -48,10 +48,12 @@ func Eval(src string) (vm.Value, error) {
 	return out, nil
 }
 
-// ReadString parses a string into a let-go Value.
+// ReadString parses a string into a let-go Value. As a single-form entry point
+// it skips leading no-value forms (comments, #_ discard) so a string that opens
+// with a ';;' comment yields the following form rather than the VOID sentinel.
 func ReadString(s string) (vm.Value, error) {
 	reader := NewLispReader(strings.NewReader(s), "<read-string>")
-	return reader.Read()
+	return reader.ReadSkipNoValue()
 }
 
 func evalInit() {
--- a/pkg/compiler/reader.go
+++ b/pkg/compiler/reader.go
@@ -178,6 +178,12 @@ func appendNonVoid(r *LispReader, vs []vm.Value, v vm.Value) []vm.Value {
 	return append(vs, v)
 }
 
+// Read reads a single form. No-value reader macros (line comments, #_ discard,
+// reader conditionals with no matching branch) return the VOID sentinel — this
+// is load-bearing: collection readers (readList/readVector/readMap/readSet)
+// call Read in element/value position and rely on seeing VOID to drop an
+// orphaned map key or splice nothing (e.g. {:a #_ 1 :b 2} => {:b 2}). Callers
+// that want a real form with leading no-value forms skipped use ReadSkipNoValue.
 func (r *LispReader) Read() (vm.Value, error) {
 	ch, err := r.eatWhitespace()
 	if err != nil {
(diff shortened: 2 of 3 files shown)


mparrett:
**Codex Findings (5.5-high)**

- **P1** [reader.go:201](/private/tmp/let-go-pr253-review/pkg/compiler/reader.go:201): looping past every `VOID` in `Read()` breaks map literals that rely on seeing `VOID` in value position. `readMap` explicitly drops an orphaned key when a reader conditional returns `VOID` at [reader.go:720](/private/tmp/let-go-pr253-review/pkg/compiler/reader.go:720), but the new loop consumes the next real form instead. Repro on the PR branch: `{:a #?(:cljs 1) :b 2}` now errors with `map literal must contain even number of forms` because `:b` is read as `:a`’s value and `2` is left as an odd key.

I verified the PR’s added focused tests pass:

`go test ./pkg/compiler -run 'TestReaderSkipsLeadingNoValueForms|TestReaderEOFAfterCommentOnly|TestReaderConditionalSkip'`

Then I added a temporary regression test in a throwaway worktree for `{:a #?(:cljs 1) :b 2}` and `{:a #_ 1 :b 2}`; both failed on the PR branch with the odd-map error. The fix likely needs to split “read one form preserving `VOID`” from the public/single-form behavior that skips leading no-value forms, so collection readers can keep their existing `VOID` handling.

---
```
package compiler

import (
	"strings"
	"testing"

	"github.com/stretchr/testify/assert"
)

func TestReaderMapDiscardDoesNotConsumeNextKey(t *testing.T) {
	r := NewLispReader(strings.NewReader("{:a #_ 1 :b 2}"), "<reader>")
	o, err := r.Read()
	assert.NoError(t, err)
	assert.Equal(t, vm.NewArrayMap([]vm.Value{vm.Keyword("b"), vm.Int(2)}), o)
}

func TestReaderMapUnmatchedConditionalDoesNotConsumeNextKey(t *testing.T) {
	r := NewLispReader(strings.NewReader("{:a #?(:cljs 1) :b 2}"), "<reader>")
	o, err := r.Read()
	assert.NoError(t, err)
	assert.Equal(t, vm.NewArrayMap([]vm.Value{vm.Keyword("b"), vm.Int(2)}), o)
}

```

nnunley:
Confirmed and fixed in 643ca515. The P1 is correct: `Read()` is shared with the collection readers, which rely on seeing `VOID` in value position (`readMap` dropping an orphaned key, splice-nothing), so looping past it consumed the next form.

Instead of changing `Read()`, I kept it VOID-preserving and added a single-form entry point `ReadSkipNoValue()` that loops past leading no-value forms, and routed `read-string` (`ReadString`) through that. Collection readers keep calling `Read()` unchanged.

Both of your regression cases are now in `reader_test.go` as `TestReaderMapDiscardDoesNotConsumeNextKey` and `TestReaderMapUnmatchedConditionalDoesNotConsumeNextKey` (using `Read()`), alongside the leading-no-value tests (now using `ReadSkipNoValue()`). End-to-end: `{:a #_ 1 :b 2}` and `{:a #?(:cljs 1) :b 2}` both read as `{:b 2}`, `(a ;; c\n b)` as `(a b)`, and `read-string ";; c\n(+ 1 2)"` as `(+ 1 2)`. Full `go test ./...` green.

nnunley:
This is already addressed in the current head (`643ca515`) — looks like the review ran against an earlier commit. The fix is the split you suggested:

- `Read()` preserves `VOID` (load-bearing — collection readers rely on seeing it).
- `ReadSkipNoValue()` is a separate single-form entry point (used by read-string) that skips leading no-value forms and never returns `VOID`.
- `readMap` drops an orphaned key when a value-position form is `VOID`.

Both repro cases are covered by passing tests on the branch:
- `TestReaderMapDiscardDoesNotConsumeNextKey` — `{:a #_ 1 :b 2}`
- `TestReaderMapUnmatchedConditionalDoesNotConsumeNextKey` — `{:a #?(:cljs 1) :b 2}`

`go test ./pkg/compiler -run 'TestReaderMapDiscardDoesNotConsumeNextKey|TestReaderMapUnmatchedConditionalDoesNotConsumeNextKey'` passes, and the built binary returns `{:b 2}` for both literals.

nooga (approved):

mparrett:
>  looks like the review ran against an earlier commit. 
Sorry for the churn. Juggling too many things!

PR #254, opened 2026-06-18
fix(ir): expand-all must expand fn-expression heads (IIFE)

nnunley:
## Summary

`expand-all` kept the head of a non-special-form list as-is, only expanding `(rest form)`. For an **immediately-invoked fn expression** — `((fn [x] ...) args)` — the head `(fn ...)` was never macroexpanded to `(fn* ...)`. The IR builder only understands `fn*`, so it misread the IIFE as a *call* with operator `fn` and arguments `[arglist]` and `[body symbols]`, then threw `ir/build: unresolved symbol <body-sym>`.

Not destructuring-specific: `((fn [x] x) 5)` fails identically with `unresolved symbol x`. Only IIFEs are affected — fns in argument position were already expanded via the `(rest form)` recursion.

## Change

Expand the head too: `(apply list (expand-all (first expanded)) (map expand-all (rest expanded)))`. For a symbol head, `expand-all` is the identity, so ordinary calls are unaffected.

## Verification

- Repro `((fn [x] x) 5)`, plus `core/distinct-step` and `check/bind` (which use IIFEs), now lower to `:lowered` (were `unresolved symbol f` / `_r`).
- `make check-generated` green (bundle lockstep + `gogen_ir` native dispatch); full untagged `go test ./...` green.
- Regenerated `core_compiled.lgb` + `generated.sums`.


mparrett:
Review was clean; no findings. 🚀

nooga (approved):

Issue #255, opened 2026-06-18
Host capabilities: generalizing the I/O decoupling to e.g. audio and graphics


mparrett:
Continues from #79 (browser-interop proposal) and the I/O arc that grew out of it: output (#231), the design doc (#233), and the typed seams for emit (#241), input (#244), and the client-owned shell (#245), all merged. That arc answers the browser-interop question #79 opened, and it raises a larger one.

§6 of #233 names graphics, audio, and controller input as possible future "peer capabilities" and claims "none is special once the I/O seams set the pattern." This issue scopes that broader surface: what the audio and graphics seams look like, where each one bends the pattern, and what is worth settling before code. The intent is to converge here, then land it as a `docs/design` doc. The finding so far is that the claim holds, with one refinement. The invariant that generalizes is the *binding discipline*, not `io.Writer`.

## The shared shape

The seams so far follow one pattern, and these two should too:

- The guest **names** the capability (`println`, `read-key`, `emit`); the host **supplies** the backend.
- The seam lives in `pkg/rt`, build-tagged, **never `pkg/vm`**, so the VM core stays host-free.
- The seam is **as thin as the operation allows** (#233 §4). It widens only where platforms differ.
- **Three bindings** per capability: native (concrete OS resource), embedder (`pkg/api` per-`Run` option), WASM (a JS-aware adapter forwarding through `window.LetGoHost`).
- The **zero value is a silent no-op**, auto-detected. Headless, CI, and replay-verification get the inert binding for free, the same way `*out*` defaults to pass-through (#233 §5.5).
- **Determinism quarantine.** Both audio and graphics are *side-effect sinks*: they read guest-visible state and never feed back into it. In a deterministic guest (xsofy replays bit-identically from seed + action-log), a capability binding must be structurally unable to perturb that state. A replay renders-or-doesn't, plays-or-doesn't, identically.
- The **rich/retained layer lives above the seam, in `.lg`** (#233 §5.6's ratatui reasoning). The runtime hosts the thin carrier; the weight stays in guest code.

What changes per capability is the *carrier*: a byte stream, a handle-based capability, or a shared buffer. Choosing the carrier is the design.

### Shape: a capability with handles, not a stream

[...]

PR #261, opened 2026-06-18
docs(design): mark runtime-I/O decoupling shipped; refresh status

mparrett:
The I/O seams `docs/design/runtime-io-host-decoupling.md` (#233) proposed have all merged, so the doc was stale at `status: planning`. This refreshes it to match.

**Frontmatter:** `planning → active` (the doc stays the design authority per the status convention), with `shipped: [#206, #207, #223, #231, #241, #244]` and `remaining-open: peer capabilities → nooga/let-go#255`.

**Body:**
- §4: "output is decoupled" → "I/O is decoupled"; `#231 open → merged`; added input (#244) and emit (#241) rows; noted the client-owned shell (#245).
- §5.2: input shape `Proposal → Decided (#244)`.
- §5.3 / §7: the input seam shipped in #244, but it **deferred the cross-platform wake protocol** as its own change, so the wake API stays an open question (not "settled"). `read-key` granularity is settled (one key per call).
- §6: input/emit shown shipped with dates; peer capabilities scoped to #255.

Design rationale (§1–3, §5.1/5.4–5.7, glossary) unchanged. `docs-frontmatter` `--check` passes.

### DIFF
--- a/docs/design/runtime-io-host-decoupling.md
+++ b/docs/design/runtime-io-host-decoupling.md
@@ -1,13 +1,15 @@
 ---
-status: planning
-last-verified: 2026-06-15
-human-verified: 2026-06-15
+status: active
+last-verified: 2026-06-18
+human-verified: 2026-06-18
+shipped: ["#206", "#207", "#223", "#231", "#241", "#244"]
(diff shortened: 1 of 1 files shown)


nnunley (approved):
Sounds good. Definitely the direction we've been moving in. Thanks for getting it down in writing.

PR #276, opened 2026-06-20
fix(wasm): non-blocking SAB key producer — newest-wins, never block the main thread

mparrett:
## Problem

The WASM input producer `_lgKey` (`pkg/rt/wasm/lg-host-core.js`, main thread) busy-waits until the worker drains the single SAB key slot:

```js
while (Atomics.load(keyInt32, 0) !== 0) { /* busy wait */ }
```

On a held key or OS auto-repeat, input outpaces the engine's turn rate, so each keypress spins the **main thread** until the worker clears the slot — freezing the page. The existing comment already flagged it: *"the pre-existing busy-wait; retiring that belongs to the input seam, #244."* #244 shipped the typed consumer seam but left this producer half.

## Fix

Drop the spin; write unconditionally (**newest-wins**). If a key is still pending, the newest overwrites it — for an interactive program the latest input is the relevant one. Non-blocking, no queue, no ring. One line removed plus a comment.

## Tradeoff

Overwriting an unread slot coalesces input only when **distinct** keys arrive within a single consumer turn; same-key auto-repeat reads back identically. A generation-counter seqlock would make even the distinct-key case lossless, but that isn't needed for the poll-one-key-per-turn consumers this targets.

## Validation

Reproduced in a SAB harness mirroring the protocol (consumer in a worker, simulated turn cost): an 8-key burst's main-thread block drops from ~566 ms to ~0.04 ms, with no rAF gap. `go test ./pkg/rt/wasm ./pkg/api` green; the four `testdata/assemble_golden*.html` fixtures are regenerated (`-update`).

### DIFF
--- a/pkg/rt/wasm/lg-host-core.js
+++ b/pkg/rt/wasm/lg-host-core.js
@@ -113,21 +113,22 @@ async function startWorkerMode() {
   const keyUint8 = new Uint8Array(sab, 8, 16);
 
   // Input is unsafe until the worker has been told to start (init posted):
-  // before that there is no consumer to drain the SAB slot, so the producer
-  // busy-wait below would spin the main thread — worst case for the whole
-  // multi-second external fetch+compile. Drop pre-start keys rather than block.
-  // (The narrow post-start window before the VM first reads a key is the
-  // pre-existing busy-wait; retiring that belongs to the input seam, #244.)
+  // before that there is no consumer to drain the SAB slot. Drop pre-start
+  // keys rather than write into a slot nothing will ever read.
   let workerReady = false;
 
   // Public input hook — backs LetGoHost.sendInput. Returns true if accepted,
   // false if dropped (worker not started yet, or input too long >16 bytes).
   // A shell's keystrokes feed through here via LetGoHost.sendInput.
+  //
+  // Non-blocking, newest-wins: write unconditionally — never wait for the
+  // consumer to drain (that wait is what froze the page on key bursts). Under
+  // load the slot coalesces to the freshest key; lossy only for *distinct* keys
+  // overlapping within one consumer read, never for same-key repeat.
   window._lgKey = function(data) {
     if (!workerReady) return false;
     const bytes = new TextEncoder().encode(data);
     if (bytes.length === 0 || bytes.length > 16) return false;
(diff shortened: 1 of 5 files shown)


mparrett:
I have A/B'ed this personally. With the patch, auto action is interrupted nearly instantly, and repeated keypress doesn't wait the game to death. The difference is appreciable, to say the least. The trade off is you can overwrite an unconsumed keypress during fast input. I haven't found this to be a problem in testing, but have tested various buffer solutions that look promising. Not worth the added complexity now, but that's what I'd reach for. Happy to host the A/B somewhere if you'd like to convince yourself. :-)

mparrett:
**Codex Findings (5.5-high)**

`P1` [[pkg/rt/wasm/lg-host-core.js:132](https://github.com/nooga/let-go/pull/276)](https://github.com/nooga/let-go/pull/276) can race with `HostKeySource.ReadKey` clearing the slot.

Removing the wait lets `_lgKey` write `keyUint8`, length, and ready flag while the worker is already consuming the previous key. The consumer then unconditionally does `Atomics.store(keyInt32, 0, 0)` after copying bytes in `pkg/rt/keysource_js_wasm.go`, so this interleaving can lose the newest key entirely:

1. Worker wakes on old key, reads `keyLen`.
2. Main thread receives a new key, writes bytes/length, stores ready flag `1`, returns `true`.
3. Worker finishes old read and stores ready flag `0`.
4. New key data remains in the byte slot, but the ready flag is cleared, so the next `read-key` waits.

That is stronger than lossy coalescing: an accepted key can disappear, and byte reads can also become mixed if overwrite happens during the copy. The producer/consumer needs ownership state, a generation/seqlock, or a small ring/double buffer. A simpler nonblocking fallback would be to drop when the slot is not safely empty rather than overwrite concurrently.

**Notes**

I did not run local tests. GitHub reports the PR’s Go workflow checks green, and I reviewed the PR diff plus the surrounding WASM key consumer code.

mparrett:
> A simpler nonblocking fallback would be to drop when the slot is not safely empty rather than overwrite concurrently.

Sounds better, will test this!

nooga (approved):

nnunley (approved):

PR #277, opened 2026-06-20
fix(wasm): host.html #terminal needs min-height:0 so client content can't overflow the frame

mparrett:
## Problem

The `lg -w` host frame (`pkg/rt/wasm/host.html`) styles the terminal as a flex item:

```css
body{display:flex;flex-direction:column}
#terminal{flex:1}
```

A flex item defaults to `min-height:auto`, which won't let it shrink below its content's intrinsic size. The default `-w-shell xterm` shell never resizes its content, so this is dormant there. But a **client-owned shell** (`-w-shell none`, #245) that makes the terminal content taller — e.g. bumping `term.options.fontSize` and re-fitting — pushes `#terminal` past the viewport, and its sibling DOM (the client's own chrome) gets shoved below the fold. Under the frame's `overflow:hidden` it's simply gone.

Measured in a client shell at 1280×800: one font bump grew `#terminal` 743px → 920px (viewport 800), moving the client UI off-screen.

## Fix

```css
#terminal{flex:1;min-height:0}
```

One line. The flex item is pinned to its viewport share; content overflow stays inside `#terminal` instead of expanding it. No behavior change for the default shell — it only makes the frame robust to clients that resize content.

## Note (out of scope here)

There's a broader design question — whether the `-w-shell none` frame should ship *any* layout CSS, or hand the client a barer/neutral mount — but that's a separate discussion; this PR is just the latent-overflow fix. `go test ./pkg/rt/wasm` green; the four `assemble_golden*.html` fixtures are regenerated.

### DIFF
--- a/pkg/rt/wasm/host.html
+++ b/pkg/rt/wasm/host.html
@@ -8,7 +8,7 @@ __LG_XTERM_CSS__
     *{margin:0;padding:0;box-sizing:border-box}
     html,body{height:100%;background:#0c0c0c;color:#e8e6df;overflow:hidden}
     body{display:flex;flex-direction:column}
-    #terminal{flex:1}
+    #terminal{flex:1;min-height:0}
     #status{color:#5a584f;font-family:monospace;font-size:13px;padding:1rem;position:absolute;
             top:50%;left:50%;transform:translate(-50%,-50%)}
     .xterm{height:100%}
--- a/pkg/rt/wasm/testdata/assemble_golden.html
+++ b/pkg/rt/wasm/testdata/assemble_golden.html
@@ -8,7 +8,7 @@
     *{margin:0;padding:0;box-sizing:border-box}
     html,body{height:100%;background:#0c0c0c;color:#e8e6df;overflow:hidden}
     body{display:flex;flex-direction:column}
-    #terminal{flex:1}
+    #terminal{flex:1;min-height:0}
     #status{color:#5a584f;font-family:monospace;font-size:13px;padding:1rem;position:absolute;
             top:50%;left:50%;transform:translate(-50%,-50%)}
     .xterm{height:100%}
--- a/pkg/rt/wasm/testdata/assemble_golden_external.html
+++ b/pkg/rt/wasm/testdata/assemble_golden_external.html
@@ -8,7 +8,7 @@
     *{margin:0;padding:0;box-sizing:border-box}
     html,body{height:100%;background:#0c0c0c;color:#e8e6df;overflow:hidden}
     body{display:flex;flex-direction:column}
-    #terminal{flex:1}
+    #terminal{flex:1;min-height:0}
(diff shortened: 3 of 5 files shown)


mparrett:
Validated in a real client shell (xterm) driving an actual app, on desktop and phone.

Repro shell: terminal mounted in `#terminal` plus a sibling chrome bar, no client-side `min-height` workaround. Without the fix, anything that makes terminal content taller than its flex slot — raising the xterm font, **or** rotating to a shorter viewport, **or** shrinking the window — leaves `#terminal` stuck at content height (the `min-height:auto` default) and pushes the sibling chrome off-screen (gone under `body{overflow:hidden}`). Measured: one font bump took `#terminal` 896px → 1568px in a 900px viewport.

With `#terminal{flex:1;min-height:0}` the item stays pinned to its viewport share through font/rotation/resize and the sibling UI survives — `#terminal` held at 840px across the same bump.

Scope note: client-side re-fit-on-resize is the client's own concern (a separate follow-up); this PR is purely the frame robustness so a client that resizes content doesn't get its layout silently broken.

mparrett:
**Codex Findings (5.5-high)**

None. I did not find a correctness issue in [[nooga/let-go#277](https://github.com/nooga/let-go/pull/277)](https://github.com/nooga/let-go/pull/277).

I reviewed the PR diff and surrounding wasm assembly code. The `#terminal{flex:1;min-height:0}` change is scoped correctly in `pkg/rt/wasm/host.html`, and the four generated golden fixtures match that template change.

Verification:
- `go test ./pkg/rt/wasm` passed locally in a temporary PR worktree.
- `git diff --check` passed.
- GitHub checks for the PR are green: lint, build, gold-differential, generated-artifacts, docs-frontmatter, and test-location.

Residual risk: This review relies on CSS behavior plus the PR author’s manual validation for the actual resize scenario; I did not run a browser/xterm visual repro locally.

nooga (approved):

nnunley:
The `min-height:0` fix is the right root-cause fix — it keeps the flex item pinned to its viewport share regardless of content size. Approving.

Two viewport-sizing gaps remain in `host.html`, separate from this change and worth a follow-up (fits the "should the frame ship layout CSS at all" question already flagged):

1. **No viewport meta tag.** `<head>` has no `<meta name="viewport" content="width=device-width, initial-scale=1">`. Without it, mobile browsers assume a ~980px layout viewport and scale the page down, so `height:100%` and the flex column don't map to the actual device viewport.
2. **`height:100%` vs dynamic viewport units.** On mobile, `height:100%` is unstable as browser chrome (URL bar) collapses/expands. `100dvh`/`100svh` pins to the actually-visible viewport.

Both harden the frame for the same client-owned-shell scenario this PR addresses, without blocking it.

mparrett:
Thanks for the review- the followup is right on and planned. Appreciate the confirmation!

PR #278, opened 2026-06-20
perf-page: relative-% timeline charts

mparrett:
Reworks the "Are we fast yet?" timeline cards from raw anchor-ratio lines into readable trend charts. Still server-rendered static SVG, with no JavaScript or chart library.

## What changed

- **Relative-% axes.** The suite is plotted against the v1.8.0 release; IR compile against the window's first run, since it postdates v1.8.0. The 0% baseline sits off-scale as a legend marker, so the axis fits the data rather than flattening to include it. The absolute charts (allocations, bytes) use one consistent scale with bare numbers and the unit in a footer line.
- **Line fidelity.** Per-run sample variance bands, and lines that break across missing snapshots instead of bridging them. Also a fix for variant-key drift: when the `bytecode`/`gogen_ir` keys were renamed to `ir_bytecode`/`aot_native`, the charts silently dropped every post-rename snapshot, truncating three of the four to the pre-rename window.
- **Readable labels.** Date x-axis, with commit SHAs moved to hover tooltips; collision-spaced ticks and annotations; inter-series Δ labels (`aot_native` vs `ir_bytecode`).
- **Table.** "Recently tightened" gains a × anchor column, comparable across machines, alongside the change vs the release.

## Screenshots

Before / after, both rendered from the same 37-snapshot timeline:

| Before | After |
|---|---|
| image | image |

| Before | After |
|---|---|
| image  |  image  |

## Notes

Diff is `cmd/perf-page` only. Rendered against the current perf-data timeline (37 snapshots); to preview, run `cmd/perf-page -timeline <dir>`. Happy to adjust the chart selection or labeling if you'd shape it differently.

### DIFF
--- a/cmd/perf-page/main.go
+++ b/cmd/perf-page/main.go
@@ -508,8 +508,10 @@ func buildChart(timeline []Snapshot, title, subtitle, unit string,
 	// the display transform below; range tracking waits until the relative
 	// basis is known.
 	series := make([]ChartSeries, 0, len(specs))
-	latestRaw, latestIdx, haveLatest := 0.0, -1, false
 	oldestRaw, oldestIdx := 0.0, math.MaxInt
+	// Latest plotted value per series, for a worst-case status across all
+	// series rather than whichever series sorts first (see chartStatus).
+	var latest []seriesLatest
 	for _, spec := range specs {
 		var pts []ChartPoint
 		for i, snap := range timeline {
@@ -531,9 +533,6 @@ func buildChart(timeline []Snapshot, title, subtitle, unit string,
 				SHA:   shortSHA(snap.Baseline.CapturedAtSHA),
 				Value: value, Low: lo, High: hi, HasBand: hasBand,
 			})
-			if i > latestIdx {
-				latestIdx, latestRaw, haveLatest = i, value, true
-			}
 			if i < oldestIdx {
 				oldestIdx, oldestRaw = i, value
 			}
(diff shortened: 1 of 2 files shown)


mparrett:
**Codex Findings (5.5-high)**

- [cmd/perf-page/main.go:534](/private/tmp/let-go-pr278/cmd/perf-page/main.go:534) / [cmd/perf-page/main.go:747](/private/tmp/let-go-pr278/cmd/perf-page/main.go:747) `latestRaw` tracks only the first series seen at the latest snapshot because it updates on `i > latestIdx`, not per series. The chart then renders one status pill for a multi-series chart from that single value. In the rendered page, “End-to-end suite” says `71% over v1.8.0`, while the other latest series point is `+86.4%`; if the first series were within budget and the second over budget, the chart header would actively misreport the regression. I’d either compute a per-series status or aggregate the worst latest delta and include the series label.

**Verification**

- `go test ./cmd/perf-page` passed.
- `go run ./cmd/perf-page -out /private/tmp/pr278-index.html` passed.
- `go test ./...` failed in `pkg/genmanifest` due stale generated artifacts; that is outside this PR diff, but the remote `generated-artifacts` check was still pending when I checked. Other remote checks were passing.

mparrett:
>  if the first series were within budget and the second over budget, the chart header would actively misreport the regression. I’d either compute a per-series status or aggregate the worst latest delta and include the series label.

Will investigate.

mparrett:
Fixed in 212817d.

The status pill on a multi-series chart was computed from `latestRaw`, which only the first series could set: the `i > latestIdx` gate never fired again once a later series hit the same snapshot index. So a within-budget first series could mask an over-budget second one, which is the suite chart case you caught.

Now each series keeps its own latest plotted value and the pill reports the worst delta, naming the series when there's more than one. The suite header reads `86% over v1.8.0 (aot_native)` instead of `71%`. Single-series charts are unchanged.

Verified: `go test ./cmd/perf-page` passes, `gofmt`/`vet` clean, and a render against the current 37-snapshot timeline shows the corrected pill.

nooga (approved):

Issue #284, opened 2026-06-20
gogen: emit concrete Go types from typeinfer instead of vm.Value everywhere


nnunley:
Deepest 'not leveraging Go' signal: the lowered tree has **21,154 `vm.Value`** occurrences and **zero** concrete numeric declarations (0 `int64`/`float64`). `typeinfer` narrows types in the IR, but emission discards that and keeps the boxed `vm.Value` interface everywhere → constant box/unbox and dynamic dispatch for arithmetic/comparisons Go could do natively.

**Ask:** thread provable typeinfer results into emission so known-typed values get concrete Go types (`int64`, `bool`, `string`, `*vm.Map`…) and native ops. Enables the largest structural quality + speed gains.

Part of epic #258.

nnunley:
Implemented incrementally across #235, #320, #321, #322, and #432. Current lowering threads inferred types into Go params, locals, returns, branches, and operations; emits concrete primitive/dtype types and native operations where proven; and retains vm.Value for genuinely dynamic or ambiguous values. Regression coverage exercises both concrete and boxed paths. Further unboxing opportunities should be tracked as measured follow-up optimizations.

Issue #281, opened 2026-06-20
gogen: elide callErr threading for non-erroring calls (~41% of lowered tree)


nnunley:
Signal from the bytes/loc ratio analysis of `core_go_lowered/` (70,328 lines).

**~41% of the entire lowered tree is callErr error-handling boilerplate:** 9,840 `if callErr != nil { return nil, callErr }` blocks (~3 lines each ≈ 29.5K lines) across 8,487 call sites. Every call is emitted as `(vm.Value, error)` + a 3-line check.

But most callees **cannot error**: pure clojure.core fns, keyword/map lookups, devirtualized method calls, single-return siblings. `function-needs-error?` exists in `lower_go.lg` but the per-call-site decision over-threads.

**Ask:** an error-effect analysis so a call to a provably-non-erroring target emits a plain assignment (no `(_, callErr)`, no check). Biggest single codegen-density win.

Part of epic #258 (leverage Go features in the target). Surfaced via fanout bytes/loc ratio investigation.

nnunley:
## Investigation — where callErr is over-threaded (grounded in `lower_go.lg`)

The elision machinery **partially exists** but is too narrow and only wired into one path:

**1. `call-error-free?` (lower_go.lg:1842) is narrow.** It returns true ONLY for deftype constructors (`*deftype-ctors*`) and total devirtualized protocol methods. It does **not** recognize:
- keyword-as-callee lookups `(:k m)` — can't error (536 `InvokeValueEC(ec, vm.Keyword …)` sites)
- known-pure/total `clojure.core` fns (get, name, str, …) routed through `rt.CachedVarFn`

**2. The trampoline path doesn't consult it per-call.** `emit-call-stmts` (1398) *always* emits the `[assign, err-check]` pair. Only `emit-typed-direct-call` (1417) branches on `(:needs-error? entry)` and elides — but that path only fires for typed registry entries (cross-ns lowered fns). The vast majority (keyword lookups + CachedVarFn core calls) take the trampoline and always get the 3-line block. That's the source of the 9,840 `callErr != nil` blocks (~41% of the tree).

## Prototype plan
1. **Extend `call-error-free?`**: recognize a keyword callee (`(:k m)`), and a conservative allowlist of total `clojure.core` fns. Must stay conservative — e.g. `/` (div0), `nth` (bounds) CAN error.
[...]

Issue #258, opened 2026-06-18
Epic: Lowering & cross-target codegen


nnunley:
Umbrella for the lowering pipeline and keeping let-go's compile targets — the bytecode VM and the `-tags gogen_ir` native-lowered tree (`core_go_lowered/`) — correct and in lockstep.

## Members
- #175 — Lowering non-determinism: global gensym churns all of `core_go_lowered` on any `core.lg` edit
- #167 — Proposal: cross-target contract tests with per-platform `:expect`
- #260 — gogen: liveness over-reports coalesced dead LoadVar loads (blocks unconditional LoadVar snapshot materialisation)
- #262 — Close gogen_ir self-AOT lowering gaps (ir-stress 94% → 100%) — tracking issue; children #263–#267
- #268 — Consolidate IR op definitions into single-source catalog (multimethods/macros)
- #269 — Rewrite PersistentVectorSeq with cached leaf/tail traversal state
- #270 — Direct-call lowering (rt.InvokeValue → direct Go calls) — tracking issue; children #271–#272
- #274 — Flat index-RPN-direct bytecode representation (Level 2: drop block-args at the source)

### Codegen quality — leverage Go features in the `-tags gogen_ir` target
Surfaced via the fanout **bytes/loc ratio** analysis (the ratio, not absolute bytes, isolates verbosity / poor Go-feature leverage):
- #281 — gogen: elide callErr threading for non-erroring calls (~41% of the lowered tree)
- #282 — gogen: direct typed `(:keyword map)` access instead of a generic dynamic invoke
- #283 — gogen: CSE the lowered output (same pure call re-invoked N×)
- #284 — gogen: emit concrete Go types from typeinfer instead of `vm.Value` everywhere

## Workflow
File lowering/codegen and cross-target parity issues here as sub-issues. Each child should stand on its own (clear repro + path to fix); this epic tracks the set and its progress.
- #326 — gogen: native Go lowering of type-dispatched defmulti/defmethod multimethods
- #327 — gogen: lower the full jank Clojure-compat surface to 100% coverage

---

## Roadmap alignment (iterative-development, 2026-06-28)
Mapping these members to the internal roadmap epics/iterations (`docs/superpowers/iterations/roadmap.md`). Source of truth is the roadmap; this mirrors it for issue-tracker visibility.

**Delivered (closed):**
- #271 → EPIC-012 native-module direct calls — delivered by #343.
[...]

PR #285, opened 2026-06-20
perf-tooling(fanout): gate on lg→go expansion ratio + portable os/ls walk

nnunley:
## Why

The absolute-byte fanout gate trips on **coverage growth** (more defns lowering = the goal of #258), which masks the signal we actually want: is the generator emitting *more Go per byte of source* — i.e. not leveraging Go features. A density investigation (callErr threading is ~41% of the lowered tree, keyword lookups as dynamic invokes, `vm.Value` everywhere) motivated reframing the gate.

## What

1. **Gate on lg→go expansion ratio** (go-output-bytes / lg-source-bytes) instead of absolute bytes. Coverage growth adds to both numerator and denominator → ratio flat → no trip; only genuine output verbosity trips it. The `.lg` source is a **stable anchor** — unlike bytes/loc-of-go, whose denominator (go LOC) is itself a codegen artifact (the goto→structured change *removed* Go lines, so a bytes/loc gate would flag that improvement as a regression). Integer cross-multiply, no floats.

2. **Portable directory walk.** Replace the shell `find` in the file walkers with a recursive `os/ls` + `os/stat` traversal — no shell dependency, so it works on targets without one (WASM, sandboxed, Windows). `os/stat`'s `:size` also gives byte counts without reading files.

3. Baseline reset to the current expansion (**4.39×**, recorded as `:source-bytes` + totals).

## Validation
- Passes at parity (4393 vs 4393, +5% band).
- Trips correctly on a simulated verbosity increase (current 4393 vs a tighter baseline → FANOUT REGRESSION).
- Coverage-growth-immune by construction.

Companion to the codegen-quality backlog under #258 (#281–#284) — this gate is what measures progress on reducing lowered-code verbosity.

### DIFF
--- a/docs/perf/fanout-baseline.edn
+++ b/docs/perf/fanout-baseline.edn
@@ -1,26 +1,30 @@
-{:total-bytes 1283546
- :total-loc 49593
- :files 25
- :modules {"_wireup" {:bytes 4585 :loc 90 :files 3}
-           "edn" {:bytes 774 :loc 30 :files 1}
-           "hash" {:bytes 1826 :loc 62 :files 1}
-           "io" {:bytes 1069 :loc 39 :files 1}
-           "ir_build" {:bytes 322345 :loc 11968 :files 1}
-           "ir_data_generated" {:bytes 33347 :loc 922 :files 1}
-           "ir_dominance" {:bytes 23479 :loc 986 :files 1}
-           "ir_dump" {:bytes 54076 :loc 1933 :files 1}
-           "ir_lower" {:bytes 109950 :loc 4438 :files 1}
-           "ir_lower_go" {:bytes 321768 :loc 13510 :files 1}
-           "ir_passes" {:bytes 2454 :loc 67 :files 1}
-           "ir_passes_constfold" {:bytes 18681 :loc 914 :files 1}
-           "ir_passes_cse" {:bytes 8275 :loc 313 :files 1}
-           "ir_passes_dce" {:bytes 7161 :loc 277 :files 1}
-           "ir_passes_infer_arg_types" {:bytes 7424 :loc 339 :files 1}
-           "ir_passes_licm" {:bytes 70743 :loc 2410 :files 1}
-           "ir_passes_mutability" {:bytes 11699 :loc 465 :files 1}
-           "ir_passes_pipeline" {:bytes 97217 :loc 3220 :files 1}
-           "ir_passes_trace" {:bytes 23302 :loc 682 :files 1}
-           "ir_passes_typeinfer" {:bytes 106481 :loc 4861 :files 1}
-           "ir_validate" {:bytes 34102 :loc 1267 :files 1}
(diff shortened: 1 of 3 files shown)


mparrett:
**Codex Findings (5.5-high)**

1. **Merge blocker: fanout tests fail on this branch**  
   In [scripts/fanout-ratchet.lg](/Users/matt/Documents/Codex/2026-06-20/let-s-review-https-github-com/let-go/scripts/fanout-ratchet.lg:128), `base-src` is multiplied before any nil/legacy-baseline check. The existing test fixture in [test/fanout_ratchet_test.go](/Users/matt/Documents/Codex/2026-06-20/let-s-review-https-github-com/let-go/test/fanout_ratchet_test.go:92) has no `:source-bytes`, so `go test ./test -run Fanout -count=1` fails with `cannot multiply let-go.lang.Int and nil`. This matches the red GitHub `build` check.

**Non-blocking, for consideration/future concerns**

2. **The new ratio does not isolate coverage growth for existing source as intended.**  
   [scripts/fanout-ratchet.lg](/Users/matt/Documents/Codex/2026-06-20/let-s-review-https-github-com/let-go/scripts/fanout-ratchet.lg:157) uses all `.lg` bytes under `pkg/rt/core` as the denominator. If future work lowers more already-existing defs in already-counted source files, `cur-go` grows but `cur-src` does not, so the gate can still trip on coverage growth. The “coverage adds to both numerator and denominator” property only holds when proportional new `.lg` source is added.

3. **The portable walker regresses missing-directory behavior.**  
   [scripts/fanout-ratchet.lg](/Users/matt/Documents/Codex/2026-06-20/let-s-review-https-github-com/let-go/scripts/fanout-ratchet.lg:70) now calls `os/ls` directly. The old `find` wrapper returned `[]` for a missing tree, and the comment still promises that behavior. Now `show/check --no-regen` against a missing tree prints an interpreter error.

**Verification**

Ran `go test ./test -run Fanout -count=1`; it fails on the PR branch. Also checked the GitHub build log, which fails in the same `github.com/nooga/let-go/test` package.

**Re-check findings 2 and 3 (1 confirmed):**

- **Coverage-growth claim**: Verified with a fixture: baseline `go=1000 src=1000`, current `go=1100 src=1000`, threshold `5` trips the new ratio gate. So the concern is real, but it’s a **design/wording caveat**, not a merge blocker. Severity: **P2/P3 depending on intended contract**. If “coverage growth” only means adding new `.lg` source, the PR is fine; if it means lowering more existing defs, the PR claim is too broad.

- **Missing directory behavior**: Verified `origin/main` returns empty metrics for a missing tree; PR emits an interpreter error from `os/ls`. This is PR-scoped, but edge-case/tooling polish. Severity: **P3**, not merge-blocking.

Recommend scoping as: **one blocking CI regression**, plus **two non-blocking follow-ups/clarifications**.

nooga (approved):

Issue #268, opened 2026-06-19
Consolidate IR op definitions into single-source catalog (multimethods/macros)


nnunley:
Sub-issue of #258. (local `lg-6r45`)

Consolidate the scattered IR op definitions into single-source structures (catalog) driven by multimethods/macros, so op metadata (arity, local-carrying, lowering arm, numeric identities) lives in one place rather than being hand-duplicated across `build.lg` / `lower.lg` / `lower_go.lg` / `typeinfer.lg`. Reduces the per-op drift that shows up as lowering gaps.

Structural enabler for the gogen_ir gap-closing in #262; tracked here so it's visible at the epic level.

nnunley:
**Plan update (2026-06-22): deftype-unified iteration**

Mechanism decided — continue the `ir.ops` / `IROp` **deftype-per-op** skeleton from #310 (which reified `:const` as the first slice), rather than the original `defmethod`-handler idea (openspec task 2.2). Each IR op becomes a self-contained `deftype …Op` carrying its behavior facets (`op-infer`, `op-lower-stmts`, plus `op-value-expr` / arity / constfold as needed); static scalar props stay the generated `op_generated.go` columns.

Key addition: a **build-time exhaustiveness check** so a missing facet is a compile/test failure, not the invisible nil that motivated this issue. Ops not yet reified stay on an explicit *legacy allowlist* that shrinks each iteration.

Plan shape:
- **Walking skeleton:** reify `:inc`/`:dec` end-to-end + stand up the exhaustiveness check. Proof: omit a facet → loud failure naming op+facet; restore → byte-identical `core_compiled.lgb` + `core_go_lowered/`, `ir-stress lower-go` `:ok` unchanged.
- **Then op-at-a-time by family:** binary arith/comparison → loads & `:block-arg` → state/control (`:set-var :tail-call :recur :pop`) → terminators → `:call` → `:try` (the proof op) → interop & closures (`:dot :make-closure :push-closed`).
- **Close-out:** delete the legacy `cond`-on-op arms; derive `purity/pure-ops` (and the other duplicated op-set literals: `numeric-ops`, `comparison-ops`, `binary-op`, `commutative`, `unary-only-ops`) from the catalog/columns.

Each iteration is gated by: byte-identity, `ir-stress` `:ok` non-decreasing, the exhaustiveness sentinel test, and per-op isolation tests.

Structural enabler for #262 (gogen_ir self-AOT gap closing); tracked under epic #258.

[...]

Issue #262, opened 2026-06-19
Close gogen_ir self-AOT lowering gaps (ir-stress 94% → 100%)


nnunley:
Tracking issue for closing the remaining **gogen_ir self-AOT lowering** gaps — the `-tags gogen_ir` native-lowered tree (`core_go_lowered/`) lowering the full `pkg/rt/core` corpus with zero failures.

Sub-issue of #258. Surfaces the local `lg-h8gv` epic, **reconciled against `main` @ `d4c36cf2` (#251) on 2026-06-19**.

## Current state (authoritative repro)

`ir-stress.lg lower-go` over the full corpus: **709/749 (94%), 40 failures.**

```
go build -o /tmp/lg-stress .
LG_STRESS_PASSES=1 LG_STRESS_TIMEOUT_MS=8000 LG_STRESS_LOG=/tmp/s.log \
/tmp/lg-stress scripts/ir-stress.lg lower-go pkg/rt/core \
  core.lg walk.lg string.lg set.lg pprint.lg edn.lg io.lg async.lg hash.lg test.lg let-go/semver.lg graph.lg \
  ir/data.lg ir/data/generated.lg ir/zipper.lg ir/passes.lg ir/dominance.lg ir/structurize.lg ir/lower.lg ir/lower_go.lg ir/validate.lg ir/lattice.lg \
  ir/passes/dce.lg ir/passes/constfold.lg ir/passes/mutability.lg ir/passes/cse.lg ir/passes/typeinfer.lg ir/passes/licm.lg ir/passes/infer_arg_types.lg \
  ir/build.lg ir/passes/pipeline.lg ir/dump.lg ir/passes/trace.lg
```

A bucket is **done** when it disappears from `=== Failure Buckets ===`, `Passed` rises, and no new bucket appears.

> ⚠️ **Harness drift (fix first):** `*ti-counters*` moved from `ir.passes.typeinfer` to the new `ir.lattice` namespace, but `scripts/ir-stress.lg:340,368` still bind the old var, so the repro above **errors on current main** until those two lines are repointed to `ir.lattice/*ti-counters*` and `ir/lattice.lg` is added to the corpus list (as shown). Small standalone fix.

## Already closed since the original audit
- `:gogen/nth-non-int` (was 13) → **0, fixed** (local `lg-di6n` closed)
- `bad-identifier`/`%1` (was 4) → **0, fixed** (local `lg-2p8r` closed)

## Remaining buckets (sub-issues)
| bucket | count | note |
[...]

Issue #257, opened 2026-06-18
Epic: Clojure semantics parity


nnunley:
Umbrella for places where let-go diverges from Clojure semantics — whether an intentional, load-bearing deviation or an incidental gap. The goal is a single, roll-up-tracked backlog so divergences are visible and prioritizable, rather than scattered across issues or tied to the transient review that happened to surface them.

## Known deviations
- #256 — `set!` mutates var root instead of throwing when there's no thread-local binding
- #195 — top-level `(do ...)` isn't evaluated subform-by-subform (breaks `(do (require 'x) (x/f))`)

## Workflow
When a review or investigation surfaces a Clojure-semantics divergence:
1. Open a focused, independently-actionable issue for it.
2. Attach it here as a sub-issue (`gh api repos/nooga/let-go/issues/<this>/sub_issues -f sub_issue_id=<child id>`).

Each child should stand on its own (clear repro + path to fix); this epic just tracks the set and its progress.

PR #286, opened 2026-06-20
feat(gogen): make the gogen macro layer usable by the IR lowering pipeline

nnunley:
Makes runtime `macroexpand` resolve macros the way the compiler (and Clojure) do, restructures the go-gen sources by concern, and runs the code generators **in-process** so generation is deterministic. Together these close the ir-stress `unresolved symbol unquote` bucket (9 → 0) and let current-namespace macros lower.

## 1. `macroexpand`: full current-ns-first resolution

`macroexpand` (pkg/rt/lang.go) resolved symbols too narrowly, so macro calls reached `ir.build` unexpanded and failed on the literal symbol `unquote`:

- **Qualified** — the namespace part of a qualified symbol resolved against a **global** alias map only, so an aliased call like `(g/gquote …)` (where `g` is a per-ns `require` alias) never expanded.
- **Unqualified** — an unqualified symbol resolved against `clojure.core` only, never the **current** namespace, so a macro defined in the current ns (e.g. `ir.passes.trace/trace-pass`, used unqualified by `optimize-fn-traced`) did not expand.

Both now resolve **through the current namespace first** (`ResolveAlias` for the qualified case, `CurrentNS.Lookup` for the unqualified case), falling back to the global alias map / `clojure.core`. The existing `IsMacro` gate is unchanged, so non-macro vars are untouched.

Validated against Clojure 1.12: `(defmacro m [] 1) (macroexpand '(m)) => 1`; `(def x 5) (macroexpand '(x)) => (x)`; undefined head → unchanged. Re-lowering picks up exactly one module — `ir_passes_trace` (`trace-pass` now expands instead of reaching `ir.build` as a bogus call). New regression test: `test/macroexpand_current_ns_test.lg`.

## 2. Restructure go-gen by concern

`examples/go-gen/` was a grab-bag. Split:
- **gogen.lg** — generic Go-AST codegen macro lib (0 IR refs) → `pkg/rt/gogen/`, beside its Go-native half (`pkg/rt/gogen.go`), on the resolver load path via a new repo-root `deps.edn` `{:paths ["pkg/rt/gogen"]}`.
- **IR-generation specs** `ir_ops`/`ir_bridge`/`ir_data.lg` → `pkg/ir/` as proper namespaces (next to `op_generated.go`; aligns with #268).
- **demos** `example`/`numeric_specialize.lg` → `examples/gogen/`.

## 3. Run the code generators in-process (determinism)

`generate.lg` shelled out (`os/sh ./lg src`) to compile-and-run each generator and capture stdout. We already run inside `lg`, so the subprocess was redundant — and the captured stdout's trailing newline was dropped **nondeterministically**, so `generated.lg` flipped between `…\n` and `…\n\n` across runs (`make generate` was not idempotent). The two Go generators are masked by a `gofmt` pass; the Lisp generator (`ir_data`) has no formatter and was the lone offender.

`lg-gen` now loads each generator with `load-string` (sequential top-level eval, so the generators' own `defmacro`s register before use — a single `(do …)` `eval` compiles every form before any runs and the macro calls fail) and captures `*out*` via `with-out-str`, restoring the namespace the generator's `(ns …)` switches away from (`in-ns`; a `binding [*ns* *ns*]` doesn't restore it because the `ns` form mutates the process-global current ns). `ir_data` is loaded and `(emit-data-file)` evaluated as the final form, written with one atomic `spit`. The Go generators' output is **byte-identical** to the subprocess; `generated.lg` loses only its trailing blank line (canonical single newline) and is now stable across runs.

## Epics

Part of **#258 — Epic: Lowering & cross-target codegen** and **#257 — Epic: Clojure semantics parity**; advances **#262** (close gogen_ir self-AOT lowering gaps). Aligns with #268 (single-source op catalog).

## Verification

- `make generate` is byte-deterministic across runs (`op_generated.go` / `ir_bridge_generated.go` identical to committed; `generated.lg` stable); `make check-generated` + the genmanifest staleness test OK.
- `go build ./...` and `go build -tags gogen_ir ./...` succeed; `go test ./test ./pkg/ir ./pkg/rt` green; the clojure-compat suite passes under `-tags gogen_ir`.
- ir-stress full corpus: `unresolved symbol unquote` 9 → 0.


nnunley:
**Corpus verification — this clears the quasiquote bucket.**

This move puts the `gogen` Lisp macro layer (`gquote`/`unquote`/`unquote-splicing` + helpers) on the resolver load path via `pkg/rt/gogen/` + repo-root `deps.edn {:paths ["pkg/rt/gogen"]}`. Before the move those 9 `ir/build: unresolved symbol unquote` / `unquote-splicing` rows were a **symptom of an unexpanded macro** (the `g/gquote` template reached `ir.build` raw because `gogen.lg` wasn't loadable when the harness lowered a corpus file), not a lowering gap.

**After (full `ir-stress.lg corpus`, `LG_STRESS_TIMEOUT_MS=15000`):** the `unquote`/`unquote-splicing` bucket is **gone** — those defns now macroexpand and lower the expansion. Broad corpus sits at **1193/1199 (99.5%)** `:ok`.

**Build pipeline intact + byte-identical:** `make generate` reads `pkg/rt/gogen`; `op_generated.go` / `ir_bridge_generated.go` / `ir/data/generated.lg` regenerate identically except the provenance-header path string; `make check-generated` OK. `pkg/ir` / `pkg/rt` / `test` green (incl. `language_test.go`, which loads gogen from the new path).

The Go-native gogen half (`pkg/rt/gogen*.go`) is untouched — it installs at startup and is why core `ir.lower-go` lowers without any `.lg` require. Only the Lisp macro half moved.

Tracking: this is the last PR in the #258 self-AOT close-out; the remaining 5 corpus residuals are reclassified (gaps #289/#290, perf #292, stale-script #293, 2 harness `:exclude` candidates) — see #262 for the full table.

mparrett:
Final review pass on latest head `fa27cf1`:

No blocking findings for the current PR changes.

One non-blocking/pre-existing follow-up remains: `macroexpand` now correctly handles qualified namespace aliases like `g/gquote`, but still does not resolve unqualified macros from the current namespace. I reproduced `(defmacro m [] 1) (macroexpand '(m)) => (m)` on both base `8087f99` and this PR head, so this is not introduced by #286. Only worth tracking separately if runtime `macroexpand` is intended to fully mirror compiler symbol resolution.

Verification run locally:
- `go run ./cmd/check-generated` passes.
- `go test ./pkg/genmanifest` passes.
- `(require '[gogen :as g]) (macroexpand '(g/gquote 1))` expands to `(gogen/int-lit 1)`.

The earlier generated-output staleness I saw on the older head is fixed in the current PR head.

nooga (approved):

PR #294, opened 2026-06-21
feat(gogen): elide callErr check for total keyword-callee calls

nnunley:
## What

A keyword callee `(:k m)` lowers to `rt.InvokeValueEC`, whose error is always nil at arity 1–2 — `vm.Keyword.Invoke` (`pkg/vm/keyword.go`) only returns a non-nil error for arity outside `[1,2]`. A `:call` node's arg-count is statically known, and a keyword callee is a `:const` (not a redefinable var), so such a call is **provably error-free**. The trampoline still emitted the 3-line `if callErr != nil { return … }` block for it.

This adds a narrow `keyword-callee-total?` predicate; when it holds, the default trampoline emits `v, _ = rt.InvokeValueEC(...)` and skips the dead check.

## Impact

Removes **462 of 9840** `callErr` blocks in the lowered tree. Behavior is identical (the elided error was always nil). Conservative first slice of #281 — only compile-time keyword constants, which can't be invalidated by redefinition; var-mediated calls are untouched, guarded by a negative test.

## Verification

- New `test/gogen_keyword_callerr_test.lg` (TDD RED→GREEN): keyword calls drop the check, a var-mediated call keeps it (no over-elision).
- `make check-generated` OK; whole-corpus `go build -tags gogen_ir ./...` OK.
- `go test ./test/ ./pkg/ir/... ./pkg/rt/...` all green.

Part of #281, epic #258.


mparrett:
Looks good. Codex found nothing.

I reviewed `nooga/let-go#294` against the PR diff and a detached worktree at `/private/tmp/let-go-pr294`. The optimization looks narrowly scoped: it only elides `callErr` for constant keyword callees with statically known arity 1 or 2, matching `vm.Keyword.Invoke`’s error behavior.

Verification run:

```text
go run ./cmd/check-generated
go test -run TestRunner/gogen_keyword_callerr_test.lg -count=1 -v ./test
go test -count=1 ./pkg/ir ./pkg/rt ./pkg/vm ./test/...
```

All passed. Residual risk is low; I did not see a behavioral regression or stale generated artifact issue.

nooga (approved):

PR #295, opened 2026-06-21
fix(ir-stress): close the (do ...) wrapper on its own line in read-all

nnunley:
## Bug

`read-all` wraps each corpus file as `(do <src>)` and `read-string`s it, but appended the closing paren with **no newline**. A file ending in a `;;` line comment without a trailing newline swallowed that `)` into the comment — leaving `(do` unclosed, so `read-string` hit EOF. The runtime file-loader doesn't wrap, so these files loaded fine standalone; only the ir-stress harness reported `READ-ERROR`.

This had been misattributed to a reader bug ("close-paren in a string in a line comment") and worked around by excluding the affected files — masking the real harness defect.

## Fix

Append the closer on its own line (`"\n)"`), keeping `(do ` inline so file line numbers still map 1:1 into the wrapped form for error reporting.

Drop `test/simple.lg`, `test/hello.lg`, `test/perf.lg` from the corpus `:exclude` list — they were excluded as "reader errors" but read correctly now (each carries 0 `defn` fixtures, which is correct). A full corpus run reports zero READ-ERRORs.

### DIFF
--- a/scripts/ir-stress-corpus.edn
+++ b/scripts/ir-stress-corpus.edn
@@ -1,9 +1,8 @@
 {:doc
  "IR-stress lower-go corpus: every repo .lg the AOT lowering should handle. Failures against :stress are real lowering gaps - forward/mutual-ref resolution, unsupported arg patterns, gogen emission, lowering-cost timeouts. Widest net: core via go:embed all:core plus tests, examples, scripts; external deps are themselves lowered .lg so they stay IN. Consumer: make ir-stress runs lower-go on :stress minus :exclude. Regenerate the file list via scripts/ir-stress.lg."
+ ;; Nothing excluded: every listed file must read and lower.
  :exclude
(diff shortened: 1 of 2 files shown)


mparrett:
Codex Findings (5.5-high): none.

I reviewed `nooga/let-go#295` against the changed harness behavior and ran verification in an isolated temp checkout.

Verified:
- `make build`
- `env LG_STRESS_PASSES=1 ./lg scripts/ir-stress.lg lower-go . test/simple.lg test/hello.lg test/perf.lg`
- `make ir-stress`

The targeted run reports all three formerly excluded files as `0/0 ok`, not `READ-ERROR`. The full corpus run completed successfully with no read errors over 245 files; remaining failures are existing lowering buckets, not introduced by this patch.

nooga (approved):

Issue #175, opened 2026-06-03
Lowering non-determinism: global gensym churns all of core_go_lowered on any core.lg edit


nnunley:
## Problem

On `main` (and branches off it, e.g. `feat/malli-example`), editing any
`pkg/rt/core/**/*.lg` file and regenerating churns the **entire**
`pkg/rt/core_go_lowered/` tree — thousands of lines across files the edit
never touches.

Root cause: `gensym` is a **global counter**, so the temp-symbol numbering in
every lowered function depends on global compilation state. Adding a single
`defn` to `core.lg` shifts the counter and renumbers temps in every
downstream-lowered namespace, producing a huge, meaningless diff.

### Evidence (from upstreaming the malli core fns, PR #172)

Adding ~9 leaf fns to `core.lg` produced a working-tree diff of **~7500 lines
across 21 `core_go_lowered/*.go` files** — none of which my fns appear in
(`core` is `skip-go`, so its fns live only in `core_compiled.lgb`). The churn
was 100% gensym renumbering. I had to `jj restore pkg/rt/core_go_lowered/`
before committing to keep the PR reviewable.

Note `core_go_lowered/` is listed in `.gitignore` (intended as a build
artifact), yet 21 of its files are still **tracked** — so the churn lands in
commits unless manually restored. (Untracking them is a related cleanup.)

## The fix already exists

A fix is implemented in the `.workspaces/direct-call-registry` workspace
(`pkg/rt/core/ir/build.lg`): a **context-scoped gensym** plus a **per-function
`lower-go` renumber**, which makes lowering output independent of global state.
With it, fresh regeneration is byte-identical to the committed artifacts
(`IDENTICAL? => true`). It just hasn't been landed on `main`.

## What landing it involves

- Isolate the ctx-gensym change from the rest of the direct-call-registry work
  (it touches `build.lg` + the per-fn renumber in `lower_go.lg`).
- Apply onto `main`; verify with `make check-generated` and a fresh-regen
  `IDENTICAL?` check.
- (Related, optional) untrack `pkg/rt/core_go_lowered/` so it matches its
  `.gitignore` intent.

The deferred dynamic-var gensym counter was judged redundant once ctx-gensym +
per-fn renumber are in place.

## Why it matters

Until this lands, every `core.lg` edit produces an unreviewable diff and a
`make check-generated` failure unless the lowered tree is hand-restored.

nnunley:
[...]

PR #299, opened 2026-06-21
test(determinism): add lowering determinism property test (#175)

nnunley:
## Summary

Adds a regression test for the lowering determinism fix to guard against future non-deterministic code generation.

The test verifies that two runs of `lgbgen --target=go` produce byte-for-byte identical `pkg/rt/core_go_lowered/` output, which is critical because the lowering encodes inferred types and generated code into Go source files. Non-determinism in token positions, type inference, or code generation would break reproducible builds.

This test is informed by the determinism fix (ctx-gensym threading + per-fn renumbering) and ensures that fix remains stable.

## Test plan

- [x] Runs `TestLoweringDeterminism` directly: `go test ./test/e2e/ -run TestLowering`
- [x] Skipped under `testing.Short()` (runs lgbgen twice, expensive)
- [x] Full test via `make test` or `go test ./test/e2e/`

---

Part of #175.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -62,14 +62,18 @@ jobs:
       # default: they run each script under let-go and compare to the committed
       # test/gold/*.out goldens. Re-deriving those goldens from a LIVE Clojure is
       # done in the dedicated `gold-differential` job (no JVM dependency here).
+      # TEMPORARY (#299): per-package test timeout raised 60s→180s so the
+      # lowering-determinism harness (runs lgbgen twice, ~60-120s on CI) fits.
+      # Revert to 60s once that test is moved to a dedicated long-timeout step.
       - name: Test with bundled bytecode stdlib
-        run: go test -v -timeout 60s ./... -skip TestClojureTestSuite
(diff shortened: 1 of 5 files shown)


nooga (approved):

nooga:
Found one issue:

- **[P2] Test rewrites the real checkout while `go test` is running.** `generateLoweredTree` runs `lgbgen --target=go` from the repo root without passing an output directory, then copies `pkg/rt/core_go_lowered` from the checkout. That means `go test ./...` can rewrite tracked generated files and wireup files while other package tests are being built or run. Even if the final tree happens to be byte-identical, this is a source-tree side effect in a test and can race with the Go test driver. `lgbgen` already supports `--target=go <out-dir>`, so the test should generate into isolated temp dirs, or run in a temporary repo copy if the fixed wireup files also need to be included.

Verification: `git diff --check` against the PR base passed, and `go test ./test/e2e -count=1 -run TestLoweringDeterminism -v` passed in about 141s in an isolated worktree. The worktree ended clean only because the generated output matched the checked-in files.

PR #300, opened 2026-06-22
perf(vm): PersistentMap.Equals walks in place instead of materializing nodeSeq

nnunley:
## What

`PersistentMap.Equals` called `m.root.nodeSeq()` to iterate, which **materializes the entire map into a `[]MapEntry` slice** (recursively allocating a slice per HAMT node) on *every* comparison. The IR typeinfer fixpoint compares maps constantly.

A heap profile of one bootstrap lowering pass (`go run -tags bootstrap ./cmd/lgbgen --target=go`) showed `hmapBitmapNode.nodeSeq` at **30 GB — 53% of all allocation**, of which **98.95% was driven by `Equals`**.

## Change

- Add an in-place `each(fn func(key, val Value) bool) bool` visitor to the `hmapNode` interface and both node types (`hmapBitmapNode`, `hmapCollisionNode`).
- Rewrite `Equals` to walk one map calling the other's `find()` per entry — **zero slice materialization**. Sizes are compared first, so one-directional containment is sufficient.
- Add a `-memprofile` flag to `lgbgen` (mirrors the existing `-cpuprofile`), used to capture the profile.

## Measured (one lowering pass)

| Metric | Before | After | Δ |
|--------|--------|-------|---|
| Allocated | 56.2 GB | 27.2 GB | **−52%** |
| Wall time | ~106s | ~50s | **−53%** |
| GC (`gcBgMarkWorker`) | 41.5s | 21.5s | **−48%** |

The generated `core_go_lowered/` tree is **byte-identical** before/after — semantics unchanged. Existing `pkg/vm` tests pass; added `TestPersistentMapEqualsDeep` exercising multi-level HAMT, value/key/size divergence.

### DIFF
--- a/cmd/lgbgen/main.go
+++ b/cmd/lgbgen/main.go
@@ -18,6 +18,7 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"runtime"
 	"runtime/pprof"
 	"sort"
 	"strings"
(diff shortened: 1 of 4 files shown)


nooga (approved):

nooga:
No issues, please rebase.

PR #303, opened 2026-06-22
feat(lg): add -cpuprofile and -memprofile flags

mparrett:
Adds optional `-cpuprofile` and `-memprofile` flags to a profiling build of `lg`, wiring `runtime/pprof` around script execution without changing the default release binary.

## What

Default builds stay size-neutral and do not expose the profiling flags:

```
make build
./lg script.lg
```

Profiling builds opt in with the `lg_profile` build tag:

```
make build-profile
./lg-profile -cpuprofile=cpu.pprof script.lg     # CPU profile of the run
./lg-profile -memprofile=mem.pprof script.lg     # heap profile after the run (post runtime.GC())
```

The profiling flags are no-ops when unset in the profiling binary, and are not compiled into the default `lg` binary.

## Why a separate profiling build

Binary size matters for `lg` itself and for `lg -b` bundled executables, which copy the base binary. The first version of this PR added about 33.6 KiB to the stripped default binary. Moving the CLI profiling surface behind `-tags lg_profile` brings the default `lg` binary back to the pre-PR size while keeping the profiling workflow available for perf work.

## Why explicit, not deferred

Profiling is scoped to the script/REPL execution section of `main()`, and started/stopped explicitly rather than with `defer`. `main()` exits the version/compile/bundle/wasm paths through `os.Exit`, which skips deferred functions. `pprof.StopCPUProfile()` must run for a CPU profile to be valid.

The profiling build also registers the stop path with `rt.AtExit`, so language-level `os/exit` and `System/exit` flush active profiles before they call `os.Exit`. Hard signals and Go-level panics still bypass it, matching normal `os.Exit` semantics. Build modes (`-c`/`-b`/`-w`) and the bundled-binary path are not profiled, as the flag help states.

## What it found

Profiling a bench that's already in the tree:

```
./lg-profile -memprofile=mem.pprof test/benches/pegbench.lg
go tool pprof -top -sample_index=alloc_space mem.pprof
```

points straight at `clojure.core/subs`, which accounted for ~90% of allocation in the run (and most of the `string->rune` CPU cost). It allocated a `[]rune` of the entire input on every call. I'll send that fix as a separate PR; this one is just the optional profiling build.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -141,6 +141,7 @@ fabric.properties
 /letgo
 /let-go
 /lg
+/lg-profile
 /lg-gogen
 /letgo.wasm
 
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,7 @@ endif
 
 # Prefer - to _ for make var names (won't conflict with env vars):
 LG := lg
+LG-PROFILE ?= lg-profile
 GOLANGCI-LINT := github.com/golangci/golangci-lint/cmd/golangci-lint
 REPORT-SCRIPT := scripts/clojure_compat_report.sh
 
@@ -48,6 +49,8 @@ run: $(LG)
 
 build: $(LG)
 
+build-profile: $(LG-PROFILE)
+
 # Bundle target. The runtime loads compiled bytecode for the core
 # namespaces from this file, NOT from the .lg sources. Anyone editing
 # a .lg under pkg/rt/core/ must regenerate the bundle or runtime
@@ -56,6 +59,7 @@ build: $(LG)
 # the bundle in lockstep with the .lg sources.
 CORE-LG-FILES := $(shell find pkg/rt/core -name '*.lg' -type f 2>/dev/null)
 LGBGEN-SOURCES := $(shell find cmd/lgbgen -name '*.go' -type f 2>/dev/null)
+ROOT-GO-FILES := $(shell find . -maxdepth 1 -name '*.go' -type f 2>/dev/null)
 pkg/rt/core_compiled.lgb: $(CORE-LG-FILES) $(LGBGEN-SOURCES) $(GO)
 	go run -tags bootstrap ./cmd/lgbgen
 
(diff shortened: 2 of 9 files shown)


mparrett:
Heads-up on something adjacent that I deliberately left out of scope here.

The `AtExit` hook this PR adds is registered only by profiling. `ShutdownAllPods` is in the same situation it solves — it's a `defer` in `main`, so `os/exit` and `System/exit` skip it too, and a script that loads a pod and then exits that way orphans the pod's child process. The hook could close that gap by registering `ShutdownAllPods` as well.

I held off because it's a behavior change, not just a flush: `Pod.Shutdown` calls `cmd.Wait()`, so routing it through the hook turns `os/exit` from an immediate exit into one that blocks until each pod subprocess dies — and a wedged pod could hang it. Profiling has no such risk, which is why I kept this PR to that.

If it's worth doing, the safe shape is a bounded wait around `cmd.Wait()` plus a test for the load-pod-then-`os/exit` path. Happy to send it as a follow-up, or leave pods as they are. Your call.

nooga:
Does this have any impact on our binary size?

mparrett:
@nooga Great question about binary size.  Please forgive my laziness. I asked codex to do the grunt work.

**TL;DR:** Yes, and enough to matter. If we want to keep this, it's almost certainly best to have a separate build for profiling.

```
Yes, but small.

For the stripped `lg` binary from `make build`, this PR changes:

13,022,546 bytes -> 13,056,146 bytes
delta: +33,600 bytes, about +32.8 KiB / +0.26%

Most of that is text/code. `runtime/pprof` was already linked before this PR via the existing `profile/cpu-start!` runtime namespace, so this is not pulling in pprof from zero. The new retained pieces are mostly the heap-profile writer path used by `-memprofile`, plus the small CLI wrapper and `rt.AtExit` hook.

What is affected:
- the normal `lg` CLI binary: +33,600 bytes
- `lg -b` bundled executables: same +33,600 bytes, because bundles copy the base `lg` binary and append the same payload
- `lg -w` generated WASM: `main.wasm` grew by +1,669 bytes in my tiny-script check; external-mode HTML/service worker were unchanged

What is not affected:
- compiled `.lgb` bytecode output; same 101-byte output in my tiny-script check, byte-identical
- module dependencies; no new external dependency
- build/compile/bundle paths are not profiled at runtime unless these flags are used
```

mparrett:
@nooga It's bigger lift and out of scope here, but did an experiment to see the impact of dropping/conditionalizing the existing `profile/cpu-start!` — shaves an additional `84 KiB` off `lg`.

mparrett:
Updated this PR to address the binary-size concern by making the CLI profiling surface opt-in:

- default `make build` no longer compiles in `-cpuprofile` / `-memprofile`
- new `make build-profile` builds `lg-profile` with `-tags lg_profile`
- default stripped `lg` is back to `13,022,546` bytes, matching the pre-PR size
- `lg-profile` is `13,056,114` bytes, so the ~33.6 KiB cost is paid only by the profiling binary
- `lg -b` bundles built from the default binary no longer inherit this PR's profiling-size delta

I also kept the `os/exit` / `System/exit` flush fix, but only in `lg_profile` builds; default builds get a no-op hook stub.

nooga (approved):

PR #304, opened 2026-06-22
perf(rt): subs without materializing the whole string to runes

mparrett:
`clojure.core/subs` allocated a `[]rune` of the entire input string on every call, then sliced and re-encoded it — O(input) allocation regardless of substring size. In a parser, where `subs` runs constantly over the source, that dominated allocation.

## Fix

Map the start/end rune indices to byte offsets in a single pass with `utf8.DecodeRuneInString`, stopping once both offsets are known, then slice the original string. Character-index semantics are unchanged.

## Numbers

A/B microbenchmark of the core (a ~6KB source, a short token sliced near the front — the common parser case):

```
BenchmarkSubsOld   7393 ns/op   27264 B/op   1 allocs/op
BenchmarkSubsNew     25.6 ns/op      0 B/op   0 allocs/op
```

The old cost is the whole-string `[]rune` (~6800 runes × 4 bytes) on each call. The new path is strictly better even worst-case: a substring at the end of a long string is an O(n) walk, but still allocates nothing, versus the old O(n) walk *plus* O(n) allocation.

How it surfaced: profiling `test/benches/pegbench.lg` with `-memprofile` (the flags from #303) showed `subs` at ~90% of the run's allocation.

## Correctness

Verified against ASCII, multibyte (`é`), and 4-byte-rune (emoji) inputs, plus the boundary cases (`start == length`, empty slice, full string) and the out-of-range / `end < start` errors. The existing suite passes.

## Tradeoff

The result now shares the source string's backing array (normal Go string-slice behavior) instead of being a fresh copy. This is what removes the allocation, but a long-lived substring keeps the whole source string alive. For parse-and-discard that's pure win; if let-go has a path that retains many small substrings of large sources indefinitely, that's the case to weigh. Happy to add a copy on the result if you'd rather not share backing.

### DIFF
--- a/pkg/rt/lang.go
+++ b/pkg/rt/lang.go
@@ -21,6 +21,7 @@ import (
 	"sync"
 	"sync/atomic"
 	"time"
+	"unicode/utf8"
 	"unsafe"
 
 	"github.com/nooga/let-go/pkg/vm"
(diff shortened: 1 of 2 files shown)


mparrett:
xsofy itself hits this, beyond the pegbench bench in the description.

The replay-code panel wraps a long code into width-bound lines by recursively re-slicing the remainder ([render.lg#L999-L1002](https://github.com/nooga/xsofy/blob/6f774be24bac3e53aeae277a266773eab40fa04c/xsofy/render.lg#L999-L1002)):

```clojure
(loop [s replay-code acc []]
  (if (<= (count s) cw)
    (conj acc s)
    (recur (subs s cw) (conj acc (subs s 0 cw)))))
```

Each recursion calls `subs` twice, and with the old whole-string `[]rune` every call re-materializes the entire remaining string, so wrapping an n-character code is O(n²). Modeling that loop on a 256-char code at width 72:

```
old   6479 ns   5200 B   15 allocs
new   1697 ns    112 B    3 allocs   (residual is the lines vector, not subs)
```

The same repo also truncates text to a panel width per frame in a few render paths (`render.lg:969`, `render.lg:1036`, `ui.lg:54/64/85` — all `(subs s 0 w)`); each drops from an input-sized `[]rune` to zero allocation.

Posting as a downstream data point in case it helps weigh the change.

mparrett:
Added regression tests for the cases the description claims to verify. `1453510` adds `test/subs_test.lg` covering:

- ASCII slicing (two-arg, one-arg-to-end, empty slice, full string),
- multibyte (`héllo`, where `é` is two bytes) and 4-byte runes (emoji), to pin the codepoint-index semantics the byte-offset walk has to preserve,
- the boundary cases (`start == length`), and the out-of-range / `end < start` errors.

They're auto-discovered and run by `test/language_test.go`'s `TestRunner`, so the rune-index equivalence is now guarded against regressions rather than only asserted in the description.

nooga (approved):

nnunley (approved):

PR #307, opened 2026-06-22
fix(term): tokenize read-key so a key burst yields one event per key

mparrett:
`read-key` did a single `os.Stdin.Read` into a 16-byte buffer and returned the
whole read as one string. A fast burst — held-key auto-repeat, or input queued
while the VM was busy — therefore arrived as one multi-char blob (`"llll"`)
that a game's key parser rejects as unknown. Symptom: holding a movement key in
a native TUI never produced continuous movement.

This is original behavior, not a regression — `read-key` has globbed reads
since before the I/O-seam work (#165 wake-pipe, #244 KeySource), both of which
preserved it byte-for-byte.

## Fix

- `nextKey` (new `pkg/rt/key_tokenizer.go`): a pure function that splits exactly
  one key off the front of a byte slice — a UTF-8 rune, a CSI/SS3 escape
  sequence, or a single byte — and reports the bytes consumed.
- `nativeKeySource.ReadKey` now buffers raw reads and hands out one token per
  call, refilling from stdin only when the buffer drains. The blocking
  poll/SIGWINCH-wake read path is unchanged, just renamed to `readRaw`.
- `KeyPending` reports buffered tokens as well as pending stdin.

The tokenizer lives in a constraint-free file (no build tags) so the wasm key
source can reuse it later.

## Behavior

- `"llll"` → four `"l"` (the fix).
- `"\x1b[C"` (arrow) stays one key; modified keys and SS3 (`"\x1bOP"`) stay whole.
- SGR mouse reports (`"\x1b[<0;10;5M"`) are kept intact as single tokens —
  decoding them is out of scope here.

## Testing

- `pkg/rt/key_tokenizer_test.go`: 14 tokenizer cases (bursts, arrows, mixed
  printable + escape + burst, UTF-8, control bytes, empty input).
- `go test ./pkg/rt/ ./pkg/api/` pass; `go vet` and `golangci-lint` clean.
- Validated end-to-end in a native TUI: a 5-`l` burst now produces multiple
  moves; held-key movement works.

## Notes

- A CSI sequence split across two reads (no final byte yet) is returned
  best-effort as the remaining buffer. Raw-mode reads deliver escape sequences
  atomically, so this is rare in practice.
- Native has a single, non-concurrent stdin, so the buffer is package-level.

### DIFF
--- a/pkg/rt/key_tokenizer.go
+++ b/pkg/rt/key_tokenizer.go
@@ -1,0 +1,124 @@
+package rt
+
+import "unicode/utf8"
+
+// nextKey splits exactly one key token off the front of b and returns the
+// token plus the number of bytes consumed.
+//
+// A single raw terminal read can carry several keys (held-key auto-repeat or
+// queued input) or one multi-byte escape sequence (arrows, SGR mouse). The
+// pre-tokenizer read-key returned the whole read as one string, so a burst
+// like "llll" arrived as one unrecognized blob. Tokenizing lets read-key hand
+// out one event per call: "llll" -> four "l", while "\x1b[C" stays one key.
+//
+// Recognized fronts:
+//
+//	ESC '[' … final(0x40-0x7E)   one CSI sequence (arrows, modified keys, SGR mouse)
+//	ESC 'O' x                    one SS3 sequence (application-cursor keys, F1-F4)
+//	ESC (alone / other)          the bare Escape key
+//	UTF-8 lead byte              the whole rune (never split)
+//	any other byte               that single byte (ASCII / control)
+//
+// A token split across two reads (an unterminated CSI, or a multi-byte rune
+// missing continuation bytes) is returned best-effort as the remaining buffer.
+// ReadKey avoids hitting that path: it checks incompleteToken and refills first
+// when more bytes are pending, so nextKey runs on a complete token.
+func nextKey(b []byte) (string, int) {
+	if len(b) == 0 {
+		return "", 0
+	}
+	if b[0] == 0x1b { // ESC
+		if len(b) >= 2 && b[1] == '[' { // CSI: ESC [ params/intermediates final
+			i := 2
+			for i < len(b) && (b[i] < 0x40 || b[i] > 0x7e) {
+				i++
+			}
+			if i < len(b) {
+				return string(b[:i+1]), i + 1 // include the final byte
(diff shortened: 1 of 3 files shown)


mparrett:
**Codex Findings (5.5-high)**

- `pkg/rt/term.go:123` / `pkg/rt/key_tokenizer.go:39`: `ReadKey` tokenizes only the current `readRaw` chunk, but `readRaw` still reads
**Findings**

- `pkg/rt/term.go:123` / `pkg/rt/key_tokenizer.go:39`: `ReadKey` tokenizes only the current `readRaw` chunk, but `readRaw` still reads into a fixed 16-byte buffer. If a UTF-8 rune or CSI/SGR sequence crosses that chunk boundary, `nextKey` emits the partial bytes as a complete token instead of waiting for the rest. Example: a queued 15 ASCII bytes followed by `é` emits the first UTF-8 byte as an invalid token; a long SGR mouse report over 16 bytes emits an incomplete `ESC[` token and then the remainder separately. This weakens the new “UTF-8 rune kept whole” / “SGR mouse kept intact” contract and will be hard for callers to recover from. I’d either make incomplete UTF-8/CSI return “need more bytes” and refill, or make `readRaw`/buffering accumulate until a complete token is available.

**Verification**

- Ran `go test ./pkg/rt ./pkg/api` in a PR worktree: passed.

mparrett:
Good catch — fixed. The tokenizer now stitches a token that straddles a read instead of emitting the partial.

`ReadKey` checks `incompleteToken` on the front token before tokenizing — an unterminated CSI/SS3, or a UTF-8 lead byte missing continuation bytes — and refills when more bytes are actually pending on stdin. Both cases you flagged now hold: a long SGR report over the buffer size, and `é` after a queued ASCII run, each come back as one token. Gating the refill on pending input also resolves the lone-ESC case without a timer — a real Escape keypress has nothing queued behind it.

I also widened the read buffer 16 → 256, but that's only fewer syscalls/refills under bursts — correctness is the refill path, not the buffer size (even a 1-byte buffer tokenizes correctly now).

Added table tests for `incompleteToken` (escape + multi-byte UTF-8 boundaries) and verified end to end: 255 ASCII bytes + `é` straddling the buffer edge stitch back to one `é`.

nooga (approved):

PR #309, opened 2026-06-22
fix(lgbgen): bundle ir.passes.purity so a fresh regen doesn't nil dce/removable? (#258)

nnunley:
## Problem

`lg-dq58-phase3b` (071622ea) split `ir.passes.purity` out as the canonical purity pass and made `dce`/`cse`/`licm` `:require` it — but never added `ir.passes.purity` to `goLoweringNSOrder` in `cmd/lgbgen/main.go`. The bytecode bundle is compiled from that explicit list (`orderedEmbeddedNS`), so a fresh `make generate` omits purity entirely. At runtime `purity/effect-free-inst?` resolves to nil and `dce/removable?` crashes with `TypeError: nil is not a function`, breaking `./pkg/ir/...`.

The committed `core_compiled.lgb` predated the `:require` edge, so it masks the bug — but `make check-generated` already reports the artifacts **STALE** on a clean checkout. Anyone running the required post-edit `make generate` hits the failure.

## Fix

Register `ir.passes.purity` in `goLoweringNSOrder`, ordered after its dep (`mutability`) and before its dependents (`dce`/`cse`/`licm`). `purity` carries `lgbgen:skip-go`, so it is bundled but not Go-lowered.

## Verification (clean jj workspace off main)

- Regenerated bundle now includes purity (32 → 33 namespaces)
- `./pkg/ir/...` passes **untagged** and **-tags bootstrap** (was failing)
- `make check-generated` green — lockstep + native dispatch
- `go build -tags gogen_ir ./...` clean
- `go test ./pkg/genmanifest/...` (freshness gate) green

## Follow-on (not in this PR)

`goLoweringNSOrder` is a hand-maintained second source of truth that drifted here. `discoverEmbeddedNS` already implements `:require`-graph topological ordering but is unused for the bundle; the lowered Go tree should use real Go imports for compile-time-enforced transitive deps + collision detection.

---

Part of epic #258.


mparrett:
**Codex Review (5.5-high) : Found one blocking issue.**

- **[P1] DCE now removes unused `quot`, suppressing divide-by-zero**: [purity.lg](/private/tmp/let-go-pr309/pkg/rt/core/ir/passes/purity.lg:41) adds `:quot` to the shared `pure-ops`, and [dce.lg](/private/tmp/let-go-pr309/pkg/rt/core/ir/passes/dce.lg:20) removes any unused `effect-free-inst?`. This changes observable behavior in the IR pipeline. A temporary repro showed:
  - normal bytecode: `(defn dead-quot [] (let [unused (quot 1 0)] :ok))` errors with `divide by zero`
  - PR Lisp IR pipeline: returns `:ok`
  
  Since let-go treats `quot` divide-by-zero as an error, DCE should not classify `:quot` as effect-free unless the project intentionally accepts dropping traps in IR-compiled code. The safer fix is to keep `:quot` out of DCE’s removable set, even if LICM/CSE keep their own narrower definition.

I didn’t find issues with the actual bundle-order fix; `mutability -> purity -> dce/cse/licm` is the right direction. Verification note: I ran a focused temporary repro in `/private/tmp/let-go-pr309` and removed it afterward; I did not run the full suite.

nooga:
Did not find anything besides @mparrett 's findings

nnunley:
Both addressed in e6a30b5.

1. quot trap: split value-purity from effect-freedom. pure-ops stays value-pure so CSE/LICM keep applying via pure-inst?; a new trapping-ops set (currently :quot) is excluded from effect-free-inst?, the DCE predicate, so an unused (quot _ 0) survives and still traps. Regression test added: dead trapping quot is kept, dead non-trapping (+ 1 2) is still removed.

2. The omission this PR patched was a silent footgun in the hand-maintained goLoweringNSOrder. validateGoLoweringOrder now fails the build, naming the dependents, whenever a listed ns :requires a bundle-able (non-lgbgen:skip) ns that is missing. Verified silent on the correct list, loud when purity is removed (flags dce/cse/licm). It touches no ordering, so lowering stays deterministic (bundle byte-identical across regens). Full topo-sort derivation is deferred on purpose: discovery skips the lgbgen:skip namespaces (core/string/set/graph) the lowering bundle still needs, so auto-derivation is a larger, separately-verified change.

nnunley:
Follow-up (a0589399): did the full topo-sort derivation, replacing the hand list + the completeness check.

The audit found the only obstacle was `graph` carrying a stale `lgbgen:skip` "runtime intern blocks" header — it has none (pure toposort/SCC functions, `:require`d by lower-go and pipeline). Removed the directive entirely: graph is now bundled (runtime-visible for the passes that call `graph/sccs` and `graph/toposort-by`) AND lowered like any normal namespace — its multi-arity defns (`toposort`, `sccs`) fall back to bytecode per-function, its single-arity defns (`toposort-by`, `cyclic?`, `has-cycle?`, `reachable`) lower to Go (`graph.go`, 4 defns). With graph discoverable, the derived set equals the old hand list exactly, so `deriveGoLoweringOrder` just topo-sorts by `:require` plus one synthetic edge `:require` cannot express: `ir.build` after the analysis/inference passes (sorted, so the bundle stays deterministic). `ir.data` stays full-skip — it has a genuine runtime re-intern block; its bundle-able form is `ir.data.generated`.

Verified: bundle deterministic across 3 regens; bytecode + gogen_ir builds and tests green.

mparrett:
All passed, and both generator runs left the checkout clean. No remaining issues from prior findings.

nooga (approved):

PR #310, opened 2026-06-22
feat(ir): reify IR ops as native deftype op-types — S3 walking skeleton (#268)

nnunley:
**Stacked on #309** (purity bundle fix). Until #309 merges, this PR's diff includes #309's commit; the S3-specific commits are the 4 listed below. Review/merge #309 first.

## What this does

Reifies one IR op (`:const`) as a native op-*type* — a `deftype` (`ConstOp`) implementing a `defprotocol` (`IROp`) — with its lower-go + typeinfer facets co-located on the type, dispatched via a closed-world op-keyword bridge. Proves the ops-as-types mechanism (S3) end-to-end **with genuine native Go lowering**, not a bytecode bridge.

## The key fix

The deftype/defprotocol→Go native-lowering machinery already existed (`lower-ns-to-go` emits interface/struct/methods + binds the ctor/protocol registries), but `cmd/lgbgen/main.go`'s `runGoTarget` only forwarded `defn`/`defn-` forms to it — silently dropping `deftype`/`defprotocol` declarations for every core namespace. So core-ns op-types bytecode-bridged instead of lowering natively. This PR forwards declaration forms, then makes the dispatch statically typed so it devirtualizes.

## Commits
1. `refactor(ir): move classify-const to ir.lattice` — breaks the require cycle so the leaf `ir.ops` ns can classify consts.
2. `feat(lgbgen): forward deftype/defprotocol decl forms to native Go lowering` — `runGoTarget` now feeds declaration forms to `lower-ns-to-go` (was defn/defn- only).
3. `feat(ir): reify IR ops as native deftype op-types with devirtualized dispatch` — `ir.ops` (IROp protocol + ConstOp), routes `:const` infer/lower through it; dispatch constructs the op-type inline so typeinfer types it `[:dtype ConstOp]` and the protocol call lowers to a native `recv.OpInfer()`/`recv.OpLowerStmts()`.
4. `build(makefile): add clean-lowered target` — removes the generated Go tree + wireup files; wired into `make clean`.

## Result (generated `pkg/rt/core_go_lowered/ir/ops/ops.go`)
`type Irop interface` + `type Constop struct{ vm.RecordBase }` + `func _Constop()` + receiver methods; dispatch is `_Constop().OpInfer(ec, …)` — **native, no `CachedVarFn`**.

## Verification
- `make check-generated` green (bundle lockstep + native dispatch)
- `./pkg/ir/...` green untagged **and** `-tags bootstrap`
- full `./...` suite green; `go build -tags gogen_ir ./...` clean
- double-regen deterministic
- `test/ir_ops_const_optype.lg` + `test/deftype_native_skeleton.lg` pass

## Follow-on
Tree-wide native deftype lowering is now reachable; the catalog/codegen could generate the op-keyword→type bridge from a single registry (eliminating the remaining drift between `op-registry` keys and the dispatch `cond`).

---

Part of #268.


nooga:
`make clean` now fails on a fresh checkout because `clean-lowered` runs `find pkg/rt/core_go_lowered ...` unconditionally, but that generated tree is not tracked and usually does not exist until after `lgbgen --target=go`. `find` exits nonzero for a missing root, so the top-level `clean` target becomes unusable in the default repo state. Use `rm -rf pkg/rt/core_go_lowered` or guard the `find` with a directory check.

nnunley:
Fixed in 55fb92f. `clean-lowered` now `rm -rf`s the (untracked, fully-generated) `pkg/rt/core_go_lowered` tree instead of `find ... -delete`, so `make clean` exits 0 on a fresh checkout where the tree does not exist yet. Verified against that state.

mparrett:
Looks good to me.

nooga (approved):

PR #311, opened 2026-06-22
fix(term): flush on piped stdout

### LINKED ISSUE #308: term/flush propagates Sync() EBADF on a piped stdout — breaks piped test/smoke runs
`term/flush` returns the error from `os.Stdout.Sync()`. When stdout is a pipe — any captured or subprocess run (CI, `lg script.lg | tail`, a harness using `subprocess.run(..., capture_output=True)`) — `Sync()` returns `EBADF` ("bad file descriptor"), and the flush throws instead of treating the sync as a no-op. A pipe can't be fsync'd.

## Repro

Any program that reaches `term/flush` with stdout on a pipe:

```
$ lg render-script.lg | cat
error: sync /dev/stdout: bad file descriptor
  --> …/title.lg:149:5   (term/flush)
```

To a TTY or a regular file it's fine — those fds are syncable.

## Impact

Found via xsofy, whose `make test` and `make smoke-lg` capture stdout through a pipe. Every render-path test throws and the smoke driver never emits its sentinel. Piped: `2317 pass, 4 error` plus a smoke failure. To a file: `2325 pass, 0 error`, smoke OK. Same code, different fd.

## Regression

The behavior changed after v1.9.0:

- **v1.9.0** (`pkg/rt/term.go:417`) — calls `os.Stdout.Sync()` and ignores the return; benign on a pipe.
- **current** (`pkg/rt/term.go:576-579`) — `err = os.Stdout.Sync()` is propagated out of `term/flush`, so it throws on a pipe.

## Suggested fix

Treat the stdout sync in `term/flush` as best-effort: swallow `EBADF`/`ENOTTY` (syncing a pipe or tty is inherently a no-op, not a program error). That restores v1.9.0 behavior while keeping the explicit flush for files that benefit from it.

mparrett:
## Summary

Fixes #308 by treating stdout fsync errors from non-syncable file-backed streams as best-effort no-ops in `term/flush`.

## Root Cause

`term/flush` propagates `Sync()` from the active file-backed `*out*`. When stdout is a pipe, `fsync` is not meaningful and can return `EBADF`/`EINVAL` instead of completing successfully. That made captured or piped runs fail even though the terminal bytes had already been written.

## Changes

- Swallow `ENOTTY`, `EBADF`, and `EINVAL` only for file-backed `term/flush` syncs.
- Preserve error propagation for real file sync failures and embedder writer `Flush()` failures.
- Add an e2e regression test that runs `lg` with pipe-backed stdout.

## Validation

- `go test -run TestTermFlushWithPipedStdout -count=1 -v ./test/e2e`
- `go test -run 'TestRunner/term_out_binding_test\.lg' -count=1 -v ./test`

Note: `go test -count=1 ./test/e2e` still fails on an existing unrelated `TestGogenAOTDiff` / `pipeline_dump_ir.lg` divergence.

Co-authored with Codex.

### DIFF
--- a/pkg/rt/term.go
+++ b/pkg/rt/term.go
@@ -579,12 +579,15 @@ func installTermNS() {
 			err = os.Stdout.Sync()
 			fileBacked = true
 		}
-		// fsync on a terminal returns ENOTTY (macOS/BSD); flushing a TTY is a
-		// no-op, so swallow it. Only for a file-backed *out* — an embedder
-		// writer's Sync goes through Flush() and must surface its own errors.
-		// A regular file's fsync never returns ENOTTY, so real I/O errors
-		// (EIO, ENOSPC, …) still propagate.
-		if fileBacked && errors.Is(err, syscall.ENOTTY) {
+		// fsync on terminals and pipes returns ENOTTY, EBADF, or EINVAL
+		// depending on the platform; flushing those fds is a no-op, so swallow
+		// them. Only for a file-backed *out* -- an embedder writer's Sync goes
+		// through Flush() and must surface its own errors. A regular file's
+		// fsync never returns these, so real I/O errors (EIO, ENOSPC, etc.)
+		// still propagate.
+		if fileBacked && (errors.Is(err, syscall.ENOTTY) ||
+			errors.Is(err, syscall.EBADF) ||
+			errors.Is(err, syscall.EINVAL)) {
 			err = nil
 		}
 		return vm.NIL, err
--- a/test/e2e/term_flush_pipe_test.go
+++ b/test/e2e/term_flush_pipe_test.go
@@ -1,0 +1,30 @@
+/*
+ * Copyright (c) 2026 let-go contributors; see CONTRIBUTORS.
+ * SPDX-License-Identifier: MIT
+ */
+
+package e2e
+
+import (
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+func TestTermFlushWithPipedStdout(t *testing.T) {
+	bin := buildLG(t)
+	script := filepath.Join(t.TempDir(), "flush.lg")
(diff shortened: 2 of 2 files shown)


mparrett:
LGTM — correct, minimal, well-tested. A couple of minor notes below.

I checked the one part that looked broad — `EBADF` is the widest errno to swallow — and confirmed it empirically on macOS: `fsync` on a pipe returns `EBADF` on darwin (`os.Stdout.Sync()` with piped stdout returns it too). So the three-errno set is grounded per platform, not defensive over-reach:

- `ENOTTY` — fsync on a TTY (macOS/BSD), the original case.
- `EBADF` — fsync on a pipe (macOS), confirmed here.
- `EINVAL` — fsync on a pipe/FIFO/socket (Linux), per `man 2 fsync`.

The test reproduces the bug: `CombinedOutput()` hands the child a pipe on fd 1, and it's double-guarded (checks exit status and that output is exactly `ok`). It's cross-platform too — EINVAL on the Linux CI, EBADF on macOS. The `fileBacked` gate keeps embedder-writer `Flush()` errors and real-file I/O errors (`EIO`/`ENOSPC`) propagating, which is the invariant that matters.

Minor:

1. The comment slightly overstates. "A regular file's fsync never returns these" isn't quite true for `EBADF` — a closed or otherwise invalid fd on a real file also yields `EBADF`, so this now silently no-ops a flush on an already-closed handle. The risk is low given flush-on-exit semantics, but since `EBADF` is the one broad enough to mask a use-after-close, I'd have the comment own that rather than imply it can't happen.

2. Trivial: the rewritten comment swaps the file's `—` for `--` and `…` for `etc.`, which drifts from the surrounding style.

Optional, not blocking: an alternative to catching errnos is to `Stat()` the fd and skip `fsync` for `ModeNamedPipe`/`ModeCharDevice` up front, which is more explicit about intent. But the errno approach mirrors the existing `ENOTTY` handling, so keeping it is the smaller, consistent change — I'd keep it.

nooga (approved):

Issue #298, opened 2026-06-21
Add a persistent `storage` I/O seam (key/value) to the host-decoupling model


nnunley:
## Summary

Add a **`storage` I/O seam** — a persistent key/value capability exposed to `.lg` code, bound per-host — extending the runtime↔host decoupling model (`*out*` / `*in*` / `js/emit`) described in `docs/design/runtime-io-host-decoupling.md`. This is the same shape as the peer capabilities umbrella in #255 (audio, graphics, controller), applied to persistent storage.

## Motivation

The motivating consumer is **xsofy** (terminal roguelike running in the browser via `lg -w`). It now has a save/load system (`xsofy/save.lg`) that persists a game-state to EDN via `spit`/`slurp`. That works natively but **not in the browser**:

- Under `GOOS=js` there is no real filesystem, so `spit`/`slurp` don't persist.
- `js/emit` is **one-way** (WASM→JS, fire-and-forget). It can *write* a save out to `localStorage` from the JS shell, but there is **no read-back seam** for arbitrary data — the only inbound channel is terminal input (`*in*` / `sendInput`).

So loading a save in the browser has no clean path today. A storage seam closes that gap and generalizes beyond xsofy.

## Proposed surface (`.lg`)

A `storage` namespace, **synchronous** string key/value (Go `syscall/js` calls are synchronous; `localStorage` is synchronous):

```clojure
(storage/get k)      ; -> string | nil   (nil when absent)
(storage/set k v)    ; -> nil            (v is a string; throws on failure)
(storage/remove k)   ; -> nil
(storage/keys)       ; -> seq of string  (optionally (storage/keys prefix))
```

Values are strings — callers serialize (xsofy hands it `pr-str` EDN). Keeping it string/string mirrors `localStorage` and keeps the seam minimal.

## Per-host binding (mirrors the Emitter model)

| Host | Binding |
|---|---|
| **WASM bundle** (`lg -w`) | `localStorage` via a `HostStorage`, analogous to `HostEmitter` (`hostemitter_js_wasm.go`) |
| **native `lg`** | filesystem-backed dir (e.g. `$XDG_DATA_HOME/<app>/`), one file per key |
| **embedder** (`pkg/api`) | injectable, e.g. `api.WithStorage(...)` over a map or custom store |
| **default** | in-memory (or no-op) |

## Open questions (for discussion with the let-go maintainers)

1. **Namespace name**: `storage` vs `kv` vs `store`.
[...]

PR #315, opened 2026-06-23
feat(rt,wasm,api): add host-bound storage seam

mparrett:
## Summary

Initial draft for #298: adds a host-bound `storage` namespace for synchronous string key/value persistence.

Surface:

```clojure
(storage/get k)      ;; string | nil
(storage/set k v)    ;; nil, v must be string
(storage/remove k)   ;; nil
(storage/keys)       ;; vector of logical string keys
(storage/keys prefix)
```

The implementation follows the existing I/O seam pattern:

- `core/*storage*` is the dynamic host binding
- `rt.Storage` is the thin Go interface
- `api.WithStorage(...)` gives embedders a per-Run binding
- native `lg` installs a file-backed store
- WASM bundles install a `localStorage`-backed `HostStorage`
- default bare-runtime behavior is in-memory

## Portability choices in this draft

- Keys are logical strings, not paths.
- Native `FileStorage` encodes logical keys before using them as filenames.
- `(storage/keys prefix)` filters decoded logical keys, not filenames.
- Browser `HostStorage` scopes logical keys under a host-selected `let-go:<store-id>:` localStorage prefix.
- Native `lg` uses a host-selected store id and a per-store root below the user config dir.

## Notes / open questions

- Store id derivation is intentionally conservative but probably deserves maintainer taste: explicit `-storage-id` wins; otherwise `main.lg` uses the current directory basename, other scripts use the script basename, bundled binaries use the executable basename.
- Native root currently uses `os.UserConfigDir()/let-go/storage/<encoded-store-id>`. If maintainers prefer a data-dir policy instead, this is the main thing to adjust.
- `storage/keys` returns a sorted vector for deterministic tests and predictable UI. If the contract should avoid ordering promises, the implementation can keep sorting while docs say callers should not rely on order.
- This keeps TTL/cache semantics out of v1. Apps can encode expiration metadata in values; a real cache can be a separate seam later if needed.

## Validation

- `go test ./pkg/rt ./pkg/api`
- `go test ./pkg/rt/wasm`
- `go test ./...`
- `go run . -storage-id codex-storage-smoke -e '(do (storage/set "smoke:slot" "ok") [(storage/get "smoke:slot") (storage/keys "smoke:") (storage/remove "smoke:slot") (storage/get "smoke:slot")])'`
- `env GOOS=js GOARCH=wasm go test -c ./pkg/rt`

The direct smoke returned:

```clojure
["ok" ["smoke:slot"] nil nil]
```

Co-authored with Codex (5.5-high)


mparrett:
## Review notes on this draft

Self-review before this moves past draft. The seam itself mirrors `keysource.go` (interface → `*storage*` var → `resolveStorageVar`/`boundStorage` → per-platform install + `api.WithStorage` per-Run rebind), and the path-traversal handling is well covered by tests. The items worth fixing are concentrated in two files.

### Worth fixing

**1. `FileStorage.Set` is non-atomic.** `os.WriteFile` truncates then writes, so a crash, interrupt, or full disk mid-write leaves a truncated existing value, not just a failed new one. The headline use case is save games, so clobbering a good save with a half-written one is the failure mode to avoid. Write to a temp file alongside the target and `os.Rename` over it — rename is atomic on the same filesystem, and a partial write never touches the old value.

**2. WASM `HostStorage` error contract is right but undocumented.** The `recover` turns a thrown `getItem`/`setItem` (Safari private mode, blocked storage) into a Go error, which surfaces as a guest exception. That is the correct behavior: silently dropping a save write or reading a broken store back as `nil` is the dangerous failure mode for a save/load feature. It also already matches `FileStorage`, where `Get` returns `nil` only for a missing file and propagates every other read error. So the contract across both backends is: missing key returns `nil`, backend error or unavailability throws. The gap is documentation, not behavior — state those three cases on `WithStorage`/`HostStorage` and cover them with tests where feasible. One refinement worth weighing: detect unavailability once at `NewHostStorage` (probe a sentinel `setItem`/`removeItem`) so private mode surfaces a single clear failure rather than throwing on every call.

**3. Two empty backends with different semantics.** `iort.go` installs `MemoryStorage` as the root, where writes persist for the process, but `boundStorage` falls back to `nopStorage{}`, where writes vanish. In `keysource.go` the nop is the genuine default; here it's nearly unreachable, and where it does fire it disagrees with the real default. Pick one answer for "no host bound": drop `nopStorage` and fall back to a shared memory store, or make the root binding the nop.

**4. `storageIDForScript` has no tests.** It's the piece flagged as needing maintainer taste, and it's the logic a reviewer will poke at first. Table tests are cheap: `main.lg` resolves to the cwd basename, `foo.lg` to `foo`, the `init`/`main`/`.` exclusions hold, `-storage-id` wins, and a bundle falls back to the executable basename.

### Minor

- Long keys base64-encode past the 255-byte filename limit, so `Set` errors out. Fine for v1; worth a note, or hash keys over some length later.
- The store-id directory is itself `k-<base64>`, so the storage root is opaque to anyone browsing it. Keeping the raw id when it is already filename-safe would make it legible. Cosmetic.
- `FileStorage` has no in-process lock while `MemoryStorage` does. The single-threaded guest makes it moot today, but a one-line comment noting it relies on the OS plus a single-writer assumption would explain the asymmetry.

### Keeping

`strconv.Quote` for the template splice, the `..`/slashes encoding test, leaving TTL and cache semantics out of v1, and `storage/keys` returning an empty vector rather than nil.

mparrett:
Pushed three commits addressing the review notes above:

- **`f43e091`** — `FileStorage.Set` now writes a temp file and renames over the target, so a crash mid-write can't truncate an existing save. `NewFileStorage` no longer creates its root eagerly; that was littering the config dir with an empty store dir on every `lg` run that never touched storage. And `nopStorage` is gone — the unbound fallback now shares a memory store, matching the default root binding instead of silently dropping writes.
- **`7ed7ad8`** — documents the error contract (absent key returns nil, backend failure throws) on `WithStorage` and `HostStorage`. The WASM behavior was already correct, so this is doc-only, plus a note about probing localStorage once at construction so private mode fails clearly rather than on every call.
- **`9760b19`** — table tests for `storageIDForScript`, since the derivation is the policy-heavy part and had no coverage.

`go test ./...`, `vet`, and `gofmt` are clean; cross-builds pass on linux, js/wasm, and plan9.

mparrett:
One scope note on the atomic `FileStorage.Set` — a deliberate call, not an oversight.

The temp-file + `os.Rename` fixes the clobbering mode: a good save is never overwritten by a partial write, since the old file is untouched until the rename. What it doesn't do is `fsync` the temp file before the rename or the directory after it, so a power loss can still lose the latest write — or, on a filesystem that doesn't order data writes ahead of the rename metadata, leave a zero-length entry. ext4 `data=ordered` happens to be safe here; that's not guaranteed across filesystems.

For save/load that feels like an acceptable v1 tradeoff: the integrity of the *prior* save is what matters most, and that's covered. The belt-and-suspenders version, if we want crash-durability of the newest write too, is `fsync` on the temp file before the rename plus an `fsync` on the directory after. Happy to add it now or leave it as a documented limitation.

mparrett:
Codex: The substantive review items are addressed. The remaining blocker is mechanical: `lg_storage_test.go` cannot live at repo root. Since root `package main` white-box tests are banned, I’d move the storage ID derivation into a testable non-root helper, probably near storage code, and leave `lg.go` as thin plumbing around flag/cwd/executable lookup. Alternative is a black-box e2e test that observes the storage directory, but that is heavier and more brittle.

The fsync note is fine as documented. I would not require fsync in v1 unless the contract promises durability of the latest write across power loss. Atomic rename protecting the previous good save is the important fix.

mparrett:
@nnunley I think this gets the job done. If you have bandwidth, would appreciate a review.

nnunley:
Took a careful pass through this (multi-agent bug-hunt + verification). It's a clean, well-tested seam — atomic temp-then-rename, base64 key encoding, the absence-vs-failure error contract, and the three-layer install ordering all hold up. No correctness or memory-safety bugs found. Three things worth a look before merge, in priority order:

**1. WASM `HostStorage` doesn't encode the store id into its prefix — native/WASM asymmetry.** `hoststorage_js_wasm.go` builds `prefix = "let-go:" + storeID + ":"` with the id raw, whereas the native `FileStorage` runs the id through `encodeStorageKey` in `NewDefaultFileStorage`. Since `:` is legal in both store ids and logical keys, the (id, key) → physical-key mapping isn't injective: store id `"a"` + key `"b:x"` and store id `"a:b"` + key `"x"` both produce `let-go:a:b:x`. Each bundle bakes one fixed id at build time, so this only bites if a host deploys two bundles with colliding ids on the same origin — but it's the same class of problem the native path already guards against, so mirroring it (`"let-go:" + encodeStorageKey(storeID) + ":"`) closes the gap cheaply.

**2. `StorageIDFrom` doc vs. code mismatch with a real consequence.** The doc comment lists the cwd basename as the fallback "when the script name says nothing about the app (empty, '.', 'main', or 'init')", but the cwd branch lives inside `if script != ""`, so an *empty* script skips cwd and goes straight to the executable basename (the `storage_id_test.go` "empty script falls back to executable" case encodes this). Bundled binaries call `storageIDForScript("")`, so per the actual behavior they key by exe basename — multiple apps shipping as `lg` would silently share one store rather than being separated by cwd. Worth deciding which half is intended (the cwd-based id seems more useful for bundles) and aligning doc + code + a test.

**3. `installPersistentStorage` degrades silently for the guest.** If `NewDefaultFileStorage` fails it warns to stderr and leaves the in-memory default in place; `storage/set` then returns success but never persists. The host operator sees the warning, but the guest program gets no signal — slightly in tension with the "never silently drop a write" goal in the `Storage` doc. Fine as a dev-time degradation; for production you might want the failure louder (non-zero exit, or a backend whose `Set` errors) so silent data loss isn't possible.

A few minor suggestions too (not blockers): `boundStorage`'s shared-global `fallbackStorage` is inconsistent with the nop defaults used by `*emit*`/`*keys*`; `resolveStorageVar` duplicates `resolveIOHandleVar` and could share a generic helper; and the `FileStorage.Keys` temp-file-exclusion invariant (relies on the `k-` prefix in `decodeStorageKey`) would benefit from a one-line comment to protect it from future refactors.

🤖 Reviewed with a multi-agent agentic-review pass.

mparrett:
@nnunley Thanks for the careful pass. All three are addressed + the branch is rebased onto current main and green.

**1. WASM store-id encoding.** Fixed in `7551ba1` — `HostStorage` now builds its prefix as `"let-go:" + encodeStorageKey(storeID) + ":"`, mirroring the native root. Since the encoded id can't contain `:`, the separator stays unambiguous and the (store-id, key) → physical-key mapping is injective again; the raw logical key still round-trips through `TrimPrefix` in `Keys`.

**2. `StorageIDFrom` doc vs. code.** You're right that they disagreed; I went the other way on which half to fix. The doc is corrected in `b6d7c0e` to match the code rather than moving the code to cwd, because cwd is the wrong key for a bundle: it's wherever the user launched the binary, so a cwd-keyed store loses a player's saves the moment they run the game from a different directory. The exe basename is stable, which is why the empty-script (bundle) path lands there.

That said, the collision you're pointing at is real: two native bundles both named `lg` would share one store. One reasonable fix is to bake the id at build time from the source basename for native bundles, which is what the WASM path already does (`buildWasm` passes `storageIDForScript(files[0])`). The native bundle path derives it at runtime from the exe. Closing that native/WASM asymmetry is a slightly larger change; I recommend we do it as a follow-up rather than widen this PR. For now a bundle author who wants a guaranteed-distinct store sets `-storage-id` explicitly.

**3. Silent degradation in `installPersistentStorage`.** Fair tension with the "never silently drop a write" line. With the lazy root, `NewDefaultFileStorage` now only fails if `os.UserConfigDir()` itself fails, which is rare, and within a run the guest still reads back what it wrote. The data is non-durable rather than dropped. I've left it as a dev-time degradation for v1. Making the failure loud for production (non-zero exit, or a backend whose `Set` errors so the guest sees it) is a deliberate v1.x option; happy to take it now if you'd rather not ship the quiet fallback.

On the minors:

- **`fallbackStorage` vs. the nop defaults** — this one's deliberate, and `b6d7c0e` adds a comment saying so: `*storage*`'s root binding is a `MemoryStorage`, so its fallback matches that, where `*emit*`/`*keys*` fall back to nops because their roots are nops. Matching each var's own root felt more correct than matching the other seams.
- **`Keys` temp-file invariant** — added the one-line comment in `b6d7c0e`: the `k-` prefix in `decodeStorageKey` is what excludes `Set`'s `tmp-*` files, and the two need to stay in sync.
- **`resolveStorageVar` / `resolveIOHandleVar` duplication** — agreed it wants a shared helper. I'd do that as its own cleanup across all four seams rather than fold it into this PR.

Separately, the save-format compatibility point you raised offline is a good one, and I think an xsofy/client-side concern. The seam stays an opaque string KV store, and the version/algorithm envelope rides inside the value. I'll track it as a follow-up there rather than here.

---

**The proposed follow-ups:**

1. let-go side: bake the native-bundle store id at build time (from the source basename), mirroring what the WASM path already does. This is the real fix for the item-2 collision you flagged (two lg-named native bundles sharing a store).
2. xsofy side: the save-envelope versioning (schema + mapgen-version header) for graceful save/replay incompatibility. 3. (possible) Fail loudly on installPersistentStorage - TBD. What do you think?

nnunley (approved):
LGTM. Looks like all the things that needed to be addressed were.

PR #316, opened 2026-06-23
fix(ir/lower-go): guard box-as-value in :try op lowering (#258)

nnunley:
## What

Wrap the body/handler/finally `box-as-value` calls in the `:try` op lowering
(`pkg/rt/core/ir/lower_go.lg`) in `(try … (catch _ nil))`, so a boxing error
during `:try` lowering yields `nil` instead of aborting the whole lowering pass.

## Why

A boxing failure on one sub-expression of a `:try` form would otherwise throw
out of the lowering pass for the entire fn. Guarding each box keeps the pass
resilient.

## Reviewer note

This **swallows boxing errors to `nil`**. That's intentional for resilience, but
worth a careful look: if a boxing error ever indicates a *real* lowering gap, this
would mask it (the fn silently lowers with a nil sub-expr instead of surfacing the
gap). Happy to switch to a narrower guard or surface-and-skip if preferred.

Independent of the jank-coverage PR; based on current `main`.

---

Part of epic #258.


nooga:
No blockers found

nooga (approved):

PR #317, opened 2026-06-23
feat(ir,scripts): jank lowering-coverage gate + canonical fn enumeration

nnunley:
## What

A coverage gate that measures whether the IR pipeline can **lower the broader
Clojure surface**, driven by the vendored jank Clojure-compat suite, plus the
canonical fn-enumeration it relies on.

- **`ir.passes.pipeline/analyze-ns-forms`** — macroexpand a namespace's forms
  *once* and return `{:fixtures :defined-names}`: the single source of truth for
  "what lowering operates on". A `defn` is itself a macro (`(def name (fn …))`),
  so this expansion-based enumeration subsumes defn-only extraction **and** catches
  macro-generated fns (e.g. `deftest` bodies); it excludes `defmacro`.
  `lowerable-fn-forms`/`defined-names` are thin accessors.
- **`scripts/ir-stress.lg`** consumes it (drops its private defn-only extraction);
  macroexpansion-based fixture enumeration is now corpus-wide.
- **`make jank-stress`** runs `lower-go` over `test/clojure-test-suite` with
  `LG_SOURCE_PATHS` so the gate measures Clojure-surface lowering coverage.
- **`docs/perf/microbench/`** — `compute-bound.lg` + README: a startup-isolated
  `gogen_ir` native-lowering perf gate (today native ≈ bytecode; the README
  documents the dynamic-dispatch root cause and fix levers).

## Baseline

`make jank-stress` → **177/272 (65%)** lower cleanly. Top gaps surfaced: ratio /
BigDecimal / `#uuid` literals, `are`-macro-context literals, `gogen func-decl`,
and a few `nth` build/typeinfer errors — seeds for follow-on work.

## Verification

`make check-generated` green (bundle lockstep + native dispatch); `go test
./pkg/ir ./pkg/genmanifest` green. Based on current `main`.


nooga:
Review findings:

1. **Read errors are dropped from the coverage denominator.** `make jank-stress` includes every `core_test/*.cljc`, including `core_test/long.cljc`. In the new target that file reports `READ-ERROR`, but `stress-file` returns no `:total`/`:err` for read errors, and the summary only reduces fixture totals. That makes the coverage gate silently exclude unreadable files from the pass-rate. In my shortened run, `core_test/long.cljc` reported `READ-ERROR`, while the run still reported `Total fixtures: 272` and exited 0.

2. **Trace mode cannot debug the macro-generated fixtures this PR now reports.** Main stress mode now enumerates macro-expanded `deftest` fixtures, but trace mode still uses the old literal `defn-forms` scanner. For example, the jank run logs `core_test/and.cljc	test-and`, but `ir-stress trace ... core_test/and.cljc test-and` says `no defn named 'test-and'`. That breaks the advertised drill-down path for the new jank coverage failures.

Checks run:
- `go test ./pkg/ir ./pkg/genmanifest -count=1`
- `go test ./test/ -count=1 -run TestRunner/macroexpand_current_ns_test.lg -v`
- `LG_STRESS_TIMEOUT_MS=1000 LG_STRESS_PASSES=1 make jank-stress`

nnunley:
Both findings fixed in cbddc3b79 (scripts/ir-stress.lg only — no bundle/Go change):

1. **Read errors now count.** Each unreadable file is counted as one failed fixture, so it lands in the denominator instead of vanishing. With your repro, `core_test/long.cljc` READ-ERROR now pulls `Total fixtures: 272 → 273`, shows up in a `:read-error` failure bucket, and the summary prints `(incl. 1 unreadable file(s), each counted as 1 failure)`.
2. **Trace mode debugs the new fixtures.** Trace now enumerates via the same canonical `ir.passes.pipeline/analyze-ns-forms` as main mode, so `ir-stress trace ... core_test/and.cljc test-and` drills into the `deftest`-generated fixture instead of reporting `no defn named 'test-and'`. (Removed the dead `defn-forms` scanner.)

Verified with your exact repros plus `TestRunner/macroexpand_current_ns_test.lg` green.

nooga (approved):

Issue #271, opened 2026-06-19
Direct-call slice 2: native-module cross-ns direct calls (seed lowered-registry)


nnunley:
**Slice 2 of #270.** Branch: `feat/native-direct-calls`. Local ticket: `lg-dq58`.

Activates the previously-dead `:native?` direct-call path end-to-end. Main already has the typed emission (`emit-typed-direct-call`) and cross-ns resolution (`resolve-call-entry` handles `:native?`); the missing piece is **seeding `*lowered-registry*`** with the native modules from `(rt/native-modules)` and emitting their Go imports.

- `lower_go.lg`: build `{[ns-sym name arity] entry}` native-registry from `(rt/native-modules)` keyed by internal-ns-sym (skips variadics); filter nil `:decl` results (decl-less import carriers).
- `pipeline.lg`: Pass-2 merges native-registry into `*lowered-registry*` (lowered siblings win), binds `*native-imports-used*`, emits accumulated native import paths into the file import block.
- `native_direct_call_test.go`: `(seq x)` lowers to a direct `corefns.Seq(...)` call.

**Effect:** eliminates `seq` trampolines at 94 sites → `corefns.Seq`. **Perf-neutral** on `BenchmarkIRCompile` (native calls are cold — the hot trampolines are Slice 3).

**State:** implemented on-branch (3 commits, 42 behind main). Needs rebase onto current main + `make generate` + `check-generated`. Verified green on its original base (`go test ./...`, native dispatch).

nnunley:
Delivered by #343 (Go-lower clojure.core + direct-call hot builtins via pkg/rt/builtins). Native-module cross-ns direct calls now seed the lowered registry. Roadmap: EPIC-012 (native-module direct calls).

PR #318, opened 2026-06-23
feat(ir): native-direct cross-ns calls into clojure.core builtins (#271)

nnunley:
## What

Cross-namespace calls into `clojure.core` Go builtins now lower to **direct `corefns.X(...)` calls** under `-tags gogen_ir`, instead of the `ec.Invoke(rt.CachedVarFn(&var, ns, name), []vm.Value{...})` var-lookup + IFn-dispatch trampoline.

## Why

A compute-bound microbench showed native lowering delivered *no* speedup over the bytecode VM because it was **dynamic-dispatch-bound**: of 7807 `CachedVarFn` sites in the lowered corpus, **5752 (74%) target `clojure.core`**, whose hottest callees are Go builtins in `lang.go`. Removing the var lookup + IFn dispatch + per-call slice alloc on those is the dominant lever.

## How

Most of the mechanism already existed on main (registry `:native?`/`:go-pkg` shape, `resolve-call-entry` cross-ns + native refer semantics, `lookup-direct-callable` import recording). This wires the missing pieces:

- **`lower_go.lg`**: `native-registry` builds direct-call registry entries from `(rt/native-modules)`; variadic fns are skipped.
- **`pipeline.lg`**: `lower-ns-to-go` seeds `native-reg` into the pass-2 registry. Native entries stay **stable across the sibling-shrinking fixpoint** — they name external Go fns that always exist, so the `select-keys reg (keys emitted)` step (which would otherwise strip them and silently revert to trampolines) only shrinks the sibling half. Accumulated native imports are carried into the emitted file.
- **`corefns`**: lifted `first`/`second`/`next`/`rest`/`count` from their `lang.go` closures (`corefns` can't import `rt`, so the bodies are duplicated; a **differential test guards against drift** vs the canonical `clojure.core` builtins).

## Result

~1100 trampolines become direct calls (First 318, Count 266, Rest 223, Next 98, Second 63, Seq 137); `clojure.core` `CachedVarFn` sites **5752 → 4598**.

> Note: param/result specs are `vm.Value` for now; typed signatures (`Sequable`/`Seq`/`Counted` contracts, nullability-orthogonal) are a planned follow-up that removes boxing and enables typed-call chaining.

## Tests

`TestNativeModuleSeedsCrossNsDirectCall`, `TestLowerNsSeedsNativeRegistry`, `TestCorefnsArity1MatchBuiltins`. Updated 3 cached-var tests (`count` → `str`, since `count` is now a direct call). Full suite green, `gogen_ir` build clean, `check-generated` OK.

---

Part of #271.


nooga:
Review finding:

1. **Native direct calls bypass `with-redefs` / Var root changes for core fns.** This PR seeds `clojure.core` native entries into the lowering registry, and `lookup-direct-callable` returns those entries without checking whether the Var can be rebound at runtime. That changes semantics for lowered code because `with-redefs` explicitly works on ordinary defns by mutating Var roots.

   I probed this shape:

   ```clojure
   (defn probe [x]
     (with-redefs [count (fn [_] 42)]
       (count x)))
   ```

   `lower-ns-to-go` emits `corefns.Count(arg0)` inside the `with-redefs-fn` body, so the temporary `count` root is ignored. Before this PR, the cached-var trampoline dereferenced the Var and preserved the override seam. This likely needs a guard that keeps native-direct disabled when a function body can rebind or mutate vars, or at least for calls to vars touched by `with-redefs` / `alter-var-root`.

Checks run:
- `go test ./pkg/ir ./pkg/rt/corefns ./pkg/genmanifest -count=1`
- `go run -tags bootstrap ./cmd/lgbgen --target=go`
- `go test -tags gogen_ir ./pkg/ir -count=1`
- `go test -tags gogen_ir ./pkg/rt ./pkg/rt/corefns -count=1`
- Targeted lowering probe for `with-redefs` + `count`

mparrett:
**Codex Findings (5.5-high)**

1. **P1: Native direct calls bypass Var root changes.**  
   [pipeline.lg](/private/tmp/let-go-pr318/pkg/rt/core/ir/passes/pipeline.lg:664) seeds native `clojure.core` entries unconditionally, and [lower_go.lg](/private/tmp/let-go-pr318/pkg/rt/core/ir/lower_go.lg:1533) resolves them straight to `corefns.X(...)`. That skips the Var deref path that `with-redefs` and `alter-var-root` rely on. Example: `(with-redefs [count (fn [_] 42)] (count x))` lowers to `corefns.Count(...)`, ignoring the temporary root. This matches the existing PR comment and is a real semantic regression.

2. **P1: Stale native import tracking can emit uncompilable Go.**  
   [pipeline.lg](/private/tmp/let-go-pr318/pkg/rt/core/ir/passes/pipeline.lg:668) creates `native-imports` once before the pass-2 fixpoint, mutates it while lowering, then emits everything in it at [pipeline.lg](/private/tmp/let-go-pr318/pkg/rt/core/ir/passes/pipeline.lg:727). If a function records a native call and then falls back or is discarded, the final file can retain an unused `corefns` import. I reproduced with a temporary probe using a function that calls `seq` before an unsupported `#uuid` quote; `lower-ns-to-go` returned only:
   
```go
package nativeimportprobe

import "github.com/nooga/let-go/pkg/rt/corefns"
```

That file will fail Go compilation due to the unused import. The import accumulator should be per fixpoint iteration and/or derived from the converged valid results only.

**Checks**

`go test ./pkg/ir ./pkg/rt/corefns -count=1` passed on the PR worktree. I also ran the targeted unused-import probe above and removed the temporary test afterward.

nooga (approved):

nnunley:
Fixed in 27eb4828a. native-direct now disabled under var-root rebinding: a dynamic `*direct-calls-disabled?*` flag gates the single direct-call chokepoint (`lookup-direct-callable`), bound true while lowering any fn body — and its nested closures, where the with-redefs body actually lives — whose dynamic extent can rebind a var root. Such calls fall back to the `rt.CachedVarFn` trampoline, which re-reads the root each call, so the seam is preserved (covers native, lowered-sibling, and cross-ns typed direct calls — all share the chokepoint).

Detection reuses `ir.passes.mutability` as the single source of truth (now an explicit `:require` of lower-go; it also learns `with-redefs-fn`, the post-expansion callee — the `with-redefs` macro symbol is gone by IR-build time).

Your probe now lowers `(count x)` to `rt.CachedVarFn(&__v_clojure_core_count, "clojure.core", "count")` instead of `corefns.Count`. Verified: `go test ./pkg/ir ./pkg/rt/corefns ./pkg/genmanifest`, `--target=go` regen, `go build -tags gogen_ir ./...`, `go test -tags gogen_ir ./pkg/ir ./pkg/rt ./pkg/rt/corefns`; bundle byte-deterministic.

Issue #270, opened 2026-06-19
Direct-call lowering: rt.InvokeValue → direct Go calls


nnunley:
Tracking issue for **direct-call lowering** — lifting the `rt.LookupVar → Var.Deref → rt.InvokeValue` trampoline into **direct Go calls** in the `-tags gogen_ir` lowered tree. Sub-issue of #258. Reconciled against `main` @ `d4c36cf2` (#251), 2026-06-19.

## Already in main
- Infrastructure: `emit-typed-direct-call`, `resolve-call-entry` (handles `:native?`), `lookup-direct-callable`, `*lowered-registry*`, `*native-imports-used*` (`lower_go.lg`/`pipeline.lg`).
- **Slice 1 — intra-namespace direct calls: MERGED in #168** (2026-06-02).

## Pending slices (sub-issues)
- **Slice 2 — native-module cross-ns direct calls** (branch `feat/native-direct-calls`, ticket `lg-dq58`). Seeds `*lowered-registry*` from `(rt/native-modules)` and emits the Go imports so the dormant `:native?` path activates end-to-end. Eliminates 94 `seq` trampolines → `corefns.Seq`. **Perf-neutral** (these native calls are cold). Implemented on-branch (3 commits); needs rebase onto current main + bundle regen.
- **Slice 3 — global-lowered-registry: hot cross-ns *lowered* direct calls** (planned follow-on, named in the lg-dq58 commit). Targets the *hot* trampolines — cross-ns calls into lowered `licm`/`typeinfer`/`build`. This is where the actual perf win is; Slice 2 deliberately leaves it.

## Adjacent (not filed here)
- **Lambda lifting** — queued *after* direct-call per the roadmap; distinct optimization, no branch yet.
- **Structured Go emission** (`ir.structurize`) is **already merged in main** — `lower-body` is structured-first (if/for/break/continue for reducible CFGs) and falls back to the goto emitter when `structurize` reports `:fallbacks > 0`. It is the *proper-emission* mechanism, not pending direct-call work. Its residual fallbacks are exactly the #263 bucket (driving `structurize :fallbacks` → 0).

## Superseded branches (archive, not filed)
- `feat/direct-call-intra-ns` → landed as #168.
- `direct-call-boundary` → dormant origin scaffold; superseded by `feat/native-direct-calls`.
- `ir-exploration` → 54-commit accumulator; mine for tests/notes, not a single issue.

nnunley:
[...]

PR #320, opened 2026-06-23
feat(ir): interface category types + collection-literal classification — typed-call foundation (#270)

nnunley:
**Stack (review/merge bottom-up):** #318 → **this** → typed-directcall-emission → crossns-directcall-emit. Built on the unmerged #318, so the diff includes #318's commit until it merges.

## What
Foundation for typed direct-call lowering — the lattice + source halves, both **behavior-neutral** (lowering output byte-identical; `check-generated` green).

- **Lattice** (`lattice.lg`): concrete collection/seq kinds (`:vector`/`:map`/`:set`/`:list`/`:cons`/`:lazyseq`) + derived vm-interface satisfaction (`seqable-type?`/`counted-type?`/`seq-type?`) as a MUST query over the kinds a value may be. Collection kinds are **non-maskable** (the `:ratio` precedent), so the proven bitset proptest is untouched (26244 pairs still ≡). Nullability is orthogonal (`:nil` is a separate member).
- **Source** (`typeinfer.lg` `classify-const`): collection literals classify as their kind; `go-type-spec` maps them to `vm.Value`.

Interfaces are derived subset-queries, **not** stored OR'd interface bits (those would be may-semantics under bit-or join — unsound for the MUST question).

## Tests
`TestLatticeCollectionKindsAndInterfaceSatisfaction`. Full suite + `check-generated` green.

---

Part of #270.


nooga:
`seq-type?` leaves `:string` out of `seq-kinds`, even though `vm.String` implements the `vm.Seq` method set (`First`, `More`, `Next`, `Cons`) and the surrounding comment says these sets are verified against VM interfaces. That makes `(ir.lattice/seq-type? :string)` a false negative, so any follow-up typed direct-call work that relies on `seq-type?` will miss valid string-to-`vm.Seq` cases.

nnunley:
Fixed in e227e4e, and the fix is broader than just :string. Checked pkg/vm: ArrayVector, Map, Set, String (and *List/*Cons/*LazySeq) all implement the full vm.Seq method set (First/More/Next/Cons), so seq-kinds now mirrors vm.Seq membership = all seven collection kinds, not Clojure's seq? predicate. That makes it coincide with seqable-kinds/counted-kinds at this granularity; they stay distinct predicates because finer vm types diverge once represented (e.g. *InfiniteRange is vm.Seq but not vm.Counted). The "verified against pkg/vm method sets" comment that read as a method-set criterion while the set encoded seq? has been rewritten, and seq-type? now has test coverage.

nooga (approved):

PR #321, opened 2026-06-23
feat(ir): typed lowered-sibling direct calls — emission half (#270)

nnunley:
**Stack (review/merge bottom-up):** #318 → #320 (foundation) → **this** → crossns-directcall-emit. Diff includes ancestor commits until they merge.

## What
The direct-call registry can now record a lowered sibling's **real typed signature**, so callers emit a direct typed call instead of the `rt.CachedVarFn`/`InvokeValue` trampoline. Previously `registry-entry-from-result` hardcoded `vm.Value` and `override-uniform-value?` excluded every typed fn.

- **Gate split**: `override-uniform-value?` fed *both* the NativeFn override wrapper and the registry. Now `override-eligible?` (uniform `vm.Value`) gates the wrapper; new `override-coercible?` gates the registry. `registry-entry-from-result` records actual `function-param-specs`/`function-return-spec`.
- **Soundness**: dropped `coerce-arg-to-type`'s dead unchecked scalar unbox-asserts (`int(x.(vm.Int))` on an unproven value would change dynamic semantics). A typed param is direct-called only when the arg is *proven* that type; else the call falls back.
- **Codegen fix**: the typed-result+error path declared its intermediate as `result, callErr = <call>` (undefined `result`; the fixed name also collided). Now `tr<nid>, te<nid> := <call>`, names keyed by the node's Indexed-RPN position (deterministic).

## Effect
Typed-**result** siblings (`const?` → bool): **7 call sites** now lower to direct `vm.Boolean(const?(ec, ...))`. Typed-**param** unboxing works end-to-end too — `(defn addone [x] (+ x 1))` → `func addone(ec, x int) int`, called directly with an unboxed int (a numeric loop lowers to pure native int Go). The compiler corpus has no typed params (node-polymorphic), so that path is capability-driven; numeric user code benefits.

## Tests
`TestLowerNsTypedResultSiblingDirectCall`. `check-generated` (compiles + dispatches natively under `-tags gogen_ir`) + full suite green.

---

Part of #270.


nooga:
No issues found

nooga (approved):

Issue #272, opened 2026-06-19
Direct-call slice 3: global-lowered-registry for hot cross-ns lowered calls


nnunley:
**Slice 3 of #270.** Planned follow-on (named in the `lg-dq58` commit). No branch yet.

Slice 2 activates *native-module* cross-ns calls, which are cold (perf-neutral). The **hot** trampolines are cross-ns calls into **lowered** namespaces — `licm`, `typeinfer`, `build` calling each other through `rt.LookupVar → Deref → InvokeValue`. A **global** lowered-registry (visible across the whole lowered tree, not just intra-file siblings) lets these resolve to direct Go calls.

**Goal:** lift the hot cross-ns lowered calls to direct calls; this is where the measurable `BenchmarkIRCompile` win is.

**Open questions to scope:** registry visibility/ordering across files in a single `--target` compile; handling of forward/mutual cross-ns refs; interaction with the bundle's frozen-order assumptions.

nnunley:
Delivered by #345 (cross-package direct calls between AOT-lowered packages). The whole-program `lower-all-ns-to-go` collector + `*cross-pkg-registry*` + `LG_` exported wrappers is the global-lowered-registry this tracked. Roadmap: EPIC-012 / ITER-0025.

PR #322, opened 2026-06-23
feat(ir): cross-ns typed direct calls — emit machinery (#272)

nnunley:
**Stack (review/merge bottom-up):** #318 → #320 (foundation) → #321 (emission) → **this** (top). Diff includes ancestor commits until they merge.

## What
Generalizes `emit-typed-direct-call` + `lookup-direct-callable` to a third callee class — a **cross-ns lowered sibling**. Package-qualification + import recording now key on `:go-pkg` (present) rather than `:native?`:

| callee | emit |
|---|---|
| intra-ns lowered (no `:go-pkg`) | `Fn(ec, args)` |
| native module (`:native?`, `:go-pkg`) | `pkg.Fn(args)` (context-free) |
| **cross-ns lowered** (`:go-pkg`, `!:native?`) | `pkg.Fn(ec, args)` (ec-passing) — **new** |

`ec` stays keyed on `:native?`. Any `:go-pkg` callee's import is recorded in `*native-imports-used*`.

**Behavior-neutral**: no cross-ns lowered entries are produced yet (a follow-up wires the corpus-wide registry), so the generated tree is unchanged (`check-generated` green). This is the emit capability + a hand-seeded test.

## Follow-up (not in this PR)
Phase 2 needs (a) a corpus-wide registry accumulated across lgbgen's dependency-ordered lowering — the call graph follows the linear order, so it's always a DAG (no Go import cycles); and (b) **exporting** lowered fns (currently `func dce`/`sum_to` are unexported, so cross-package calls won't compile). Cross-ns var refs are already fully-qualified by the resolver, so no alias handling is needed.

## Tests
`TestCrossNsLoweredDirectCallEmit`. `check-generated` + full suite green.

---

Part of #272.


nooga:
Found one issue:

- **[P2] Record cross-package imports only after direct-call emission succeeds.** `lookup-direct-callable` records `:go-pkg` imports before `emit-typed-direct-call` has proved the call can actually be emitted. If a cross-ns lowered entry has a scalar param such as `"int"` and the caller passes an unproven `vm.Value`, `coerce-arg-to-type` returns nil and the call correctly falls back to cached-var dispatch, but the import has already been added to `*native-imports-used*`. Once phase 2 wires cross-ns lowered entries, that fallback path can render an unused package import and make the generated Go fail to compile. I reproduced this with a temporary hand-seeded registry entry whose typed param forced fallback while still leaving the package import in the rendered file.

The fix should be to record the `:go-pkg` import only after `emit-typed-direct-call` succeeds, or otherwise have the emitter return/import-record atomically with the direct-call statements.

nnunley:
Fixed in 8cd849f. Moved the :go-pkg import recording out of lookup-direct-callable (which runs before coercion) into emit-typed-direct-call, gated on emission actually producing statements. So when a typed param forces the cached-var fallback, no import is recorded and the generated Go has no unused import. Added a regression test that seeds a cross-ns entry with an int param to force the fallback and asserts *native-imports-used* stays empty; verified it fails on the old recording path.

mparrett:
Reviewed the #322-specific top commits (`020cdcb` and `8cd849`) only. I treated the earlier commits from #318/#320/#321 as stack context, since GitHub currently includes them in this stacked diff. Rechecked upstream before posting: PR #322 head is still `8cd849f1f3520b4e0fd75f8525dd9514748133de`.

Findings: no blocking issues found in the current PR behavior.

PR-scoped follow-up / footgun: the new cross-ns lowered-call emission derives the selector prefix from the final `:go-pkg` path segment and records the import as a bare `{:path p}`. That works for the hand-seeded `otherns` test and for current native `corefns`, but Phase 2 can hit lowered namespaces whose package leaf collides with another import or lowered namespace leaf, e.g. `data` vs `ir.data`, `passes` vs `ir.passes`, or a generated package leaf colliding with existing aliases like `vm`/`rt`. In that case generated Go would either have duplicate local import names or qualify the call through the wrong/ambiguous selector. Since #322 does not produce cross-ns lowered entries yet, I’m calling this out as a Phase 2 footgun rather than a current regression; the registry/import carrier probably needs to carry a stable import alias instead of relying on `pkg-alias-from(:go-pkg)`.

Verification: `go test ./pkg/ir -run 'TestCrossNs(LoweredDirectCallEmit|DirectCallFallbackDoesNotLeakImport)$'`

nooga (approved):

PR #324, opened 2026-06-24
feat(ir): lower the full jank Clojure-compat surface — coverage gate 65% → 100% (#327)

### LINKED ISSUE #327: gogen: lower the full jank Clojure-compat surface to 100% coverage
Part of epic #258 (Lowering & cross-target codegen). Continues the merged coverage-gate work in #317.

Drive the jank Clojure-compat suite's `lower-go` coverage from the committed 65% baseline to 100%, closing the per-construct lowering gaps the gate surfaces.

Tracked work (PR #324).

nnunley:
## What

Follow-on to the jank lowering-coverage gate (#317): six IR-pipeline fixes that take `make jank-stress` from the **65% baseline to 100% (273/273)**. Every let-go fixture in the vendored jank Clojure-compat suite now lowers cleanly through the IR→Go pipeline.

Each fix was root-caused from a failure bucket, has a focused test (TDD), and keeps `make check-generated` green (the lowered core tree is unchanged except where noted).

## The fixes

| Commit | Bucket | Root cause → fix | Gate |
|--------|--------|------------------|------|
| `lower BigDecimal/Ratio/UUID` | `build-form: unrecognized form` | Not a data-reader gap — the reader already materialises these values; `build-form` just had no `:const` arm. Added arms + gogen emission via `vm.Must{BigDecimal,Ratio}FromString` / `vm.ParseUUID`. Also `pr-str` (not `str`) in the throw so `0.0M` stops masquerading as a float. | →183 |
| `lower BigInt` | `gogen/func-decl: expected seq, got ExceptionInfo` | `1N` satisfies `integer?` but not `int?`, so it was typed `:int` and gogen's `int-lit` choked (leaking an ExceptionInfo into the fn body). Gate the native-int path on `int?`; new `bigint?` arm → `vm.MustBigIntFromString`. | →254 |
| `reader: comment before matched #? value` | `:read-error` | A comment between a reader-conditional branch keyword and its value surfaced as VOID from the matched `r.Read()`, desyncing the form into EOF. Loop past VOID for the matched value. | →255 |
| `two nth crashes` | `nth index out of bounds` / `must be an integer` | (a) zero-arg `(+)`/`(*)`/`(-)` folded an empty arg vector → fall back to a normal call (runtime gives `0`/`1`/arity-error). (b) empty `do`/`let` body returned a bare `nil` that leaked into a branch block-arg → `build-body-result` materialises a nil-const. | →266 |
| `faithful :def IR op` | `unresolved symbol def` | Nested `def` was unsupported in the IR pipeline. Added a real `:def` op: bytecode reuses `SET_VAR` (as the bytecode `defCompiler` does — there is no `OP_DEF`); gogen interns at runtime via a new `rt.InternVar` (vs `LookupVar`). | →271 |
| `empty fn body + (Name.) ctor` | `unsupported result type` / `unresolved symbol X.` | (a) `(defn f [])` emitted `:return []` → route fn-body builders through `build-body-result` so it returns `vm.NIL`. (b) the trailing-dot record-constructor shorthand `(Name. args)` had no `build-list` case → mirror the compiler's rewrite to `(->Name args)`. | **→273** |

## Verification

- `make jank-stress` → **273/273 (100%)**, zero failure buckets.
- `go test ./pkg/...` green; `make check-generated` green (bundle lockstep + native dispatch).
- New tests in `pkg/ir/lisp_lower_go_test.go` and `pkg/compiler/reader_conditional_skip_test.go` cover each fix.

Based on current `main` (post-#317).

---

Closes #327.


nooga:
Review finding:

1. **`def` lowering drops var metadata/docstring semantics.** The new IR path unwraps `(with-meta sym meta)` but discards the metadata, accepts any 3-arg form as `(def sym _ value)`, and the Go lowering only emits `rt.InternVar(...).SetRoot(...)`. The bytecode compiler does more: it requires the second arg of a 3-arg `def` to be a string, merges it into `:doc`, applies metadata to the Var, and sets `^:dynamic` / `^:private` flags.

   This is observable for lowered code such as:

   ```clojure
   (defn install []
     (def ^:dynamic probe 1))
   ```

   After the lowered function runs, `probe` is interned and rooted, but the Var metadata/flags are missing compared with bytecode compilation. Same for `(def probe "doc" 1)`, where `(:doc (meta #'probe))` should be preserved. The IR `:def` op probably needs to carry the doc/meta information and emit equivalent runtime setup, or these forms need to remain on the bytecode path until that support exists.

Checks I ran:
- `go test ./pkg/ir -count=1`
- `go test ./pkg/compiler -run TestReaderConditionalSkipsCommentsBeforeMatchedValue -count=1`
- `go test ./pkg/vm -count=1`
- `make check-generated`

`go test ./pkg/... -count=1` only failed in `pkg/nrepl` because the sandbox cannot bind localhost ports. I did not run `make jank-stress`.

nnunley:
Thanks — you're right, the `:def` op was dropping var metadata. Fixed in two commits on this branch.

**`cbdcd29d4` — preserve metadata/docstring/flags through `:def`** (faithful to `defCompiler`):
- `build-def` now captures the `(with-meta sym meta)` map instead of discarding it, requires a `String` docstring for the 3-arg form (merged into `:doc`, error otherwise), and applies the meta to the build-time-interned var via a new `apply-def-meta!` builtin — the bytecode path has no `SET_META` opcode, so this mirrors `defCompiler` setting it at compile time.
- The meta rides along as the `:def`'s Aux so the gogen path can re-apply it at runtime with a new `rt.ApplyVarMeta` (the build-time var doesn't exist in the lowered program).
- `rt.ApplyVarMeta` is the shared helper mirroring your `defCompiler` block (`SetMeta` + `:dynamic`/`:private`).

Your `(def ^:dynamic probe 1)` example now lowers to:
```go
v = rt.InternVar("core", "probe")
rt.ApplyVarMeta(v, vm.NewPersistentMap([]vm.Value{vm.Keyword("dynamic"), vm.TRUE}))
v.SetRoot(vm.Int(1))
```
and `(def probe "doc" 1)` carries `:doc "doc"` into the var meta. Type hints survive too (`^Integer x` → `{:tag (quote Integer)}`), which was the motivation to keep the metadata rather than gate these forms back to the bytecode path. New tests cover dynamic-meta render, docstring→`:doc`, plain-def-omits-`ApplyVarMeta`, build-time var flag/meta application, and the 3-arg non-string-docstring rejection.

**`d99a45aea` — perf fix found while chasing the lone `jank-stress` miss.** `core_test/plus.cljc`'s `test-+` was timing out at ~15.6s. Profiling showed LICM was 12.4s of it (and removed zero insts): `back-edges` calls `dominates?` per `(header, pred)` pair, and `dominates?` rebuilt the full immediate-dominator fixpoint on every call — ~O(blocks²×edges), driving ~24GB of allocation (45% GC). Fixed by computing the dominator tree once per `back-edges` pass (new `dominates-idom?` against a precomputed `idom`, mirroring what `structurize.lg` already does). Not memoized on `f` by default since `f` is mutable and a cached tree would go stale. `test-+` now lowers in 3.5s (4.5×), and `BenchmarkIRCompile` improved ~46%.

Verification: `make jank-stress` **273/273 (100%)**, `go test ./pkg/...` green (614), `make check-generated` green, `make ratchets` green (0 regressions; IRCompile −46%).

nooga (approved):

PR #325, opened 2026-06-24
feat(ir): native Go lowering of type-dispatched defmulti/defmethod multimethods (#326)

### LINKED ISSUE #326: gogen: native Go lowering of type-dispatched defmulti/defmethod multimethods
Part of epic #258 (Lowering & cross-target codegen).

Lower user-level type-dispatched multimethods to a native Go `switch v.(type)` (with a runtime `vm.MultiFn` default arm) instead of a runtime hash-dispatch, while open/late/cross-ns/value-dispatched methods stay correct on the runtime path.

Tracked work (PR #325):
- Capture defmulti/defmethod pre-expansion; accumulate a `*defmulti-dispatchers*` registry during dep-ordered lowering.
- Type-dispatch fast path: `(fn [x] (type x))` → native type-switch with method-fn arms + runtime default.
- Value/`:unknown`-dispatch: tagged in the registry, trampoline to the runtime MultiFn.

Follow-ons: closed-set coverage detector (`check-multimethod-coverage!`); cross-package method-fn native imports.

nnunley:
> Implements #326 · part of Epic #258 — Lowering & cross-target codegen.

## What

Type-dispatched multimethods now **devirtualize to a native Go type-switch** instead of always falling to runtime `vm.MultiFn` hash-dispatch. A call to a same-ns multifn whose dispatch fn is `(fn [x] (type x))` lowers to:

```go
switch dispatchArg.(type) {
case vm.ArrayVector: v = _mm_<multi>_0(ec, args...)   // per confidently-mapped method
case *vm.List:       v = _mm_<multi>_1(ec, args...)
default:             v, err = ec.Invoke(rt.CachedVarFn(&__v, ns, name), args)  // runtime MultiFn safety net
}
```

## Pieces
- **`gogen/type-switch-stmt`** — new `*ast.TypeSwitchStmt` constructor (`switch [x :=] e.(type){…}`), unit-tested.
- **lgbgen `isMultimethodForm`** — forward `defmulti`/`defmethod` forms to `lower-ns-to-go`.
- **`pipeline.lg` `lower-ns-to-go`** — recognize the `(fn [x] (type x))` dispatch shape, lower each method body to a package-level `_mm_*` Go fn, map dispatch-vals to real `vm` Go types (`vm.ArrayVector`, `*vm.List`, `vm.Set` — only unambiguous single-representation literals), build the `*defmulti-dispatchers*` registry with `:home-ns` (var-resolved, robust to the core/ns-name duality) and a reserved `:dispatch-type` field for the next iteration.
- **`lower_go.lg`** — extract `trampoline-call-stmts` (reused verbatim for the switch's `default:` arm); `recognize-type-dispatch-fn` + `dispatch-val-to-go-type`; `defmulti-call-stmts` emits the switch, per-arm ABI adapting to each method's `:needs-error?`.

## Conservative by design
Unrecognized dispatch fn / unmappable dispatch-val / non-uniform-`vm.Value`-ABI method / cross-ns call → return nil → runtime trampoline. The `default:` arm is the safety net for every other Go representation and open/late methods, so dispatch stays correct. `print-method` falls back (its `pr-sequential` bodies don't lower to the uniform ABI), so `core_go_lowered/` is **byte-unchanged**.

## Tests
- `pkg/ir/TestDefmultiNativeTypeSwitchLowering` — anti-stub: greps the rendered Go for the `.(type)` switch + `_mm_*` method-fn arms + the runtime default (fails on a nil-stub).
- `test/defmulti_native_lowering_test.lg` — runtime semantics incl. the default-arm fallback for uncovered types.

## Verification
`make check-generated` OK · `go test ./...` green (both bytecode and `-tags gogen_ir`) · double-`make generate` **byte-identical** (determinism) · `ir-stress` 1213/1219 (≥ baseline).

---
**Note:** this branch stacks on the open `#318`-family PRs (`#318`/`#320`/`#321`/`#322`) + the `defmulti`/`defmethod` capture commit, so the diff against `main` is cumulative — the new work is the top commit (`feat(ir): native Go lowering of type-dispatched defmulti/defmethod`). Review/merge bottom-up.

---

Closes #326.


nooga:
Found one issue:

- **[P1] Native arms bypass later `defmethod` updates for mapped types.** `defmulti-call-stmts` emits `case <GoType>:` arms that call the baked `_mm_*` method function directly. That means once a type has a native arm, e.g. `(type [])`, any later `(defmethod mm (type []) ...)` update is ignored for that type. The runtime path is mutable: `defmethod` expands to `set!` of the multifn var, and `MultiFn.AddMethod` returns a new method map. The switch default only re-enters the runtime `MultiFn` for unmapped Go representations, so it does not protect mapped types from stale native arms. This breaks normal open/late multimethod semantics under `gogen_ir`.

Verification I ran:
- `go test ./pkg/ir ./pkg/rt -count=1`
- focused new `.lg` defmulti/gogen tests
- `go run -tags bootstrap ./cmd/lgbgen --target=go pkg/rt/core_go_lowered`
- focused `-tags gogen_ir` defmulti tests

Full `go test ./...` hit sandbox-only bind failures in `pkg/nrepl` / `os_free_port_test.lg`; unrelated packages had already passed before that.

nnunley:
Fixed in 21f57493b. Native arms are now guarded by a runtime currency check.

`MultiFn.AddMethod` already returns a fresh `*MultiFn`, so identity is the signal. `vm.MultiFn` gains a `frozen` flag; the multifn is frozen once at namespace-load completion (`rt.ApplyGoOverrides`, which runs after bytecode replay creates the var and applies all build-time defmethods). The switch is now gated:

```go
if rt.MultiFnNativeFrozen(&__v_…tdnative, "clojure.core", "tdnative") {
    switch arg0.(type) { case vm.ArrayVector: _mm_tdnative_0(ec, arg0); … default: <runtime> }
} else {
    <runtime trampoline>
}
```

A late `(defmethod …)` set!s the var to a new, unfrozen `*MultiFn`, so the guard falls to the runtime trampoline and the override is honored — for mapped types too, not just unmapped Go representations. Per call it's one deref + bool read (no map lookup), so native speed is preserved while no late defmethod has occurred. The pipeline emits `rt.RegisterNativeMultiFns(ns, …)` so each ns freezes its native-baked multimethods at load.

(Distinct from #318's with-redefs: that's lexically scoped so a body-local guard sufficed; defmethod is globally open, hence the runtime check.)

Tested: vm freeze primitive; rt freeze-at-load + immediate-apply + an end-to-end guard-flow (freeze → set!(AddMethod) → reports not-frozen); render asserts both the guard and the registration init. Full `pkg/{vm,rt,ir}`, `-tags gogen_ir` build + tests + e2e green; bundle byte-deterministic.

Follow-up noted in a code comment: a closed-world AOT build (no `eval` + bound `*main*`) has a closed defmethod set, so the guard can be elided for full devirtualization.

nooga:
Found one issue:

- **[P1] Duplicate native type-switch arms for repeated `defmethod` dispatch values.** `lower-ns-to-go` builds a native arm for every captured `defmethod` whose dispatch value maps to a Go type, and `defmulti-call-stmts` emits those arms directly as `case <GoType>:` clauses. Runtime multimethod semantics are last-write-wins: a later `(defmethod mm (type []) ...)` replaces the earlier method in the `MultiFn` method map. The generated Go instead emits duplicate cases, e.g.

```go
case vm.ArrayVector:
    v = _mm_tddup_0(ec, arg0)
case vm.ArrayVector:
    v = _mm_tddup_1(ec, arg0)
```

That does not type-check as Go, so a valid source namespace with repeated `defmethod` definitions can break `gogen_ir` generation. Mixed-arity repeats can also select an older native arm after the call-site `:param-count` filter, while runtime dispatch would use the latest method and then apply normal arity behavior. The registry should collapse native arms by dispatch value / Go case type with runtime-compatible last-wins semantics before emitting the switch.

Verification: I reproduced this on `origin/pr/325` by lowering a `defmulti` with two `(type [])` methods; the rendered output contained two `case vm.ArrayVector:` arms. The PR’s focused defmulti tests otherwise passed locally.

nnunley:
Second P1 (duplicate native arms) fixed in fd7a883d1.

`lower-ns-to-go` now collapses native arms by `:go-case-type`, keeping the LAST arm per dispatch value at its last source position — matching runtime last-write-wins. Two `(defmethod mm (type []) …)` now emit a single `case vm.ArrayVector:` (the latest method) instead of duplicate cases; a stale-arity earlier arm can no longer survive the call-site `:param-count` filter. No-duplicate case is byte-identical (source order unchanged). Regression: `TestDefmultiNativeDuplicateDispatchCollapsed`.

nooga (approved):

PR #331, opened 2026-06-24
feat(lg): LG_CPUPROFILE/LG_MEMPROFILE env fallback for profiling builds

nnunley:
## What

Adds an `LG_CPUPROFILE` / `LG_MEMPROFILE` environment-variable fallback to the `lg_profile`-tagged profiling build, alongside the existing `-cpuprofile` / `-memprofile` flags.

## Why

The flags work fine for the REPL and simple scripts, but they break scripts that read **raw** `os/args` with a fixed positional layout. For example `scripts/ir-stress.lg` does:

```clojure
;; os/args is [binary script user-arg user-arg...]
(def user-args (vec (drop 2 (seq os/args))))
```

Passing `-cpuprofile X -memprofile Y` shifts those positionals (the flags land in `os/args`), so the script reads `-memprofile` as its `<mode>` and aborts. There's no way to profile those harnesses today.

The env fallback sidesteps the positional layout entirely:

```sh
LG_CPUPROFILE=/tmp/cpu.prof LG_MEMPROFILE=/tmp/mem.prof \
  ./lg scripts/ir-stress.lg trace <dir> <file> <defn>
```

Explicit flags still win when both a flag and its env var are set. This was used to root-cause the LICM dominator recompute investigated in #324 (ir-stress `trace` + heap profile showed `dominators` rebuilt per `dominates?` call → 24GB churn / ~45% GC).

## Scope

- Build-tag gated (`lg_profile`); zero effect on normal/bundled builds.
- One file (`lg_profile.go`): env read in `startProfiling`, flag help mentions the fallback.
- Smoke-tested: `LG_CPUPROFILE=… LG_MEMPROFILE=… lg -e '(println (reduce + (range 1000)))'` writes both profiles with no flags.

### DIFF
--- a/lg_profile.go
+++ b/lg_profile.go
@@ -24,9 +24,11 @@ var (
 func init() {
 	flag.StringVar(&cpuProfile, "cpuprofile", "",
 		"write a Go CPU profile of script/REPL execution to this file "+
-			"(build modes -c/-b/-w and the bundled-binary path are not profiled)")
+			"(build modes -c/-b/-w and the bundled-binary path are not profiled; "+
+			"env LG_CPUPROFILE is the flag-free fallback)")
 	flag.StringVar(&memProfile, "memprofile", "",
-		"write a Go heap profile to this file after script/REPL execution")
+		"write a Go heap profile to this file after script/REPL execution "+
+			"(env LG_MEMPROFILE is the flag-free fallback)")
 }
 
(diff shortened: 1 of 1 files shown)


mparrett (approved):
Nice, looks good. 

No automated test for the env fallback itself, but the implementation is small and the manual coverage matches the intent. :shipit:

PR #332, opened 2026-06-25
feat(ir): AOT native lowering of user namespaces — cross-package direct calls (#258)

nnunley:

## AOT native lowering of user namespaces — cross-package direct calls

Compile a **set** of user `.lg`/`.clj(c)` namespaces to native Go with direct
cross-package calls, make the ns→Go-package naming a single source of truth in
the gogen layer (no hardcoded Go bindings), and let `lgbgen` load namespaces
from a classpath-style source-path list. Plus two rt-layer Clojure-compat fixes
surfaced along the way.

The motivating target is AOT-compiling a multi-namespace Clojure program (the
YamlStar pipeline) where a package leaf collides with a like-named local var.

### What's here

**Cross-package direct calls** (`pkg/rt/core/ir`)
- `*cross-package-registry*` + `*export-lowered-fns*` in `lower_go.lg`: a call
  into another lowered package emits `pkg.Fn(ec, …)` and imports that package
  instead of a runtime `CachedVarFn` trampoline. Both default **off**, so core
  lowering stays byte-identical.
- `pipeline.lg` merges the cross-package registry into the pass-2 sibling
  registry (sibling last, so a ns's own fns stay intra-package).
- `scripts/lg-compile`: two-pass driver (build registry, then emit) lowering a
  file set to a Go package tree. Variadic/non-coercible fns stay on trampolines.

**Collision-resistant import aliasing**
- `pkg-alias-from` derives a `pkg_<last-two-segments>` alias (e.g.
  `.../parser/parser` → `pkg_parser_parser`) so a bare package leaf never gets
  shadowed by a ubiquitous local of the same name (`parser, _ := pkg.Make()`),
  which previously broke `pkg.Fn` resolution.

**gogen-driven ns→Go-package naming (single source of truth)**
- `gogen/ns->go-pkg`, `gogen/ns->go-reldir` in `gogen.lg` — the canonical
  leaf-segment package + nested-dir convention, consumed by **both** `cmd/lgbgen`
  (via the runtime) and `scripts/lg-compile`. `lgbgen`'s `nsToGoPkgName`/
  `nsToGoRelDir` now delegate; no parallel Go copy.

**lgbgen classpath source-paths (no hardcoded paths)**
- `--source-paths`: classpath-style list of `.lg` source roots (OS path-list
  separator), searched beyond the embedded core, threaded through every caller
  via a single `LGBGEN-SOURCE-PATHS` make var.
- Resolver: explicit `--source-paths` roots are searched **before** the embedded
  bundle, with `LG_PREFER_EMBEDDED_NS` as override. No-op when no source paths
  are set, so `./lg` and tests are unaffected.
- Resolver fix: a path file provides ns `X` only if it actually declares
  `(ns X)` — prevents a sibling file from being mistaken for a namespace's source.

**rt Clojure-compat fixes**
- `clojure.string/split` returns a **vector** (`NewArrayVector`), not a
  `PersistentList` — matching Clojure's contract for index/conj callers.
- `slurp` reads any reader-coercible value — an `IOHandle` like `*in*`, a boxed
  `LGReader`/`LGBuffer`, an `io.Reader`, or an `IReadable` — via `coerceReader`,
  not just a filename `String`. Mirrors Clojure's `(slurp *in*)` / `(slurp
  reader)`; a `String` still reads the named file. Closes the `*in*` gap.

### Tests
- `pkg/ir` go-build **behavior test** (`TestCrossPackageDirectCallCompiles`):
  lowers a two-namespace lib+app via `lg-compile` and `go build`s the result
  end-to-end — proving cross-package calls qualify, import under the right alias,
  and produce no undefined symbols / unused imports / alias collisions. Skipped
  under `-short`.
- `test/ions_test.lg` `slurp-reader`: covers `slurp` of a reader, of a file-path
  String, and of a file reader.

### Verification
- `make check-generated`: bundle + lowered tree in lockstep (content-based);
  core lowering byte-identical with both flags off.
- `go build ./...`, `pkg/rt` + `pkg/ir` + `test/` suites green.
- `gogen_ir` lowered tree compiles + dispatches natively.


mparrett:
**No findings (Codex 5.5-high)**

Reviewed PR nooga/let-go#332 against `origin/main`, focusing on the AOT cross-package lowering, resolver source-path precedence, generated artifacts, and runtime `slurp`/`split` changes. I didn’t find a concrete blocking bug.

Verification run in a temp checkout:

- `go test ./pkg/ir -run TestCrossPackageDirectCallCompiles -count=1`
- `go test ./pkg/rt ./test -run 'Test|slurp|Slurp|Ions' -count=1`
- `go test ./pkg/resolver -count=1`
- `make check-generated`
- `git diff --check origin/main...HEAD`

All passed, and the temp checkout stayed clean after `make check-generated`. Residual risk: I did not run full `go test ./...`, but the targeted coverage plus `check-generated` covers the main behavioral surface changed by this PR

nooga (approved):

nnunley:
Superseded. The cross-package direct-call machinery here landed via the ir-stack — #343 (Go-lower clojure.core), #344 (build args once), #345 (cross-package direct calls) — all merged to main. main's `*cross-pkg-registry*` + `lower-all-ns-to-go` replace this PR's parallel `*cross-package-registry*` mechanism. The genuinely-new user-namespace AOT tooling (scripts/lg-compile + examples/aot) is salvaged, ported onto the merged mechanism, in #348. Closing.

PR #335, opened 2026-06-25
feat(ir): native Go lowering of multi-arity + variadic defns (#258)

nnunley:
> Part of Epic #258 — Lowering & cross-target codegen. (iterative-development: ITER-0014)

## Native Go lowering of multi-arity + variadic defns

Top-level multi-arity defns `(defn f ([a] …) ([a b] …))` and variadic defns
`(defn g [a & xs] …)` now lower to native Go under `-tags gogen_ir` instead of
falling back to the bytecode trampoline. This unblocks AOT-lowering namespaces
whose hot paths are multi-arity/variadic (e.g. the YamlStar PEG parser driver
`call` and the `any`/`all` combinators).

### What changed
- **`compile-form*`** reconstructs the whole multi-arity defn and lowers it
  **once** into a single `rt.MakeNativeMultiArity` dispatch value (tagged
  `:multi-arity?`), reusing the existing `fn*` multi-arity lowering. The old
  per-arity `:multi-fn-template` flatten path — which emitted two colliding
  `func foo` decls (a Go redeclaration, the reason these were gated out) — is
  removed.
- **`registry-entry-from-result`** keeps multi-arity/variadic fns OUT of the
  direct-call sibling registry; callers reach them via the registered override
  value, never a dangling direct call.
- **`cmd/lgbgen`**: the `isSingleArityDefn` gate is removed so the driver admits
  multi-arity defns. `defmacro` stays excluded (native `defmacro` lowering is a
  separate follow-up).
- **`rt.MakeNativeMultiArityDeferred`** (new): a multi-arity constructor lowers to
  `func name(ec) vm.Value` whose arity branches capture `ec`, so it cannot be
  registered eagerly from `init()` (no ec) and a construction-time ec would be
  unsound for a value invoked later. Registration defers it — rebuilding the
  dispatch value with the **live** ec on each call. Still fully native (no
  bytecode `Frame.Run`); the per-call rebuild can be memoized later when arity
  bodies are ec-independent.
- **`vm.MakeMultiArity`**: treat a variadic `*NativeFn` as the `rest` branch (it
  previously only recognized `*Func`/`*Closure`). Without this a native variadic
  arity is mis-indexed as a fixed arity and rejects any call with more than its
  declared arg count. (Bytecode path is unaffected — it uses `*Func`/`*Closure`.)

### Result
8 multi-arity core fns now lower natively (was 0): `pprint`, `pprint-str`,
`lower-body`, `value-ident`, `value-name`, `infer-arg-types`, `typeinfer`,
`read-string`.

### Evidence
- `pkg/ir`: `TestLowerNsToGoMultiArityRegistersDispatchingValue`,
  `TestLowerNsToGoVariadicMultiArityLowers` (lowering shape: one decl,
  `MakeNativeMultiArity`, deferred registration, `BoxRestArgs`).
- `pkg/rt`: `TestMakeNativeMultiArityDeferredDispatch` (runtime: arity-1 vs
  variadic dispatch + rest-arg packing — this is what caught the
  `MakeMultiArity` variadic-`*NativeFn` bug).
- `make check-generated` green (bundle lockstep + lowered tree compiles +
  dispatches natively under `-tags gogen_ir`); `go test ./...` clean;
  `go build -tags gogen_ir ./...` clean; generation deterministic.

### Stacking note
This is the next layer of the AOT back-end stack and builds on
**#332** (`feat/aot-crosspkg-directcall`) plus the `feat/aot-toplevel-def`
commits. The diff against `main` is therefore cumulative; the new work in this
PR is the final commit, `feat(ir): native Go lowering of multi-arity + variadic
defns (ITER-0014)`. Review/merge after the lower layers land.


mparrett:
**Codex Findings (5.5-high)**

- [P1] [pkg/vm/func.go](/private/tmp/let-go-pr335-review/pkg/vm/func.go:257): variadic multi-arity dispatch rejects empty rest calls. `NativeFnType.Box` stores `arity: declArgs`, so a lowered branch like `func(arg0 vm.Value, args ...vm.Value)` reports arity `2`, while its minimum accepted argument count is `1`. `MultiArityFn.invokeIn` checks `le >= l.rest.Arity()`, so a native multi-arity defn whose variadic branch should accept an empty rest tail, e.g. `([x & more] ...)`, rejects `(f 1)` unless there is a separate fixed arity-1 branch. I confirmed this with a temporary local test; it failed with `doesn't have a 1-arity variant`. The new test only exercises `(g 1)` through the fixed branch and `(g 1 2 3)` through the rest branch, so it misses this edge.

**Verification**
- `go run -tags bootstrap ./cmd/lgbgen --target=go --source-paths pkg/rt/core:pkg/rt/gogen`
- `go test ./pkg/rt -run TestMakeNativeMultiArityDeferredDispatch -count=1`
- `go test ./pkg/ir -run 'TestLowerNsToGo.*MultiArity|TestCrossPackageDirectCallCompiles' -count=1`
- `git diff --check origin/main...HEAD`
- Temporary reproduction test failed as described, then was removed; temp checkout is clean.

---

**Specific concerns:**

- It breaks normal Clojure variadic semantics for a function whose only matching branch is variadic: `([x & more] ...)` must accept one argument with `more = ()`.
- The bug is in the dispatch predicate, not just a test gap: `NativeFn.Arity()` returns the Go function’s declared parameter count, including the variadic slice parameter, but the rest branch minimum arity is one less.
- It will affect generated native code whenever a multi-arity/variadic lowered function relies on its variadic arity for the minimum case and does not also have a fixed branch at that arity.

A minimal fix would be to make `MakeMultiArity` compare variadic `*NativeFn` rest branches using `Arity()-1`, or expose a `MinArity()`/`IsVariadic()` helper and use that in `MultiArityFn.invokeIn`.

nnunley:
Fixed the variadic min-arity bug in b630a0164, with the helper approach you suggested: `restMinArity(Fn)` returns `Arity()-1` for a variadic `*NativeFn` (its stored arity is the declared Go param count, fixed + the `...slice`) and `Arity()` otherwise, used at both rest-dispatch sites — `MultiArityFn.invokeIn` and the `Closure` multi-arity path. Test: a variadic-only native multi-arity now accepts `(f 1)` with `more = ()`.

Verifying that fix surfaced a second, worse regression in this PR that your repro didn't hit, so flagging it: lowering `filter` (a multi-arity defn) to native Go for the first time exposed a latent `ir.build` defect — `build-call` built each call's arguments **twice** (a discarded copy emitted only to switch to the join block before building the head, then the real copy in `build-call-with-head`). Harmless for pure ops (DCE drops the dead copy), but a dead duplicate of an **effectful** call still executes: `(chunk-cons (chunk b) (filter f (chunk-rest s)))` emitted `(chunk b)` twice, and `chunk` consumes its mutable buffer, so the dead first call drained it and the live one got an empty chunk. Net effect: `(filter pred (range n))` (the chunked-seq path) silently returned an empty seq under `-tags gogen_ir`, e.g. `(reduce + 0 (map inc (filter odd? (range 10))))` errored instead of `30`. `TestGogenAOTDiff`'s `seq.lg` caught it.

Fixed in 9acc6e68b by dropping the redundant pre-build — `build-call-with-head` already threads the head id through any block joins via `bind-local!`/`lookup-local`, so building the head first and the args once is correct (matches the existing const/number/string-head caller). The bytecode compiler is a separate path and was never affected.

Verification: new `build` dedup test (one Inc, not two); new variadic-min test; `TestGogenAOTDiff` green; full bytecode suite + `-tags gogen_ir` build/e2e green; bundle byte-deterministic; `make check-generated` clean.

nooga (approved):

nnunley:
Superseded. The "build args once" and "enable core Go-lowering" changes here landed via #344 and #343 (merged to main); multi-arity native lowering is covered by the merged ir-stack. Unique AOT tooling salvaged in #348. Closing.

PR #336, opened 2026-06-25
feat(bundle,lg): bake storage store id into native bundles (v3 trailer)

mparrett:
## Summary

Follow-up to #315. Native bundles derived their storage store id at runtime from the executable basename, so two bundles both named `lg` shared one store, and renaming the binary moved a player's saves out from under them. The WASM build already bakes the id at build time from the source basename; this closes that asymmetry by baking it for native bundles too.

## Why a trailer change

A native bundle is built by copying the `lg` binary and appending a trailer — there's no recompile step, unlike the WASM path which generates and compiles a fresh `main.go` and can inject a build-time constant. So the appended trailer is the only channel from build to runtime, and carrying a baked id means a new trailer variant.

I ruled out the two cheaper-looking alternatives: a build-time Go constant is impossible without a recompile, and deriving the id from the program content (a hash of the bytecode) would orphan every save on any code change, which breaks the "saves survive app updates" expectation. A baked source basename is both stable across updates and distinct per app.

## What changed

A v3 (`LGB3`) trailer variant:

```
v3: [lgb][resArc][store id][8:lgbSize][8:resSize][8:idSize][4:'LGB3']
```

- A non-empty id selects v3; the resource archive stays optional. A resource-only bundle still writes v2, and an id-less, resource-less bundle keeps the byte-identical legacy trailer.
- Readers recognize all three formats, so bundles built before this read back unchanged with an empty id — at which point the runtime keeps the old exe-basename behavior.
- `bundleBinary` bakes `storageIDForScript(src)` — an explicit `-storage-id`, else the source basename. At startup a bundle prefers the baked id over the exe fallback.
- The payload-fits-file overflow guard subtracts the third size step by step like the existing two, so a crafted `idSize` can't wrap the sum or drive a bad seek or allocation.

## Validation

- `go test ./...`, `go vet ./...`, gofmt clean; linux, js/wasm, and plan9 builds; the generated-artifacts check.
- New `pkg/bundle` tests: v3 round-trip with and without resources, a corrupt-`idSize` guard, and v3 `payloadFitsFile` cases.
- End to end: bundled a `coolgame.lg`, renamed the binary to `lg`, ran it from an unrelated directory — the store keyed by `coolgame` (the baked id) rather than `lg` or the working directory.

Behavior for bundles that set neither `-storage-id` nor carry a v3 trailer is unchanged.

### DIFF
--- a/lg.go
+++ b/lg.go
@@ -146,10 +146,12 @@ func checkBundledLGB() (lgb []byte, res []byte, storeID string) {
 func bundleBinary(ctx *compiler.Context, nsRes *resolver.NSResolver, src string, dst string, basePath string) error {
 	ctx.SetSource(src)
 
-	// Snapshot the resource roots and output path *before* compiling — and
-	// absolutize them against the current cwd — because CompileMultiple runs
-	// the program's top-level forms, which may change the working directory.
-	// Relative roots resolved afterward would point at the wrong place.
+	// Snapshot the resource roots, output path, and storage store id *before*
+	// compiling — and absolutize the paths against the current cwd — because
+	// CompileMultiple runs the program's top-level forms, which may change the
+	// working directory. Relative roots resolved afterward would point at the
+	// wrong place, and the store id (which falls back to the cwd basename for
+	// main.lg/init.lg) would key off the changed directory.
 	resRoots := buildResourcePaths()
 	for i, r := range resRoots {
 		if abs, aerr := filepath.Abs(r); aerr == nil {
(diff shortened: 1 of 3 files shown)


nooga (approved):

mparrett:
**Codex Findings (5.5-high)**

- `P2` [lg.go:239](/private/tmp/let-go-pr336-review/lg.go:239): `storageIDForScript(src)` is called after `ctx.CompileMultiple(f)`, but compilation runs top-level forms and this repo already accounts for those forms changing cwd. For `main.lg` / `init.lg`, `StorageIDFrom` falls back to the cwd basename, so a compile-time `(syscall/chdir ...)` can bake the wrong store ID into the v3 trailer. Snapshot the bundle store ID before compilation, alongside `resRoots` and `dstAbs`, then pass that saved value to `AppendTrailer`.

**Notes**

The overall direction looks right: the v3 trailer is narrowly scoped, older formats still read, and the overflow guard remains conservative. I did not find other blockers.

Validation I ran locally in a temp checkout:
- `go test ./pkg/bundle`
- `go test ./...`

nnunley (approved):

PR #338, opened 2026-06-25
feat(docs): docs-status report for the judgement layer

mparrett:
`docs/frontmatter-hook.md` lists a follow-up under "Out of scope": a `docs-status` report over the judgement-layer frontmatter fields. This adds it.

The convention already has a mechanical floor — the pre-commit hook stubs and bumps `last-verified:`, and the `docs-frontmatter` CI check rejects missing or unparseable frontmatter. Nothing drives the fields the hook deliberately leaves to a human: whether a doc is still current, whether a supersession link has its other half, whether two docs claim the same `authoritative-for:` topic, whether the README index routes to a doc at all.

## The tool

`scripts/docs_status.py` is a read-only report over `docs/**/*.md` frontmatter. Stdlib-only, like the existing hook — nothing to install. It never mutates; it reports and leaves judgement to a reader. Six checks:

- stale `last-verified:` — older than a threshold; the hook bumps it on every touch, so a stale value means the doc is untouched, not wrong
- aged `human-verified:` — a human vouched, but long enough ago to be worth refreshing
- invalid date values — a `last-verified:`/`human-verified:` that's present but not a valid `YYYY-MM-DD`, so a typo'd date surfaces instead of being silently ignored
- supersession integrity — a `superseded-by:` / `supersedes:` link that dangles or isn't mirrored on the other doc
- `authoritative-for:` clashes — one topic claimed by more than one doc
- docs missing from the README index

`human-verified:` is treated specially: a blank value is reported only as a count, never per-doc. It records that a human vouched for a doc and is set only by explicit human action, so the tool never writes it.

## CI

A non-blocking `docs-status` job runs on PRs that touch `docs/`, scans the whole tree (the supersession, clash, and index checks are cross-document), and writes the report to the run's step summary plus a one-line annotation — a notice when clean, a warning otherwise. It never fails the build; `docs-frontmatter` stays the only gate. Staleness and the judgement fields are reviewer signals, not blockers.

## Index updates

Routes four docs the index didn't reach: `frontmatter-hook.md`, `docs-status.md`, `exec-context-threading.md`, and `regenerating-generated-artifacts.md`.

The tool also flags `docs/perf/microbench/README.md` as unindexed. I left that one alone — whether the top-level index should route to a sub-area README is your call, and the tool's job is to surface it, not decide.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -232,7 +232,12 @@ jobs:
             echo "No docs/**/*.md changed in this PR; skipping docs-status report."
             exit 0
           fi
-          report=$(python3 scripts/docs_status.py || true)
+          # Capture the exit code without failing the step (the report is
+          # non-blocking). rc is non-zero when a doc is unreadable/malformed.
+          set +e
+          report=$(python3 scripts/docs_status.py)
+          rc=$?
+          set -e
           headline=$(printf '%s\n' "$report" | head -1)
           {
             echo '## docs-status report'
@@ -245,10 +250,11 @@ jobs:
             echo '```'
           } >> "$GITHUB_STEP_SUMMARY"
           # Surface the headline as an annotation + log line so the result is
-          # legible from the PR without opening the run summary: a notice when
-          # clean, a (non-blocking) warning when there's anything to look at.
+          # legible from the PR without opening the run summary. "Clean" requires
+          # both a clean exit (no read/parse errors) and zero findings — an
+          # errors-only run still prints "0 findings", so the headline alone lies.
           echo "$headline"
-          if printf '%s' "$headline" | grep -q ' 0 findings'; then
+          if [ "$rc" -eq 0 ] && printf '%s' "$headline" | grep -q ' 0 findings'; then
             echo "::notice title=docs-status::${headline} — clean"
           else
             echo "::warning title=docs-status::${headline} — see the run summary"
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -208,6 +208,52 @@ jobs:
           echo "$changed"
           echo "$changed" | xargs python3 scripts/docs_frontmatter_hook.py --check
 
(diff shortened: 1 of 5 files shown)


mparrett:
**Codex Findings (5.5-high)**

- [P2] [scripts/docs_status.py](/private/tmp/let-go-pr338-review/scripts/docs_status.py:176) does not count `rep.errors`, while CI masks the script exit with `|| true` in [.github/workflows/go.yml](/private/tmp/let-go-pr338-review/.github/workflows/go.yml:235) and decides “clean” from the headline at line 251. A malformed-doc-only run prints `docs-status: 1 docs scanned, 0 findings` plus an `## errors` section, so CI would emit a clean notice despite the documented failure path. Count errors in the headline or make the workflow warn when the script exits non-zero.
- [P3] [scripts/docs_status.py](/private/tmp/let-go-pr338-review/scripts/docs_status.py:134) collapses invalid dates and blank/missing dates to `None`. That means `last-verified: not-a-date` gets no finding at all, and `human-verified: not-a-date` is counted as blank. Since the existing frontmatter check only verifies key presence, this report should surface invalid date values as errors or findings.

**Checked**
- Ran `python3 scripts/docs_status.py`: `31 docs scanned, 1 findings` for the intentional unindexed nested README.
- Ran `git diff --check`: clean.
- Refreshed PR checks: all currently successful.

mparrett:
Addressed both findings in 2407a4b:

- **P2** — the CI notice now keys "clean" off the script's exit code *and* zero findings, not the headline string. An errors-only run (e.g. a malformed doc) exits non-zero and the headline carries the error count (`… 0 findings, 1 errors`), so it renders as a warning, not a clean notice.
- **P3** — `parse_date` no longer collapses invalid and blank dates together. A present-but-unparseable `last-verified:`/`human-verified:` is now an `invalid-date` finding rather than being dropped or counted as never-vouched.

Verified: an errors-only run emits a warning; a typo'd date produces a finding; the real tree is unchanged (the one intentional `perf/microbench` finding).

nnunley (approved):

PR #339, opened 2026-06-26
feat: add js/url-param for reading the launch URL from a WASM bundle

mparrett:
## What

Adds `(js/url-param name)` to the `js` namespace: reads a query parameter from the
page URL that loaded a WASM bundle, returning its value as a string, or nil if the
param is absent or the code is running off-browser.

## Why

A WASM bundle has no first-class way to read its own launch URL, so parameterizing
one — a seed, a mode flag — means hand-writing JS around the host. `js/url-param`
lets `.lg` boot code do it directly:

```clojure
(let [seed (or (some-> (js/url-param "seed") read-string) (default-seed))]
  ...)
```

It sits in the same host-seam family as `js/emit` (#241) and the storage seam
(#315): a small, capability-scoped bridge to the page.

## How

- `pkg/rt/js.go` registers the primitive on the unified `js` namespace. Bad arity
  or a non-string name returns nil rather than erroring, so an optional-param read
  in boot code can't crash.
- `urlparam_default.go` (non-wasm) is a nil-returning stub, keeping the primitive
  callable in native dev and tests.
- `urlparam_js_wasm.go` parses the page URL. WASM runs in a Web Worker whose own
  `location` is the worker script, not the launching page, so the host shell
  forwards `window.location.search` into the worker as the `_lgUrlSearch` global
  (`lg-host-core.js`), and the primitive reads it via `URLSearchParams`.

## Tests

`urlparam_test.go` covers the native stub contract and the primitive's registration
plus argument handling (arity, non-string name, absent name). The wasm parsing path
runs in a browser and isn't unit-tested here, matching how the emit and storage host
seams are tested. The assemble golden fixtures are updated for the new host
forwarding.

### DIFF
--- a/pkg/rt/js.go
+++ b/pkg/rt/js.go
@@ -75,5 +75,25 @@ func installJSNS() {
 	})
 	ns.Def("emit", emitFn)
 
+	// (js/url-param name) -> string-or-nil. Reads a query parameter from the
+	// page URL that loaded the bundle; nil if absent or off-browser. Lets .lg
+	// boot code parameterize from the URL (?seed=42, ?bench=fast) without JS
+	// plumbing. The page URL reaches the worker via the host seam below
+	// (see hostURLParam).
+	urlParamFn, _ := vm.NativeFnType.Wrap(func(vs []vm.Value) (vm.Value, error) {
+		if len(vs) != 1 {
+			return vm.NIL, nil
+		}
+		name, ok := vs[0].(vm.String)
+		if !ok {
+			return vm.NIL, nil
+		}
+		if v, ok := hostURLParam(string(name)); ok {
+			return vm.String(v), nil
+		}
+		return vm.NIL, nil
+	})
+	ns.Def("url-param", urlParamFn)
+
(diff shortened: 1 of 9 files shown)


mparrett:
Reviewed PR #339 at head `a48f1ac`.

No blocking issues found.

The implementation looks consistent with the existing host-seam pattern: native/off-browser returns nil, wasm reads the forwarded page search string, and the host JS updates cover worker and main-thread modes plus the assembler goldens.

Verification run in `/private/tmp/let-go-pr339`:

- `go test ./pkg/rt ./pkg/rt/wasm`
- `env GOOS=js GOARCH=wasm go test -c -o /private/tmp/let-go-pr339-rt.wasm.test ./pkg/rt`
- `go test ./...`

All passed. The only residual gap is that I did not run a live browser smoke for `(js/url-param ...)`; the wasm path was compile-checked and the generated host HTML was covered by the existing golden tests.

nooga (approved):

nnunley (approved):

Issue #259, opened 2026-06-18
Epic: Host integration & interop


nnunley:
Umbrella for let-go interoperating with the world beyond its own core — host-platform capabilities, Go-package interop, and running external Clojure/ecosystem code on let-go.

## Members
- #255 — Host capabilities: generalizing the I/O decoupling to e.g. audio and graphics
- #97 — Roadmap: Go-package interop (lg-gen, smart wrappers, Mode B AOT lowering)
- #134 — Tracking: run stock metosin/malli on let-go (no fork, no new default deps)
- #49 — Running yamlscript atop let-go

## Workflow
File host-integration and interop issues here as sub-issues. Each child should stand on its own; this epic tracks the set and its progress. (The ecosystem-compat items #134/#49 may split into their own epic if this umbrella gets crowded.)

PR #340, opened 2026-06-26
feat(wasm): add -w-host-eval to call into a live image via LetGoHost.eval

mparrett:
# feat(wasm): add `-w-host-eval` to call into a live image via `LetGoHost.eval`

## The gap

`lg -w` bundles run the program's `main` and exit. That's the right default for a
batch or TUI program, but it means JS can only receive data *out* of the runtime
(`js/emit` → `LetGoHost.onEmit`) — it can't call *into* the loaded image to trigger
work. The standalone `wasm/main.go` does expose `window.Eval`, but it embeds no
program (only a bare `user` namespace), so the two capabilities never compose: you
can have your namespaces loaded, or a callable eval, but not both.

That's the missing piece for any "live wasm" let-go app (epic #259): a page that
boots the runtime once and drives it from JS (regenerate on a control change, eval
into the image for live-coding) rather than reloading per action.

## What this adds

A `-w-host-eval` flag for `-w`. When set, the generated main, *after* running the
program's main chunk (typically a set of `defn`s), installs the eval bridge and
parks so the image stays callable. The public surface is on `LetGoHost`, matching
the contract from #79/#245 (`_lg*` are internal hooks, `LetGoHost` is the one public
object):

```js
await LetGoHost.eval(code)   // run code in the loaded image
                             // → Promise<string> (value, or a formatted error)
```

Structured results still come back out-of-band through `(js/emit …)`; the string
return is for triggering and for simple values. Intended pairing is `-w-shell none`:
the client owns the UI and drives the runtime through `LetGoHost.eval`.

```
lg -w out -w-shell none -w-host-eval app.lg
# out/index.html boots the runtime, then LetGoHost.eval(code) is live
```

## Both boot modes, ready-gated

**Works in both boot modes.** `LetGoHost.eval` dispatches a direct `_lgEval` call on
the main thread, or a worker `postMessage` round-trip when the bundle is
cross-origin isolated (worker mode). So eval is reachable whether or not the page
got COI — no constraint on how the bundle is served, and the eval'd program never
calls `read-key`, so worker mode's input ring is untouched. (An earlier cut set a
bare `window.Eval` from the runtime, which only worked on the main thread and forced
the bundle to avoid COI; this replaces it.)

**Ready-gated, so there's no race.** `onReady` fires when the glue is wired, *before*
the wasm runs — and the eval hook only exists after the program's main returns. The
runtime calls `_lgRuntimeReady` once the hook is installed; `LetGoHost.eval` awaits
that, so a call made before boot queues instead of hitting an undefined hook.

One limitation by construction: the program's `main` must return. The eval hook is
installed after the main chunk runs, so a program that blocks in its own `main` (a
TUI key loop) never reaches it. Host-eval suits programs whose top level is
definitions.

## Implementation

The generated-main template, its `-w-host-eval` splice, and `RenderMain` live in
`pkg/rt/wasm` (next to `AssembleHTML` and the `lg-host-core.js` it substitutes into);
its test sits there too, clearing the repo-root test-file ban. The host-eval splice
follows the existing `__LG_STORAGE_ID__` placeholder approach: with the flag off, the
marker lines drop out whole, so the default bundle's generated main is unchanged.
`RenderMain` is unit-tested (including a `go/format` parse of the output) without
invoking the `GOOS=js` toolchain.

## Verification

Built a generator bundle (`-w-shell none -w-host-eval -w-wasm external`) and ran it
in both boot modes: served plain with the service worker blocked
(`crossOriginIsolated === false`, main-thread), and served with COOP/COEP headers
(`crossOriginIsolated === true`, worker). Both:

- `LetGoHost.eval('(+ 1 2)')` → `'3'`
- `LetGoHost.eval('(app/generate 42 2)')` triggers generation; floors arrive via
  `LetGoHost.onEmit`
- a `defn` redefinition through eval persists, so the next call uses it (live-coding)
- a bad form resolves to the formatted error string

`gofmt` clean; `go vet`, the `pkg/rt/wasm` tests (golden bundle refreshed for the
`lg-host-core.js` change), the root-test-location check, and `GOOS=linux`/`plan9`
builds pass. (`GOOS=windows` fails in `pkg/rt/term.go` on `unix.SIGWINCH` etc. —
pre-existing on `main`, unrelated to this change.)

## Scope

Deliberately minimal: a string-in / string-out trigger plus the existing emit
channel, enough to drive a live runtime. A richer return that builds real JS values
instead of a stringified one is a possible future enhancement; nothing tracks it yet.

Marked draft for design feedback on the surface (`LetGoHost.eval` shape, the
`_lgRuntimeReady` signal) before polish (a worker-mode test in CI, docs).

### DIFF
--- a/lg.go
+++ b/lg.go
@@ -321,7 +321,7 @@ func init() {
 	flag.StringVar(&wasmOutput, "w", "", "build .lg file into a WASM web app (specify output directory)")
 	flag.StringVar(&wasmShell, "w-shell", "xterm", "shell for -w: 'xterm' (default) or 'none' (emit core only; client supplies its own shell via window.LetGoHost)")
 	flag.StringVar(&wasmPayload, "w-wasm", "inline", "wasm delivery for -w: 'inline' (default; gzip-base64 baked into index.html) or 'external' (emit a separate main.wasm the loader fetches + streams)")
-	flag.BoolVar(&wasmHostEval, "w-host-eval", false, "for -w: expose window.Eval(code) to call into the loaded image and keep it live (park after the program's main returns); pair with -w-shell none. Main-thread only — serve without cross-origin isolation")
+	flag.BoolVar(&wasmHostEval, "w-host-eval", false, "for -w: expose LetGoHost.eval(code) to call into the loaded image and keep it live (park after the program's main returns); works in both boot modes. Pair with -w-shell none")
 	flag.StringVar(&storageID, "storage-id", "", "logical storage store id for the storage namespace (default: script name, or current directory for main.lg)")
 	flag.StringVar(&sourcePaths, "source-paths", "",
 		"namespace search paths separated by the OS path-list separator "+
--- a/lg.go
+++ b/lg.go
@@ -302,6 +302,7 @@ var bundleBase string
 var wasmOutput string
 var wasmShell string
 var wasmPayload string
+var wasmHostEval bool
 var storageID string
 var sourcePaths string
 var resourcePaths string
@@ -320,6 +321,7 @@ func init() {
 	flag.StringVar(&wasmOutput, "w", "", "build .lg file into a WASM web app (specify output directory)")
 	flag.StringVar(&wasmShell, "w-shell", "xterm", "shell for -w: 'xterm' (default) or 'none' (emit core only; client supplies its own shell via window.LetGoHost)")
 	flag.StringVar(&wasmPayload, "w-wasm", "inline", "wasm delivery for -w: 'inline' (default; gzip-base64 baked into index.html) or 'external' (emit a separate main.wasm the loader fetches + streams)")
+	flag.BoolVar(&wasmHostEval, "w-host-eval", false, "for -w: expose window.Eval(code) to call into the loaded image and keep it live (park after the program's main returns); pair with -w-shell none. Main-thread only — serve without cross-origin isolation")
 	flag.StringVar(&storageID, "storage-id", "", "logical storage store id for the storage namespace (default: script name, or current directory for main.lg)")
 	flag.StringVar(&sourcePaths, "source-paths", "",
 		"namespace search paths separated by the OS path-list separator "+
@@ -589,7 +591,7 @@ func main() {
 			fmt.Fprintf(os.Stderr, "error: -w-wasm must be 'inline' or 'external', got %q\n", wasmPayload)
 			os.Exit(1)
 		}
-		if err := buildWasm(context, nsResolver, files[0], wasmOutput, wasmShell == "xterm", wasmPayload == "external", storageIDForScript(files[0])); err != nil {
+		if err := buildWasm(context, nsResolver, files[0], wasmOutput, wasmShell == "xterm", wasmPayload == "external", wasmHostEval, storageIDForScript(files[0])); err != nil {
 			fmt.Fprintf(os.Stderr, "error: %v\n", err)
 			os.Exit(1)
 		}
--- a/pkg/rt/wasm/assemble_test.go
+++ b/pkg/rt/wasm/assemble_test.go
@@ -31,14 +31,16 @@ func TestAssembleHTMLGolden(t *testing.T) {
 		name     string
 		shell    bool
 		external bool
+		hostEval bool
 		golden   string
 	}{
-		{"xterm shell, inline wasm", true, false, "assemble_golden.html"},
-		{"shell-less core, inline wasm", false, false, "assemble_golden_shellless.html"},
-		{"xterm shell, external wasm", true, true, "assemble_golden_external.html"},
-		{"shell-less core, external wasm", false, true, "assemble_golden_shellless_external.html"},
+		{"xterm shell, inline wasm", true, false, false, "assemble_golden.html"},
+		{"shell-less core, inline wasm", false, false, false, "assemble_golden_shellless.html"},
+		{"xterm shell, external wasm", true, true, false, "assemble_golden_external.html"},
+		{"shell-less core, external wasm", false, true, false, "assemble_golden_shellless_external.html"},
+		{"shell-less core, external wasm, host-eval", false, true, true, "assemble_golden_hosteval.html"},
 	} {
-		got := AssembleHTML(wasmExecJS, wasmGzB64, tc.shell, tc.external)
+		got := AssembleHTML(wasmExecJS, wasmGzB64, tc.shell, tc.external, tc.hostEval)
 		goldenPath := filepath.Join("testdata", tc.golden)
 
 		if *updateGolden {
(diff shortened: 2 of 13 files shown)


mparrett:
**Findings (Codex 5.5-high):**

Changes needed. The minimal evaluator logic is reasonable, but the browser-facing approach conflicts with the existing host contract and has lifecycle failures.


- **[P1] CI is failing because `wasm_test.go` is at the repository root.** The project explicitly forbids root test files via [`[test-location](https://github.com/nooga/let-go/blob/ba856ba2a049f24d3a12ed398977df233190731b/.github/workflows/go.yml#L26-L36)`](https://github.com/nooga/let-go/blob/ba856ba2a049f24d3a12ed398977df233190731b/.github/workflows/go.yml#L26-L36). The helper likely needs extracting into an internal package so its tests can move out of root. [[Failing check](https://github.com/nooga/let-go/actions/runs/28271640683/job/83770043918)](https://github.com/nooga/let-go/actions/runs/28271640683/job/83770043918).

- **[P1] There is no reliable “Eval is ready” signal.** `LetGoHost.onReady` fires before WASM instantiation and execution, while `window.Eval` is installed only after the program’s main chunk returns. A client following the existing API can therefore call Eval while it is undefined. See [`[_ready('main')](https://github.com/nooga/let-go/blob/ba856ba2a049f24d3a12ed398977df233190731b/pkg/rt/wasm/lg-host-core.js#L273)`](https://github.com/nooga/let-go/blob/ba856ba2a049f24d3a12ed398977df233190731b/pkg/rt/wasm/lg-host-core.js#L273), [`[go.run](https://github.com/nooga/let-go/blob/ba856ba2a049f24d3a12ed398977df233190731b/pkg/rt/wasm/lg-host-core.js#L345)`](https://github.com/nooga/let-go/blob/ba856ba2a049f24d3a12ed398977df233190731b/pkg/rt/wasm/lg-host-core.js#L345), and [[Eval installation](https://github.com/nooga/let-go/blob/ba856ba2a049f24d3a12ed398977df233190731b/wasm.go#L129-L146)](https://github.com/nooga/let-go/blob/ba856ba2a049f24d3a12ed398977df233190731b/wasm.go#L129-L146).

- **[P1] Reusing an output directory can permanently defeat host-eval.** The build merely skips writing `coi-serviceworker.js`; it does not remove an existing one. Meanwhile host-core still registers that path. Rebuilding a normal bundle into the same directory with `-w-host-eval` can therefore keep activating COI, boot the worker, and leave `window.Eval` unavailable. See [[service-worker omission](https://github.com/nooga/let-go/blob/ba856ba2a049f24d3a12ed398977df233190731b/wasm.go#L309-L318)](https://github.com/nooga/let-go/blob/ba856ba2a049f24d3a12ed398977df233190731b/wasm.go#L309-L318) and [[unconditional registration](https://github.com/nooga/let-go/blob/ba856ba2a049f24d3a12ed398977df233190731b/pkg/rt/wasm/lg-host-core.js#L10-L17)](https://github.com/nooga/let-go/blob/ba856ba2a049f24d3a12ed398977df233190731b/pkg/rt/wasm/lg-host-core.js#L10-L17). Server-provided COI causes the same fundamental failure.

Approach recommendation:

Use `window.LetGoHost.eval(code)` returning a Promise, with an explicit runtime-ready state. Relay evaluation through worker messages when isolated and invoke directly in main-thread mode. This preserves input/COI and keeps one API across both boot modes.

That fits the documented rule that `LetGoHost` is the browser’s single public contract. [[Issue #255](https://github.com/nooga/let-go/issues/255)](https://github.com/nooga/let-go/issues/255) is mainly about guest-to-host capabilities, not structured Eval results, but its binding discipline reinforces this direction; it does not support adding another raw global.

Targeted `go test . ./pkg/rt/wasm` passed locally.

mparrett:
**Codex 5.5-high:**

Re-review verdict: substantially improved; all three previous blockers are fixed.
 
 One remaining finding:
- **[P1] `LetGoHost.eval` hangs forever on bundles built without `-w-host-eval`.** The method is installed unconditionally, but `runtimeReady` is never resolved in ordinary bundles, so the intended “not built with host-eval” error is unreachable. Feature detection also incorrectly reports Eval as available. See [`[lg-host-core.js](https://github.com/nooga/let-go/blob/d9c28d5e9f31b4ddf294e8a6bf1677785e4109b0/pkg/rt/wasm/lg-host-core.js#L107-L123)`](https://github.com/nooga/let-go/blob/d9c28d5e9f31b4ddf294e8a6bf1677785e4109b0/pkg/rt/wasm/lg-host-core.js#L107-L123).

I reproduced this in a generated default bundle: after three seconds, the Promise remained pending. Either omit `LetGoHost.eval` when disabled or make it reject immediately; this likely requires passing `hostEval` into HTML assembly.

Verification:

- Worker/COI Eval worked, including calls made before runtime readiness.
- Targeted tests and test-location guard passed.
- All current CI checks are green.
- The new `LetGoHost` API and dual-mode relay now align with the project’s interop direction.

mparrett:
Fixed in c55d774.

`hostEval` now threads through `AssembleHTML` as a `HOST_EVAL` flag, and
`LetGoHost.eval` is gated on it, so the method exists only in `-w-host-eval`
bundles. A default bundle reports `typeof LetGoHost.eval === 'undefined'` (no hang,
honest feature detection); host-eval bundles still eval in both boot modes. Added a
host-eval golden case and the `__LG_HOST_EVAL__` marker to the marker-removal test.

nooga (approved):

mparrett:
cc: @nnunley Does this shape look reasonable?

Motivation, if not already expressed: it lets a JS host drive a live in-page let-go image (`LetGoHost.eval` into an image that stays resident instead of run-and-exit). My immediate use is a wasm-mode dungeon testbed. My current local prototype works locally against a native `lg` HTTP/nREPL server, and that makes the front-end lightweight. It's great to keep as an option. This PR's interop would let the same live-redefine loop run fully in-browser with no server (so it's hostable/shareable and resets on reload). The testbed motivation is a bit niche, but I think the interop primitive is worth having regardless.

Also, I'm thinking this might be the last big piece of browser-side/interop to include in the 1.11.0 cut. Thoughts/opinions?

nnunley:
Shape looks reasonable. The `-w-host-eval` flag wiring doesn't perturb the non-wasm build, and the `__LG_HOST_EVAL_*__` template substitution produces valid Go in both the host-eval-on and -off paths (covered by the rendermain tests).

One robustness call-out for the worker path: in `pkg/rt/wasm/lg-host-core.js` the `evalPending` Map entry is never deleted if a worker eval never gets a response (worker crash / dropped message), so it leaks. A timeout that rejects + removes the pending entry would harden it. Not blocking for a draft.

mparrett:
Thanks — good catch on the worker-path leak. Fixed in 8a3186c.

Each worker eval now arms a 30s timer; on timeout it rejects the promise and removes the `evalPending` entry, so a worker crash or dropped message no longer leaks the entry or hangs the caller. The worker runs eval synchronously, so this is a lost-response net rather than a slow-eval killer — a reply that arrives after the timeout is harmless, since its id is already gone from the map and the result is dropped. The `delete(id)` guard on the timeout path means a result/timeout race can't double-settle the promise.

I left the 30s as a fixed constant rather than a per-call option — adding an argument changes the `LetGoHost.eval` contract, which seemed heavier than this hardening needs. Easy to make it configurable later if you want that.

Main-thread mode calls `_lgEval` directly with no pending map, so it's unaffected. Goldens regenerated since the core JS embeds into every bundle.

nnunley:
The `evalPending` leak call-out is resolved by `wasm: time out worker host-eval to bound pending-map leak`. The 30s timeout `delete()`s the entry and rejects the promise, `clearTimeout` runs on the success path so completed evals don't leak timers, and a late response after timeout no-ops on the missing entry — all three paths (success / timeout / late-response) clean up correctly. Looks good.

PR #341, opened 2026-06-27
feat(rt): add set-rand-seed! for reproducible rand, on math/rand/v2

mparrett:
`rand`, `rand-int`, `rand-nth`, `shuffle`, and `math/random` all drew from Go's package-level `math/rand` functions. Those are auto-seeded at startup and expose no usable `Seed`, so there's no way to pin a reproducible sequence for a test, a benchmark, or a bug reproduction.

This routes all of them through a single package-local `math/rand/v2` generator and adds `(set-rand-seed! n)` to reseed it.

## Why `math/rand/v2`

v2 has no global `Seed` by design: the way to get a reproducible stream is to construct your own source, which is what `set-rand-seed!` does. The [math/rand/v2 announcement](https://go.dev/blog/randv2) covers the rationale. These five primitives were also the only `math/rand` (v1) users in the tree, so the switch removes v1 entirely.

## Concurrency

A `*math/rand/v2.Rand` is not safe for concurrent use, unlike the package-level functions it replaces. A `sync.Mutex` guards every draw and the reseed, so the primitives keep the concurrency safety they had before. The default generator is seeded from the auto-seeded top-level v2 source, so it needs no `time.Now()` and keeps full entropy; `set-rand-seed!` swaps in a deterministic source.

## Scope

The trailing bang marks the process-wide mutation. This is for tests, benchmarks, and reproducible bug-repros, not a gameplay or simulation determinism mechanism. Code that needs that should thread explicit state instead. A given seed reproduces a sequence only when the rand calls happen in a deterministic order.

## Testing

`rand_seed_test.go` asserts that reseeding with the same value reproduces the sequence (it compares sequences, not specific generator outputs, which are an RNG implementation detail that can change across versions) and runs a `-race` stress over concurrent draws and reseeds. `go test -race ./...`, `go vet`, and `gofmt` are clean; linux, wasm, and plan9 builds pass.

### DIFF
--- a/pkg/rt/lang.go
+++ b/pkg/rt/lang.go
@@ -13,7 +13,7 @@ import (
 	"fmt"
 	"math"
 	"math/big"
-	"math/rand"
+	"math/rand/v2"
 	"os"
 	"sort"
 	"strconv"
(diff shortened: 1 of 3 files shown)


mparrett:
No actionable findings in [[PR #341](https://github.com/nooga/let-go/pull/341)](https://github.com/nooga/let-go/pull/341).

Verified:

- RNG access and reseeding are mutex-protected.
- All five intended primitives use the shared generator.
- `go test ./pkg/rt/...` and race tests pass.
- CI checks pass except the aggregate build still pending.
- Full local suite only hit sandbox-restricted network-binding failures unrelated to this PR.

Residual gap: tests exercise private RNG helpers, not `set-rand-seed!` through the language runtime. This is worth improving but not merge-blocking.

mparrett:
Codex: No actionable findings in [[PR #341](https://github.com/nooga/let-go/pull/341)](https://github.com/nooga/let-go/pull/341).

Verified:

- RNG access and reseeding are mutex-protected.
- All five intended primitives use the shared generator.
- `go test ./pkg/rt/...` and race tests pass.
- CI checks pass except the aggregate build still pending.
- Full local suite only hit sandbox-restricted network-binding failures unrelated to this PR.

Residual gap: tests exercise private RNG helpers, not `set-rand-seed!` through the language runtime. This is worth improving but not merge-blocking.

mparrett:
Addressed the residual coverage note: added a test that drives `set-rand-seed!` and `rand-int` as resolved `core` primitives (not the private Go helpers), so registration and wiring are covered end to end: reseeding to the same value through the runtime reproduces the `rand-int` sequence. Folded into the existing commit.

nooga (approved):

nnunley (approved):

PR #342, opened 2026-06-27
fix(ir): empty-list () lowers to the empty-list constant, not invoke-nil (#258)

nnunley:

## Problem

In the native IR pipeline, the empty list `()` is a self-evaluating constant (the empty list), not a call with no operator. Two bugs conspired to lower it to an invoke of a nil callee (`rt.InvokeValueEC(ec, vm.NIL, …)`), which faulted at runtime ("invoke of nil") once `clojure.core` was Go-lowered — e.g. `(defn reverse [coll] (reduce conj () coll))`:

1. `pipeline/expand-all` reconstructed `()` as `(apply list (expand-all nil)) = (nil)` — a call of nil.
2. `ir.build/build-list` had no empty-list case, so `()` fell through to `build-call`.

## Fix

- `expand-all` guards the `(seq? form)` branch so an empty list is returned untouched.
- `build-list` mirrors the bytecode compiler (`OP_LOAD_CONST vm.EmptyList`) with a `:const` Inst for the empty list (defense in depth).

## Verification

New regression test `TestLowerGoEmptyListLowersToEmptyListConstNotNilInvoke`; full pkg/ir + clojure suites green in both bytecode and gogen_ir modes.

---
First of a 3-PR stack (empty-list → clojure.core lowering → build-call dedup).


mparrett:
*Codex Findings (5.5-high)*

No scoped correctness findings.

I reviewed `origin/main...5fe6527`, including both empty-list guards and the generated-artifact updates. The focused validation passed:

```text
go test ./pkg/ir ./pkg/rt/...
```

I did not treat unrelated pre-existing behavior as findings.

nooga (approved):

PR #343, opened 2026-06-27
feat(ir,rt): Go-lower clojure.core + direct-call hot builtins via pkg/rt/builtins (#258)

nnunley:

## Summary

Enables AOT Go-lowering of `clojure.core` itself (previously `lgbgen:skip-go`-excluded) and adds a clean, vm-only `pkg/rt/builtins` leaf package whose exported Go functions back the hottest core builtins, so lowered code direct-calls them instead of trampolining through `rt.InvokeValue`.

## What's here

- **clojure.core lowers natively** — removed the `skip-go` exclusion; the goto-over-declaration codegen bug (typed-direct-call temps) is fixed by giving those temps function scope.
- **`pkg/rt/builtins`** — a vm-only leaf package (seq/first/rest/next/count/vector/not/cons/contains?, …) registered as a native module via `(rt/native-modules)`, so cross-ns calls into them lower to direct `builtins.X(…)` / `corefns.X(…)` Go calls.
- **`list` intrinsic** — `(clojure.core/list a … n)` lowers to `vm.EmptyList.Cons(n)…Cons(a)` (no args-slice allocation; keeps `list?` true), written in the .lg lowering layer.

## Verification

`make check-generated` green; pkg/ir, pkg/rt, and the clojure suites (bytecode + gogen_ir) all pass.

---
Stacked on #342 (review/merge after it). The diff includes #342's changes until that merges.


mparrett:
*Codex Findings (5.5-high)*

1. **[P1] The new `list` intrinsic bypasses var-root rebinding semantics.** `call-assign-stmts` invokes `list-intrinsic-call-stmts` before `lookup-direct-callable`, so it does not honor `*direct-calls-disabled?*`. That flag exists specifically so `with-redefs`, `alter-var-root`, and `intern` continue dispatching through the Var.

Concrete repro:

```clojure
(defn probe [x]
  (with-redefs [list (fn [& _] :overridden)]
    (list x)))
```

Lowering this PR emits the following inside the callback:

```go
v2 = vm.EmptyList.Cons(arg0)
```

Therefore `probe` returns `(x)` instead of `:overridden`. I reproduced this with `ir.passes.pipeline/lower-ns-to-go`; this is not speculative. A direct fix is to make `list-intrinsic-call-stmts` return nil when `*direct-calls-disabled?*` is true (or move intrinsic selection behind the same guard), then add a regression parallel to `TestWithRedefsDisablesNativeDirect` using `list`.

2. **Future footgun (not a current semantic regression): `pkg/rt/builtins` is not yet the claimed single source for interpreter and AOT behavior.** `installLangNS` still registers the separate local implementations (`not`, `vector`, `cons`, and `contains`), while `registerBuiltinsModule` only registers string metadata for AOT. The new package header says these functions back both paths, but they currently do not. Either wire the interpreter registrations to these exports or label this as transitional and add differential tests like `pkg/rt/corefns/corefns_diff_test.go`; otherwise the two paths can silently drift.

Validation after materializing generated output:

```text
make check-generated
go test ./pkg/ir ./pkg/rt/...
go test -tags gogen_ir ./pkg/ir ./pkg/rt/...
```

All passed; the first finding requires the targeted rebinding case above.

nnunley:
Both findings addressed in 3ddf7643b.

P1 (`list` intrinsic bypassed `*direct-calls-disabled?*`): `list-intrinsic-call-stmts` now returns nil when the flag is set, so a `(with-redefs [list …] …)` body (and `alter-var-root`/`intern`) falls back to the cached-var trampoline instead of the baked `vm.EmptyList.Cons(…)` chain. Regression: `TestWithRedefsDisablesListIntrinsic`.

Footgun (builtins not yet single-source): completed the stated design instead of just documenting it — `installLangNS` now delegates `not`/`vector`/`cons`/`contains?` to `builtins.{Not,Vector,Cons,Contains}` (arity check stays in the wrapper). One body each, drift impossible.

(The earlier `gogen-diff` red was a transient `golang.org/x/sys` proxy fetch error, not code; the new push re-runs it.)

nooga (approved):

PR #344, opened 2026-06-27
feat(ir): build args once in build-call — eliminate dead duplicate call nodes (#258)

nnunley:

## Problem

`build-call` pre-built the argument list via `build-args` purely for its block-switching side effect, **discarded the result**, then let `build-call-with-head` build the same args a second time and emit the call. Every argument expression containing a non-DCE'd op (every nested `:call`) was therefore built **twice**: one dead copy plus the live copy.

`lower-go` keeps dead `:call` nodes conservatively (calls may be effectful), so the duplicates survived into the generated Go as pairs like:

```go
_, callErr = corefns.Seq(arg0)      // dead duplicate
arg__9, callErr = corefns.Seq(arg0) // live
```

doubling call work and allocations for ~half of all calls in the lowered tree.

## Root cause (minimal repro)

IR for `(count (seq form))` showed two `Call seq` nodes — one dead (`v2`), one live (`v5`) — from a single source `(seq form)`. The `arg__N` names are exactly `build-args`' gensyms, confirming the double build.

## Fix

Build the args **once**, bind each result to an `arg__` gensym local across the head build so any block change the head introduces re-threads them (the symmetric inverse of how `build-call-with-head` threads the head through the arg build), then emit the `:call` directly.

## Effect

- Discarded calls in generated `core.go`: **356 → 14** (the 14 are genuine effect-position calls, e.g. `push-binding!`).
- gogen_ir `IRCompile` allocations: **138,898 → 132,672** (−6,226/op), narrowing the gap to bytecode (129,565) from 9,333 → 3,107.

## Verification

- Minimal IR dump: `(count (seq form))` now builds `(seq form)` once.
- Control-flow-in-args cases (`(+ (if x 1 2) (if x 3 4))`, `(str (if x (seq x) []))`) still thread correctly through join blocks with no cross-block refs — the exact case the pre-build existed for.
- `make check-generated` green; pkg/ir (155), pkg/rt (137), and the clojure suites (bytecode 465 + gogen_ir 465) all pass; bytecode path unaffected (identical allocations).

---
Stacked on #343 (review/merge after it). The diff includes #342 + #343 until those merge.


mparrett:
*Codex Findings (5.5-high)*:

**[P1] Preserve callee-before-argument evaluation**

The new `build-call` implementation builds arguments before the computed callee. This changes observable evaluation order whenever an argument creates control flow.

For example:

```clojure
(defn order-test [x]
  ((do (println "head") identity)
   (if x (do (println "arg") 1) 2)))
```

The lowered Go from this PR executes the `"arg"` print inside the branch before the `"head"` print. The bytecode path evaluates the callee first, so this is a semantic divergence.

The existing `build-call-with-head` helper already preserves a callee across argument CFG joins. Building `fn-id` first and then calling that helper once avoids duplicate argument construction while retaining the required callee-before-arguments order. I tested that direction against the control-flow repro and focused `pkg/ir` call/lowering tests; it emitted `head` before `arg` and preserved CFG threading.

This case should also get a regression test because the current suite and CI are green despite the ordering change.

nnunley:
Fixed by building the head first, then `build-call-with-head` once (4a6f5405f).

`build-call-with-head` already threads the head across argument control-flow joins, so the dead pre-build of args was unnecessary — this keeps the dedup win (args built once) AND restores callee-before-arguments order. Confirmed the args-first variant emitted the argument's effect ahead of the callee's in the lowered Go; `TestBuildCallEvaluatesCalleeFirst` asserts HEADMARK lowers before ARGMARK (verified to fail under the old variant).

nooga (approved):

PR #345, opened 2026-06-27
feat(ir): cross-package direct calls between AOT-lowered packages (#258)

nnunley:
> Part of Epic #258 — Lowering & cross-target codegen. (iterative-development: EPIC-012)

## Summary

A lowered package now calls another lowered package's exports **directly** (`pkg.LG_<go>(ec, …)`) instead of routing through the `rt.CachedVarFn` trampoline. Before this, the only cross-package direct calls were into the hand-written native shims (`corefns`/`builtins`); every call between AOT-lowered `.lg` packages — and the ~2,800 `clojure.core` calls — trampolined, even though the callee was already lowered Go sitting in the sibling package.

Example — `clojure.edn`'s `deep_sort` now emits:
```go
arg__3, callErr = core.LG_keys(ec, arg0)
tr17, te17 = core.LG_vector_QMARK_(ec, arg0)
```

## Mechanism (mostly in the `.lg` lowering layer)

- **`ir.passes.pipeline/lower-all-ns-to-go`** — a whole-program driver. **Phase 0** (`collect-ns-exports`) runs pass-1 lowering of every namespace to discover its direct-callable exports + the cross-namespace fns it references; **phase 1** re-lowers each namespace with the global cross-package registry bound. Two *separated* phases ⇒ emission order is irrelevant and no whole-program fixpoint is needed for the acyclic case.
- **`exported-wrapper-decl`** — emits a thin exported `func LG_<go>(ec, …)` wrapper for each cross-referenced fn (the `LG_` prefix is always exported + valid even for `_dt_*` go-names). Gated so only fns actually called cross-package get a wrapper — no dead code.
- Reuses the existing `emit-typed-direct-call` `:go-pkg` path (qualified call + collision-resistant aliased import), which already anticipated cross-ns lowered siblings.
- `lower_go/internal-ns-sym` made public so the wrapper-target match normalizes both the recorded refs (`clojure.core`) and the lowered ns (`core`) — the alias-remapped namespaces otherwise emit no wrappers.
- `cmd/lgbgen` `runGoTarget` gathers per-namespace specs and makes one `lower-all-ns-to-go` call; per-namespace lowering resilience preserved.

## Soundness

Registry visibility only turns trampolines into direct calls (or the same trampoline on coercion failure) — it never causes a phase-0 emitter to bail in phase 1 — so every confirmed export's wrapper is emitted. `make check-generated`'s compile step is the hard backstop against any dangling cross-package reference (it caught exactly one real bug during development: the alias-remapped-ns wrapper-target mismatch, now fixed).

## Scope

Only `vm.Value`-uniform / coercion-supported callees go direct; typed and variadic callees still trampoline by design (a typed param is direct-called only when the arg is *proven* that type). Cyclic cross-package call graphs are deferred (they'd need a whole-program fixpoint). Broadening direct-call coverage is the concrete-vm-type specialization track (follow-on work).

## Results / verification

- `clojure.core` trampoline census **2802 → 2542**; **394** cross-package direct calls (was 0); 96 exported wrappers.
- `make check-generated` green (compiles the whole lowered tree + bundle lockstep + native dispatch); **double-regen byte-identical**.
- Both clojure suites **465/465** in bytecode AND gogen_ir (semantic parity — direct-vs-trampoline must not change behavior).
- New hermetic tests: `TestCrossPackageDirectCallEmitted`, `TestExportedWrapper{Emitted,Suppressed}` (pkg/ir).

---
Stacked on #344 (review/merge after it). The diff includes #342 + #343 + #344 until those merge.


mparrett:
*Codex Findings (5.5-high)*:

**[P1] Exclude cyclic package edges from the direct-call registry**

`lower-all-ns-to-go` currently merges every phase-0 export into `global-reg`, then enables all matching cross-package calls in phase 1. A two-namespace mutual call therefore generates:

```go
// package cyca
import "example/cycb"
v = cycb.LG_from_b(ec, arg0)

// package cycb
import "example/cyca"
v = cyca.LG_from_a(ec, arg0)
```

Go rejects that output as an import cycle. I reproduced this with two otherwise valid single-arity namespaces passed to `lower-all-ns-to-go`; both returned source successfully, so the per-namespace error handling does not catch it.

The PR describes cyclic call graphs as deferred, but the current implementation actively lowers them into uncompilable packages. Phase 0 should detect strongly connected namespace components and omit at least those cyclic edges from the cross-package registry, leaving those calls on `CachedVarFn` until cycle support exists (or reject the input before emitting files).

**[P2] Do not treat an empty wrapper-target set as “export everything”**

The wrapper predicate currently contains:

```clojure
(or (empty? *wrapper-target-names*)
    (contains? *wrapper-target-names* [ns-int fn-name]))
```

When `lower-all-ns-to-go` processes a program with no cross-package references, `targets` is the empty set, so every direct-callable function receives an `LG_` wrapper. I reproduced a one-namespace program with no external calls and it still emitted `func LG_only_here(...)`.

This contradicts the stated gating (“only fns actually called cross-package”) and adds exported dead API. Use a distinct sentinel for the unit-test/manual “emit all” mode; an explicitly collected empty set should match no functions.

The existing and full `pkg/ir` suites remain green with both repros, so regression coverage is needed for these cases.

nooga (approved):

nnunley:
Both findings fixed in 094a517a9.

**P1 (import cycle):** `lower-all-ns-to-go` now builds the candidate cross-package call graph, takes its transitive closure, and gives each package a registry view that drops any callee able to reach back to it — the cycle-closing edges stay on `CachedVarFn`. Acyclic programs are unaffected (every callee passes the reach test). A 2-cycle conservatively drops both edges (sound). Regression: `TestCrossPackageCyclicEdgesStayTrampolined` (verified to fail without the filter).

**P2 (empty set ≠ export-all):** `*wrapper-target-names*` now uses an explicit `:all` sentinel for the manual/single-ns emit-all mode; `lower-all-ns-to-go` always binds the concrete collected set, so a program with no cross-package refs emits no wrappers. Regression: `TestWholeProgramNoCrossRefEmitsNoWrappers`.

PR #347, opened 2026-06-28
ci(release): isolate slow lowering e2e tests in a dedicated step

mparrett:
## Why

The release workflow runs goreleaser, and its before-hook is `go test -timeout 60s ./...`. Since #299 the lowering-determinism harness compiles the stdlib twice and takes ~60–120s on CI (~200s on a contended laptop), so a `v*` tag push fails the release even though `main` is green. CI only passes because #299 temporarily raised the per-package timeout to 180s in the `go.yml` lanes, and that bump never reached the goreleaser hook. This blocks cutting v1.11.0.

## What

- Pass `-short` in the two fast CI lanes and the goreleaser before-hook, and revert them to a 60s per-package timeout.
- Run the expensive determinism and deftype-skeleton lowering harnesses in a dedicated CI step with a 600s ceiling.
- Keep the AOT diff harness in its existing `gogen-diff` job, which regenerates the lowered tree before testing.
- Drop the TEMPORARY (#299) timeout bumps in `go.yml` and the Makefile.

## Why it's safe

- All three harnesses already self-skip under `testing.Short()`, so `-short` is the existing intended lever, not a new gate.
- Each builds its own `-tags`-selected binary via subprocess, so the outer `go test` build tag doesn't change them. That's why a single untagged CI run of each harness reproduces the coverage the old bytecode and bootstrap fast lanes gave between them — collapsing two runs into one loses nothing.
- The existing `gogen-diff` job continues to gate the AOT diff harness without running it twice.

## Follow-up

Tracked separately, not in this PR. The determinism harness is the slow one, and ~200s warm is heavier than the #299 estimate. It compiles the whole stdlib twice concurrently; caching the first lowered tree or diffing a cheaper artifact would cut local `make test` time. Worth a separate issue rather than holding this fix.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -63,7 +63,7 @@ jobs:
       # test/gold/*.out goldens. Re-deriving those goldens from a LIVE Clojure is
       # done in the dedicated `gold-differential` job (no JVM dependency here).
       # -short skips the expensive lowering harnesses (determinism, AOT diff,
-      # deftype skeleton); they run in the dedicated step below so this lane
+      # deftype skeleton); they run in dedicated CI steps below so this lane
       # stays inside a 60s per-package timeout.
       - name: Test with bundled bytecode stdlib
         run: go test -short -v -timeout 60s ./... -skip TestClojureTestSuite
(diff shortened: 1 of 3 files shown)


mparrett (commented):
No blocking findings.

I verified that `testing.Short()` is used only by the three named lowering harnesses, so the fast lanes and GoReleaser hook are not silently dropping unrelated coverage. The dedicated CI step ran successfully at this commit, and all current checks are green. The subprocess build tags also support the claim that one untagged invocation preserves the intended harness coverage.

Non-blocking: `TestGogenAOTDiff` now runs both in this combined step and in the existing `gogen-diff` job. Keeping both is harmless, but removing it from one location would avoid duplicate work if there is no intentional distinction between testing the committed lowered tree here and the regenerated tree in `make gogen-diff`.

nnunley:
Reviewed the CI isolation — mechanism is sound: the three harnesses self-skip under `-short`, the dedicated 600s step is appropriately sized, and the YAML is well-formed.

One call-out: `TestGogenAOTDiff` now runs **twice** per CI run — once in the new isolated e2e step (via the `-run` regex) and once in the existing `gogen-diff` job (`make gogen-diff`). Either drop it from the new step's regex, or fold the standalone `gogen-diff` job into the isolated step, so the slow test runs in exactly one place.

nnunley (approved):

mparrett:
Addressed in 80bf3e8 — `TestGogenAOTDiff` now runs only in the existing `gogen-diff` job.

nnunley:
The duplicate-run call-out is resolved by `80bf3e8` — `TestGogenAOTDiff` is dropped from the new e2e step's `-run` regex and still runs once via the `gogen-diff` job (`make gogen-diff`). All three slow tests now run exactly once and self-skip under `-short`; timeout hierarchy (60s fast lanes / 600s e2e / Makefile default for gogen-diff) is consistent. Looks good.

PR #348, opened 2026-06-28
feat(aot): lg-compile AOT driver + examples on lower-all-ns-to-go (#258)

nnunley:
Part of Epic #258 — Lowering & cross-target codegen.

Salvages the user-namespace **AOT compile tooling** from the now-superseded #332/#335 "aot" stack, ported onto the cross-package lowering that landed via #343/#344/#345.

## What

- **`scripts/lg-compile`** — a thin CLI over `ir.passes.pipeline/lower-all-ns-to-go`. Given a set of `.lg` files it parses each into a lowering spec, interns every namespace so cross-ns refers resolve, runs the whole-program lowering (direct `pkg.Fn(ec,…)` calls + imports + exported wrappers between AOT-lowered packages), and writes one Go package per namespace.

  ```
  LG_SOURCE_PATHS=pkg/rt/core:pkg/rt/gogen \
    ./lg scripts/lg-compile <out-dir> <import-prefix> <file.lg>...
  ```

- **`gogen/ns->go-pkg` + `gogen/ns->go-reldir`** — the canonical namespace→Go package/dir manglers, so `cmd/lgbgen` and `lg-compile` name packages identically. Not used by core lowering, so `core_compiled.lgb` is unchanged (genmanifest staleness test stays green).

- **`examples/aot/`** — `cross-package/` (minimal `lib`+`app` demo), `ys-on-let-go/` (YS shim + programs), `yamlstar/` (lowerability diagnostics), and a README. Generated Go (`examples/aot/*/go/`) stays gitignored.

## Verification

`lg-compile` on the cross-package example emits `lib.LG_greet(ec, …)` as a **direct cross-package call** from `app`, and `go build ./examples/aot/cross-package/go/...` compiles clean.

Supersedes #332 and #335 (their cross-package registry / build-args / core-lowering changes landed via #343/#344/#345; this keeps only their genuinely-new tooling).

### DIFF
--- a/examples/aot/README.md
+++ b/examples/aot/README.md
@@ -1,0 +1,25 @@
+# AOT compilation examples
+
+Trials for compiling let-go `.lg`/`.clj(c)` namespaces **ahead-of-time to native
+Go**, via `scripts/lg-compile` (the multi-namespace AOT driver with direct
+cross-package calls).
+
+Driver:
+
+```
+./lg scripts/lg-compile <out-dir> <import-prefix> <file.lg>...
+```
+
+It lowers each single-arity `defn`/`defmulti`/`defmethod` to a native Go func,
+emits one Go package per namespace (canonical `gogen/ns->go-pkg` naming: leaf
+segment + nested dir), and wires direct `pkg.Fn(ec, …)` calls between the
+lowered packages. Variadic / non-coercible fns stay on runtime trampolines.
+
+| Directory | What it is |
+|---|---|
+| `cross-package/` | Minimal two-namespace demo (`lib` + `app`) exercising a direct cross-package call. The smallest reproducible AOT example. |
+| `ys-on-let-go/`  | The YS-on-let-go shim: a small YAMLScript-style stdlib plus example programs. See its own `README.md`. |
+| `yamlstar/`      | Diagnostics for AOT-compiling the real YamlStar Clojure pipeline (`ys-coverage.lg` — per-file lowerability tally; `ys-lower-report.lg` — per-fn lower/fallback report). |
+
+Generated Go output (`*/go/`, `*/go_lowered/`) is git-ignored — regenerate with
+`lg-compile`.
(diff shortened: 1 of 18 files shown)


mparrett:
Review findings, scoped to changes introduced by this PR:

1. **[P1] Make multi-namespace compilation independent of input order.** In `scripts/lg-compile:82-84`, forms are evaluated one input file at a time. With the included example and the documented source paths, passing `app.lg lib.lg` (the natural order from `src/*.lg`) fails while evaluating `app`'s `ns` form because `lib` has not been loaded yet; `lib.lg app.lg` succeeds. Since the CLI accepts a set of files and lowers them as a whole program, it should discover/load dependencies before evaluating namespace forms rather than requiring callers to topologically sort inputs.

2. **[P1] Avoid executing top-level user code during compilation.** The same loop evaluates every source form, including arbitrary top-level expressions. For example, compiling the newly included `01_hello.lg` prints `hello from let-go via yamlscript` during compilation by executing `(apply main ARGS)`, while that invocation is absent from the generated Go (only the lowered function/override remains). This gives compilation arbitrary runtime side effects without preserving those forms in the output.

I reproduced both against `57c7dda`. The documented Go build, targeted `pkg/ir` and `pkg/genmanifest` tests, and current CI otherwise pass.

nnunley:
Both fixed in 19206f3e3.

1. **Input-order independence:** `lg-compile` now loads inputs in dependency order — it parses each file's `ns`-form `:require`/`:use` and top-level `(require …)`/`(use …)` clauses, builds the sibling-dependency graph, and DFS-topologically orders the load (marking-before-recursing, so a `require` cycle terminates rather than loops). `app.lg lib.lg` and `lib.lg app.lg` now both succeed.

2. **No top-level user code at compile time:** only *definitional* forms (`ns`/`require`/`use`/`def`/`defn`/`defmulti`/`defmethod`/`declare`/…) are evaluated during loading; bare top-level expressions like `(apply main ARGS)` are skipped (they were never part of the lowered Go output). Compiling `01_hello.lg` no longer prints during compilation.

Verified both orders emit `lib.LG_greet(ec, …)` and `go build ./…/go/...` passes.

nooga (approved):

mparrett:
LGTM

PR #351, opened 2026-06-28
docs: use tap install command for Homebrew

nooga:
Updates the README and GitHub Pages install card to use the primary Homebrew command:

```bash
brew install nooga/tap/let-go
```


mparrett (approved):
Thanks!

PR #356, opened 2026-06-29
perf(startup): recover cold-start regression (16ms → 5.1ms)

nnunley:
## Summary
Recovers the cold-start regression tracked in #353. Two parts:

1. **Unbundle the `ir.*` AOT pipeline** from the encoded `core_compiled.lgb`. The IR/lowering pipeline (~10k lines) was compiled into the bundle and eagerly decoded on every process start, even for `lg -e nil`. `isIRBundleSkipped` (cmd/lgbgen) excludes `ir.*` from the bytecode bundle only — it stays generated for `-tags gogen_ir` (`core_go_lowered/ir/**`) and loadable from embedded source at runtime via `(require 'ir.passes.pipeline)`. Bundle: **711,268 → ~233,697 bytes**.
2. **Rewrite the startup bundle decode**: materialize bundle bytes directly into live VM structures (no temp `ChunkData`/const intermediates); pre-size `CodeChunk`, `SourceMap`, local-var tables, and the const pool; decode the string table with a single-allocation `unsafe.String`. Pause GC for the short ~5MB transient-garbage decode burst. Adds an env-gated `LG_BOOT_TIMING` profiler.

`ReserveLocalVars`/`ReserveSourceMap` preserve existing entries (capacity-only contract), with regression coverage in `pkg/vm/decode_prealloc_test.go` (addresses @mparrett's review).

## Measured impact (`lg -e nil`)
| metric | upstream/main | this PR |
|---|---:|---:|
| startup | ~16ms | **5.1ms ± 0.3ms** |
| `BenchmarkInitFromLGB` | 7.05ms (33,944 allocs) | 1.55ms (8,930 allocs) |
| `BenchmarkDecodeCore` | 4.02ms (27,825 allocs) | 1.30ms (9,935 allocs) |
| `core_compiled.lgb` | 711,268 B | 233,697 B |

## Scope note
This PR was split to keep it single-concern. Diagnostics moved to #360 (stacked on this); transient-map lowering moved to #361 (independent, lands after this). Together they cover the previous combined head — nothing dropped.

## Verification
- `go test ./...` (default + `-tags gogen_ir`)
- `make check-generated`
- `go build -tags gogen_ir .`
- runtime `(require 'ir.passes.pipeline)` + `*ir-compile*` still optimizes (constant-folds `(+ 1 2)` → one `LOAD_CONST`)


mparrett:
I found one issue in the new preallocation API:

**`ReserveLocalVars` discards existing entries** (`pkg/vm/vm.go:154-158`). When the requested capacity exceeds the current capacity, replacing the slice with `make([]LocalVar, 0, n)` resets its length to zero. That contradicts the method's capacity-only contract and can silently remove debug metadata if it is called after `AddLocalVar`.

Please preserve the existing values, as `ReserveSourceMap` does, and add a regression test that reserves after adding an entry.

Otherwise the change looks sound. Validation completed successfully:

- `go test ./...`
- `make check-generated`
- default build plus `test/gold-aot/ir_pipeline.lg` (output: `12`)

GitHub reported no CI statuses on the reviewed head commit.

nnunley:
This branch is back in the acceptable range using the same reporting frame as PR #353: real `lg -e nil` startup, stripped binary size, and startup RSS.

On my machine:

| metric | PR #353 report | this branch |
|---|---:|---:|
| stripped binary | `13M` | `13.0M` (`13,007,762` bytes) |
| startup (`lg -e nil`) | `15.3ms ± 0.4ms` | `5.1ms ± 0.3ms` |
| startup RSS | `22.4MB` | `~15.1MB` |

Measured with the same style as the benchmark harness:

```bash
go build -ldflags='-s -w' -o /private/tmp/letgo-pr353 .
hyperfine --warmup 3 --runs 10 --export-json /private/tmp/pr353-startup.json \
  -n let-go '/private/tmp/letgo-pr353 -e nil'
```

The `~15.1MB` RSS comes from `hyperfine`'s recorded `memory_usage_byte` samples in the exported JSON.

Supporting low-level numbers versus `upstream/main` from sequential `-count=5` runs:

- `BenchmarkInitFromLGB`: `7.05ms -> 1.55ms`
- `BenchmarkInitFromLGB` allocs/op: `33,944 -> 8,930`
- `BenchmarkInitFromLGB` B/op: `6,817,751 -> 1,015,573`
- `BenchmarkDecodeCore`: `4.02ms -> 1.30ms`
- `BenchmarkDecodeCore` allocs/op: `27,825 -> 9,935`
- `BenchmarkDecodeCore` B/op: `5,619,211 -> 1,409,399`
- `pkg/rt/core_compiled.lgb`: `711,268 -> 233,697` bytes

So the branch both recovers the user-visible startup regression and removes most of the decode-path overhead that had accumulated on `main`.

mparrett:
Rechecked the force-pushed head `dcb2e21fa87a01d5bd5d712daeebe47fba587231`.

Blocking finding remains:

`ReserveLocalVars` still drops existing debug entries when it has to grow (`pkg/vm/vm.go:154-158`):

```go
c.localVars = make([]LocalVar, 0, n)
```

That resets the slice length to zero instead of preserving entries already added via `AddLocalVar`, so a later reserve can silently erase local-variable metadata. The new test only covers reserving an empty table (`pkg/vm/decode_prealloc_test.go:25-34`), so it does not catch the destructive grow case.

Please preserve existing entries, matching the previous implementation / the `ReserveSourceMap` pattern, and add a regression test that adds at least one local var before reserving a larger capacity.

Validation on this head:

- `go test ./...`
- `make check-generated`
- `go build -o /private/tmp/lg-pr356-recheck .`
- `/private/tmp/lg-pr356-recheck test/gold-aot/ir_pipeline.lg` -> `12`
- `BenchmarkInitFromLGB` and `BenchmarkDecodeCore` smoke runs completed successfully

nooga (approved):

mparrett:
Just came here to say LGTM! Glad you merged it.

nnunley:
### Scoped this PR down to startup-only (split out diagnostics + transient-map work)

Heads up on why the head moved from `c8abaacc` to `cb2516ac`: while iterating, this branch had accumulated two unrelated bodies of work on top of the startup recovery. To keep this PR single-concern and reviewable, I split them into their own PRs. **No work was dropped** — the three together cover exactly what `c8abaacc` had:

- **This PR (#356)** — startup cold-start recovery only: exclude the `ir.*` AOT pipeline from the encoded `core_compiled.lgb`, rewrite the bundle decode (materialize directly into live VM structures, pre-size `CodeChunk`/`SourceMap`/local-var/const pool, single-alloc `unsafe.String` string table), and pause GC during the boot decode. Startup `lg -e nil`: **~16ms → 5.1ms ± 0.3ms**.
- **#360** — the env-gated diagnostics (`LG_DECODE_TAG_STATS`, `LG_LOOKUP_STATS`), stacked on this PR.
- **#361** — transient-map IR lowering + determinism-harness gating, independent (lands after this one; both regenerate the binary bundle).

@mparrett — your `ReserveLocalVars` finding is addressed **in this PR** at the new head `cb2516ac`: it now preserves existing entries (capacity-only contract, matching `ReserveSourceMap`), with the add-then-grow regression test in `pkg/vm/decode_prealloc_test.go` (`TestCodeChunkReserveLocalVarsPreservesExistingEntries`). Re-review of the slimmed PR welcome when you have a moment.

PR #359, opened 2026-06-29
examples: host-eval browser demo (client-owned shell)

mparrett:
## What

A worked example of the `-w-host-eval` bundle feature (#340): a plain `lg -w -w-shell none -w-host-eval` build where the embedding page owns the shell and binds the runtime only through the public `window.LetGoHost` surface.

It complements the hosted REPL under `wasm/`, which is a bespoke Go-WASM build with its own `window.Eval`. This is the `lg -w` path — what you'd use to embed a let-go image in your own page.

## What it shows

Both host seams, together:

- `LetGoHost.eval(code)` runs a form in the live image and resolves to its value
  — the REPL pane shows returned values.
- `*out*` routes through the `HostWriter` to `onOutput` — the stdout pane shows
  what the image prints.

The `time` and `fizzbuzz` chips exercise both at once: a return value in the
REPL pane and printed lines in the stdout pane.

## Files

- `main.lg` — minimal program; prints a banner, then the image stays live for eval.
- `shell.html` — the client shell: an inline REPL plus a read-only stdout pane,
  bound only via `window.LetGoHost`.
- `build.sh` — builds the bundle and grafts the shell into one self-contained page.

## Build

```
LG=lg ./build.sh
```

Serve with cross-origin isolation (`COOP: same-origin`, `COEP: require-corp`) — the wasm input ring is `SharedArrayBuffer`-backed, so the page must be `crossOriginIsolated`, or boot fails on input.

image

### DIFF
--- a/examples/host-eval/shell.html
+++ b/examples/host-eval/shell.html
@@ -82,10 +82,10 @@
     .box { height: 240px; }
   }
   /* iOS Safari auto-zooms a focused field under 16px. Target touch devices
-     directly — a width query (max-width) misses iOS's wider CSS viewport, which
-     is why the earlier attempt didn't take. Bump the whole REPL to 16px so the
-     input never trips the trigger and input + history stay matched. (#expr
-     inherits via font:inherit.) Pairs with the maximum-scale viewport patch. */
+     directly — a width query (max-width) misses iOS's wider CSS viewport. Bump
+     the whole REPL to 16px so the input never trips the trigger and input +
+     history stay matched (#expr inherits via font:inherit). This alone prevents
+     the zoom — we leave the page's pinch-zoom untouched. */
   @media (pointer: coarse) {
     #log { font-size: 16px; }
   }
@@ -128,16 +128,6 @@
   </div>
 </div>
 <script>
-  // iOS Safari auto-zooms when focusing a sub-16px field. maximum-scale=1 stops
-  // that automatic zoom; since iOS 10 Safari ignores the cap for user-initiated
-  // pinch, so manual zoom still works (no accessibility regression). The bundle
-  // owns the <head> viewport meta (#291), so patch it here rather than the head.
-  (function () {
-    var vp = document.querySelector('meta[name="viewport"]');
-    if (!vp) { vp = document.createElement("meta"); vp.name = "viewport"; document.head.appendChild(vp); }
-    vp.setAttribute("content", "width=device-width, initial-scale=1, maximum-scale=1");
-  })();
-
   const term = new Terminal({
     fontFamily: '"JetBrains Mono", ui-monospace, "SF Mono", Menlo, monospace', fontSize: 13,
     // stdout is read-only output, not an input: disable stdin and hide the
--- a/examples/host-eval/shell.html
+++ b/examples/host-eval/shell.html
@@ -110,6 +110,15 @@
       <div id="evalio" class="box"></div>
     </section>
   </div>
+  <template id="line-template">
+    <div data-ref="row" data-text="text"></div>
+  </template>
+  <template id="echo-template">
+    <div class="echo" data-ref="row">
+      <span class="ps">user=&gt;</span>
+      <span class="src" data-text="src"></span>
+    </div>
+  </template>
   <div class="examples" id="examples">
     <div class="exrow">
       <span class="label">value</span>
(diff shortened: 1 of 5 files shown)


examples/host-eval/shell.html
```diff
@@ -0,0 +1,231 @@
...
+
+  // Insert a transcript line just above the live prompt (which stays last).
+  function emit(text, cls) {
+    const d = document.createElement("div");
+    if (cls) d.className = cls;
+    d.textContent = text;
+    logEl.insertBefore(d, promptline);
+  }
+  // Freeze the submitted input as a history line: "user=> <expr>".
+  function echo(src) {
```
nnunley, line 188: Can you just register some html templates for these so you don't have to do as much source manipulation?

nnunley (approved):
LGTM. There's some followup clean-up possibilities, but those are non-blocking.

examples/host-eval/shell.html
mparrett: Thanks for the review. Are you thinking something along these lines?

```
<template id="echo-template">
  <div class="echo"><span class="ps">user=&gt;</span><span class="src"></span></div>
</template>
```

```
function echo(src) {
  const d = document.getElementById("echo-template").content.firstElementChild.cloneNode(true);
  d.querySelector(".src").textContent = src;
  logEl.insertBefore(d, promptline);
}
```

nnunley:
Follow-up idea I’m considering building on top of this: keep `window.LetGoHost` as the public seam, but grow it from a thin `eval(code)` bridge into a more general request/response protocol for browser tooling.

Concretely, something like `LetGoHost.request({op, session, ns, code, options}) -> Promise<{ok, value, error, warnings, output, metrics, artifacts}>`, with `eval()` as a convenience wrapper.

That would let a client-owned page ask for structured compiler products instead of scraping printed output: `eval`, `compile`, `disassemble`, `ir-build`, `ir-optimize`, `lower-bytecode`, `lower-go`, later `describe` / `complete` / docs, etc.

The Godbolt-style explorer could then just be one consumer of that protocol: forms in through the real REPL compile path, then panes for bytecode disasm, IR, optimized bytecode, lowered Go, timings, warnings, and so on.

So I’m thinking of this PR as establishing the right embed seam, and a follow-up would generalize that seam rather than baking a special-purpose explorer API into the runtime.

examples/host-eval/shell.html
nnunley: Yes. You should be able to also bind inputs and outputs, so it ends up just setting a property on the template.

mparrett:
> Follow-up idea I’m considering building on top of this: keep `window.LetGoHost` as the public seam, but grow it from a thin `eval(code)` bridge into a more general request/response protocol for browser tooling.
> 
> Concretely, something like `LetGoHost.request({op, session, ns, code, options}) -> Promise<{ok, value, error, warnings, output, metrics, artifacts}>`, with `eval()` as a convenience wrapper.
> 
> That would let a client-owned page ask for structured compiler products instead of scraping printed output: `eval`, `compile`, `disassemble`, `ir-build`, `ir-optimize`, `lower-bytecode`, `lower-go`, later `describe` / `complete` / docs, etc.
> 
> The Godbolt-style explorer could then just be one consumer of that protocol: forms in through the real REPL compile path, then panes for bytecode disasm, IR, optimized bytecode, lowered Go, timings, warnings, and so on.
> 
> So I’m thinking of this PR as establishing the right embed seam, and a follow-up would generalize that seam rather than baking a special-purpose explorer API into the runtime.

Great idea. I was thinking of something similar (and maybe complementary) the other day — more of an async remote debugging model. I love the "compiler as a service" angle, though. I think what you proposed here would be powerful!

examples/host-eval/shell.html
mparrett: Addressed in https://github.com/nooga/let-go/pull/359/changes/b0cc99f3aefd92a8db41c8e0cbf995c74f184be9 — I realize there are a few ways to go about this. I think this is a decent pattern to start and materially better than the previous version.

mparrett:
@nooga Nothing earth shattering here, but it does demonstrate the new host capabilities. Let me know if you have feedback!

nooga:
Super useful 👌  merging it!

PR #360, opened 2026-06-29
perf(diagnostics): env-gated startup decode + lookup stats

nnunley:
Splits the diagnostics out of #356 into their own PR.

Two env-gated diagnostics used to investigate the startup-decode cost. Both default **off** with zero overhead unless the env var is set:

- `LG_DECODE_TAG_STATS` — per-tag / string / var-ref counts during bundle decode (`pkg/bytecode/decode_stats.go`).
- `LG_LOOKUP_STATS` — process-lifetime symbol/namespace lookup stats, emitted at process end (`pkg/vm/lookup_stats.go`, `lg.go` `emitRuntimeStats`).

Finding from these: runtime symbol lookup was **not** the startup regression — the cost was chunk/debug materialization (handled in the parent startup change).

**Stacked on #356** (`perf/unbundle-ir-startup`). The diff shows the startup commit too until #356 merges; review only the diagnostics commit. Purely additive (+448 / -0).


mparrett:
Grain of salt here; some uncertainty. I realize you might be in the process of squashing or rebasing, and I'm not sure if this is accurate given the disclaimer below about not reviewing the stacked startup commit. Sharing this in case it's helpful, not intended to be a blocker.

---

**Findings**

[P2] Default-off lookup stats still add hot-path locking 
[lookup_stats.go](/private/tmp/let-go-pr360-review/pkg/vm/lookup_stats.go:100), [namespace.go](/private/tmp/let-go-pr360-review/pkg/vm/namespace.go:271), [symbol.go](/private/tmp/let-go-pr360-review/pkg/vm/symbol.go:53)

`noteLookup` and `noteNamespaced` always take `lookupStatsMu`, even when `LG_LOOKUP_STATS` is unset. These are now called from `Namespace.Lookup` and `Symbol.Namespaced`, so the default production path pays a mutex lock/unlock on every lookup / namespace split. That contradicts the PR’s “default off with zero overhead” goal and could regress the exact startup/runtime paths this diagnostic is investigating. I’d make the enabled flag a cheap `atomic.Bool` fast path before the mutex/map work.

**Verification**

I reviewed only the diagnostics commit `264340350d834ac15265ff13b0a745dbd392317d`, not the stacked startup commit from `#356`.

Ran focused tests:

```sh
go test ./pkg/bytecode ./pkg/vm
```

Both passed. No pushes or repo mutations beyond the temporary review checkout in `/private/tmp/let-go-pr360-review`.

nnunley:
Superseded: #356 was merged as the combined (pre-split) change, so this PR's content (diagnostics / transient-map work) is already in `main`. Closing as redundant. The remaining unmerged piece — the ratchet refresh, now expanded to a machine-profile-partitioned + machine-independent-alloc-gated baseline — lives in #362.

PR #361, opened 2026-06-29
feat(ir): transient-map lowering + gate slow determinism harness

nnunley:
Splits the transient-map work out of #356 into its own PR.

- Support `TransientMap` through the IR lowering pipeline (`passes/typeinfer`, `lower_go`, `lattice`) and runtime type dispatch (`lang.go` case, `core.lg` type check), with a `reduce-kv` regression test.
- Gate the slow lowering-determinism e2e harness behind an opt-in.
- Add persistent/transient map microbenchmarks.

**Independent of #356**, but both regenerate the binary `core_compiled.lgb`, so they will conflict at merge. Recommend landing **after #356**, then rebasing this branch to pick up the ir-excluded bundle (a `make generate` after rebase).


nnunley:
Superseded: #356 was merged as the combined (pre-split) change, so this PR's content (diagnostics / transient-map work) is already in `main`. Closing as redundant. The remaining unmerged piece — the ratchet refresh, now expanded to a machine-profile-partitioned + machine-independent-alloc-gated baseline — lives in #362.

PR #362, opened 2026-06-29
perf(ratchet): machine-profile baseline + machine-independent alloc gate + interactive multi-machine dashboard

nnunley:
Makes the benchmark ratchet correct for our **mixed hardware** (Apple M1/M2/M3 — all arm64 but different microarchitectures — plus amd64 CI), and rebuilds the "Are we fast yet?" dashboard to actually visualize it. Two ratchet failure modes it fixes: timing numbers from different CPUs were being MIN-merged/compared through one anchor, and a representation-keying mismatch left stale duplicate entries.

### Multi-machine baseline (schema v2)
`docs/perf/baseline.json` is now partitioned by **machine profile** — the raw `arch/cpu_model` key (e.g. `arm64/Apple M3`), so M1/M2/M3/amd64 each get their own sub-baseline with its own anchor and full `Machine` fingerprint (provenance). **Timing** (`ns/op`, `ratio_to_anchor`) is paired/merged **only within a profile**, so an M1 run never tightens an M3 bar. v1 baselines migrate on read.

### Machine-independent deterministic gate
`allocs/op` and `bytes/op` are deterministic — they come from the **arch-independent generated code** (lgbgen `.lgb` / lowered-Go are identical bytes on any arch), so they're gated against the **global-min across all profiles** (2% budget). This catches an allocation regression on any machine, *including one with no timing profile yet* — closing a blind spot in pure per-machine partitioning.

### Representation alignment
Every benchmark carries a uniform `[representation]` label (`bytecode`/`ir_bytecode`/`gogen_ir`/`aot_native`); legacy mixed-convention keys migrate via an explicit alias table. The obsolete suite `[gogen_ir]` entry is **dropped** (it measured the pre-native-dispatch tautology — a different, lower-alloc workload that would have wrongly tightened the deterministic bar).

### Interactive multi-machine dashboard (`perf-page`)
The "Are we fast yet?" page is rebuilt on a single declarative **d3 / Observable Plot** stack (both pinned + SRI; replaces the old hand-rolled SVG, after a Vega-Lite detour that couldn't render the dense per-row views). Every view reads one generated `explorer.json` (timeline rows including retained per-run samples, deduped where deterministic):

- **Explore over time** (Observable Plot) — pick a benchmark + metric; one line per CPU model with its per-run min/max spread band and **every individual gathered sample as a dot**. Compact SI axis, CPU legend, units + direction labels.
- **Trend sparklines** (canvas) — one row per benchmark × CPU, each a real trend sparkline scaled to its own range (so tiny and huge benchmarks are both legible), with first/last endpoint rings and the endpoint Δ. Canvas, not SVG/Plot: the dense grid is tens of thousands of marks and needs an independent y-scale per row (which faceted Plot can't give).
  - **Scale families merged** — variants that differ only by a scaling factor (`…/10`, `/100`, `/1000`) collapse into one row; their lines are overlaid and each indexed to its own first value (% change from a shared 0% baseline) so you compare how each scale moved, with per-scale Δ pills.
  - **Hover** — crosshair + tooltip with the actual value and % change at the snapshot under the cursor (per scale for grouped rows; value + sample spread for single rows).
  - **Filters** — search by name, hide single-point series, hide unchanged (Δ≈0), with a live count.

### Notes
- Metrics are mixed-architecture; the dashboard compares trends **per CPU** (the anchor ratio only normalizes within one CPU, not across).
- The dashboard immediately surfaced a real, separate regression worth investigating (tracked as a follow-up, **not** part of this PR): the `BenchmarkVarDeref*` family (`…BoundParallel`/`RootParallel`/`DistinctParallel`/`Bound`/`Root`) blew up ~5–50× in `ratio_to_anchor` across the timeline on the amd64 tiers — e.g. `VarDerefBoundParallel` 0.25 → 13.3 (+5184%).
- Verified: `go build ./...`, `go test ./cmd/bench-ratchet ./cmd/perf-page`, `make check-generated`, the ratchet gate (timing 0 regressions + deterministic OK), and the rendered page in Chrome against the live 105-snapshot perf-data timeline (4 CPU tiers).


nnunley:
Added the **per-CPU-over-time dashboard** to this PR (folds the multi-machine work through to the page): `perf-page` now emits the whole timeline as a tidy table and renders an interactive **Vega-Lite** explorer (loaded via CDN) — pick any benchmark + metric (`ratio_to_anchor` / `allocs_per_op` / `bytes_per_op`), one line per CPU model, click the legend to isolate a CPU, hover for values. This replaces the old single-mixed-series timeline (which zig-zagged across the 4 amd64 CI CPUs). Verified by rendering against the real `perf-data` timeline (104 snapshots, ~28k tidy rows). The curated SVG charts stay; `index.html` is still deploy-generated and not committed here.

nooga (approved):

PR #363, opened 2026-06-29
ci(perf): arm64 timeline leg + auto-backfill of release tags

nnunley:
Splits the perf **CI** work out of #362 (which is the ratchet + dashboard *code*).

### arm64 timeline leg (`perf-timeline.yml`)
Every perf workflow currently runs only on `ubuntu-latest` (amd64), so the timeline has **no arm numbers** — developers' Apple-Silicon runs never reach `perf-data`. This adds a `macos-14` (Apple M1) leg via a **serialized** matrix (`max-parallel: 1`, since both legs push the single `perf-data` branch). Snapshot filenames gain an **arch suffix** so amd64/arm64 snapshots of the same commit don't collide, and the timestamp is computed **portably** (`date -r` on BSD/macOS vs `date -d @` on GNU). With the machine-profile partitioning from #362, the two coexist and the explorer gains an Apple-M line.

### Release auto-backfill (`perf-backfill.yml`, new)
Weekly + dispatchable. Scans tags, keeps only those that carry the anchor harness (**floors at v1.8 / #119**; older releases — including the by-date-older `v2.0.x` line — are skipped), finds which lack a `perf-data` snapshot, and dispatches `perf-timeline` for each, so new releases self-backfill.

**Needs a secret:** `PERF_DISPATCH_TOKEN` (fine-grained PAT, `actions:write`) — the automatic `GITHUB_TOKEN` can't trigger `workflow_dispatch` (recursion guard), so without the PAT the dispatch step enqueues nothing.

**Out of scope:** carrying the harness *backwards* onto pre-v1.8 tags (overlay today's bench files on the old tree) — feasible but compile-compat-risky, would need its own job.

> Note: depends conceptually on #362's multi-machine ratchet (so arm + amd64 don't contaminate each other). Can merge independently, but the arm numbers only render meaningfully once #362 lands.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -17,10 +17,15 @@ jobs:
       - uses: actions/setup-go@v5
         with:
           go-version-file: go.mod
-      - name: Install golangci-lint
-        run: go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@latest
+      # Use the official action: it installs a cached, prebuilt golangci-lint
+      # binary from GitHub Releases instead of `go install`-ing it through
+      # proxy.golang.org, which intermittently drops the HTTP/2 stream
+      # mid-download and fails the job (INTERNAL_ERROR received from peer).
       - name: Run golangci-lint
-        run: golangci-lint run --timeout=5m
+        uses: golangci/golangci-lint-action@v9
+        with:
+          version: v2.12
+          args: --timeout=5m
 
   # Enforce the "no test files at the repo root" guard. The pre-commit hook of
   # the same name is the local equivalent, but jj does not run git hooks, so CI
--- a/.github/workflows/perf-backfill.yml
+++ b/.github/workflows/perf-backfill.yml
@@ -1,0 +1,90 @@
+name: Perf Backfill Releases
+
+# Backfills the perf timeline for tagged releases that don't yet have a
+# snapshot on the perf-data branch. Floors naturally at the release where the
+# benchmark harness landed (#119 / v1.8): a tag is only eligible if it carries
+# pkg/vm/bench_ratchet_anchor_test.go, so older releases (which predate the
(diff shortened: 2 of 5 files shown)


mparrett:
@nnunley Can/should this PR also add an M2 larger-runner leg, or should that be separate? If we add it here, I think the snapshot naming needs one more dimension than `uname -m`, since M1 and M2 are both `arm64` and would otherwise collide for the same commit.

nooga (dismissed):

nnunley:
Let's keep the M2 leg separate — scoping this PR to the arm64 timeline + tag backfill. You're right that `uname -m` collides M1/M2; the M2 PR should extend the snapshot key with a micro-arch dimension (append a CPU-model token) rather than `uname -m` alone. Happy to review when you open it.

nnunley:
Changed my mind — addressed this here in #363 rather than deferring, so the arm leg lands collision-safe from day one (`4ce1590`).

bench-ratchet now has a `machine-key` subcommand that prints a filesystem-safe slug of `perfdata.MachineKey` (the exact `<arch>/<CPUModel>` key the timeline already partitions snapshots by). Snapshots are named `<stamp>-<sha>-<machine-key>.json`, so M1 (`arm64-apple-m1`) and M2 (`arm64-apple-m2`) get distinct filenames and can't overwrite each other for a commit — and the filename dimension now tracks the content partition exactly. Bonus: it uses `runtime.GOARCH`, fixing a latent mismatch where the filename said `x86_64` (uname -m) while the snapshot content says `amd64` (two different amd64 runner SKUs would also have collided). Backfill of pre-existing tags falls back to a shell slug of the same shape, since their bench-ratchet predates the subcommand.

So when you open the M2 leg, it's purely additive — no snapshot rename/migration needed.

nooga (approved):

nooga (approved):

Issue #312, opened 2026-06-22
Investigate stamping version/commit metadata for go install @sha builds


mparrett:
Context from nooga/xsofy#80: xsofy may want to use `let-go.semver/require-letgo` to assert that it is running on a specific let-go commit until the required WASM glue lands in a tagged release.

The caveat is that a binary installed with `go install github.com/nooga/let-go@<sha>` appears to report `lg dev` and expose commit `none`, because `main.version` / `main.commit` are only stamped by release ldflags. If that is true, a SHA-shaped `require-letgo` guard cannot enforce under common CI installs: it sees an unknown commit and warn-and-passes instead of failing on the wrong build.

Things to investigate:

- Confirm the behavior for `go install github.com/nooga/let-go@<sha>` and `go install github.com/nooga/let-go@<tag>`.
- Decide whether non-release installs should expose enough metadata for `require-letgo` SHA checks to work, or whether docs should explicitly say SHA checks require a stamped binary.
- Consider a supported CI pattern for pinned source builds, e.g. build from a checked-out commit with `-ldflags "-X main.version=dev -X main.commit=<sha>"`.
- Consider whether `require-letgo` should phrase the skip warning more pointedly when a SHA pin is requested but commit metadata is `none`.

This matters because `require-letgo` supports both semver ranges and bare commit SHAs, but the SHA path is only useful when the running binary actually knows its commit.

mparrett:
Investigation notes:

I reproduced the issue with isolated installs under `/private/tmp/letgo-issue312` using Go `1.26.3 darwin/arm64`.

Confirmed behavior:

- `go install github.com/nooga/let-go@v1.11.1` builds a binary that reports `lg dev`.
- `go install github.com/nooga/let-go@25ccf5505588cfd6fd0c1ed7e870446ae1ec4a2a` also builds a binary that reports `lg dev`.
- `go version -m` does contain useful module metadata:
  - tag install: `mod github.com/nooga/let-go v1.11.1`
  - untagged SHA install: `mod github.com/nooga/let-go v1.10.1-0.20260628062305-25ccf5505588`
- But `lg` does not read Go build info. It only propagates `main.version` / `main.commit` into `rt.Version` / `rt.Commit`, and those remain the defaults `dev` / `none` unless release/build ldflags stamp them.

Practical effect confirmed:

```clojure
(require-letgo "25ccf55")  ;; warns and passes
(require-letgo "deadbee")  ;; also warns and passes
```

[...]

PR #366, opened 2026-06-30
[codex] Derive lg metadata from Go build info

mparrett:
## Summary

This adds a best-effort metadata fallback for `lg` binaries that are built without release ldflags. At startup, `lg` now keeps stamped `main.version` / `main.commit` when present, otherwise reads Go build info to populate missing metadata.

The fallback covers:

- `go install github.com/nooga/let-go@<tag>`: derive the runtime version from `debug.ReadBuildInfo().Main.Version`.
- source checkout builds: derive the runtime commit from `vcs.revision`.
- `go install github.com/nooga/let-go@<untagged-sha>`: derive a 12-character commit prefix from the module pseudo-version suffix.

Dirty source builds intentionally keep version as `dev`, while still exposing the base VCS revision, so local modified builds do not masquerade as a clean release.

## Impact if left as-is

- `require-letgo` SHA pins remain non-enforcing for common CI installs using `go install ...@<sha>`.
- A wrong pinned binary can warn-and-pass when commit metadata is `none`, so source-level compatibility checks can give false confidence.
- Tagged `go install` binaries report `lg dev`, so semver-based `require-letgo` guards also skip instead of enforcing.
- Consumers need a documented custom build step with `-ldflags` to make SHA pins meaningful.

## Impact of this fix

- Tagged `go install` binaries can satisfy semver guards without release artifacts.
- Untagged `go install @sha` binaries can satisfy 7-12 character SHA-prefix guards when Go exposes the pseudo-version.
- Local checkout builds expose `vcs.revision`, so SHA guards work for checked-out builds even without manual ldflags.
- Release builds remain unchanged because ldflags still take precedence.

Full 40-character SHA enforcement for module-proxy `go install @sha` is still not guaranteed unless Go embeds `vcs.revision`; in my repro it did not. For that case, a stamped/source-checkout build is still the exact solution.

## Validation

- `go test .`
- Built patched `lg` locally and verified:
  - `lg -version` reports `lg dev (79b96e5)` for a dirty source checkout.
  - `require-letgo "79b96e5"` returns `nil`.
  - `require-letgo "deadbee"` hard-fails with the running commit.

Addresses #312.

### DIFF
--- a/lg.go
+++ b/lg.go
@@ -13,6 +13,7 @@ import (
 	"maps"
 	"os"
 	"path/filepath"
+	runtimeDebug "runtime/debug"
 	"strings"
 
 	"github.com/nooga/let-go/pkg/bundle"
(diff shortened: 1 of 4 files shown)


nnunley (approved):
You have a go linter error that you should fix. Looks generally good.

nooga (approved):

PR #367, opened 2026-06-30
feat(rt): spit/write! accept a byte-array (write raw bytes)

mparrett:
A let-go `String` stores raw bytes, although text-producing operations encode characters as UTF-8. Consequently, `(str (char 200))` produces the two bytes `0xC3 0x88`, so it cannot be used to construct an arbitrary single byte such as `0xC8`. Byte arrays can represent arbitrary binary data, but the existing output sinks could not write their contents directly.

Specifically:
- `spit` requires a `String` (`spit expected String`).
- `write!` falls back to `vs[1].String()` for a non-String, i.e. a byte-array serializes to its `#byte-array[10 20 …]` *repr*, not its bytes.

So emitting any binary — encoding a PNG, a wasm asset, a zip, a checksum digest — was impossible from pure let-go without shelling out to an external decoder. This is asymmetric: the **read** side already moves raw bytes through a `String` (`read-bytes`/`read-file`/`slurp` return `vm.String(rawbytes)`), and `(bytes "str")` coerces String → byte-array. Only the inverse write was missing.

## Change

Add a package-level helper — the inverse of the `bytes` coercion — and route the two binary sinks through it:

```go
// asBytes returns the raw bytes of a String or a byte-array (ArrayByte TypedArray).
func asBytes(v vm.Value) ([]byte, bool) {
	switch c := v.(type) {
	case vm.String:
		return []byte(string(c)), true
	case *vm.TypedArray:
		if c.Kind() == vm.ArrayByte {
			return c.Unbox().([]byte), true
		}
	}
	return nil, false
}
```

- `spit` (`pkg/rt/lang.go`): `contents, ok := asBytes(vs[1])` instead of a String assertion; error becomes `spit expected String or byte-array`.
- `write!` (`pkg/rt/iort.go`): try `asBytes` first, else fall back to the printed form (unchanged for non-binary values).

Net: `(spit "x.png" (byte-array data))` and `(write! h (byte-array data))` write the bytes verbatim. Existing String behavior is unchanged. `spit` newly accepts byte-arrays, while `write!` intentionally changes byte-array handling from writing the printed `#byte-array[…]` representation to writing its raw bytes.

## Why a helper, not just per-sink code

`asBytes` keeps `spit`/`write!` one-liners and is the spot to later extend the same widening to `io/spit` (`pkg/rt/ions.go`) and `syscall/write-file` (`pkg/rt/syscall_*.go`), which still take String-only. Left those out to keep this PR tight — easy follow-up.

## Tests

`pkg/rt/asbytes_test.go` — `TestAsBytes`: String (incl. a high byte), byte-array (PNG signature + `0xFF`/`0xC8`), and a non-byte typed array (not coercible). Full `pkg/rt` and `pkg/vm` suites pass; `go vet ./pkg/rt/` clean.

## Verified downstream

With a locally-built lg carrying this, a path-tracer demo's headless bench writes a valid `out.png` directly — no temp file, no `openssl` — and `xxd` shows byte-exact output (`8950 4e47 00ff c80d 0a00` for `[137 80 78 71 0 255 200 13 10 0]`).

## Design question

Widen the existing sinks rather than introduce a `bytes->string` constructor. This keeps binary data typed as a byte-array through the I/O boundary and avoids requiring callers to construct a raw-byte String.

### DIFF
--- a/pkg/rt/asbytes_test.go
+++ b/pkg/rt/asbytes_test.go
@@ -2,6 +2,8 @@ package rt
 
 import (
 	"bytes"
+	"os"
+	"path/filepath"
 	"testing"
 
 	"github.com/nooga/let-go/pkg/vm"
@@ -26,5 +28,49 @@ func TestAsBytes(t *testing.T) {
 	// A non-byte typed array is not byte-coercible.
 	if _, ok := asBytes(vm.NewIntArrayFrom([]int64{1, 2})); ok {
 		t.Fatal("int-array should not be byte-coercible")
+	}
+}
+
+func TestSpitWritesByteArrayVerbatim(t *testing.T) {
+	want := []byte{0x89, 0x50, 0x4e, 0x47, 0x00, 0xff, 0xc8}
+	path := filepath.Join(t.TempDir(), "out.bin")
+
+	spit := LookupCoreVar("spit")
+	if spit == nil {
+		t.Fatal("core/spit not found")
+	}
+	if _, err := spit.Invoke([]vm.Value{
+		vm.String(path),
+		vm.NewByteArrayFrom(append([]byte(nil), want...)),
+	}); err != nil {
+		t.Fatalf("spit: %v", err)
+	}
+
+	got, err := os.ReadFile(path)
+	if err != nil {
+		t.Fatalf("read output: %v", err)
+	}
+	if !bytes.Equal(got, want) {
+		t.Fatalf("spit wrote %v, want %v", got, want)
+	}
+}
+
+func TestWriteWritesByteArrayVerbatim(t *testing.T) {
+	want := []byte{0x89, 0x50, 0x4e, 0x47, 0x00, 0xff, 0xc8}
+	var dst bytes.Buffer
+
+	write := LookupCoreVar("write!")
+	if write == nil {
+		t.Fatal("core/write! not found")
+	}
+	if _, err := write.Invoke([]vm.Value{
+		vm.NewBoxed(NewWriterHandle("test", &dst)),
+		vm.NewByteArrayFrom(append([]byte(nil), want...)),
+	}); err != nil {
+		t.Fatalf("write!: %v", err)
+	}
+
+	if got := dst.Bytes(); !bytes.Equal(got, want) {
+		t.Fatalf("write! wrote %v, want %v", got, want)
 	}
 }
--- a/pkg/rt/asbytes_test.go
+++ b/pkg/rt/asbytes_test.go
@@ -1,0 +1,30 @@
+package rt
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/nooga/let-go/pkg/vm"
+)
+
+// asBytes backs the binary file/stream sinks (spit, write!). The byte-array case
+// is the one that used to be impossible: a byte-array handed to spit/write! would
+// stringify to its #byte-array[…] repr, so bytes >127 could never be written.
+func TestAsBytes(t *testing.T) {
+	// String → its bytes verbatim, including a high byte.
+	if b, ok := asBytes(vm.String("hi\xff")); !ok || string(b) != "hi\xff" {
+		t.Fatalf("String: got %q ok=%v", b, ok)
+	}
+
+	// byte-array → raw bytes, high bytes preserved (PNG signature + 0xFF/0xC8).
+	want := []byte{137, 80, 78, 71, 0, 255, 200}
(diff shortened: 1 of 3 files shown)


nnunley (commented):
Reviewed the mechanism against the String type and the existing byte/rune handling — the code is correct, one framing/prose issue to fix.

**Code is right, and encoding-agnostic.** `asBytes` on the String case is `[]byte(string(c))` — raw bytes, no UTF-8 interpretation. It matches what `String.getBytes` (no-arg) already does (`pkg/vm/string.go:85`), and it round-trips read-side bytes verbatim. No behavior change for existing String callers. Good.

**But the 'a let-go String is UTF-8' framing overstates the invariant.** String is a raw-byte container that's UTF-8 *by convention when built from text*, not by guarantee:
- `read-bytes`/`slurp` deliberately push arbitrary (possibly non-UTF-8) bytes through a String — the PR relies on this itself.
- `key_tokenizer.go:63-71` explicitly handles invalid UTF-8 ('invalid byte — emit as-is, never stall').
- The type is dual-natured: `count` = `len` (**bytes**, string.go:47) while `seq`/`nth`/`ValueAt` index `[]rune` (runes). That split *is* the byte/rune handling.

The real invariant the code depends on is narrower and always true: **String storage is raw bytes; `asBytes` returns them verbatim (inverse of `getBytes`/`bytes`).** Suggest reframing the rationale around that rather than a UTF-8 guarantee.

**Factual fix:** `(str (char 200))` yields `0xC3 0x88`, not `0xC2 0x88` (U+00C8 UTF-8; `0xC2 0x88` is U+0088). The point stands, the bytes are wrong.

**On the open question (widen vs `bytes->string` constructor) — worth knowing this diverges from Clojure by design.** Clojure's `spit` is:
```clojure
(with-open [^java.io.Writer w (apply jio/writer f options)]
  (.write w (str content)))
```
i.e. character-only: it `(str content)`s (a byte-array becomes its `[B@…` object repr — never its bytes) and encodes through a Writer with charset/`:append`/`:encoding` options. Clojure has no raw-byte `spit`; you'd reach for `io/output-stream`+`io/copy`. `write!` has no clojure.core analog at all. So widening `spit` to write bytes is a deliberate, reasonable improvement over the Clojure wart — I'd take the widening — but flagging that (a) it means let-go `spit` intentionally does NOT match Clojure's character/charset semantics, and (b) there's still no `:append`/`:encoding` option path, if that matters later.

Non-blocking — code LGTM, just asking for the prose/rationale correction.

nnunley:
Re: the open question (widen sinks vs `bytes->string` constructor) — worth recording exactly how this sits against Clojure, since `spit` shares the name. Clojure's is character-only (core.clj):

```clojure
(defn spit [f content & options]
  (with-open [^java.io.Writer w (apply jio/writer f options)]
    (.write w (str content))))
```

Three concrete divergences this PR introduces/keeps:

1. **Character stream vs raw bytes.** Clojure writes through a `java.io.Writer` (charset-encoded). let-go `spit` here writes raw bytes via `os.WriteFile` — byte-exact, no charset layer; a String goes out as its verbatim UTF-8 bytes rather than platform-charset-encoded.
2. **Byte-arrays.** Clojure `(str content)`s everything, so a byte-array serializes to its `[B@…` object repr — never its bytes. Clojure has no raw-byte `spit`; you drop to `io/output-stream` + `io/copy`. This PR makes let-go do what Clojure deliberately can't, which is the right call IMO — just noting it's an intentional divergence, not a compatibility match.
3. **Options.** Clojure forwards `:append` / `:encoding` to `jio/writer`; let-go `spit` takes neither (always truncate + UTF-8 bytes). Fine for now, but that's the gap a future follow-up would close.

Also: `write!` has no `clojure.core` analog at all (nearest is `.write` on a Writer/OutputStream, or `io/copy`) — so there's no Clojure semantics to match there; you're free to define it.

Net recommendation stands: take the widening. Just recommend the PR description frame it as 'raw-byte write, intentionally unlike Clojure's character/charset spit' rather than leaning on a 'String is UTF-8' invariant.

mparrett:
Updated the description to frame Strings as raw-byte storage, corrected the `(char 200)` byte sequence, and documented that widening `spit` intentionally extends let-go’s existing raw-byte semantics rather than matching Clojure’s character-writer semantics. I also closed the design question in favor of widening the sinks.

Also, added and pushed sink-level tests for exact byte output:
- `spit` writes byte-arrays verbatim to disk.
- `write!` writes byte-arrays verbatim through an I/O handle.
- `go test ./pkg/rt/...` passes.

nnunley (approved):

PR #368, opened 2026-06-30
feat(rt): base64-encode / base64-decode builtins

mparrett:
Native base64 in the runtime — `base64-encode` and `base64-decode`. There's no base64 in let-go today (only JSON/EDN encode), so any program that needs to move bytes through a text-only channel has to hand-roll it.

**Stacked on #367.** This builds on the `asBytes` helper #367 adds, so it includes that commit too — only the second commit (base64) is new here. Merge #367 first and I'll rebase this down to the base64 change alone.

## Change

Thin wrappers over Go's `encoding/base64` (`StdEncoding`):

- `(base64-encode x)` — `x` is a String or byte-array (via `asBytes`) → padded base64 String.
- `(base64-decode s)` — String → byte-array.

## Why

base64 is the standard way to carry arbitrary bytes through a text seam, which let-go needs because a String is UTF-8 and can't hold bytes >127 built from ints. With these builtins that's a one-liner instead of ~15 lines of hand-rolled bit-twiddling. It enables:

- **Images in the terminal** — base64 a PNG into an iTerm2/kitty inline-image escape (what prompted this PR: a path-tracer demo emits one per frame).
- **Data URIs** — `data:image/png;base64,…` to embed a render in generated HTML, or a download link.
- **Decode-and-write** — paired with #367's byte-array `spit`, decode a base64 blob (an embedded asset, an API response) straight to a binary file.
- **Binary over text channels generally** — `js/emit` JSON payloads, the `storage` namespace, URL params, logs — anywhere raw bytes can't travel.

Together with #367 this closes the loop: #367 writes bytes to disk, base64 carries them through everything that isn't disk.

## Tests

`TestBase64` (`pkg/compiler/base64_test.go`) — encode of a byte-array and a String, plus a decode round-trip — run through the compiler the way `reader_bb_test` tests other rt builtins. Full `pkg/rt`/`pkg/vm`/`pkg/compiler` suites green; `go vet ./pkg/rt/` clean.

### DIFF
--- a/pkg/compiler/base64_test.go
+++ b/pkg/compiler/base64_test.go
@@ -1,0 +1,45 @@
+/*
+ * Copyright (c) 2026 let-go contributors
+ * SPDX-License-Identifier: MIT
+ */
+
+package compiler
+
+import (
+	"testing"
+
+	"github.com/nooga/let-go/pkg/vm"
+)
+
+// base64-encode/decode are rt builtins, so exercise them through the compiler the
+// way reader_bb_test does for other rt features.
+func TestBase64(t *testing.T) {
+	// byte-array → padded base64 String (PNG signature → "iVBORw==").
+	if v := evalStr(t, `(base64-encode (byte-array [137 80 78 71]))`); v != vm.String("iVBORw==") {
+		t.Fatalf("encode byte-array: got %v", v)
+	}
+	// String input encodes too.
+	if v := evalStr(t, `(base64-encode "hi")`); v != vm.String("aGk=") {
+		t.Fatalf("encode string: got %v", v)
+	}
+	// Round-trip: decode(encode(bytes)) preserves the byte count.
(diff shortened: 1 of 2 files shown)


nooga (approved):

mparrett:
**[P2] Keep the base64url APIs interoperable.**

The new `base64url-*` functions use unpadded `RawURLEncoding`, while the existing `io/encode :base64url` and `io/decode :base64url` paths use padded `URLEncoding`. As a result:

- `(base64url-decode (io/encode :base64url "x"))` fails on the padding.
- `(io/decode :base64url (base64url-encode "x"))` fails because padding is absent.

Please either use one representation consistently or, to preserve existing output while retaining URL-friendly raw output here, make both decoders accept padded and unpadded forms.

The focused `compiler`, `rt`, and `vm` suites otherwise pass on the PR head and when merged with current `main`.

PR #371, opened 2026-07-01
Multi-arity direct calls + self-host coalesce fix, os/exec* streaming, VM Box fast-path

nnunley:
A 3-commit stack. Opening the multi-arity lowering gate so native direct calls
happen for multi-arity `clojure.core` callees, plus the fixes that made it
landable (a native self-host coalescing bug and a compile-time alloc regression),
plus a supporting tooling improvement.

## Commits

**1. `feat(ir/lower-go): multi-arity direct calls + native self-host coalesce fix`**
- Opens the multi-arity gate: multi-arity fns lower to native per-arity helpers
  (`<name>_<arity>`) + a `len(args)`-switch adapter, with arity carried on the
  ir-fn form (`:multi-arity?`/`fn-arity`), not a dynamic var.
- Fixes the native self-host regression it exposed: `oos-unsafe-names` now
  detects **edge-copy vs in-block-def** interference. `when-let`/`if-let`
  auto-gensym temps (`temp#`) are not unique across sibling/nested
  macroexpansions, so two nested `when-let`s share one source name while both
  live; the out-of-form coalescing merged them onto one Go local, clobbering
  the branch cond in lowered `lower_tree`. Versioning the interfering name
  restores Indexed-RPN value identity.
- Adds `*strict-structured?*` (`LG_STRICT_STRUCTURED`): hard-fail the goto-CFG
  fallback (→ bytecode) so a fn the structurizer can't recover can't silently
  ship a mis-lowered body.

**2. `feat(os): route os/exec* through *out*/*err*; surface generate.lg errors`**
- `os/exec*` streams the child through the dynamic `*out*`/`*err*` bindings
  (fd fast-path preserved when unbound, so interactive children keep their TTY).
- `generate.lg` merges child stderr into the logged stream so an lgbgen
  `CompileError` can't hide behind a bare `exit 1`.

**3. `perf(vm): non-reflection fast path for func(Value…N)(Value,error) natives`**
- `theNativeFnType.Box` now type-asserts the common lowered-lambda shapes and
  dispatches directly, skipping the per-invocation `[]reflect.Value`, arg
  boxing, and `reflect.Call`. Fixes the compile-alloc regression the gate
  exposed (higher-order core calls moved onto reflection-boxed native paths).

## Verification
- `ir-stress lower-go` self-host: **698/698 under `gogen_ir`**, parity-identical
- jank suite: 5657/5657 both engines; `gogen-diff` PASS; `check-generated` OK
- `BenchmarkIRCompile [gogen_ir]` allocs **247k → 226k** (+9.8% → +0.53% vs
  baseline), bytes below baseline, ~12% faster; deterministic-regression
  section empty
- `BenchmarkClojureTestSuite [aot_native]` runtime **-32% to -38%**

Note: the codegen commit carries the alloc regression that the VM commit fixes,
so the three are bisect-clean only as a set.


mparrett:
**Findings (codex, high-5.5):**

- **[P1] Invalid Go identifiers remain possible** — [lower_go.lg:3176](/private/tmp/let-go-pr371/pkg/rt/core/ir/lower_go.lg:3176). The export munge map omits valid Lisp-symbol characters such as `@`, `#`, `.`, brackets, and Unicode. For example, `foo@bar` emits `Foo@bar`, which Go rejects.

- **[P1] Export-name collisions cause duplicate declarations** — [lower_go.lg:3202](/private/tmp/let-go-pr371/pkg/rt/core/ir/lower_go.lg:3202). The PascalCase conversion is not injective: `vector?` and `vector-qmark` both emit `VectorQmark`. I reproduced two `func VectorQmark(...)` declarations from one namespace, making generated Go uncompilable. The naming scheme needs collision-safe encoding or detection/fallback.

- **[P2] Import-cycle regression test is now ineffective** — [lisp_crosspkg_wrapper_test.go:114](/private/tmp/let-go-pr371/pkg/ir/lisp_crosspkg_wrapper_test.go:114). It still searches for `cycb.LG_`/`cyca.LG_`, but this PR removes that prefix. Both checks are always false, so reciprocal direct calls would no longer fail the test.

Focused tests pass: `go test ./pkg/vm ./pkg/ir`.

---
Alignment check: we’re aligned with issue #270.

The discussion defines PR #371’s goal as per-arity direct calls and registry entries. It does not establish lossy PascalCase naming as an accepted constraint. In fact, it describes helpers as `<go-name>_<arity>` and `reduce_2`, which implies retaining the existing safe munged name.

Therefore:

- The multi-arity/direct-call implementation is consistent with #270.
- The first two findings concern the additional exported-PascalCase naming switch, not the direct-call design.
- The stale cycle test remains a valid current-diff issue because the test still expects removed `LG_` names.

I would retain all three findings. The naming findings could be presented as one P1: “export naming must remain valid and collision-free.”

nnunley:
Confirmed all three against the branch head:
- **P1 collisions** — reproduced; `vector?` and `vector-qmark` both PascalCase to `VectorQmark`, so one namespace can emit two `func VectorQmark`. Real.
- **P1 invalid idents** — the export munge map omits `@ # .` etc.; corpus-safe today (as the comment notes) but not guaranteed, so I won't lean on that.
- **P2 stale test** — right, the `cycb.LG_`/`cyca.LG_` grep is dead now that exports are PascalCase; both `Contains` checks are always false and the cycle `Fatalf` can't fire.

Agree with your alignment read: all three ride on the export-*naming* switch, not the multi-arity direct-call design #270 asked for.

Keeping the PascalCase exports — the readable Go output is worth it — and making the scheme safe rather than reverting:
- **Collisions:** add a per-unit collision pass — when two source names map to the same PascalCase, fall back to the injective safe munge for the whole colliding group, so output stays deterministic regardless of encounter order.
- **Invalid idents:** make the munge map total — any non-`[A-Za-z0-9]` char with no explicit token gets a deterministic escape (e.g. `U<hex>`), so an unmapped char can never reach an illegal identifier.
- **P2 test:** re-point it at the actual emitted PascalCase names so the import-cycle guard fires again.

nooga (approved):

PR #372, opened 2026-07-01
feat(ir): type math intrinsic call results as :float in typeinfer

mparrett:
## What

Type the result of a call to a float-returning `math` intrinsic as `:float` in the `typeinfer` pass, so the `gogen_ir` lowering can unbox it to native `float64` and let contagion flow through the surrounding arithmetic.

## Why

`call-core-type` only knew a handful of `clojure.core` builtins. A call to `math/sqrt` (and the rest) fell through to `:unknown`, so its result lowered as a boxed `vm.Value` — even though these functions always return a `Float`.

That starves numeric kernels of a float *source*. With no `:float` origin, contagion never fires, and arithmetic that should be native stays boxed:

```
sd = Invoke(math/sqrt, disc)      // result boxed
r  = rt.SubValue(b, sd)           // stays boxed, no native float64
```

## The change

Type the `:call` result of the float-returning math fns `:float`. The set is exactly the `mathFn1`/`mathFn2` registrations in `pkg/rt/math.go` (which return `vm.Float` unconditionally), matched under both the `math` and `clojure.math` namespaces.

It deliberately excludes:
- `round` — returns `long` (matching Clojure; `pkg/rt/math.go` says so too)
- `abs` — `clojure.core`-shadowing and int-preserving
- the `*-exact` ops — operate on longs

## Scope and soundness

- Affects the `gogen_ir` lowering only. The bytecode interpreter is untouched, so bytecode semantics don't change.
- The `:float` result unboxes via the existing `float64(x.(vm.Float))` path in `lower-go`. That's sound because the runtime fn always returns `Float` — the same var-identity assumption `call-core-type` already makes for the `clojure.core` builtins. If a `:float` name drifts from `math.go`, the cost is a missed unbox (stays boxed), never a miscompile.
- One pre-existing caveat, not introduced here: if user code redefines e.g. `math/sqrt` to return a non-Float, the AOT unbox would panic — the same tension the direct-call AOT optimization already carries. The bytecode path is immune.

## Testing

`test/gold-aot/math_float.lg` exercises `sqrt`/`floor`/`pow`, float contagion (`(+ (math/sqrt 16.0) 1)` => `5.0`), the `clojure.math` alias, and the `round`-stays-integer exclusion. `TestGogenAOTDiff` confirms the bytecode and `gogen_ir` engines agree. `go test -short ./pkg/ir/... ./pkg/compiler/... ./test/e2e/...` is green.

## Open design questions

1. The `:float` set is a hand-synced mirror of `math.go`. Would you prefer `math.go` export the float-returning name set (or a per-fn flag) so the IR pass can't drift? I kept it simple here to match the existing `call-core-type` style, but I could alternatively wire the exported-set version.
2. Natural follow-on, out of scope here: emit a native `math.Sqrt` call instead of unboxing a boxed `Invoke`. Bigger — `lower-go` needs a math-intrinsic emission path — so I left it for a separate PR. This change already gets the contagion benefit.


nnunley:
Nice, this is the right layer for it and the exclusions (`round`, `abs`, `*-exact`) are correct.

One gap: `IEEE-remainder` is missing from `float-returning-math-fns`. It's registered as `mathFn2(math.Remainder)` (`pkg/rt/math.go:325`), and since every `mathFn1`/`mathFn2` wrapper returns `vm.Float` unconditionally, it's float-returning just like `hypot`/`copy-sign`/`next-after` which you did include. It's the only such registration left out of the set. One-line add.

(This is exactly the hand-sync drift your design note called out — might be worth a follow-up test that asserts the set is a subset of the actual `mathFn1`/`mathFn2` registrations so this can't silently rot.)

nooga (approved):

mparrett:
Thanks for the feedback! Fixed in 8b5d172: added `IEEE-remainder` to `float-returning-math-fns` and covered it in the AOT fixture.

Also added a cheap drift test that parses `pkg/rt/math.go` for `mathFn1`/`mathFn2` registrations and compares that exact set against `float-returning-math-fns`, so future hand-sync misses should fail in `go test ./pkg/ir`.

Checks run:
- `go run ./cmd/check-generated`
- `go test -run TestTypeInferFloatReturningMathFnsMatchRuntimeRegistrations -count=1 ./pkg/ir`
- `go test -short ./pkg/ir/... ./pkg/compiler/... ./test/e2e/...`
- `make gogen-diff`

nnunley (approved):
Regenerate the conflicted files?

mparrett:
Done — 01707a1 merges `origin/main` and regenerates `pkg/rt/generated.sums` from the merged sources (`make generate`; the digest is a single content hash, so it can't be hand-merged). `core_compiled.lgb` regenerated in lockstep. `make check-generated` is OK and `TestGogenAOTDiff` stays green. PR is mergeable again.

PR #374, opened 2026-07-01
fix(ir): chained-comparison lowering + ec/callErr name collision in gogen

nooga:
## What broke

Found while AOT-compiling a real-world multi-namespace program (a small
terminal editor, ~14 namespaces) to native Go via `scripts/lg-compile`.
Both bugs are in the IR build/lowering pipeline that backs the AOT Go
backend (`ir.build` / `ir.lower-go`), not anything specific to that
program — any `.lg` source hitting these patterns would miscompile the
same way.

### 1. Chained comparisons lower to a nonsensical fold

`(<= 97 c 122)`-style range checks (e.g. alnum classification) go through
`build-builtin-op`'s n-ary fallback, `fold-binary-chain`, for any op with
3+ args. That fold is correct for associative arithmetic (`+`/`-`/`*`):
`[a b c d] → (((a op b) op c) op d)`, feeding the previous result back in
as an operand. It is **wrong for comparisons** — Clojure's `(<= a b c)`
means "pairwise adjacent comparisons, ANDed" (`(and (<= a b) (<= b c))`),
not "chain the boolean result of the first comparison into the second as
if it were a number." Concretely, the generated Go for `(<= 97 c 122)`
was:

```go
v6 = rt.LeValue(vm.Int(97), c)   // bool
or__x_7 = v6 <= 122               // comparing a bool against an int literal
```

— a type error in Go (caught at `go build`), and semantically wrong even
where the types might have lined up.

### 2. `ec`/`callErr` aren't collision-checked against user locals

Every lowered function's Go signature hardcodes `ec *vm.ExecContext` as
its first parameter, and every function body declares a shared `callErr
error` local. Lisp-level local names munge straight to their Go
identifier with no check against these two fixed, non-gensym'd names.
A source local literally named `ec` (in the test case: a destructured
`end-column`, a very reasonable name to pick) redeclares the function's
own `ec` parameter in the same block — a hard `go build` error, and every
`ec.Invoke(...)` call after the shadowing assignment is now operating on
the wrong thing.

## The fixes

- **`pkg/rt/core/ir/build.lg`**: added `build-comparison-chain`, invoked
  from `build-builtin-op` when the op is one of `:lt :lte :gt :gte :eq`
  with 3+ args. It builds one comparison Inst per **adjacent pair** of the
  original args and joins them with genuinely short-circuiting `if`s — the
  same shape the `and` macro already expands source-level `and` into. No
  new IR op, no bytecode-lowering (`ir.lower`) or opcode changes needed;
  it's built entirely out of existing, already-tested `build-if`/
  `build-form` machinery, by gensym-binding each pairwise comparison to a
  local and feeding a synthesized nested-`if` form back through
  `build-form`.
- **`pkg/rt/core/ir/lower_go.lg`**: extended the existing
  `go-reserved-words` set (already used to suffix real Go keywords that
  collide with a Lisp identifier) with `"ec"` and `"callErr"` — the two
  fixed identifiers the lowering itself injects into every function's
  scope.
- Regenerated the precompiled core bundle (`go generate ./pkg/rt/`) since
  `pkg/rt/core/ir/*.lg` changed.

`go test ./... -count=1` passes. Verified end-to-end by AOT-compiling and
`go build`-ing the real program that surfaced both bugs (all 14
namespaces, direct cross-package calls) — it now builds and boots
correctly.

## Possible followups (not done here, flagging for visibility)

- The unary case `(= 1 (count args))` in `build-builtin-op` returns
  `(nth args 0)` (identity) for comparisons too, per an explicit existing
  comment ("comparisons → identity"). That means `(<= x)` returns `x`
  itself rather than `true`, which doesn't match Clojure (`(<= 5)` => `true`).
  Pre-existing, not touched here — didn't want to change documented-as-
  intentional behavior as a drive-by in a bugfix PR, and it wasn't
  exercised by the program that surfaced the two bugs above.
- The `ec`/`callErr` collision fix is a targeted allowlist of the two
  identifiers the lowering currently hardcodes. If a future change adds
  another fixed non-gensym'd identifier to every function's scope (or to
  some other shared scope), it'll need adding to `go-reserved-words` too —
  there's no structural guard preventing a new collision class from being
  introduced the same way.
- Didn't attempt to regenerate `pkg/rt/core_go_lowered/` (the separate
  gogen_ir-tagged pre-lowered core tree) — out of scope for this fix, and
  `go test ./...` already passes without it since that tree wasn't stale
  in a way any test exercises.

### DIFF
--- a/pkg/rt/core/ir/build.lg
+++ b/pkg/rt/core/ir/build.lg
@@ -412,6 +412,14 @@
 ;; Ops whose arity-1 form is the canonical shape (no n-ary folding).
 (def unary-only-ops #{:inc :dec})
 
+;; Comparison ops are NOT associative the way +/-/*/= (arithmetic) are, so
+;; they can't go through fold-binary-chain's left fold: (<= a b c) means
+;; "b is between a and c" — pairwise checks between ADJACENT original args,
+;; ANDed — NOT (((a <= b) <= c) ...), which would feed a comparison's bool
+;; RESULT back in as an operand of the next comparison (wrong type, and
+;; wrong even if the type happened to line up). See build-comparison-chain.
+(def comparison-ops #{:lt :lte :gt :gte :eq})
+
 (defn build-args [forms ctx]
   ;; Build each arg, AND register its result as a temporary live local
   ;; under a gensym so any subsequent arg's build-if symmetric pass-
@@ -482,6 +490,28 @@
                op-kw [acc (nth args i)] nil)
              (inc i)))))
 
+(defn build-comparison-chain [op-kw args ctx]
+  "[a b c d] → (and (op a b) (op b c) (op c d)): one comparison Inst per
+   ADJACENT pair of original args, joined with real short-circuiting `if`s
+   (same shape the `and` macro expands to) instead of fold-binary-chain's
+   left fold. Reuses build-if/build-form for the join logic, so this needs
+   no new IR op — just gensym'd locals naming each pairwise comparison so
+   the synthetic `if` forms below can refer back to them by symbol."
+  (push-locals! ctx)
+  (let [pair-syms (mapv (fn [i]
+                          (let [pid (add-inst! ctx (ctx-block ctx) op-kw
+                                      [(nth args i) (nth args (inc i))] nil)
+                                sym (ctx-gensym! ctx "cmp__")]
+                            (bind-local! ctx sym pid)
+                            sym))
+                        (range (dec (count args))))
+        chain-form (reduce (fn [acc sym] (list 'if sym acc false))
+                           true
+                           (reverse pair-syms))
+        result (build-form chain-form ctx)]
+    (pop-locals! ctx)
+    result))
+
 (defn build-builtin-op [op-kw form ctx]
   ;; Builtin ops are binary (stack-effect 2→1). Fold n-ary calls.
   ;; Unary-only ops (:inc, :dec) take exactly one ref.
@@ -512,6 +542,8 @@
         :else (nth args 0))
       (= 2 (count args))
       (add-inst! ctx (ctx-block ctx) op-kw args nil)
+      (contains? comparison-ops op-kw)
+      (build-comparison-chain op-kw args ctx)
       :else
       (fold-binary-chain op-kw args ctx))))
 
--- a/pkg/rt/core/ir/lower_go.lg
+++ b/pkg/rt/core/ir/lower_go.lg
@@ -373,7 +373,13 @@
   #{"break" "default" "func" "interface" "select" "case" "defer" "go"
     "map" "struct" "chan" "else" "goto" "package" "switch" "const"
     "fallthrough" "if" "range" "type" "continue" "for" "import"
-    "return" "var"})
+    "return" "var"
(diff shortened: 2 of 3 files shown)


mparrett:
Reviewed this end-to-end rather than just reading the diff, and it holds up. Verdict: correct and minimal, both bugs are real, ship it.

What I ran, off a worktree at the PR head:

- `go test ./... -count=1` passes, including the genmanifest staleness test, so the regenerated bundle matches the `.lg` sources on disk.
- Compiled a source file exercising `(<= 97 c 122)`, `(< a b c d)`, `(= x y z)`, and a `let`-bound local named `ec`, on both the base and the fix:
  - On the base commit, `go build` fails exactly as described: `v5 = v4 < arg2 (mismatched types bool and int)` for the chain, and `ec redeclared in this block` plus `cannot use arg1 - arg0 ... as *vm.ExecContext` for the collision. Both reproduce.
  - On the fix, the chain lowers to pairwise-ANDed `if`s, the local suffixes to `ec_` while the param stays `ec`, and the package builds clean.

The approach reads right too: reusing `build-if`/`build-form` with gensym'd locals, no new IR op or lowerer change. The locals threading is sound, since `build-if` rebinds outer captures in place, so the `push`/`pop` in `build-comparison-chain` only discards the frame it owns.

Three small notes, none blocking:

1. The comment on `comparison-ops` says comparisons aren't associative "the way +/-/*/= are" — but `=` isn't associative either (`(= a b c)` isn't `((a=b)=c)`), and the code correctly puts `:eq` in `comparison-ops` and handles it pairwise. Worth dropping `=` from that parenthetical so the comment doesn't read as if `=` should have gone through the fold.

2. "Short-circuiting" is only half true. The boolean joins short-circuit, but the pairwise comparison instructions are all emitted eagerly before the chain — in the generated `between3`, `cmp__3 = arg2 < arg3` is computed before `cmp__2` is tested. Harmless for pure comparisons and it matches Clojure's arg-eval order, so this is just a wording nit on the docstring, not a code change.

3. Deferring the unary `(<= x)` → identity case is the right call — pre-existing, documented as intentional, and off the path that surfaced these two.

nnunley (approved):
Approving — traced both fixes, they're correct.

**Chained comparisons:** the reduce builds `(if p0 (if p1 (if p2 true false) false) false)` = `p0 && p1 && p2` in correct left-to-right order, i.e. `(and (<= a b) (<= b c))`. Old fold fed a bool back into the next comparison (real `go build` type error). Including `:eq` is right too — pairwise-adjacent, not `(= (= a b) c)`.

**ec/callErr:** minimal and correct; complements (doesn't overlap) the coalesce-map package-alias collision fix in #371 — that one handles local-vs-imported-alias, this handles local-vs-injected-param. Both needed.

One small note, not blocking: the pairwise comparison Insts are emitted eagerly into the entry block before the short-circuiting `if`s, so both comparisons always execute — only the boolean *fold* short-circuits. Safe (operands pre-evaluated, comparisons pure), just slightly more than the description's 'genuinely short-circuiting' implies.

Heads up: this and #371 both regen core_compiled.lgb + generated.sums, so whichever lands second will conflict on those artifacts — resolve with `make generate`, not a hand-merge of the .lgb. Otherwise orthogonal.

PR #377, opened 2026-07-02
feat(wasm): -w-shell accepts a custom HTML template

nnunley:
## What

Generalizes `lg -w -w-shell` from `xterm` | `none` to also accept a **custom HTML template path**. When given a template, `lg` splices the assembled host runtime JS into it natively at the `__LG_HOST_JS_BODY_PLACEHOLDER__` marker — no external post-processing step needed.

Motivation: building a client-owned browser shell (e.g. a compiler-explorer UI) currently requires `-w-shell none` followed by an out-of-band script that scrapes the emitted `<script>` and injects it into a custom page. This lets `lg` do that splice itself.

## Change

- `pkg/rt/wasm/embed.go`: extract `buildHostJS(...)` (the wasm_exec + gz-b64 + host-core + eval assembly) shared by `AssembleHTML` and the new `AssembleHTMLWithTemplate(template, ...)`, which injects host JS into a caller-supplied template at `HostBodyMarker` and does not touch the built-in `__LG_XTERM_*` markers. `AssembleHTML`'s output is byte-identical (existing golden tests unchanged).
- `lg.go`: `-w-shell` accepts a template path; new `resolveShell` validates it (must exist + contain exactly one marker) and emits a clean `error:` + exit 1 on failure (never a panic).
- `wasm.go`: `buildWasm` gains a `customShellTemplate` param and routes to `AssembleHTMLWithTemplate` when set.

## Tests

- `pkg/rt/wasm/embed_test.go`: `TestAssembleHTMLWithTemplate` (host JS injected, marker consumed, no xterm assets, custom markup preserved); existing goldens stay byte-identical.
- `wasm_test.go`: `TestResolveShell` covers xterm / none / missing / no-marker / valid.
- `go build ./...`, `go vet` clean.

### DIFF
--- a/lg.go
+++ b/lg.go
@@ -23,6 +23,7 @@ import (
 	"github.com/nooga/let-go/pkg/nrepl"
 	"github.com/nooga/let-go/pkg/resolver"
 	"github.com/nooga/let-go/pkg/rt"
+	wasm "github.com/nooga/let-go/pkg/rt/wasm"
 	"github.com/nooga/let-go/pkg/vm"
 
 	_ "github.com/nooga/let-go/pkg/rt/corefns"
@@ -329,7 +330,7 @@ func init() {
 	flag.StringVar(&bundleOutput, "b", "", "bundle .lg file into a standalone executable (specify output path)")
 	flag.StringVar(&bundleBase, "bundle-base", "", "path to target-platform lg binary for cross-OS bundling (defaults to current executable)")
 	flag.StringVar(&wasmOutput, "w", "", "build .lg file into a WASM web app (specify output directory)")
-	flag.StringVar(&wasmShell, "w-shell", "xterm", "shell for -w: 'xterm' (default) or 'none' (emit core only; client supplies its own shell via window.LetGoHost)")
+	flag.StringVar(&wasmShell, "w-shell", "xterm", "shell for -w: 'xterm' (default), 'none' (emit core only; client supplies its own shell via window.LetGoHost), or a path to a custom HTML template containing __LG_HOST_JS_BODY_PLACEHOLDER__")
 	flag.StringVar(&wasmPayload, "w-wasm", "inline", "wasm delivery for -w: 'inline' (default; gzip-base64 baked into index.html) or 'external' (emit a separate main.wasm the loader fetches + streams)")
 	flag.BoolVar(&wasmHostEval, "w-host-eval", false, "for -w: expose LetGoHost.eval(code) to call into the loaded image and keep it live (park after the program's main returns); works in both boot modes. Pair with -w-shell none")
 	flag.StringVar(&storageID, "storage-id", "", "logical storage store id for the storage namespace (default: script name, or current directory for main.lg)")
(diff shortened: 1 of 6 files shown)


mparrett:
@nnunley I was thinking about this yesterday, thanks for materializing it! I'll get back to you with feedback soon.

nnunley:
Heads-up on an overlap with #378 ("feat(wasm): neutral #app host mount"): both PRs edit `pkg/rt/wasm/embed.go`, and #378 changes the 5 `pkg/rt/wasm/testdata/assemble_golden_*.html` files — which this PR's `TestAssembleHTML` goldens assert byte-identical.

So the two will conflict on `embed.go` + the goldens; whichever lands second should rebase and (a) re-resolve `embed.go` (this PR extracts `buildHostJS` + adds `AssembleHTMLWithTemplate`; #378 changes the host mount), and (b) re-baseline the goldens. Both diffs are small, so it's a quick resolve either way. Flagging so the golden churn stays contained.

mparrett:
Big +1 on the direction — we're a direct consumer of this on the xsofy side, so it's close to home.

Today we build a client-owned browser shell with `-w-shell none` and then a post-build step that grafts our page in before `</body>`. That hand-edit-the-generated-HTML step is the fragile pattern this replaces; we've hit the same head-vs-body class of bug the host-eval demo ran into. So we'd happily retire the graft and author the shell as a `-w-shell <template>` instead — motivated to see this in, and glad to adapt to it.

Two things from the adopter seat:

**1. Coordinate with the neutral-mount work.** There's parallel movement to shift the mount from `#terminal` to a neutral `#app`. That and this custom-template contract both touch the shell/mount boundary, so they're worth sequencing: a custom template makes assumptions about where it mounts, and we'd rather those survive the neutral-mount change than hit the collision after both ship.

**2. Make the template marker a public contract.** Right now it's the internal `__LG_HOST_JS_BODY_PLACEHOLDER__` assembly placeholder. A consumer's shell is a deploy artifact: whatever marker we bake in ships to production and stays there, so a later rename of that internal name breaks us. A deliberate public token — a comment-style `<!-- letgo:runtime -->`, or an exported const documented as the stable template contract — would make it safe to depend on.

Bigger picture, not a blocker: we see this as a stepping stone toward lg vending a standalone host-core JS you `<script src>` from a page you own outright, the end state we'd want for a rich touch shell. This doesn't foreclose that; the ask is only that the contract stay forward-compatible with it.

We also ran it locally and have a couple of small implementation notes (nothing structural) — happy to drop them as an inline review if useful. Whenever it's in, we're glad to be first adopter and port our shell over to report back.

nnunley:
Both asks land from the adopter seat:

1. **Marker as public contract** — agreed. I'll promote `__LG_HOST_JS_BODY_PLACEHOLDER__` to a documented, stable token, leaning toward a comment-style `<!-- letgo:runtime -->` marker so it's inert in the shipped HTML, and treat it as *the* template contract. Keeps it forward-compatible with the standalone host-core `<script src>` end state you're pointing at.
2. **Sequencing with neutral-mount (#378)** — agreed; whichever of us lands second rebases so the template's mount assumptions survive the `#terminal` → `#app` move.

Yes on the inline impl notes — drop them.

mparrett (approved):
Thanks for the collab. Go ahead and land it at your convenience!

PR #378, opened 2026-07-02
feat(wasm): neutral #app host mount

mparrett:
## Context

`lg -w`'s generated frame emits a `#terminal` div plus terminal-presentation CSS (`#terminal{flex:1}`, `.xterm{height:100%}`) in the neutral host page. That's a leak from the runtime into the host frame: a `-w-shell none` client rendering to a canvas or audio surface is still handed a "terminal" div with terminal layout baked in. Worse, that mount ships `display:none` and is unhidden only by the xterm shell, so a shellless client (which omits that shell) renders into an invisible element.

Rebased onto #377, which added the custom `-w-shell <template>` path. A neutral, visible mount is exactly what a custom-template client binds to.

## Change

Make the frame a neutral, generically-named mount, with terminal presentation traveling with the xterm shell instead of the frame:

- **`host.html`**: the mount is `<div id="app">` with a neutral full-viewport rule (`#app{flex:1;min-height:0}`), no terminal-specific CSS. It also drops the inline `display:none`. The old `#terminal` was hidden and only the xterm shell unhid it, so a `-w-shell none` bundle rendered into an invisible mount; it now ships visible, and a shellless client gets a usable mount out of the box.
- **xterm shell CSS**: the one xterm-specific rule (`.xterm{height:100%}`) now ships in `xtermCSS`, assembled in only when the xterm shell is included. A `-w-shell none` bundle gets the bare `#app` mount with no xterm styles.
- **`lg-shell-xterm.js`**: binds `#app`, and on boot hides only the loading status. The `display` toggle is gone, since the frame no longer hides the mount for the shell to unhide.
- Goldens regenerated (all 5 variants).

Net: the frame's contract is "a visible, full-viewport mount named `#app`", and whatever shell mounts into it owns its own presentation.

## Why the rename

The mount id and CSS were the visible half of the leak, so `#app` is the fix, not a cosmetic. A canvas / audio / repl client mounting into it is no longer handed terminal semantics. This also finishes the `min-height:0` frame work from #277: the neutral `#app` carries it, and the terminal-fill behavior that used to be frame-global is scoped to the shell that wants it.

## Consumer impact

A client that mounts the frame element by id `#terminal` (rather than supplying its own container) binds `#app` instead, a one-line change per consumer, made when it updates to this let-go version. Since a build pins one let-go, its shell always matches the mount that build emits, so no compatibility shim is needed. Clients that render into their own container (a canvas demo, or a repl mounting its own div) are unaffected.

## Testing

- Goldens regenerated; the assemble test passes across all 5 variants (xterm / shell-less × inline / external wasm, plus host-eval). The shell-less goldens now carry a bare, visible `<div id="app"></div>`.
- `go test ./...` green; `gofmt` clean; builds on `linux`, `js/wasm`, and `plan9/amd64`.
- Built a `lg -w` bundle and confirmed in a browser that the default xterm shell fills the viewport on the `#app` mount with no console errors; separately confirmed a `-w-shell none` client that mounts its own div renders unaffected.

### DIFF
--- a/examples/host-eval/shell.html
+++ b/examples/host-eval/shell.html
@@ -52,8 +52,8 @@
   /* overflow:hidden clips xterm's pre-fit default grid so it can't transiently
      widen the flex row before fit() shrinks it. */
   .box { height: 300px; border: 1px solid var(--rule); border-radius: 6px; box-sizing: border-box; overflow: hidden; }
-  /* NOT #terminal — the bundle's host page already owns that id; a dupe sends
-     xterm into the hidden one. Use a unique id so our visible box gets output. */
+  /* NOT the host mount (#app) — the bundle's host page already owns that id; a
+     dupe sends xterm into the hidden one. Use a unique id so our box gets output. */
   #evalio { padding: 6px; background: var(--term-bg); }
 
   /* REPL pane: one scroll flow; the prompt is the last line (no separate box). */
--- a/pkg/rt/wasm/embed.go
+++ b/pkg/rt/wasm/embed.go
@@ -32,8 +32,9 @@ var lgShellXtermJS string
 var htmlTemplate string
 
 const (
-	xtermCSS = `  <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@xterm/xterm@5.5.0/css/xterm.min.css">`
-	xtermJS  = `  <script src="https://cdn.jsdelivr.net/npm/@xterm/xterm@5.5.0/lib/xterm.min.js"></script>
+	xtermCSS = `  <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@xterm/xterm@5.5.0/css/xterm.min.css">
+  <style>.xterm{height:100%}</style>`
+	xtermJS = `  <script src="https://cdn.jsdelivr.net/npm/@xterm/xterm@5.5.0/lib/xterm.min.js"></script>
   <script src="https://cdn.jsdelivr.net/npm/@xterm/addon-fit@0.10.0/lib/addon-fit.min.js"></script>`
 )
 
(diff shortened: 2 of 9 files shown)


nnunley:
Heads-up on an overlap with #377 ("-w-shell accepts a custom HTML template"): both PRs edit `pkg/rt/wasm/embed.go`, and this PR updates the 5 `assemble_golden_*.html` fixtures that #377's `TestAssembleHTML` asserts byte-identical.

They'll conflict on `embed.go` + goldens; whichever merges second should rebase and re-resolve. #377 refactors `AssembleHTML` into a shared `buildHostJS` helper + adds `AssembleHTMLWithTemplate` (custom-template path), so a merge there is mechanical — just fold the `#app` mount change into `buildHostJS`/`host.html` and regenerate the goldens. Both are small; flagging for merge ordering.

mparrett:
**Codex Findings (5.5-high)**

- [P2] [pkg/rt/wasm/host.html](/private/tmp/let-go-pr378-review/pkg/rt/wasm/host.html:19) keeps the new `#app` mount hidden with inline `display:none`. The xterm shell fixes this by setting `termEl.style.display = 'block'`, but shellless bundles do not include `lg-shell-xterm.js`, so a `-w-shell none` client that follows the new contract and renders into `#app` gets invisible output unless it also knows to unhide the generated mount. That undermines the “neutral full-viewport mount” behavior. I’d move this visibility toggle into the xterm shell or remove the inline `display:none` from the mount.

**Verification**

- `go test ./pkg/rt/wasm`
- `go test ./...`

mparrett:
Good catch, fixed in 3d166b5.

The `display:none` was inherited from the old `#terminal` mount, and you're right that only `lg-shell-xterm.js` unhid it — so a `-w-shell none` bundle (which drops that shell) rendered into an invisible mount. Removed the inline `display:none` from `#app` and dropped the now-redundant `termEl.style.display = 'block'` from the xterm shell (it still hides the loading status on boot). The mount ships visible, so a shellless client sees output out of the box; the xterm path is visually unchanged — the empty full-viewport mount sits under the centered "loading…" until the terminal opens. Regenerated the 5 goldens; the shellless golden now carries a bare `<div id="app"></div>`.

Verified: `go test ./pkg/rt/wasm`, `go test ./...`, and cross-builds on linux / js·wasm / plan9.

nooga (approved):

mparrett:
@nnunley I plan to merge this soon, but let me know if you have any concerns.

PR #380, opened 2026-07-02
nrepl: use pre-1.25 WaitGroup form

mparrett:
## Motivation

As of 2026-07-05, an agent sandbox (claude.ai, Sonnet 5, high) claimed it only had access to go1.24.4 and was unable to cleanly build let-go.

## Proposal

Remove an avoidable use of a go1.25-only API. `sync.WaitGroup.Go` (added in 1.25) was the only call in the codebase that required a 1.25+ language version; rewriting it to the pre-1.25 `Add`/`go`/`Done` form is behavior-identical.

## What this does not do

An earlier version of this PR claimed it dropped the toolchain floor to 1.22. That was wrong, and @nnunley was right to push on it. The go directive is pinned at 1.25.0 by dependencies, not source: `x/term v0.41.0` (imported by `pkg/rt`, so it ships in the `lg` binary), `x/sys v0.44.0`, and `x/perf` all declare `go 1.25.0`, so `go mod tidy` holds the module at 1.25.0 and a `go1.24.4` build still fails.

So this stands on its own as a small cleanup: it drops the only source-level 1.25 dependency, with no compat claim attached. Lowering the actual floor means downgrading those deps (e.g. `x/term` → v0.40, which declares `go 1.24`), which I'll do as a separate PR so the dependency change gets its own review.

`go.mod` is unchanged.

---

**From original motivating transcript**
(Sonnet 5, high, ~2026-07-02)

> Good news: your actual question is how to fix this properly — let me go pull the real source and see what's really there, and that'll also answer what my sandbox can do with it.Go 1.22.2 is installed. Now let's clone the real repos and see what's actually there.Good — github.com is reachable for git clones. Let's see what's actually in them.This actually checked out — the "reality engine" and "dead language man pages" lines are real, straight from the actual README (I got lucky extrapolating, not skilled — same caveat applies to anything else I say until I've verified it). More importantly: **xsofy has no `go.mod`** — it's pure `.lg` source files run by the `lg` interpreter binary from `nooga/let-go`, and the README says let-go has **no external dependencies**. That last part matters a lot for your sandbox question. Let me check let-go itself and pull up that spell-dsl doc.That's my answer on `go.mod`, but I can dig one step further — let's see whether the code actually *needs* 1.26 features, or if that's just a conservative version pin.Good — the GitHub-hosted deps (`chzyer/readline`, `hashicorp/go-uuid`, `zeebo/*`, `klauspost/cpuid`) fetched fine via direct mode. The real blockers are `golang.org/x/*` packages needing `golang.org` itself for redirect discovery, and the code genuinely uses Go 1.23's `iter` package (not just a conservative version pin). Let me try one more thing — `x/sys` and `x/term` have official read-only mirrors on `github.com/golang/*`, which *is* reachable.That `iter` and now `sync.WaitGroup.Go` — that's genuinely a Go 1.25 stdlib API, not a version-pin quirk. This codebase really does need a modern toolchain. Before giving up on native builds, let me check if there's a prebuilt release binary — the Makefile mentions Homebrew taps, which usually means GitHub Releases with prebuilt artifacts.One more useful, cheap thing before I wrap up — you already have a `spell-dsl.md` design doc in the actual xsofy repo, which beats any more guessing from me.Bottom line: I can't actually **run** `lg` here right now, but I've got real, verified access to the source — not guesses.

> **What's blocked:** let-go's `go.mod` wants Go 1.26. Ubuntu's apt tops out at 1.24.4 (installed it). I confirmed this isn't just a conservative version pin — lowering the directive to test it, the code genuinely calls `sync.WaitGroup.Go` (a real Go 1.25 stdlib addition) and imports the `iter` package, so 1.24 hard-fails on real compile errors, not just a gate. Getting the real 1.25/1.26 toolchain needs `proxy.golang.org` / `golang.org` / `go.dev`, none of which are on my network allowlist. I could build Go itself from source (`github.com/golang/go` clones fine, and 1.24 could bootstrap it) but that's a multi-minute compile for a payoff I can't guarantee — say the word if you want me to push through it.
> Also checked: no cut GitHub Releases for let-go yet, so a prebuilt `lg` binary isn't an option either.

> **What works:** `git clone` against `github.com`/`codeload.github.com` is wide open, so I pulled real xsofy, legmacs, and let-go source. Concretely useful finds:

> - GitHub-hosted deps (`chzyer/readline`, `zeebo/xxh3`, etc.) resolve fine under `GOPROXY=direct`. Only `golang.org/x/*` needs the blocked discovery redirect (workaround exists via `replace` to `github.com/golang/sys` etc., but doesn't fix the toolchain gap above).

> Practically, that means the useful thing I can do right now is **write real `.lg` source against the actual codebase** — new rune/reality primitives following the spell-DSL's existing shape, real function names, real module paths — that you run and iterate on locally where `lg` already works (or in the WASM playground). I can't self-test it here yet.

> Want me to start drafting that against the real xsofy source — e.g. a small `reality` namespace with `collapse-run!`/`defuse!` as actual let-go functions in the shape spell-dsl.md already uses — or do you want to spend the sandbox time getting a working toolchain first?

### DIFF
--- a/pkg/nrepl/server.go
+++ b/pkg/nrepl/server.go
@@ -97,7 +97,9 @@ func (n *NreplServer) Start(port int) error {
 
 	fmt.Printf("nREPL server started on port %d on host 127.0.0.1 - nrepl://127.0.0.1:%d\n", port, port)
 
-	n.wg.Go(func() {
+	n.wg.Add(1)
+	go func() {
+		defer n.wg.Done()
 		for {
 			select {
 			case <-n.stop:
@@ -115,7 +117,7 @@ func (n *NreplServer) Start(port int) error {
 				go n.handleConn(conn)
 			}
 		}
-	})
+	}()
 	return nil
 }
 


nnunley:
WaitGroup rewrite is correct — `Add(1)` before the goroutine, deferred `Done()`, no capture issue.

But the stated goal (compat down to older toolchains) isn't actually reached while `go.mod` still declares `go 1.26` — that directive alone makes 1.24/1.22 toolchains refuse to build. If the intent is real 1.22+ compat, the `go` line needs to drop too; if you're leaving that as a maintainer call, worth saying so explicitly in the PR body so it isn't merged thinking the compat gap is closed.

mparrett:
You're right, and it goes a bit further than the go.mod line. Even with the directive dropped, `go mod tidy` pins it back to 1.25.0: `x/term v0.41.0` (pulled in by `pkg/rt`), `x/sys v0.44.0`, and `x/perf` all declare `go 1.25.0`. So the rewrite removes the only source-level 1.25 use, but the effective floor is set by deps and a `go1.24.4` build still fails.

I've corrected the body to drop the 1.22 claim — this now stands as a plain cleanup with no compat promise. Actually lowering the floor means downgrading those deps, which I'd rather land as a separate PR so the dependency change gets its own review.

mparrett:
Asked the original reporting Claude to re-check his assertions:

--- 

> That team feedback actually makes complete sense once you account for one variable: network egress. Let me get fresh, direct evidence rather than just re-asserting the old finding.Confirmed — and it's not a let-go problem at all, it's this sandbox specifically:

```
proxy.golang.org → 403, x-deny-reason: host_not_allowed
golang.org       → 403, x-deny-reason: host_not_allowed
go.dev           → 403, x-deny-reason: host_not_allowed
github.com       → 200  (for comparison)
```

> Here's what's actually going on: Go's toolchain manager (since 1.21, `GOTOOLCHAIN=auto` by default) auto-downloads a newer toolchain the moment it sees `go.mod` wants a version newer than what's installed. Any environment that can reach `proxy.golang.org` just transparently fetches 1.26 on first build and never notices the version gap — that's almost certainly what your team's Claude sandboxes are doing, since their network egress allowlists apparently aren't as locked down as this one. Mine can't make that call, so I was stuck manually working around it with an older local toolchain, and the `WaitGroup.Go` patch only exists because of that workaround.

> So the underlying fact is still real — that's a genuine compiler error I hit directly, not a guess — but I scoped and wrote up the PR wrong. It reads like "let-go doesn't build," when the honest claim is "let-go doesn't build *in network-restricted environments that can't auto-fetch the toolchain*" (locked-down CI, air-gapped builds, egress-limited agent sandboxes like this one). That's a much narrower audience than the PR implies, and probably exactly why the team's confused — most of them never touch that path.

> A few honest options from here:

> 1. Rewrite the PR description to scope it correctly instead of implying a universal build failure.
> 2. Decide the audience is narrow enough that it's not worth carrying even a trivial diff, and close it.
> 3. Keep it but as a documented fallback (a CI comment or README note: "toolchain auto-fetch needs `proxy.golang.org`; see #NNN if that's unavailable") rather than a default-path change.

> Is locked-down/air-gapped building an actual scenario your team cares about, or was this one worth closing?

mparrett:
Re-tested:

```
I'll clone and build let-go for you.The repo needs Go 1.26. Let me check what's available in the 
environment.Go isn't installed. The good news is `core_compiled.lgb` is already committed, so I 
don't need the bootstrap toolchain — I just need a Go 1.26 compiler. Let me get one. First, let me 
see what's reachable.Here's the situation: apt only has Go 1.22, but let-go requires Go 1.26, and 
Go's toolchain auto-downloader (go.dev / proxy.golang.org) is blocked by the network policy. 
However, `github.com` is allowed — and that's where the Go source lives. So I'll bootstrap Go 1.26 
from source using the apt Go 1.22 as the bootstrap compiler. Let me kick that off.Bootstrap Go 1.22 
is in. Now let me find the latest Go 1.26 tag on GitHub and build it from source.Latest is 
go1.26.4. Let me shallow-clone that tag and build it (this takes a few minutes).Go 1.26 needs a Go 
1.24.6+ bootstrap, and 1.22.2 is too old to jump straight there. I'll build a chain: 1.22.2 → 1.23 
→ 1.24 → 1.26. Let me get the intermediate tags.Chain confirmed: 1.22.2 → 1.23.12 → 1.24.13 → 
1.26.4. Building the first link now (1.23.12).The build ran past the tool's time limit. Let me run 
these builds in the background and poll for completion.The build process was killed on timeout. Let 
me restart it detached in the background so it survives, and poll.Go 1.23.12 ready. Now building 
1.24.13 with it as bootstrap.Go 1.24.13 ready. Now the final link — building Go 1.26.4 (I already 
cloned its source earlier) using 1.24.13 as bootstrap.Go 1.26.4 is built. Now I'll build the `lg` 
binary directly with it (the committed `core_compiled.lgb` bundle means no bootstrap codegen is 
needed). I'll invoke `go build` directly to bypass the Makefile's Go auto-installer.Dependencies 
must be fetched, but the default Go module proxy is blocked. All these deps live on GitHub, which 
is allowed. For `github.com/...` modules I can use `GOPROXY=direct` (clones straight from GitHub). 
The two `golang.org/x/...` modules resolve via a blocked host, so I'll redirect them to their 
GitHub mirrors with `replace` directives. Let me confirm the mirror tags exist.Both tags exist. Let 
me add the redirect directives and build with direct VCS fetching.`golang.org/x/perf` is also 
pulled in. Same fix — redirect to its GitHub mirror.Two more vanity-hosted (indirect/test) modules 
resolve through blocked hosts. Redirecting them to GitHub as well.`gopkg.in/yaml.v3` can't map to a 
`github.com/...` path at `v3.0.1` (Go's major-version path rule). I'll clone that one locally and 
use a filesystem replace, which sidesteps the rule.The `lg` binary built — 13.5 MB, in line with 
the project's ~12 MB claim. Let me verify it actually runs.Working. Let me run one richer program 
to confirm the runtime (higher-order fns, lazy seqs, data structures) is healthy, then copy the 
binary out for you.Done — let-go is fetched, built, and verified working. The `lg` binary is above.

The one wrinkle worth knowing about: the build needed some detours because of the sandbox's network 
allowlist.

**Toolchain.** `go.mod` pins `go 1.26`, but the only apt-installable Go here is 1.22, and Go's 
toolchain auto-downloader (go.dev / proxy.golang.org) is blocked. So I bootstrapped 1.26.4 from 
source off GitHub, which required chaining through intermediate compilers since each Go needs a 
fairly recent bootstrap: **1.22.2 → 1.23.12 → 1.24.13 → 1.26.4**.

**Dependencies.** `GOPROXY`'s default is blocked too, so I built with `GOPROXY=direct` to clone 
modules straight from GitHub. Three module groups don't live on GitHub under their import path and 
had to be redirected via `replace`:
- `golang.org/x/{sys,term,perf}` → their `github.com/golang/*` mirrors (matching tags/commit)
- `gopkg.in/check.v1` → `github.com/go-check/check` (pseudo-version)
- `gopkg.in/yaml.v3` → a local clone (a `github.com/...` path can't carry a `v3.0.1` tag, so a 
filesystem replace was the clean way around it)

None of that touched real source — only module resolution — and `core_compiled.lgb` was already 
committed, so no bootstrap codegen was needed. The binary is 13.5 MB and the runtime checks out: 
threading macros, lazy seqs, persistent maps, and interop-free core all evaluate correctly.

Two notes for reuse:
- The `replace` lines I appended to `go.mod` are local build scaffolding — you'd want to drop them 
(or keep them out of any commit) since they only exist to dodge the allowlist.
- If you'd rather avoid the whole source-bootstrap dance next time, adding `go.dev` + 
`proxy.golang.org` (or just `storage.googleapis.com`) to the sandbox's egress settings would let 
the standard toolchain download work directly.

Want me to build any of the other targets — the WASM web page, a standalone bundled binary via 
`-b`, or run the test suite?
```

mparrett:
@nnunley @nooga I honestly don't have a great read on whether it's useful to be able to target an older golang and whether this change is worth keeping. The agent foibles are documented and closed out in #390, I'm tempted to close this too, and revisit if a concrete use case arises. Let me know if you have strong opinions, otherwise I'll clean it up.

nnunley:
I don't think it's worth that much to force swimming against the current here. As long as we're protected going forward, I think we're good.

nooga:
I don't think this is worth keeping. We can always dig it up if needed in the future.

PR #383, opened 2026-07-03
feat(vm): carry metadata on functions (with-meta/meta)

### LINKED ISSUE #381: Functions don't support metadata (with-meta/meta drop it)
## Summary

Functions don't support metadata: `with-meta` on a function returns it unchanged and `meta` reads back `nil`.

```clojure
(meta (with-meta (fn [x] x) {:tag :t}))
;; expected: {:tag :t}
;; actual:   nil
```

This affects **every** function-value type, not just plain fn literals:

```clojure
(meta (with-meta + {:a 1}))                 ; native builtin        => nil (expect {:a 1})
(defmulti mm identity)
(meta (with-meta mm {:a 1}))                ; multimethod (fn? true) => nil
(defprotocol P (p [x]))
(meta (with-meta p {:a 1}))                 ; protocol method        => nil
```

## Cause

The function value types (`*Func`, `*Closure`, `*MultiArityFn`, `*NativeFn`, `*MultiFn`, `*ProtocolFn`) don't implement `IMeta`, so `with-meta` falls through its non-`IMeta` pass-through in `rt/lang.go` (the one that lets value-position type hints like `^long x` compile) and returns the function untouched. `meta` then finds nothing.

A subtlety worth flagging for a fix: a capture-free fn compiles to a bare `*Func` constant pushed by `OP_LOAD_CONST`, so it's shared across every evaluation of the literal — `with-meta` there must be copy-on-write, not an in-place mutation.

## Why it matters

[circleci/bond](https://github.com/circleci/bond) (a small, widely-used spying/stubbing library for tests) wraps a fn as `(with-meta (fn [& args] …) {::calls <atom>})` and reads that atom back with `(meta f)`. With metadata dropped, every assertion falls into bond's "not a spied function" throw, so the library is unusable under let-go.

## Notes

- In Clojure, functions (including core fns and protocol methods) carry metadata; multimethods are not `IObj` there, but let-go's `fn?` already includes `MultiFn`, so covering it keeps metadata uniform across all `fn?` values.
- I have a fix ready (copy-on-write `IMeta` on all six function value types, with tests) and can open a PR.
- Found alongside a separate gap — `ns-publics` is missing (filed separately). The two together unblock bond.

**Environment:** `main` (also reproduces on 1.11.1).

abogoyavlensky:
Resolves: #381

### DIFF
--- a/pkg/vm/func.go
+++ b/pkg/vm/func.go
@@ -28,6 +28,7 @@ type Func struct {
 	arity       int
 	isVariadric bool
 	chunk       *CodeChunk
+	meta        Value // IMeta support; nil until (with-meta fn m) copies one in
 }
 
 func MakeFunc(arity int, variadric bool, c *CodeChunk) *Func {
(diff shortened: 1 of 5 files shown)


nnunley (approved):
LGTM

PR #384, opened 2026-07-03
feat(rt): add ns-publics

### LINKED ISSUE #382: Missing ns-publics (and ns-interns/ns-map)
## Summary

`ns-publics` is missing:

```clojure
(ns-publics 'clojure.core)
;; expected: a map of symbol -> var for the namespace's public vars
;; actual:   CompileError: Can't resolve ns-publics in this context
```

`ns-interns` / `ns-map` are likewise absent.

## Why it matters

This is more severe than a typical missing function. [circleci/bond](https://github.com/circleci/bond) (a small, widely-used spying/stubbing library for tests) references `ns-publics` from a **top-level** `defn` (`ns->fn-symbols`). A single unresolved symbol in a top-level form fails the *entire* namespace compile, so even the parts of bond that never touch `ns-publics` (`spy`/`calls`/`with-spy`/`with-stub`) become unloadable:

```
CompileError: Can't resolve ns-publics in this context
```

## Notes

- `the-ns` / `find-ns` / `ns-name` already exist, and `Namespace` already has an enumerable registry, so this is a small additive function: return a `symbol -> var` map of a namespace's non-private interned vars, accepting either a namespace or a symbol naming it, and resolving the same `clojure.core`-style aliases (`clojure.core` -> `core`, `clojure.string` -> `string`, …) the rest of the runtime honors.
- I have a fix ready (`ns-publics` core fn + a `Namespace.PublicVars()` accessor, with tests) and can open a PR.
- Found alongside a separate gap — functions don't carry metadata (filed separately). The two together unblock bond.

**Environment:** `main` (also reproduces on 1.11.1).

abogoyavlensky:
Resolves: #382

### DIFF
--- a/pkg/rt/lang.go
+++ b/pkg/rt/lang.go
@@ -5353,6 +5353,38 @@ func installLangNS() {
 		return vm.Symbol(ns.Name()), nil
 	})
 
+	// ns-publics returns a map of symbol -> Var for the public (non-private)
+	// interned vars of a namespace, given either the namespace or a symbol
+	// naming it (matching Clojure, which passes its arg through the-ns).
+	nsPublics, _ := vm.NativeFnType.Wrap(func(vs []vm.Value) (vm.Value, error) {
+		if len(vs) != 1 {
+			return vm.NIL, fmt.Errorf("wrong number of arguments %d", len(vs))
+		}
+		var ns *vm.Namespace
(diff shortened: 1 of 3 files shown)


nnunley (approved):
Thanks! This looks good.

PR #385, opened 2026-07-03
perf(vm): monomorphic inline cache for protocol dispatch

mparrett:
## What

A monomorphic inline cache for protocol method dispatch.

`ProtocolFn.invokeIn` resolved the implementation on every call via `Protocol.Lookup`: a map index by the receiver's type, then a `PersistentMap` probe for the method keyword. Protocol call sites are overwhelmingly monomorphic in steady state, so that resolution repeats identically call after call.

This caches the last `(type -> impl)` resolution on the `ProtocolFn` behind an `atomic.Pointer`. The guard is a `ValueType` identity compare, which is a pointer compare because value types are singletons, and a hit skips `Lookup` entirely.

## Correctness

A generation counter on the `Protocol` (bumped by `Extend`/`ExtendNil`, recorded in each cache entry) is the watchpoint: any change to the dispatch tables strands stale entries, so a redefined or newly-added impl is never served from cache. `TestProtocolPICInvalidatesOnRedefine` pins this.

## Bounding the downside

A single slot cannot serve a polymorphic site, and storing on every miss would allocate an entry per call, which is worse than no cache. So a site latches "megamorphic" on the second live type it sees and falls back to plain `Lookup` with no further caching, capping the cost at baseline.

## Numbers

Microbenchmark, two environments:

| environment | monomorphic | polymorphic |
|---|---|---|
| Apple M2, `benchstat` n=12 | 106.6 -> 22.6 ns, -78.8% time, 1 -> 0 alloc (p=0.000) | 110.5 vs 112.3 ns, no significant change (p=0.325) |
| x86 linux-amd64 (Intel Xeon), anchor-normalized | -80.0%, flagged improved | within the 12% budget, neutral |

The benchmark uses trivial method bodies to isolate dispatch cost, so the end-to-end win on real code scales with how dispatch-bound it is. The monomorphic result reproduces across Apple silicon and x86, so it is not an M2 artifact.

## Interaction with AOT, IR, and wasm

The cache lives in the bytecode VM's dispatch path (`ProtocolFn.invokeIn`), so it composes with the other execution paths as follows.

**AOT (`gogen_ir`).** Where the receiver type is statically known, lowering devirtualizes a protocol call to a direct Go method call, bypassing `invokeIn`; the cache is inert there, adding neither benefit nor cost. Where the type is not known, or a frozen native baseline was invalidated by a late `extend`, lowered code falls back through `InvokeValueEC` -> `ec.Invoke` -> `invokeIn`, and the cache accelerates that fallback. Static devirtualization and the dynamic cache cover disjoint sites; they do not conflict.

**IR passes.** Constant folding, CSE, DCE, LICM, and type inference run on the IR before lowering and do not touch runtime dispatch, so they are orthogonal to the cache. The one link: stronger type inference feeds more AOT devirtualization, which substitutes for the cache at the margin (fewer sites reach the runtime fallback).

**wasm.** `lg -w` ships the bytecode VM interpreting the bundle, with no lowering in that path, so wasm dispatch runs through `invokeIn`, the path the cache targets. Measured under `js/wasm` on Node, monomorphic dispatch goes from 397.9 ns / 1 alloc to 70.9 ns / 0 alloc (-82%, p=0.000, n=8). The relative win is a touch larger than native: per-op interpretation is costlier on wasm, so removing the method-table probe is a bigger fraction of the total.

## Follow-ups

Not in this PR:

- N-way PIC to serve stably-polymorphic sites (the bimorphic case).
- Site-indexed cache slot. The cache currently lives on the `ProtocolFn`, so it is per-method rather than per-call-site.
- Same treatment for multimethods, where the baseline also re-runs the dispatch fn on every call.

### DIFF
--- a/pkg/vm/protocol.go
+++ b/pkg/vm/protocol.go
@@ -5,7 +5,10 @@
 
 package vm
 
-import "fmt"
+import (
+	"fmt"
+	"sync/atomic"
+)
 
 // Protocol defines a named set of methods with type-based dispatch.
 type Protocol struct {
@@ -14,6 +17,14 @@ type Protocol struct {
 	impls   map[ValueType]*PersistentMap // type → {method-name → fn}
 	nilImpl *PersistentMap               // implementation for nil
 	meta    Value                        // IMeta support
+
+	// gen is the dispatch-table generation, bumped on every extend. It is the
+	// watchpoint backing ProtocolFn's inline cache: any change to impls strands
+	// every cached (type → fn) entry, so a redefinition or a newly-added impl
+	// can never be served stale. Pointer-typed so a WithMeta copy (which shares
+	// the dispatch tables) shares the counter, and so the struct stays copyable
+	// without tripping vet's copylocks check on the atomic.
+	gen *atomic.Uint64
 }
 
 func NewProtocol(name string, methods []Symbol) *Protocol {
@@ -21,6 +32,7 @@ func NewProtocol(name string, methods []Symbol) *Protocol {
 		name:    name,
 		methods: methods,
 		impls:   make(map[ValueType]*PersistentMap),
+		gen:     new(atomic.Uint64),
 	}
 }
 
@@ -50,11 +62,13 @@ func (p *Protocol) WithMeta(m Value) Value {
 // implMap is a PersistentMap of {method-keyword → fn}.
 func (p *Protocol) Extend(vt ValueType, implMap *PersistentMap) {
 	p.impls[vt] = implMap
+	p.gen.Add(1) // invalidate every ProtocolFn inline cache over this protocol
 }
 
 // ExtendNil adds implementations for nil.
 func (p *Protocol) ExtendNil(implMap *PersistentMap) {
 	p.nilImpl = implMap
+	p.gen.Add(1)
 }
 
 // Lookup finds the implementation of a method for a given value's type.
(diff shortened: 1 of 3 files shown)


nooga (approved):

mparrett:
Before merging I wanted to confirm this helps real code, not just the
micro-benchmark. It doesn't fire on xsofy at all.

I instrumented protocol dispatch per method and ran two real workloads on this
branch:

- engine benchmark (entity AI, pathfinding, lighting, fire — the per-turn hot
  path): **0** dispatches
- full test suite (281 tests, 2733 assertions): **0** dispatches

The instrumentation is sound (a synthetic `defprotocol` loop records hits/misses
as expected), so the zero is real, and it's structural:

- the hot collection ops (`count`, `nth`, `seq`, `conj`, `get`, `first`) are Go
  builtins, not protocols, so they never reach `ProtocolFn.invokeIn`
- the only core protocols are `IFn`/`IDeref`/`IROp`, and xsofy defines none of
  its own on the hot path

So the ~12 ns/0-alloc fast path is real but measures a path production code
doesn't take. On the flagship app the cache is pure overhead (the extra `gen`
counter, the atomics, the stats scaffolding on the dispatch path) for no
realized win.

This doesn't prove no program benefits: a hot, globally-monomorphic user
protocol would. But I'd want to see that workload before carrying the machinery,
so I'm leaning toward holding this unless we can point at a realistic
beneficiary.

If we keep it, the earlier issues still stand:

- the new benchmarks aren't in the perf CI filter, so nothing validates the
  optimization
- the `FrameDispatch` improvement in the perf report is unrelated noise; that
  benchmark does no protocol dispatch, and the diff only touches `protocol.go`
- "inline cache" implies per-call-site; this is a per-protocol-function cache
- the megamorphic latch never resets, so one transient polymorphic call disables
  that method's fast path for the rest of the process

mparrett:
Closing this. The measurement settles it for me: zero protocol dispatches across
xsofy's engine benchmark and full suite, and structurally the hot ops are
builtins, so the fast path never fires on real code. Not worth the machinery
without a workload that exercises it. Happy to reopen if a protocol-heavy,
monomorphic use case turns up.

PR #386, opened 2026-07-03
perf(ci): wasm A/B benchmark gate (perf-wasm label)

mparrett:
## What

A `perf-wasm` label gate: a same-runner base-vs-head microbenchmark A/B run under `GOOS=js/wasm`, mirroring the existing `perf` gate but measuring the wasm execution path. Two pieces: a `-wasm` mode on `bench-ratchet`, and `perf-wasm.yml`.

## Why a separate wasm gate

`lg -w` ships the bytecode VM interpreting the bundle, so what runs in the browser is the wasm build, not the native binary. The native gate can't see wasm-only regressions: different codegen, single-threaded, and a GC where allocation pressure hits harder. A change can be flat natively and move under wasm.

## How it works

`bench-ratchet -wasm` runs the selected profile under the `go_js_wasm_exec` shim (Node) instead of natively. Same modes (`update`/`check`/`snapshot`), same anchor normalization — the `RatchetAnchor` bench runs under wasm too, so ratios cancel Node/runner tier differences. `perf-wasm.yml` reuses the `perf-pr.yml` structure: resolve the PR's merge-base and head, bench both on one runner, post the delta to the job summary. Informational, never gates. Opt-in via the `perf-wasm` label, separate from `perf` so wasm's longer, noisier runs don't ride every perf PR.

Three design choices that aren't obvious from the diff:

- **Trimmed environment.** The shim forwards the whole environment into the wasm sandbox via Node argv, which overflows Node's arg-length limit when the parent env is large (CI, dev shells). The `GOOS=js` run keeps only what the toolchain and shim need.
- **Own machine profile.** Wasm runs report the machine as `js/wasm`, so their numbers key to a separate baseline and never mix with native ones.
- **`-tags` off.** The wasm bundle ships the bytecode VM, not the lowered-Go path, so the wasm run drops the `gogen_ir` tag (which also sidesteps building the lowered tree under `js/wasm`), and there is no `make lowered` step.

## Budget

The wasm run uses a 30% budget, wider than the native gate's 12%. Wasm micro-benchmarks under V8 have a higher noise floor: in a same-runner A/B, untouched Go-map lookups swung past 50% run to run. At 12% the report flags phantom movers on code the PR never touched; 30% suppresses the bulk of that while still surfacing real regressions. It only tunes which rows the informational report lists, not a pass/fail.

## Not in this PR

The `-wasm` mode composes with `snapshot` and arbitrary-ref dispatch, so a wasm timeline lane and a persistent wasm ratchet are natural follow-ups. Both need the machine key to also capture the Node/V8 version (a Node bump would otherwise move a persistent baseline), which the same-runner A/B here does not require.

### DIFF
--- a/.github/workflows/perf-wasm.yml
+++ b/.github/workflows/perf-wasm.yml
@@ -1,0 +1,190 @@
+name: Perf WASM
+
+# Same-runner A/B microbenchmark check under GOOS=js/wasm, run via the
+# go_js_wasm_exec shim on Node. Mirrors Perf PR (base vs head, one runner, no
+# gate) but measures the wasm bundle's execution path — the bytecode VM that
+# `lg -w` ships. Native perf can't see wasm-only regressions: different codegen,
+# single-threaded, and a GC where allocation pressure hits harder.
+#
+# Wasm is ~4-8x slower per-op and noisier than native, so this is opt-in via the
+# `perf-wasm` label, separate from the native `perf` gate. Add the label to run;
+# remove it to stop. Informational only.
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, labeled, unlabeled]
+    paths:
+      - 'pkg/**'
+      - 'cmd/**'
+      - '.github/workflows/perf-wasm.yml'
+  workflow_dispatch:
+    inputs:
+      pr:
+        description: 'PR number to benchmark'
+        required: true
+
+permissions:
+  contents: read
+
+concurrency:
+  group: perf-wasm-${{ github.event.pull_request.number || github.event.inputs.pr }}
+  cancel-in-progress: true
+
+jobs:
+  bench:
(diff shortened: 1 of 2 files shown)


nooga (approved):

mparrett:
Review finding:

[P1] `.github/workflows/perf-wasm.yml` runs `go run ./cmd/bench-ratchet ... -wasm` after checking out `$BASE` in the merge-base benchmark step. The merge-base for this PR does not have the newly added `-wasm` flag, so the base half of the workflow will fail before benchmarking. This also breaks labeled long-lived PRs whose fork point predates this change.

Suggested fix: build `bench-ratchet` from the PR head once, store that binary under `$RUNNER_TEMP`, then run that binary against both the base and head checkouts.

mparrett:
Resolved in `96159fc`: the workflow now builds the PR-head `bench-ratchet` once and uses that binary for both the merge-base and head wasm benchmark runs.

PR #387, opened 2026-07-04
feat(ir): lowering strategy pattern + unified ir.coverage

nnunley:
## What

Refactor the IR pipeline's `compile-form*` so a **lowering-strategy registry** replaces the `*target*` conditional, add a **composite** that runs every lowering against one shared interim IR, and consolidate the coverage-failure classifier into a shared **`ir.coverage`** namespace.

This is behavior-preserving groundwork: it makes the lowering back-end pluggable and mutually consistent, so a single build+optimize can feed all lowerings (bytecode + Go) at once instead of recompiling per target.

## Why

- Today `compile-form*` conflates the shared front half (build → optimize) with a `(= *target* :go)` conditional back half. Anything that wants *both* lowerings from one form (e.g. an inspection/emission tool) has to run the whole pipeline twice, which is redundant and can diverge under gensym-counter nondeterminism.
- The coverage-failure classifier (`classify-error`) lived only in `scripts/ir-stress.lg`, so the stress gate, and anything else that needs the same taxonomy, would have to duplicate it.

## Changes (4 commits)

1. `refactor(ir): route single-arity lowering through a strategy registry` — introduce `lowerings` `{:bytecode … :go …}`; the single-arity leaf becomes `((lowerings *target*) ir-fn)`. Strategies return the backend's **native** value (a `CodeChunk` for `:bytecode`, `{:status :lowered :decl …}` for `:go`) — no normalization, so callers are unaffected.
2. `refactor(ir): route multi-arity lowering through the strategy registry` — same seam for the multi-arity leaves; the `:multi-fn-template` / `make-multi-arity` wrapping is preserved.
3. `feat(ir): add lower-fn-all composite` — `(lower-fn-all ir-fn)` runs every strategy against one shared interim IR, returning `{:bytecode … :go …}`. Additive; no existing caller touched.
4. `refactor(ir,scripts): extract classify-error into shared ir.coverage ns` — move the classifier verbatim into `pkg/rt/core/ir/coverage.lg`; `scripts/ir-stress.lg` requires it (exactly one definition remains). The `cond` arm ordering (special-form heads before the generic `unresolved symbol` regex) is preserved character-for-character.

## Verification

- **Byte-identity:** `make check-generated` green after every commit — the registry dispatch is the structural twin of the old conditional, so single-target output is unchanged.
- `go test ./pkg/ir ./pkg/genmanifest` — 174 tests pass (4 new).
- **`make jank-stress` — 273/273 (100%), baseline unchanged** — the classifier move is behavior-identical and lowering output is byte-identical, so the coverage set is preserved.

## Notes

- Two new tests are intentionally narrow (the single-arity registry test and the composite consistency test assert the `:go` path; `:bytecode` is ~100% covered by the existing jank-stress gate). Can broaden on request.
- This is the foundation for a follow-on single-pass emission/inspection engine (the composite is what lets it project the IR pane and both lowering panes from one interim IR).


mparrett:
Reviewed PR #387. I found no actionable issues. I checked the IR pipeline registry changes, ir.coverage extraction, runtime namespace loading, and generated artifact consistency. Verification passed: go test ./pkg/ir -run TestCoverageClassifierBuckets, go test ./pkg/ir -run TestLoweringsRegistryExists, make check-generated, and go test ./pkg/ir ./pkg/genmanifest.

nooga:
This looks good but there's lint. Please fix before merging.

nooga (approved):

PR #390, opened 2026-07-05
go.mod: lower toolchain floor to go 1.24

mparrett:
Stacked on #380 — merge that first. #380 removed the only go1.25-only *source*
API (`sync.WaitGroup.Go`); this removes the remaining blocker, the dependency go
directives, so the tree builds on a go1.24.x toolchain (e.g. go1.24.4) without
downloading a newer Go.

## What changed

Three deps declared `go 1.25.0`, which pinned the module there via `go mod tidy`.
Rolled each back to the newest version declaring `go <= 1.24`:

| dep | before | after | declares |
|---|---|---|---|
| `x/sys` | v0.44.0 | v0.40.0 | go 1.24.0 |
| `x/term` | v0.41.0 | v0.36.0 | go 1.24.0 |
| `x/perf` | @20260512 | @20250305 | go 1.23.0 |

`x/perf` is used only by `cmd/bench-ratchet` (`benchfmt`) and `cmd/perf-page`
(`benchunit`); both APIs are unchanged across the rollback. The go directive is
now `go 1.24.0`.

## Verification

On go1.26.3 with `-lang go1.24`: `go build ./...`, `GOOS=js` wasm, linux + plan9
root+pkg, `go vet ./...`, and the full `go test ./...` (including bench-ratchet,
perf-page, and e2e) all pass.

## Forward compatibility

The `go` directive is a floor, not a ceiling — newer toolchains still build the
tree with no downgrade (verified on go1.26.3 with default `GOTOOLCHAIN`, no
toolchain download). The only effect is that go.mod now gates the *language
version* at 1.24, so adopting a go1.25+ language feature later means bumping the
directive back up.

The dependency rollback is inert for let-go's used surface, checked at the file
level (not just by the passing build): the terminal path via x/term
(`MakeRaw`/`Restore`/`GetSize`/`IsTerminal`) and the direct `x/sys/unix` usage on
Linux (the `SYS_*` mount/namespace constants in `pkg/rt/syscall_linux.go`) both
resolve to code that's unchanged across the ranges. Per-dependency breakdown in
a comment below.

## Why 1.24, not lower

1.24 matches the motivating toolchain (go1.24.4) with the smallest dependency
rollback. It isn't the source floor: range-over-func (`iter.Seq`, a go1.23
feature) is used in `cmd/lginterop`, `examples/`, and `pkg/{api,rt,vm}`, so
**1.23** is the lowest reachable without rewriting iterator code. Going to 1.23
rolls `x/sys`/`x/term` back further for toolchains outside the motivation — happy
to target 1.23 instead if you'd rather have the wider range.


mparrett:
Breakdown of what each rollback drops, mapped to what let-go imports.

**`x/term` v0.41.0 → v0.36.0** — 9 commits, 6 of them `go.mod` bumps plus the
`[generated]` go-directive raise to 1.25.0 (the raise is what pinned the floor).
The 3 functional commits (`handle transpose`, `handle delete key`, comment fixes)
touch only `terminal.go`, the `term.Terminal` line editor, which let-go doesn't
import (it uses chzyer/readline). `MakeRaw`/`Restore`/`GetSize`/`IsTerminal` are
byte-identical across the range.

**`x/perf` @20260512 → @20250305** — almost entirely `go.mod` bumps and
go-directive raises. In `benchfmt` and `benchunit` (the only packages let-go
imports, from `cmd/bench-ratchet` and `cmd/perf-page`) the sole change is a
comment fix.

**`x/sys` v0.44.0 → v0.40.0** — 21 commits. let-go imports `x/sys/unix` directly
on Linux (`SYS_*` mount/namespace constants in `pkg/rt/syscall_linux.go`) and
transitively via x/term (termios/winsize ioctls), so this one got a closer look:

- 7 `windows:` and 6 `cpu:` commits: let-go imports neither package (`x/sys/cpu`
  import count is 0; CPU feature detection goes through klauspost/cpuid).
- `unix/` changes are `IoctlSetString` (new), `CPUSetDynamic` (new), netlink
  `Prefixmsg`/`PrefixCacheinfo` structs (new), plus a `Utime` nil-deref fix and a
  `Recvmsg` error-path fix.
- The Linux syscall-number tables (`zsysnum_linux_*.go`) that let-go reads are
  not among the changed files, so the `SYS_*` constants are identical in v0.40.0.
  Same for the termios path via x/term.

The one real regression surface: the `Utime` and `Recvmsg` fixes exist only above
v0.40.0, but in syscalls let-go doesn't call. If that changes later, bumping
x/sys picks them up — the floor only constrains the go directive x/sys declares,
not its version.

mparrett:
Closing this as too narrow to be worth carrying.

On a fresh look, the build failure that motivated the stack isn't a let-go
problem: with `GOTOOLCHAIN=auto` (the default since 1.21), Go transparently
downloads the toolchain go.mod asks for, so `go 1.26` builds fine anywhere with
network access. It only fails where the toolchain fetch is blocked —
egress-restricted CI, air-gapped, or sandboxed agent environments that can't
reach `proxy.golang.org`. Anyone hitting that can install a go1.26+ toolchain
or allow the proxy; it doesn't warrant pinning `x/sys`/`x/term`/`x/perf` back and
gating the language version at 1.24 for everyone.

The dependency changelog above stays here for reference if the question comes up
again. #380 (the `WaitGroup.Go` rewrite) stands on its own as a small cleanup and
stays open.

PR #391, opened 2026-07-05
async: core.async parity — buffer policies, closed-channel puts, thread, alts!!

nnunley:
Brings the async namespace to parity with clojure.core.async's documented semantics.

## New surface
- `buffer` / `dropping-buffer` / `sliding-buffer` + `(chan buf)`: dropping completes puts and discards the NEW value when full; sliding completes puts and evicts the OLDEST. `vm.Chan` stays a raw Go channel — the policy lives in a side table consulted by every put path (`>!`/`>!!` in lang.go, `offer!` in async.go); entries are cleaned up on `close!`.
- `thread` macro (async.lg): go under the core.async-parity name — every let-go go block is a real goroutine, so blocking ops are safe in both.
- `alts!!` registered (alias of `alts!`; all let-go channel ops block the calling goroutine).

## Parity fixes
- `>!` / `>!!` on a closed channel now return `false` (previously: panic — send on closed Go channel).
- `offer!` returns `nil` when the put can't complete (previously `false`), including on closed channels — matches the documented `true true nil` example.
- `close!` is a no-op on an already-closed channel (previously: panic).

## Hybrid-namespace fix (compiler/eval.go)
async is the first namespace that is both natively registered and carries bundled lg source. The bundle decoder interned its lg-defined vars as nil stubs into the already-live native ns, and qualified symbol resolution never triggers the lazy chunk loader — so `async/thread` was a nil stub unless something ran `(require 'async)` first. Bundle chunks for namespaces that pre-existed the decode (i.e. native-backed) now execute eagerly at bundle load.

## Tests
`pkg/rt/async_buffers_test.go` — each test cites the core.async documented behavior it encodes (drop-newest, evict-oldest, closed-put false, offer nil, close-twice no-op, puts-never-park). Full `pkg/rt` + `pkg/compiler`: 185 pass.


mparrett:
Status check on #194:

- Issue #194 is still open.
- `alts! :default` landed in #250 and is present on `main`.
- `offer!` and `poll!` are also present on `main` and registered in `async`; there is existing async namespace coverage for `offer!` + `poll!`.
- The remaining gap, reading #194 strictly, is the empty-value contract for `poll!`. Current `poll!` returns `nil` both when nothing is immediately ready and when the channel yields/closes with `nil`, so it still does not provide the requested distinct empty signal / `[val ok]`-style disambiguation.

So I would not treat #194 as fully closed by this PR unless this PR also changes `poll!` to return an unambiguous result shape or we explicitly decide that nil-on-empty is the accepted let-go behavior.

mparrett:
@nnunley Codex thoughts for consideration...

Follow-up on #194 / `poll!` semantics after thinking through the single-threaded wasm case:

I still think the clean fix is to make `poll!` return a shaped result instead of a bare value:

```clojure
(async/poll! ch)
;; nil           => nothing ready right now; returned immediately
;; [v true]      => value was taken, including v = nil
;; [nil false]   => channel is closed
```

That maps directly to Go's non-blocking receive:

```go
select {
case v, ok := <-ch:
    return [v ok]
default:
    return nil
}
```

This satisfies the strict ask in #194 because a delivered `nil` is distinguishable from "nothing ready". It also preserves closed-channel information.

The wasm-specific constraint reinforces this shape: the `nil` return should mean "idle frame, do other work now", and any vector should mean "a channel state transition happened without blocking". The implementation should stay as a true non-blocking `select default`: no ctx receive, no timer, no helper path that can park even briefly. `poll!` does not need cancellation awareness because it never blocks.

For the multi-channel case, `alts! :default` should keep the same rule: include `reflect.SelectDefault` and do not add the ctx `Done` case on the default path. Blocking `alts!` can remain ctx-aware.

This is a semantic change from today's `poll!`, which returns the raw value or `nil`, but `poll!` is still new enough that I would fix the contract now rather than preserve an ambiguous API. If we need compatibility, add a separate raw-value helper; otherwise make `poll!` the precise `[value ok]` API requested by #194.

mparrett:
This should be an easy/surgical follow-up if you want it in this PR.

The code change is just `poll!`’s receive branch returning a vector result instead of the raw value:

```go
select {
case v, ok := <-ch:
    return vm.NewArrayVector([]vm.Value{v, vm.Boolean(ok)}), nil

default:
    return vm.NIL, nil
}
```

Then update the existing expectations:

- value ready: `:val` / `vm.Int(1)` becomes `[:val true]` / `[1 true]`
- closed channel: `[nil false]`
- nothing ready: still `nil`
- optional extra test: enqueue a real `nil` value and assert `(poll! ch)` returns `[nil true]`

Happy to push a small patch for that if useful.

nooga (approved):

nooga:
Approved for merging, with or without the poll! fix.

PR #392, opened 2026-07-05
glplat: minimal GL platform layer (experiment) — GLFW/OpenGL backend, font primitives, screenshot

nnunley:
Experimental platform layer backing xsofy's GL frontend spike (see the companion xsofy `gl-frontend` branch). Draft — for visibility and review of the seams, not necessarily for merging as-is.

## What's here
- `pkg/glplat`: pure-Go public API delegating through an interface/registry to `internal/native` (cgo: GLFW 3.3 + OpenGL 4.1 core). Contract: 9-float vertices (x y z u v r g b a), `SubmitTriangles` draws immediately with the current column-major MVP, texture 0 = untextured white.
- Font primitives (`FontLoad` ttf/otf/ttc, `FontHasGlyph`, `FontRasterizeCell`, `SaveGlyphAtlasPNG`) — minimal rasterization surface so atlas-baking policy lives in lg code (xsofy `tools/bake_atlas.lg`), per 'as minimal as possible in Go'. Replaces the earlier `cmd/lgatlas` CLI (deleted).
- `Screenshot(path)`: glReadPixels of the back buffer → PNG, for headless frame verification.
- Hand-written rt bindings in `pkg/rt/interop_glplat.go` (lginterop style; the source importer can't typecheck cgo packages).

## Notes
- Fixes found while standing up the frontend: MVP uploaded without transpose (column-major contract) and viewport set from the framebuffer size each frame (Retina).
- Browser backend (WebGL2 via syscall/js) is designed but not built; the registry seam is where it lands.
- Depends on nothing in #391, but xsofy's branch runs on a build containing both.

---
**Part of #259** (Epic: Host integration & interop) — realizes the graphics host-capability seam scoped in **#255**.


nooga (changes requested):
This is a solid spike and the seams (registry/interface, pure-Go API vs internal/native) look right. Before this lands non-draft, the cgo/GLFW dependency needs to be behind a build tag (e.g. `//go:build glplat` on `internal/native` and a stub/no-op registration otherwise) so `go build ./...` and the wasm target don't require GLFW/OpenGL/cgo on every machine and CI runner. Right now `pkg/rt` unconditionally pulls in the native backend, which would break headless CI and the existing `GOOS=js` wasm build.

nnunley:
Rebased onto current main (`8c31a809`) to refresh CI — the prior failures ran against a stale base. Locally on this branch: `pkg/glplat` builds (GL/GLFW cgo), the non-glplat build is clean, and `make check-generated` is green (so the generated-artifacts / gogen-diff jobs should pass now). Still a draft/experiment — just refreshing CI.

nnunley:
Build-tag gate implemented in 2382162: `//go:build glplat` on `internal/native` and on the blank-import shim (`pkg/glplat/init.go`). Without the tag no backend registers and the pure-Go API returns its existing `no backend registered` errors; `pkg/rt/interop_glplat.go` only touches that pure API, so nothing in the default build needs cgo.

Verified all three configurations locally:
- `go build ./...` (no tag) — succeeds, no GLFW/cgo required
- `GOOS=js GOARCH=wasm go build ./pkg/...` — succeeds
- `go build -tags glplat ./pkg/glplat/...` — succeeds with the native backend compiled in

Staying draft for now — it's still a spike — but the unconditional-cgo blocker is gone.

nnunley:
Rebased onto current main (`a1a3866e`) and CI is now fully green on the build-tag gate commit — lint, build, gogen-diff, and generated-artifacts all pass without GLFW/OpenGL/cgo, and the `GOOS=js` wasm build is clean.

mparrett (commented):
Looked at the tip (`489d5c62`) with the build-tag gate in. The seams (registry, pure-Go API, `-tags glplat`) look right, and the unconditional-cgo blocker is gone. A few concrete issues to fix before this leaves draft:

### P1: `LoadTextureRGBA` buffer length

`LoadTextureRGBA` passes `w`, `h`, and `gl.Ptr(pixels)` to `gl.TexImage2D` with no `len(pixels) >= w*h*4` check (`pkg/glplat/internal/native/native.go`). A short slice (easy from the lg seq→byte unbox) is a C-side over-read. Reject up front in the public API and/or the native backend.

### P2: texture Y orientation

`LoadTextureFile` / `LoadTextureRGBA` upload the first row first. OpenGL treats that as the texture bottom, while the spike UVs put `v=0` at the top of the quad. Atlases from `SaveGlyphAtlasPNG` / `FontRasterizeCell` are top-down, so glyphs render upside-down unless something else compensates. Flip on upload (or invert V in the contract) and say which convention the API owns.

### P2: `WindowSize` frozen at `Init`

`BeginFrame` refreshes the viewport from the framebuffer each frame (good for Retina), but `WindowSize()` returns the dimensions cached in `Init`. After a resize, ortho/UI math that trusts `WindowSize` drifts from the viewport. Read `GetWindowSize()` (and/or a size callback).

### P2: `Terminate` leaves a live window pointer

`window.Destroy()` runs but `b.window` is not nil'd, and the texture map still holds deleted GL IDs. Later `ShouldClose` / `BeginFrame` / `SubmitTriangles` is use-after-destroy. Nil the window, clear maps/IDs, and decide whether re-`Init` is supported.

### P2: font registry locking vs `sfnt.Buffer`

`FontHasGlyph` / `FontRasterizeCell` unlock, then call `GlyphIndex` on the shared `entry.buf`. Concurrent has-glyph/rasterize (or overlap with `getFace` mutating `faces`) races. Hold the mutex across buffer use, or give each call its own `sfnt.Buffer`.

### P2: `getFace` swallows `NewFace` errors

Failed face creation stores nil; later `DrawString` panics. Easy to hit when width-fit scales `int(newSize)` to `0`. Propagate the error and reject non-positive sizes.

### Smaller / residual

- No Go or lg tests for the contracts that matter (vertex stride, MVP column-major layout, screenshot-before-swap, font cell fit).
- Spike examples don't say that `lg` needs `-tags glplat`.
- Depth-test + alpha blending will bite overlapping translucent glyphs.
- Unknown texture IDs fall through to GL texture `0` instead of the white fallback.
- Hand-written `pkg/rt/interop_glplat.go` is labeled "Code generated by `lginterop` / DO NOT EDIT"; fine that it's hand-rolled (cgo), but the header will mislead the next editor.

Fonts usable without GLFW is the right cut of the surface. If this starts competing with a `surface`-style seam, spell out how it relates to #255's host-owned graphics model.

mparrett:
We rebased this branch onto current `main` to build an `lg` we could measure the xsofy GL frontend against, and three things came out of it that seem useful here.

## The rebase is nearly free

The branch was 42 commits behind `origin/main` when we took it on 2026-07-24. The entire rebase produced exactly one conflict, in `.gitignore`, resolved by keeping both sides. Everything else applied untouched. Whenever this comes off draft, that part is cheap.

## Two notes from driving the input queue

Both only really show up from the client side, so they are easy to miss from inside the package.

**One physical keypress arrives as two events.** `keyCallback` appends `key:<name>` and `charCallback` appends `char:<name>` to the same queue, so a client dispatching on both acts twice per press. We settled on consuming `char:` and ignoring `key:` for printable keys, since `char:` is the one that carries shift state. That works fine once you know it — mostly flagging it in case it is worth a note in the docs.

**`keyCallback` takes GLFW's `mods` and does not forward it.** With no modifier information in the emitted string, a client cannot tell Shift+Left from Left, which is what put shift-to-run on arrow keys out of reach for us from the game side. Encoding modifiers into the event string, or emitting them alongside, would open that up whenever it matters.

## A WebGL backend looks closer than expected

`registry.Backend` plus `SetBackend` already is the seam, and the build tag comment already names `GOOS=js` as a target that must not need cgo. Reading it as a porting surface:

- The lg-facing API is 12 distinct functions, with a single runtime
caller in our frontend. A conforming backend means the game code changes zero lines.
- The shaders are `#version 410 core`; GLES 3.00 wants `#version 300 es`
and precision qualifiers. `layout(location = ...)` on attributes is valid there, so the port is mechanical. It is textured coloured quads and two shaders.
- Vertices already cross as a flat `[]float64`, which is a direct
`Float32Array` conversion, and there are only four to six `SubmitTriangles` calls per frame — few calls with large buffers, which is the shape that survives a wasm-to-JS boundary.
- `LoadTextureRGBA(pixels, w, h)` is already on the interface, so the
synchronous `LoadTextureFile` problem has an escape hatch: preload the atlas and hand over bytes.

The hard part is the loop. A frame loop that never returns starves the browser's event loop, and this one is built to block, with `EndFrame` doing the pacing. The way out that preserves the design is a Web Worker with an `OffscreenCanvas`, where blocking is allowed; the alternative is inverting the loop to be callback-driven, which pushes the change into game code and breaks the symmetry with the terminal frontend. A Worker backend would also have to supply the pacing itself, since there is no blocking swap to lean on.

We have not built this, so treat the shape as an assessment rather than a result.

## One question before the seam sets

Should a browser backend live behind this package's own registry, or behind the surface capability from #255 as reshaped in #572? Those are two different extension points for the same kind of thing, and picking one is cheaper now than after both have implementations. That reads like a maintainer call rather than ours.

The two input findings stand on their own and are worth folding in whenever this branch next moves. The WebGL sketch can wait on that question — happy to work it up in detail, or to drop it if the answer makes it moot.

mparrett:
We built a second `registry.Backend` implementation, on Ebitengine, to see how the seam holds up — partly to chase the browser question from our earlier comment with code instead of assessment. Results below; the branch is `glplat-ebiten`: this PR's commits rebased onto current main (the same one-conflict rebase as our earlier note), plus two of ours.

## Conformance

The xsofy GL frontend runs on the ebiten backend with zero game-code changes:

- `XSOFY_GL_DUMP` output is byte-identical to the GL backend, and deterministic across runs.
- Screenshot pixel diff is 0.3% (glyph edge rounding).
- CPU and memory over a 10s run are at parity; peak RSS is flat from 30s to 60s on both backends, with ebiten carrying a constant ~11MB overhead.

## The loop inversion

Ebiten owns the main thread (`RunGame`), which inverts this package's caller-owned loop. The backend handles it internally: `main` routes through a boot trampoline, the lg program runs on a goroutine, and `EndFrame` hands each frame's batches to ebiten's `Update` over an unbuffered channel — the rendezvous doubles as frame pacing. Game code and the glplat API are untouched.

## Browser

`GOOS=js GOARCH=wasm go build -tags glplat_ebiten` links (29MB, versus 26MB untagged), and a probe page running an lg draw loop renders correctly in Chromium with a clean exit.

The part we expected to be hard was not. Our earlier comment flagged the blocking frame loop as the obstacle to a browser backend, with a Worker plus OffscreenCanvas as the likely workaround. Neither is needed: the lg goroutine blocks in `EndFrame` on a channel, the Go wasm scheduler parks it, and ebiten's requestAnimationFrame loop never starves. The blocking `EndFrame` contract survives wasm untouched, which reads as evidence for the current design.

## Interface findings

Two places where the interface makes a promise the docs don't state, and one open question this gives evidence for:

1. **Occlusion is an undocumented contract.** Vertices carry z and the GL backend depth-tests, so "nearer hides farther" is a de facto promise. Ebiten has no depth buffer; our backend currently draws in submission order, which is exact for the top-down frontend (coplanar under ortho) but mis-sorts first-person geometry. We plan a CPU back-to-front sort keyed off the per-batch matrix (perspective batches sort, ortho batches keep draw order) — but whichever way it lands, a sentence on `registry.Backend` saying whether z occludes would tell the next backend author what they are signing up for.
2. **Input events.** The mods gap and the key:/char: duplication from our earlier comment are cheaper to fix now than after more consumers arrive: any event-format change should move both backends together.
3. **The seam question.** Whether a browser backend belongs behind this registry or behind the #572 surface capability now has data: the registry route is implemented and conformant. Still your and nooga's call, but deciding it while both seams are young stays cheaper than deciding it later.

## Offer

Our default: fold the two commits into this branch — they are stacked on it already and rebase clean. If a PR against your fork's `glplat` branch is the easiest thing to merge from, say so and we'll open one; keeping them as a follow-up PR once this lands works too. The draft PR is #744, for CI and reference in the meantime, marked review-the-tip-only.

mparrett:
To save you a pass when you pick this back up: #744 stacks on this branch but doesn't modify it, so all six findings from the 2026-07-24 review still apply at `489d5c62` — nothing got fixed as a side effect of the ebiten work.

A few notes from that spike that may be useful when you do:

- **Y orientation got an empirical data point.** The conformance run rendered atlas-textured quads through both backends and they agreed to within glyph-edge rounding (~0.3% of pixels), so something in the xsofy frontend evidently compensates for GL's bottom-up upload — the "unless something else compensates" case from the review. The convention still wants writing down, and it now matters more as a two-backend consistency guarantee.
- The ebiten backend independently handles three of the analogous risks, so it may be useful as reference: it rejects short pixel buffers in `LoadTextureRGBA`, falls back to a white 1×1 texture for unknown IDs, and `Terminate` blocks until the run loop unwinds instead of leaving live pointers.
- The two `font.go` findings (mutex released across `sfnt.Buffer` use, `getFace` storing nil on error) are the ones I'd take first: that code is shared by every backend, so both apply to #744 builds too — the spike examples just don't call the font natives.

If you'd rather review than write them, I can send the font fixes (or any of the others) as PRs against your branch.

nooga:
Status check: still parked as a draft. The P1 at `pkg/glplat/internal/native/native.go:260` (no `len(pixels) >= w*h*4` check before `TexImage2D`) needs fixing before this leaves draft, and none of mparrett's six findings from 07-24 have been touched since the last push. Also noting `pkg/glplat/font.go` has no build tag, so `x/image` and `x/text` end up in every default binary. mparrett offered to send the font fixes as PRs against this branch — nnunley, want to take that up, or should this wait until you're back on it?

mparrett:
@nnunley the offer stands, whenever suits you. The one I would send first is the `pkg/glplat/font.go` build tag nooga notes above: without it `x/image` and `x/text` land in every default binary, tagged or not, which is the same class of problem the backend tags already solve elsewhere in the package. I can send that plus the rest of my 07-24 findings as a PR against this branch rather than as review comments, so you can take or drop them individually.

No urgency from my side beyond #744 sitting behind this one. If you would rather hold it all until you are back on the branch, say so and I will leave it alone.

mparrett:
Sent as nnunley/let-go#5 against `glplat`, two commits to take or drop.

**The two font.go findings.** The glyph-lookup race and `getFace`'s swallowed error. One thing worth recording: the call-local `sfnt.Buffer`, the lighter of the two options in my 07-24 note, did not fix the race on its own. The shared `font.Face` and its `*sfnt.Font` are the larger hazard, and the concurrent test still panicked inside `sfnt.LoadGlyph` until glyph work was serialized per font. These are the package's first tests, and they run cgo-free.

**The P1 texture over-read.** Guarded in the public API before any backend sees the slice, and again in the native backend.

I left the `font.go` build tag out on purpose. It is a design call, since `font.go` is the headless-font cut: `interop_glplat.go` is untagged and pulls `x/image` into every default binary through it. The PR body lays out three shapes. My preference is a separate `glplat_fonts` tag if headless fonts stay on the roadmap, or gating the whole native surface if it should all be opt-in. Your pick and I will send it.

The GL-backend state findings, WindowSize frozen at Init and Terminate's live window pointer, want a running context to verify behavior, so those come after.

nnunley:
Architecture decision (2026-09-07):

- This work should use the merged #572 `surface` host-capability model rather than making the `glplat` registry the supported public extension seam. Re-scope the native backend around that boundary.
- Headless font primitives and their dependency should be separately opt-in behind a `glplat_fonts` build tag. They should not remain in default binaries or force the whole graphics surface to be enabled.
- A backend-independent contract must be written and agreed before this lands. It must cover at least texture/frame orientation and depth/occlusion semantics, resize behavior, input events/modifiers, and terminate/re-init lifecycle.

This is direction, not clearance. The reported safety/API defects—including texture-buffer over-read, stale `WindowSize`, post-`Terminate` live state, font concurrency, and swallowed font-face errors—remain unresolved blockers. The fixes in nnunley/let-go#5 may be useful input after re-scoping, but they are not sufficient to land this PR.

Therefore #392 should remain unmerged until it is re-scoped to #572, the contract exists, the safety findings are fixed, and current-base CI/review pass.

nooga (changes requested):
About half of the earlier review findings were fixed in the September pass (the P1 short-buffer texture overread and the P2 font-registry locking/`getFace` error-swallowing) — confirmed by reading the diff. But `make check-default-deps` currently fails: `pkg/glplat/font.go` (and the untagged `pkg/rt/interop_glplat.go` that imports it) pull `golang.org/x/image/font/sfnt`, `opentype`, `vector`, and `x/text/encoding/*` into the *default untagged build* and into `./cmd/lg-runtime` — the same class of problem the first review flagged for cgo/GLFW, now surfacing for the font deps instead. `make check-generated-manifest` also fails (stale `core_compiled.lgb`/`core_go_lowered`).

Requesting changes:
1. Gate `pkg/glplat/font.go`'s `x/image`/`x/text` usage (and/or the `pkg/rt/interop_glplat.go` registration) so font-related deps don't leak into the default untagged build/`cmd/lg-runtime`, mirroring how `internal/native` is already gated.
2. Run `make generate` and commit the refreshed manifest/lowered-core artifacts.
3. Still open from the earlier review: texture Y-orientation (glyphs may render upside-down), `WindowSize()` frozen at `Init` instead of tracking resizes, `Terminate` leaving a dangling window pointer/stale texture IDs, and no tests for vertex stride/MVP layout/screenshot-before-swap.

This looks complementary to #744 (same registry-backend seam, this PR's own tests already anticipate a `-tags glplat_ebiten` sibling), not competing — makes sense to land #744 first and rebase this on top once CI is green here.

mparrett:
For future reviews, making a note here about this specific comment with open questions on #744 : https://github.com/nooga/let-go/pull/744#issuecomment-5702994616

nnunley:
Closing: this PR's code is already in `main`. #744 was stacked on it and carried its commits in when it merged (`7148f4ad`), so there is nothing left to land.

Verified per file rather than by assuming. All 15 files this PR touches exist in `main`, and diffing each one against the branch, the branch has **zero** functional lines `main` lacks. The only source difference is the header comment in `pkg/glplat/glplat.go`, which `main` rewrote to describe all three build tags and the provisional status — strictly better than the version here. The `go.mod`/`go.sum` differences are this branch being stale: it still declares `go 1.26`, where `main` is on 1.27 after #910.

@nooga — your 09-16 review ended with *"makes sense to land #744 first and rebase this on top once CI is green here."* That was right when it was written, but #744's squash took this PR's commits with it, so a rebase now produces an empty diff. Hence closing rather than rebasing.

Two of your three asks are satisfied in `main`: `pkg/glplat/font.go` is behind `//go:build glplat_fonts` with a `font_stub.go` fallback, so the `x/image` / `x/text` deps no longer reach the default untagged build, and the generated artifacts are current.

Your item 3 is **not** resolved, and since the code is now in `main` those findings are live defects with no tracker once this PR closes. They are filed as **#931**, each re-checked against `main` rather than copied forward:

- `WindowSize()` still frozen at `Init` — `windowWidth` is assigned once and never updated on resize.
- `Terminate` still leaves `b.window` non-nil after `Destroy()` and never clears `b.textures`.
- Texture upload still has no vertical flip while `Screenshot` does, with the comment *"GL rows are bottom-up"* — one direction of the same convention mismatch handled, the other not. That is the likely source of the upside-down glyphs, though confirming it needs a GL context.
- Still one test in `glplat_test.go`; nothing pins vertex stride, MVP layout, or screenshot-before-swap — and there are now two backends implementing that contract.

#931 also records the WebGL2 backend as designed-but-unbuilt, so that intent does not disappear with this PR.

PR #394, opened 2026-07-05
feat(ir): combinator inlining + closure devirtualization (opt-in via *enable-inline*)

nnunley:
## What

Adds combinator inlining and higher-order **closure devirtualization** to the Go-lowering pipeline, entirely **opt-in** behind `*enable-inline*` (default `false`). With the flag off, the core lowered tree is byte-unchanged and the full gate stays green.

This wires up the previously-built-but-unwired inline pass (the plan's step **B7**) and then adds the closure-devirt lever it enables — collapsing PEG/combinator-style dispatch into straight-line code.

## Pieces

- **B7 wiring** — `ir.passes.inline` gains `*inline-registry*` + `*enable-inline*` dynamic vars. `optimize-fn` runs `inline-with` when both are bound; `lower-ns-to-go` seeds a same-ns `defn` registry (`build-inline-registry`). Named same-ns calls splice inline.
- **Closure devirtualization** — a call whose callee is a locally-constructed closure literal is spliced inline instead of dispatched via `rt.InvokeValueEC`:
  - `splice-closure!` — single-block closures.
  - `splice-closure-general!` — **multi-block, general (non-tail) position**: splits the caller block into a continuation that receives the closure result as a block-param, clones the closure's blocks in (`:load-arg`→operands, `:load-closed`→captures), and rewires the closure's returns to branch into the continuation.
  - Reuses `lower-go/closure-info` (made public) to resolve the closure + captures.
- **New IR primitives** (`ir.data`, interned into `ir`): `move-inst!` (re-home an inst preserving its id/refs), `reparent-terminator!`, `replace-block-pred!`.
- Bundles the **lambda-lift** pass (capture-free closure hoisting) this builds on.

## Tests

New coverage in `pkg/ir`:
- `TestOptimizeFnInlinesWhenRegistryBound`, `TestLowerNsInlinesSameNsCall`, `TestLowerNsInlinesTypedSibling` — the B7 wiring + registry.
- `TestDevirtualizesLocalClosureCall` — single-block closure splice (`(fn [y] (+ y x)) 10` → native add, no `InvokeValueEC`).
- `TestDevirtualizesMultiBlockClosureCall` — capturing, multi-block, non-tail closure devirtualizes fully.

`make generate` + `make check-generated` green (bundle + lowered tree in lockstep, compiles + dispatches natively); full `go test ./...` passes.

## Not included / follow-up

Full PEG end-to-end collapse needs one more layer: **interprocedural devirt into nested closures** (combinator calls that live inside a returned closure, where the combinator is a *capture*). That requires recursing inline into nested fn-templates and threading a capture-environment — a distinct follow-up.

### DIFF
--- a/pkg/ir/lisp_inline_test.go
+++ b/pkg/ir/lisp_inline_test.go
@@ -7,8 +7,13 @@
 package ir_test
 
 import (
+	"fmt"
 	"strings"
 	"testing"
+
+	"github.com/nooga/let-go/pkg/compiler"
+	"github.com/nooga/let-go/pkg/rt"
+	"github.com/nooga/let-go/pkg/vm"
 )
 
 func TestInlineEligibilityUsesMutabilityAndSize(t *testing.T) {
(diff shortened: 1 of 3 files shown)


nnunley:
Rebased onto current main and fixed the IR-lowering regression the coverage gate was flagging.

**Root cause:** when a form lambda-lifts into a main fn + lifted siblings, lowering returns `{:kind :multi-fn-template :fns […]}` — a *successful* lowering, but with no `:status`. The internal flatten path (`lower-ns-to-go`) keys off `:kind`, but the public `compile-form` wrapper returned it raw, so callers that check `:status` (the gogen tests **and the lowering-coverage gate**) read a lowered form as unlowered (`:status nil`). This is why `gogen_binding`/`gogen_percent_identifier` failed — the `binding` and reused-`%` forms were lowering fine, just via the status-less template shape.

**Fix:** `compile-form` now surfaces a `:multi-fn-template` result as `:status :lowered`. The internal flatten path is untouched (it uses `compile-form*` directly).

Also folded in the mechanical merge with #371: `:private?` marking threads through the new `built`→lambda-lift→lower path so exported names stay PascalCase, and the inline/lambda-lift/direct-call test assertions were updated for #371's exported naming.

Verified locally: `make check-generated` OK; `pkg/ir`, `test/` (incl. both gogen tests), and `test/e2e` all green.

nooga:
Nice work getting B7 wired up and pushing through to closure devirt — the multi-block splice logic (block split + CFG clone + validate-fn! after) is careful and well tested.

One thing needs fixing before we turn this on for real: `self-recursive?` in `inline.lg` compares `(ir/aux nid f)` directly to a symbol, but for a normal `defn` built via `build.lg`, a `:load-var`'s aux is a `Var` object, not a symbol — `call-head-var` right above it even documents this and normalizes with `(try (symbol aux) (catch e nil))`. `self-recursive?` skips that step, so it'll basically never return true in the common case, and `inline-eligible?` will happily mark a self-recursive fn as inlineable. `max-inline-rounds` bounds the round count, not the size, so a splice-into-itself can blow up IR size across those rounds instead of hard-looping.

This matters here specifically because the target use case is PEG/combinator code, which is exactly where self-recursive helpers show up. `TestInlineTerminatesOnMutualRecursion`'s own docstring admits it never constructs a real self-recursive callee — so this path is untested and currently broken.

Please fix the normalization (mirror `call-head-var`'s `(symbol ...)` conversion) and add a test that actually runs a self-recursive `defn` through `inline-eligible?`/`inline-with` before we flip `*enable-inline*` on anywhere. Everything else here looks mergeable — flag defaults off so this isn't blocking the build, just don't want to forget about it once we start turning this on for the combinator bench.

nooga (approved):

nnunley:
Addressed the `self-recursive?` feedback and rebased onto current main (was conflicting).

**`self-recursive?` fix** (`inline.lg`): it now normalizes the `:load-var` aux with `(symbol …)` before comparing to the fn name — mirroring `call-head-var` — so a normal `defn`'s `Var`-valued aux actually matches. The registry keys by the qualified `ns/name` while `fn-name` is unqualified, so the compare is by `(name …)`. Also narrowed the op check to `:load-var` (a `:call`/`:invoke`'s aux isn't the callee name).

Two real tests replace the one that admitted it never built a self-recursive callee:
- `TestInlineEligibilityRejectsSelfRecursive` — a genuine `(defn fact [n] … (fact (dec n)))` → `inline-eligible?` is `false` (fails on the old code).
- `TestInlineWithLeavesSelfRecursiveCallUnspliced` — end-to-end: `inline-with` leaves the call to a self-recursive helper intact.

**Heads-up on the rebase:** moving from the old base (#372) to current main (#405) re-measures the coverage baseline to 2374/20 (was 2347/11). The +9 are pre-existing #405-base lowering gaps surfaced by the newer corpus (incl. the `clojure-numeric-equality-test` `:stress/timeout`), not the inline code — `*enable-inline*` is still off so the lowered tree is unchanged. That NaN-fixpoint timeout is fixed in the follow-up strategy PR (#387, via `cp-const=`).

nnunley:
Rescoped after #387 merged: the combinator-inlining + closure-devirt work (and `cp-const=`) is already in main via #387, so this PR is now **just the `self-recursive?` fix** you asked for — the one review item that landed in main without it.

Diff is now `inline.lg` (the normalization fix) + the two real tests (`TestInlineEligibilityRejectsSelfRecursive`, `TestInlineWithLeavesSelfRecursiveCallUnspliced`). Rebased onto current main; `make check-generated` + full `pkg/ir` green.

PR #395, opened 2026-07-07
chore(perf): re-baseline ir-stress corpus after upstream drift (2347 -> 2366)

nnunley:
## What

Re-records `docs/perf/ir-stress-baseline.edn` for the current corpus: **2366 forms** (baseline said 2347, captured 2026-06-30).

## Why

`main` currently fails its own `make ir-stress-gate`: the corpus grew by 19 forms in a post-capture merge, and since the gate is deliberately not in CI (pre-push only), the baseline was never re-recorded. Every jj/pre-push user now hard-fails on:

```
GATE ERROR: corpus size changed — 2366 forms now vs 2347 baselined.
```

## Not a coverage regression

Measured at `main` (e2e3f9e2): total **2366**, passed **2355**, failed **11** — the failure count and buckets are identical to the baselined ones (`unresolved symbol n` ×7, `#inst` ×1, `read-json` ×1, `unsupported function body shape` ×1, `unresolved symbol a` ×1). Pure corpus growth.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #396, opened 2026-07-07
perf(vm): box-free KeywordLookup fast path (-4% native suite allocs)

nnunley:
> **Stacked on** #395 — one bookkeeping commit below this branch; it disappears from the diff once that merges.

## What

Adds a box-free fast path for keyword lookup — the #1 allocation site in the AOT-native (gogen_ir) profile.

`(:kw m)` previously went through `Keyword.Invoke`, which boxed the keyword hash lookup on every call (5.44M allocations in the jank-suite native profile — the single largest native allocator). This PR introduces a `KeywordLookup` interface with box-free implementations and routes `Keyword.Invoke` through it:

- `PersistentMap`: direct box-free hash lookup
- `Record`: field-index lookup without boxing (plus a `KeywordLookup` fast path used by lowered code)
- `Keyword.Invoke` prefers the fast path whenever the receiver implements it

## Measurements

| metric | before | after | Δ |
|---|---|---|---|
| jank suite `aot_native` allocs/op | 14,173,243 | 13,604,309 | **−4.0%** |
| `IRCompile` gogen_ir allocs/op | 226,367 | 209,554 | **−7.4%** |
| `BenchmarkKeywordInvoke` allocs/op | 1 | 0 | box eliminated |

`vm.Keyword.Invoke` disappears from the native allocation profile entirely. This closes ~10% of the native-vs-bytecode allocation gap on the jank suite.

A pointer-interning variant was also prototyped and measured as a NO-GO (marginal on top of this fast path); the measurement note is in the last commit.

## Testing

- `go test ./...` (default and `-tags gogen_ir` for pkg/vm, pkg/ir) green
- jank clojure-test-suite pass count unchanged
- `make check-generated` clean

### DIFF
--- a/docs/perf/ir-stress-baseline.edn
+++ b/docs/perf/ir-stress-baseline.edn
@@ -1,7 +1,7 @@
 {:failed 11
- :total 2347
- :passed 2336
- :captured "2026-06-30"
+ :total 2366
+ :passed 2355
+ :captured "2026-07-06"
  :how "LG_STRESS_PASSES=1 make ir-stress (corpus = every shipped/test/example .lg)"
  :note
  "ITER-0021 lowering-coverage ratchet. `make ir-stress-gate` fails when the
--- a/pkg/vm/keyword.go
+++ b/pkg/vm/keyword.go
@@ -66,6 +66,12 @@ func (l Keyword) Invoke(pargs []Value) (Value, error) {
 		}
 		return NIL, nil
 	}
+	if kl, ok := as.(KeywordLookup); ok {
+		if vl == 1 {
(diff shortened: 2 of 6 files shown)


mparrett:
Reviewed the changes in this PR and I don’t have any blocking findings.

I specifically checked the new `Keyword.Invoke` dispatch, the `PersistentMap` `findKeyword` path against the existing `find`/`valueEquiv` behavior, record keyword/default handling, and collision-node traversal. I also verified the branch locally at `617d8eb` with:

- `go test ./pkg/vm`
- `go test ./...`

nooga (approved):

PR #397, opened 2026-07-07
perf(vm): chunked ArrayVectorSeq (growing-from-1) + orderless maps + alloc-free symbol parsing (-34.8% suite allocs)

nnunley:
> **Stacked on #396** (keyword-lookup fast path) — the first 4 commits belong to that PR and will disappear from this diff once it merges.
>
> **Draft until** the companion clojure-test-suite PR (jank-lang/clojure-test-suite#927) merges and the submodule pointer is bumped: the orderless-map change below makes 4 order-dependent suite tests (`cons`, `cycle`, `mapcat`) need the same `:lg` reader-conditional branches Basilisp already has (`:lpy`).

## What

Three allocation levers on the seq/collection layer, measured together on the bytecode jank suite: **13,066,797 → 8,516,059 allocs/op (−34.8%)**.

### 1. Chunked `ArrayVectorSeq` (with CoW-embedded chunk + growing-from-1 windows)

`ArrayVectorSeq` now satisfies `IChunkedSeq`, so chunk-aware consumers (`reduce`, `map`, `appendSeqValues`, …) walk vectors via `Nth` windows instead of allocating a seq node per element.

Two refinements over plain 32-wide chunking:

- **CoW-embedded chunk**: the chunk struct is embedded in the seq and reused; `ChunkedFirst` overwrites it in place instead of allocating a fresh `ArrayChunk` per window.
- **Growing-from-1 chunk sizes** (1, 2, 4, 8, … clamped to the remaining length): the size-1 head chunk keeps `first`/`take 1` element-wise (realizes exactly 1 element — no over-realization for side-effecting lazy pipelines), while longer pulls amortize toward chunk-granular cost. Instrumented histogram over the suite: ArrayVectors average 1.42 elements (68.6% are size 1), so a fixed 32-quantum would over-realize the common case; geometric growth wastes at most the trailing partial window and the final chunk is clamped to the exact remainder.

Laziness contract (documented in `test/lazy_map_test.lg`): `take 1` realizes 1, `take 2` realizes 3 (1+2), matching the geometric schedule; Range/PersistentVector chunking is unchanged (32).

### 2. Orderless array-map construction

`NewArrayMap` now builds through the transient path instead of preserving literal insertion order. Clojure documents **no ordering guarantee for maps** — insertion order on small maps is an emergent artifact of `array-map` that silently vanishes past 8 entries when the type promotes to `hash-map`. Depending on it is the same trap Google's Abseil/SwissTable deliberately randomizes against. The jank clojure-test-suite already accommodates unordered implementations via reader conditionals (Basilisp `:lpy` takes the same branches).

### 3. Allocation-free symbol namespace parsing

`Symbol.NamespacedRaw()` splits `ns/name` with `IndexByte` + substring sharing — no boxing, no `strings.Split` slice. `Namespace.Lookup` and the `Name()`/`Namespace()` accessors route through it; the boxing `Namespaced()` API is kept for callers that need `Value`s.

| step | allocs/op |
|---|---|
| before series | 13,066,797 |
| after (all three levers) | 8,516,059 (**−34.8%**) |

## Testing

- `go test ./...` green (default + `-tags gogen_ir` pkg/vm, pkg/ir)
- .lg language suite (170 tests) green; chunked-laziness contract tests added (`test/lazy_map_test.lg`, `pkg/vm/array_vector_chunked_test.go`)
- Map-order-dependent assertions in `test/coll_test.lg`, `test/builtins_test.lg`, `test/for_test.lg` made order-agnostic
- jank suite pass count 5652 → 5656 with the companion suite PR applied


nooga:
Good work — the growing-from-1 chunking is well-calibrated (the corpus histogram justifying it over a fixed 32-quantum is a nice touch), and the `NamespacedRaw` split is a clean, behavior-preserving win. Two things I want addressed before this comes off draft:

**Dead `order` field.** `NewArrayMap` was the only call site that ever originated a non-nil `PersistentMap.order` (the other three sites in `Assoc`/`Dissoc` only propagate an existing one, never create it — checked with `grep -n "order:" pkg/vm/*.go`). Once `NewArrayMap` builds via transient, nothing sets `order` again, ever — the `if m.order != nil` branches in `Assoc`, `Dissoc`, and `Seq` become permanently dead. Please fold the field and its dependent branches out in this PR rather than leave a stale abstraction lying around.

**`ArrayVectorSeq.ChunkedFirst()` mutates shared node state.** It writes into an embedded `s.chunk` field and hands back `&s.chunk`, which is a new pattern — `PersistentVectorSeq.ChunkedFirst`/`Range.ChunkedFirst` both return a fresh `*ArrayChunk` on every call, never touching the seq node itself. That means an `*ArrayVectorSeq` node is no longer safely shareable across goroutines the way the rest of our persistent-collection API is, if two callers ever call `ChunkedFirst()` on the same node concurrently. Please run the suite under `-race` and either confirm it's clean or add an explicit "not safe to call ChunkedFirst concurrently on the same node" note on the type.

Also flagging, non-blocking: `t.Assoc`/`t.Persistent()` errors in `NewArrayMap` are silently discarded — given every map literal in every program flows through this, I want to be sure that's provably infallible rather than assumed. And obviously this stays draft until the companion clojure-test-suite PR lands per your own note.

Rest of the diff (the #396 pieces) I'm treating as already covered by that PR's review.

mparrett:
Focused correctness pass, keeping this to the PR's change surface:

1. **Blocking:** `ArrayVectorSeq.ChunkedFirst()` now mutates `s.chunk` and returns `&s.chunk`, so the same seq node is no longer read/share safe. I reproduced this with `go test -race` using two goroutines repeatedly calling `s.ChunkedFirst().Nth(0)` on the same `*ArrayVectorSeq`; the race reports concurrent writes at `pkg/vm/vector.go:252` and read/write through `pkg/vm/chunk.go:63`. The cleanest unblock is to make `ChunkedFirst()` return an independent chunk value again (or otherwise make the returned chunk immutable/thread-safe) and add that race probe as a regression test.

2. **Lower priority:** `applyJankMapOrderPatch()` applies directly into the checked-out `test/clojure-test-suite` submodule. Until the companion suite PR lands and the pointer is bumped, this can leave test runs with a dirty submodule. Applying to a temporary copy or reverting in `TestMain` cleanup would avoid test-induced worktree state.

I didn't see other change-surface correctness blockers in the growing chunk schedule or `NamespacedRaw` path. Focused package tests passed (`go test ./pkg/vm`, `./pkg/rt`, `./pkg/compiler ./pkg/bytecode`); the existing VM race suite passes, but it does not cover the shared-node `ChunkedFirst()` case above.

nnunley:
Rebased onto current `main` (the merged #395/#396 commits dropped out; bundle regenerated) and addressed the feedback:

**`ChunkedFirst()` shared-node race (@nooga, @mparrett)** — now returns a fresh `*ArrayChunk` on every call instead of writing an embedded `s.chunk` and handing back `&s.chunk`, so `*ArrayVectorSeq` stays read/share-safe like `PersistentVectorSeq`/`Range`. The embedded-chunk alloc trick is gone; I kept it off the hot path rather than pursue a construction-time populate, since that would tax element-wise `Next()` (far more common than chunk-walks). Added `TestArrayVectorSeqChunkedFirstRace` — 8 goroutines × 2000 `ChunkedFirst().Nth(0)` on one node; `go test -race ./pkg/vm` is clean.

**Dead `PersistentMap.order` field (@nooga)** — confirmed `NewArrayMap` was its only origin (grep repo-wide; `EmptyPersistentMap` and `TransientMap.Persistent()` never set it), so removed the field and the now-dead `if m.order != nil` branches in `Assoc`/`Dissoc`/`entries()`. Also fixed the stale "preserves insertion order" doc on `NewArrayMap`.

**Discarded transient errors (@nooga)** — `Assoc`/`Persistent` in `NewArrayMap` now fail loud with the infallibility proof in a comment: both error only via `ensureEditable()` (use-after-persistent / cross-goroutine), neither reachable for a fresh, single-owner, finalize-once transient. A non-nil error now panics with context instead of silently building a half map.

**Submodule left dirty (@mparrett)** — `TestMain` now reverts the jank map-order patch after `m.Run()`, so the vendored submodule worktree is left clean (verified). The companion `clojure-test-suite#927` has since been accepted upstream, which clears the draft gate; a follow-up will bump the submodule pin and delete the patch mechanism entirely (independent check shows the bump is ~140 files of suite drift, so it's its own PR — jank-stress lowering holds at 285/287, the 2 fallbacks being pre-existing `#inst`/`clojure.lang.LazySeq` gaps, not drift breakage).

bench-ratchet baseline regenerated (`-full`); ir-stress is unaffected by this PR (its gate stops on a 2369-vs-2366 corpus-size drift from other merged PRs, not lowering coverage).

nooga (approved):

PR #398, opened 2026-07-07
perf(rt): ArrayVector direct-index fast paths — eliminate the seq-protocol tax (-19.7% suite allocs)

nnunley:
> **Stacked on #397** (chunked ArrayVectorSeq + orderless maps) — earlier commits belong to the PRs below this one in the stack and will disappear from this diff as they merge.

## What

Eliminates the "seq-protocol tax" for eager Go-native consumers of `ArrayVector`: builtins that only need to walk a flat `[]Value` slice were allocating an `ArrayVectorSeq` (via `seqOf`) — and often per-element successors — just to iterate. Profiling showed `concat`/`appendSeqValues` alone accounted for **21.9% of all `alloc_objects`** on the jank suite.

The collections already expose O(1) indexing (`Indexed`), but nothing used it for iteration — there is no `IReduce`-style protocol. This PR adds direct-slice fast paths at the consumer level:

| builtin | before | after |
|---|---|---|
| `reduce` | `seqOf` + chunk walk | direct index loop, zero seq/chunk allocs |
| `first` / `second` | `Seq()` alloc + `First()` | `av[0]` / `av[1]`, zero allocs |
| `next` / `rest` | `Seq()` + `Next()` (two allocs) | new `ArrayVector.SeqFrom(i)` (one alloc) |
| `concat` | seq + chunk walk per arg | bulk `append(ret, av...)` + pre-sized result |

Second commit: `reduce`'s remaining chunked/tail loops allocated a fresh `[]vm.Value{acc, x}` per element (3.36M objects — the largest single site after the fast paths). Both loops now reuse a two-arg buffer, the same pattern `some` already uses.

`ArrayVector.SeqFrom(i)` is added because `ArrayVectorSeq`'s fields are unexported; it lets `pkg/rt` step into a vector at an index with a single allocation.

| step | allocs/op | Δ |
|---|---|---|
| before (top of stack PR) | 8,516,059 | — |
| + direct-index fast paths | 7,339,761 | **−13.8%** |
| + reduce fargs reuse | 6,841,735 | **−6.8%** |

ns/op improved ~4–5%. The fast paths live in shared `pkg/rt`, so both VM variants benefit — the ratchet baseline (final commit) records:

- suite `aot_native`: 14,155,188 → 11,240,805 allocs/op (**−20.6%**), ns/op −40.4%
- suite `ir_bytecode`: −20.6% allocs, −9.6% ns
- `IRCompile` gogen_ir: −16.7% allocs; bytecode: −23.7% allocs

Cumulative across the three stacked PRs: **13,066,797 → 6,841,735 allocs/op (−47.6%)** on the bytecode suite.

## Testing

- New behavior tests: `pkg/rt/reduce_arrayvector_test.go` (incl. `reduced` short-circuit), `pkg/rt/seqfastpath_test.go` (31 cases across first/second/next/rest/concat)
- `go test ./...` green; .lg language suite (170) green; jank suite pass/fail counts unchanged
- `make check-generated` clean; ratchet anchor within determinism gate

## Note

The ratchet `baseline.json` tightening in the final commit assumes the whole stack; if this PR is reordered, re-run `go run ./cmd/bench-ratchet update` instead of taking the file as-is.

## Also included: bench iteration-isolation fix

While validating, we root-caused a `fail 1.000` metric that appeared on every *counted* bench iteration (but never in single runs): vars are interned in the process-global ns registry, so `add_watch.cljc`'s var watches — including a throwing err-watch — survive suite re-runs in the same process and contaminate the next iteration's fresh `state`. `runSuiteOnce` now snapshots `rt.AllNSes()` and drops namespaces created during the iteration (new `rt.RemoveNS` helper), restoring the fresh-runtime contract the suite assumes. Verified: counted iterations now report 0 fail / 5657 pass, matching the warmup iteration.

### DIFF
--- a/pkg/rt/lang.go
+++ b/pkg/rt/lang.go
@@ -402,6 +402,19 @@ func AllNSes() map[string]*vm.Namespace {
 	return result
 }
 
+// RemoveNS drops a namespace from the registry so a subsequent require/def
+// materializes a FRESH Namespace (and fresh Vars). Test-isolation helper:
+// vars are interned process-globally, so anything attached to them (e.g.
+// watches added by a test) survives re-running a suite in the same process
+// unless its namespaces are removed between runs.
+func RemoveNS(name string) {
+	canonical := resolveNSAlias(name)
+	nsMu.Lock()
+	delete(nsRegistry, canonical)
+	delete(nsNeedsLoad, canonical)
+	nsMu.Unlock()
+}
+
 func FuzzyNamespacedSymbolLookup(currentNS *vm.Namespace, s vm.Symbol) []vm.Symbol {
 	sns := s.Namespace()
 	var ns *vm.Namespace
@@ -2760,6 +2773,13 @@ func installLangNS() {
 		if vs[0] == vm.NIL {
 			return vm.NIL, nil
 		}
+		// ArrayVector fast path: index the flat slice directly, no seq alloc.
+		if av, ok := vs[0].(vm.ArrayVector); ok {
+			if len(av) == 0 {
+				return vm.NIL, nil
+			}
+			return av[0], nil
+		}
 		if seq, ok := vs[0].(vm.Seq); ok {
 			return seq.First(), nil
 		}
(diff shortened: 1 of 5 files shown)


nooga:
lgtm

nooga (approved):

PR #399, opened 2026-07-07
perf(rt): Range fast path in reduce + set via transient — series close (-48% suite allocs total)

nnunley:
> **Stacked on #398** (ArrayVector direct-index fast paths) — earlier commits belong to the PRs below this one in the stack and disappear from this diff as they merge.

## What

Closes out the allocation-reduction series with three small pieces:

### 1. Range arithmetic fast path in `reduce`

Reducing a `Range` walked it as a chunked seq — each `ChunkedFirst` materializes a fresh 32-slot buffer. A range is `(start, end, step)` arithmetic, so `reduce` now iterates it directly via a new `Range.Bounds()` accessor: zero seq/chunk allocation, same reused two-arg buffer as the ArrayVector fast path. Empty ranges never reach the fast path (`NewRange(0,0,1)` returns `EmptyList`; the `RawCount` early-return catches the rest), and `InfiniteRange` is a different type that stays on the seq path (terminating only via `reduced`, as before).

### 2. `set` builds via transient; sets are `transientable?`

`TransientSet` existed, but `transientable?` never included sets — so `into` onto a set, and `(set coll)` itself, built through per-element persistent `Conj`. `transientable?` now covers sets and `(set coll)` routes through `transient`/`conj!`/`persistent!`.

### 3. Ratchet baseline re-record

`bench-ratchet update` (tighten-only) after the above.

6,841,735 → **6,794,600 allocs/op (−0.7%)**.

An honest note: the profile had promised more — `alloc_objects` showed a 1.65M `Range.ChunkedFirst` bucket, but pprof totals span warmup + compile + all iterations, so it was ~250K per counted op; part of the win is also offset by #398's iteration-isolation fix legitimately rebuilding namespaces per iteration. The fast path does what it says (reduce-attributed Range chunk allocation is zero in the post-profile); the workload just contains less bare `(reduce f (range …))` than the profile suggested.

**Series total across the stack (#396→here): 13,066,797 → 6,794,600 allocs/op (−48.0%)** on the bytecode suite, with the bench `fail` metric now a clean 0.

What remains in the profile is `.lg`-driven (vector literals in hot fns, `conj` accumulation loops, `next`-stepping seq walks) — follow-up work needs `.lg`-caller allocation attribution tooling before more rewrites, not more Go-side point fixes.

## Testing

- New `pkg/rt/reduce_range_test.go` (9 cases: both arities, negative step, empty/singleton ranges, `reduced` short-circuit)
- `go test ./...` green; .lg suite (170) green; jank suite 5657 pass / 0 fail
- `make generate` + `make check-generated` clean (core.lg change regenerates both artifacts)


nooga:
Range fast path looks solid — traced the arithmetic by hand against all 9 cases including the negative-step and reduced-short-circuit ones, no issues. Appreciate the honest write-up on the -0.7% vs. what the profile initially promised; that's the right way to close out a series.

One gap before merge: `transientable?` picking up `set?` means `(set coll)` *and* both arities of `into` now route sets through `TransientSet`/`conj!`/`persistent!` instead of per-element persistent `Conj`. That's a plumbing change to a very commonly used idiom (`(into #{} ...)` especially), and there's no new test — Go or `.lg` — targeting it specifically. We're relying entirely on the existing suite staying green to catch a dedup/count regression. Please add a small explicit case, e.g. `(into #{} [1 1 2 2 3])` → `#{1 2 3}` with the right count, and one for `(set coll-with-dupes)`, same as we did for the Range and ArrayVector fast paths elsewhere in this stack.

Rest is fine; still blocked on the #396/#397/#398 items further down the stack as before.

mparrett:
Outstanding from previous review:

 the set/transient behavior change has no direct regression coverage. `set` and `into` now use `TransientSet`, but the new test file only exercises Range reduction. Consider explicit tests for duplicate handling and count, e.g. `(set [1 1 2 2 3])` and `(into #{} [1 1 2 2 3])`.

mparrett:
Should CI should run <code>make ir-stress-gate</code> as a required check? It took about two minutes locally and is explicitly intended as the lowering-coverage regression gate. If so, it would come after this updated baseline, I think.

nnunley:
Superseded: this branch's runtime changes (pkg/rt/lang.go, pkg/vm/range.go, reduce_range_test.go) are byte-identical to main, and the core.lg delta is fully absorbed — rebasing onto current main yields an empty commit. The 'series close' perf work already landed in main. Closing as redundant.

PR #400, opened 2026-07-07
Various bug fixes

ingydotnet:
Here are 6 bugs I found and fixed on my gloathub/let-go fork as I try to make the gloat automation tool work with multiple clojure->go engines.

I'll keep pushing new fixes here as I go.


mparrett:
@ingydotnet Thanks for the PR! Shipped one of your improvements in https://github.com/nooga/let-go/pull/406

nnunley (approved):
Reviewed all six fixes — approving. Each is correct, Clojure-aligned, and carries a focused regression test: `condp` arg order → `(pred test-val expr)`; `range` no longer panics on non-Int bounds (JVM-like float-end + `##Inf`); `sort` survives multibyte strings (RawCount-bytes vs Seq-runes); `shuffle` empty-lazy → `[]`; `re-seq` returns capture-group vectors; and the variadic 0-arity tail-call fix (`boxRest`, empty rest → nil). Nice root-cause comments throughout.

One call-out before merge — **it only needs a rebase onto current `main`, and the `read-line` no-arg commit (dd6652d6) should be dropped**: that landed independently via #406, so it's now redundant (verified against main). After rebasing, `make generate` to refresh the bundle/generated.sums. Everything else merges cleanly — I peeled the stack onto current main locally to confirm there are no source conflicts.

(Non-blocking, for later: `range` rejects float *start/step* with an error where Clojure produces a float range; and `re-seq`'s non-participating group yields `""` not `nil` — consistent with `re-find`, tracked in #426.)

ingydotnet:
rebased on origin/main now. 

`feat(rt): read-line with no args reads stdin` commit dropped.

PR #401, opened 2026-07-07
perf(tooling): byte-level .lg-caller allocation attribution (LG_ALLOC_ATTR=1)

nnunley:
> **Stacked on #399** (Range fast path + set via transient) — earlier commits belong to the PRs below this one in the stack and disappear from this diff as they merge.

## Why

Go's alloc profile bottoms out at builtins: every `(assoc …)` in `.lg` code is the same one rt closure, so pprof can tell you `hmapBitmapNode.cloneAndSet` allocated 2.8 GB but not *which `.lg` caller* drove it. Two earlier perf slices targeted the highest **call-count** mutators and under-delivered (−11%, −2%) because call volume ≠ bytes. This PR adds the missing attribution axis.

## What

`LG_ALLOC_ATTR=1` enables byte-level `.lg`-caller allocation attribution (`pkg/vm/allocattr.go`):

- `Frame.Run` maintains a shadow stack of interpreter frames (gated push/pop; disabled cost is one predictable branch).
- Instrumented allocation sites — `PersistentMap` `cloneAndSet`/`nodeSeq`/`entries`, `NewArrayVector`, `ArrayVector` `Seq`/`Next`, `NewLazySeq` — record bytes+count keyed by the top-3 `.lg` frames (`CodeChunk`+ip), resolved to `file:line` via the chunk source maps at dump time (distinct ips on one line aggregate).
- `pmapv` serializes to 1 worker while enabled so the shadow stack stays coherent.
- lgbgen dumps the report to stderr at exit: totals per instrumented kind, then top call stacks by bytes.

Usage: `LG_ALLOC_ATTR=1 ./lgbgen --target=go 2> report.txt`

## Did it work?

First use attributed 1.37 GB to a single `(nth string i)` scan in `pkg-alias-from` and ~540 MB to a `(seq map)` emptiness test — both invisible in pprof, both fixed in the next PR up the stack (−13% total allocation).


mparrett (commented):
**Codex (5.6-terra/high):**
P1: Account for the allocations made while building entries in pkg/vm/persistent_map.go:286. nodeSeq grows the slice with append (including child slices), so Go allocates and copies multiple backing arrays as capacity grows. Recording len(entries)*32+24 once only models the final logical length; it misses intermediate allocations and can differ from final capacity. This systematically under-reports map-sequencing bytes and can mis-rank the call sites the new tool is intended to identify. Please record each backing-array allocation at its allocation site, or change construction/accounting so the report reflects allocations actually performed.

mparrett:
Accidentally approved this prematurely, sorry about that. P1 finding still present.

nnunley:
Superseded: the alloc-attribution tooling (pkg/vm/allocattr.go + the instrumentation hooks) is already present in main — this branch's allocattr.go is byte-identical to main's and the only remaining difference is a stale generated.sums. Closing as redundant. The concurrency-doc correction for allocattr.go lives in #439 (still pending).

PR #403, opened 2026-07-07
perf(ir/typeinfer): worklist membership via in-place transient-map flags (−1GB cloneAndSet)

nnunley:
> **Stacked on #402** (string-nth + registry emptiness fixes) — earlier commits belong to the PRs below this one in the stack and disappear from this diff as they merge.

## What

Third finding from the LG_ALLOC_ATTR attribution (#401): the typeinfer solver's `queued` worklist-membership set was persistent — `(disj queued entry)` per pop and `(conj queued entry)` per enqueue each path-copied a HAMT node. ~1 GB of `cloneAndSet` per `lgbgen --target=go` run, the top remaining stack after the bitset-lattice/positional-state work (which fixed the *joins*, not the worklist bookkeeping).

## Why a transient map, not a transient set

The obvious `(transient #{})` + `disj!` does **not** fix it — measured first: `TransientMap.Dissoc` has no transient path; it runs the persistent `dissoc` (`removePair`/`cloneAndSet` per removal). Only `Assoc` has the in-place `editAndSet` path. So `queued` is now a transient **map** `entry → in-queue?`: enqueue sets `true`, pop sets `false`, both mutating owned nodes in place — zero allocation per flip after an entry's first insertion. The drain is single-threaded; the map's key count is bounded by the function's insts+params (the same peak the old set reached). Also drops the `[queue queued]` pair-vector allocation per enqueue.

## Measurements (full `lgbgen --target=go`, alloc_space)

- `cloneAndSet` flat: **2733 MB → 1689 MB**
- Run total: 18.2 GB → **16.9 GB**; cumulative across this three-PR series: **21.0 GB → 16.9 GB (−19.5%)**

ir-stress baseline: corpus +1 (this adds a defn to a corpus file), failure count unchanged at 11 — gate re-verified. Full `go test ./...` green (18 packages).

### DIFF
--- a/docs/perf/ir-stress-baseline.edn
+++ b/docs/perf/ir-stress-baseline.edn
@@ -1,6 +1,6 @@
 {:failed 20
- :total 2396
- :passed 2376
+ :total 2397
+ :passed 2377
  :captured "2026-07-10"
  :how "LG_STRESS_PASSES=1 make ir-stress (corpus = every shipped/test/example .lg)"
  :note
--- a/pkg/rt/core/ir/passes/typeinfer.lg
+++ b/pkg/rt/core/ir/passes/typeinfer.lg
@@ -443,34 +443,45 @@
           (not= :bottom (state-type state (nth source 2))))
         (get (:param-sources deps) [bid param-pos])))
 
-(defn- enqueue-entry [queue queued entry]
+(defn- enqueue-entry! [queue queued entry]
+  ;; queued is a TRANSIENT map entry→in-queue? (the drain is
+  ;; single-threaded). Membership flips via assoc! true/false: transient
+  ;; assoc mutates owned HAMT nodes IN PLACE, so after an entry's first
+  ;; insertion its enqueue/pop cost zero allocation. (A transient SET
+  ;; would not help here: disj!/Dissoc has no transient path — it clones
+  ;; the node array per removal, which is exactly the ~750MB cloneAndSet
+  ;; churn this replaces.) Returns the queue; queued mutates by side
+  ;; effect.
   (ti-inc! :enqueue-attempt)
-  (if (contains? queued entry)
-    [queue queued]
+  (if (get queued entry)
+    queue
     (do
       (ti-inc! :enqueue)
-      [(conj queue entry) (conj queued entry)])))
+      (assoc! queued entry true)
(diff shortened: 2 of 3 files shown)


mparrett (approved):
**Review: Approve (non-blocking notes below)**

I went through the diff and couldn't find a correctness issue in the worklist change itself. The transient-map-as-membership-flags approach is sound, and the reasoning for a transient map over a transient set (no in-place `Dissoc` path) is correct. Two notes, neither blocking.

**1. `into` on sets now drops collection metadata. This matches Clojure, noting it in case it surprises anyone.**

Adding `set?` to `transientable?` routes metadata-bearing sets through `(transient c)`, and `TransientSet.Persistent()` rebuilds without `meta`, so `(meta (into (with-meta #{1} {:x true}) [2]))` goes from `{:x true}` to `nil`. That's a behavior change for sets.

It's a change toward Clojure, though, not away from it. Clojure's `into` doesn't preserve collection metadata either, because it uses transients (CLJ-1651): `(meta (into ^:foo [] [1]))` is `nil`. And `into` on maps and vectors here already dropped meta the same way, since `TransientMap.Persistent()` doesn't carry it. So before this PR, sets were the odd one out that preserved meta via `reduce conj`; this brings them in line with maps, vectors, and Clojure.

Only actionable if let-go means `into` to preserve metadata as a deliberate divergence from Clojure, in which case the fix isn't to exclude sets, it's to carry `meta` through all three transients (maps and vectors are already "broken" by that standard). I read the intent as Clojure-faithful, so I'd land this as-is, maybe with a one-line changelog note.

**2. The `attrStack` global is single-goroutine by design, and the only in-VM concurrency source is already handled.**

A single global shadow stack would be wrong under concurrent VM frames, but `pmapv` is the only path that interprets `.lg` frames on spawned goroutines, and this PR serializes it to one worker while `LG_ALLOC_ATTR` is on. The other goroutines in `rt` (pods, term, and os pumps) are subprocess RPC and I/O, not in-VM interpretation, so they don't touch this stack. The single-execution assumption is documented in the file comment too. Fine as-is for an env-gated diagnostic. The thing to revisit is if in-VM `future`/`go` primitives ever land, at which point attribution state would need to go goroutine-local.

Approving.

PR #404, opened 2026-07-07
feat(ir-stress): tool-maintained coverage baseline — make ir-stress-rebaseline

nnunley:
> **Stacked on #403** (typeinfer worklist) — earlier commits belong to the PRs below this one in the stack and disappear from this diff as they merge.

## What

The ir-stress coverage baseline (`docs/perf/ir-stress-baseline.edn`) was maintained by hand — the gate's own error message prescribed "run the census, then update the file". This adds writer mode so the ratchet maintains its own baseline:

- `LG_STRESS_REBASELINE=1` flips the gate into writer mode: rewrites the baseline EDN from the live census (counts, failure buckets, capture date), preserving the committed `:how`/`:note` prose. Stable diff-friendly layout, round-trips via `read-string`.
- `make ir-stress-rebaseline` is the entry point (injects the capture date via `date +%F`).

## The writer earned its keep immediately

Its first draft used `(. (now) 'String)` for the date — which itself fails ir/build lowering, and since this script is part of its own corpus, the written census honestly reported `:failed 12`. A hand edit would have glossed right over that; the tool surfaced it. (Fixed by injecting the date from the Makefile.)

Baseline written by the tool: `:failed 11 :total 2368 :passed 2357` (+2 total = this change's own defns plus the worklist PR's; failure count unchanged). `make ir-stress-gate` green against it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -369,6 +369,17 @@ ir-stress-gate: build
 	  LG_STRESS_BASELINE=docs/perf/ir-stress-baseline.edn \
 	  ./lg scripts/ir-stress.lg corpus scripts/ir-stress-corpus.edn
 
+# Rewrite the committed coverage baseline from a fresh census (tool-maintained;
+# never hand-edit the EDN). Run after an intentional corpus or coverage change,
+# review the diff, and commit it with the change that caused it.
+ir-stress-rebaseline: build
+	LG_STRESS_PASSES=1 \
+	  LG_STRESS_TIMEOUT_MS=$${LG_STRESS_TIMEOUT_MS:-15000} \
+	  LG_STRESS_BASELINE=docs/perf/ir-stress-baseline.edn \
+	  LG_STRESS_REBASELINE=1 \
+	  LG_STRESS_DATE=$$(date +%F) \
+	  ./lg scripts/ir-stress.lg corpus scripts/ir-stress-corpus.edn
+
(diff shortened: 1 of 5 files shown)


mparrett:
Code review found no blocking concern in #404’s own changes: the rebaseline writer and bucket-based gate are internally consistent. After resolving the stack conflicts and rerunning CI, I’d be comfortable approving.

nnunley:
@mparrett — stack conflicts resolved: rebased conflict-free onto current `main` as a single self-contained commit (the merged perf-series ancestors it used to sit on are already in main). The tool-maintained baseline regenerates cleanly and the bucket-based gate passes at 20/20 (I also corrected one flaky `:stress/timeout` the rebaseline run had captured). `check-generated`, `gogen-parity`, and full `go test ./...` are green. Should be good for the approval you offered.

nooga (approved):
Solid improvement — per-bucket ratchet is more precise than the old total-count gate, and the classify-error idempotency fix is a real, well-caught bug. Two non-blocking suggestions: guard against silently blanking `:note` when the previous baseline fails to parse, and consider wiring `ir-stress-gate` into CI (or filing a follow-up) since right now nothing enforces it gets run before merge.

mparrett:
Looks good. Rebase is clean onto current main, the generated artifacts and coverage gate are consistent, and CI is green.

PR #406, opened 2026-07-07
feat(rt): zero-arg (read-line) reads current *in*

mparrett:
## Summary

- add zero-arg `(read-line)` support that reads from the current `*in*` binding
- keep the existing explicit handle form `(read-line handle)` working
- add e2e coverage for process stdin, EOF, explicit `*in*`, and rebound `*in*`

## Why

PR #400 surfaced the useful Clojure-compatible behavior that `(read-line)` should read stdin. This keeps that behavior while matching let-go's existing dynamic I/O model: standard streams are Vars, so a rebound `*in*` should be honored the same way `*out*` is honored by printing.

## Validation

- `go test -timeout 600s -count=1 -v ./test/e2e -run TestReadLineStdin`
- `go test -timeout 600s -count=1 -v ./test/e2e`

### DIFF
--- a/test/e2e/read_line_stdin_test.go
+++ b/test/e2e/read_line_stdin_test.go
@@ -1,0 +1,76 @@
+/*
+ * Copyright (c) 2026 let-go contributors; see CONTRIBUTORS.
+ * SPDX-License-Identifier: MIT
+ */
+
+package e2e
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+const (
+	userRead  os.FileMode = 1 << 8
+	userWrite os.FileMode = 1 << 7
+	groupRead os.FileMode = 1 << 5
+	otherRead os.FileMode = 1 << 2
+
+	testFilePerm = userRead | userWrite | groupRead | otherRead
+)
+
+// TestReadLineStdin verifies (read-line) with no args reads the current *in*,
+// like Clojure, while the 1-arg handle form keeps working.
+func TestReadLineStdin(t *testing.T) {
+	bin := buildLG(t)
+
+	run := func(t *testing.T, stdin string, body string) string {
+		t.Helper()
+		p := filepath.Join(t.TempDir(), "app.lg")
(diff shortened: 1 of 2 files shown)


mparrett:
Reviewed. Looks good — the change lines up with let-go's dynamic-var I/O model: resolving `*in*` via its core Var and deref'ing the current binding is the same path `*out*` takes on print, so a rebound `*in*` is honored correctly. Arity change is additive, so existing `(read-line handle)` call sites are unaffected, and the shared `stdinHandle` backs both the fallback and the `*in*` root, so there's no double-buffering. E2E coverage hits the cases that matter (stdin, EOF, explicit handle, rebound `*in*`).

Minor, non-blocking: with `*in*` always installed, the `src := stdinHandle` initializer is effectively an unreachable fallback since the zero-arg path always derefs the Var. Harmless as defensive code.

test/e2e/read_line_stdin_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+// TestReadLineStdin verifies (read-line) with no args reads the current *in*,
+// like Clojure, while the 1-arg handle form keeps working.
+func TestReadLineStdin(t *testing.T) {
+	bin := buildLG(t)
+
+	run := func(t *testing.T, stdin string, body string) string {
+		t.Helper()
+		p := filepath.Join(t.TempDir(), "app.lg")
+		if err := os.WriteFile(p, []byte(body), 0644); err != nil {
```
nnunley: Can this be written with the constants | together, rather than hardcoding the numeric constant?

nnunley (approved):
LGTM. I just have a minor nit on the test code.

test/e2e/read_line_stdin_test.go
mparrett: Good suggestion. Addressed in https://github.com/nooga/let-go/commit/e149fea7c07f545f5dc67750c4ef8a290e76fe48

mparrett:
Postscript: Credit goes to @ingydotnet for finding and initial fix submission. Thank you!

PR #407, opened 2026-07-07
feat(vm): replace `trace` with dynamically-scoped *lg-trace* (signature-keyed opcode migration)

nnunley:
## The bug

```clojure
(defn trace [s] (println s))
(trace "Hello")
```
prints a bytecode frame dump instead of `Hello`. `trace` was a **compiler special form**, and special forms are matched before var resolution — so a user `(defn trace …)` was silently unreachable at every call site: `(trace x)` compiled to `OP_TRACE_ENABLE` → evaluate args → `OP_TRACE_DISABLE`, never invoking the user's function. It is not a Clojure special form (Clojure keeps special forms to irreducible compiler primitives; tracing lives in `clojure.tools.trace` as ordinary vars) and had **zero call sites** anywhere in the repo — a vestigial debug hook occupying a common name in the one unshadowable table.

## The change

**Remove the `trace` special form** (`pkg/compiler`) so the name resolves normally.

**Add a dynamically-scoped `*lg-trace*` var** as the replacement debug hook. When it resolves truthy in a frame's execution context the frame traces, and because the binding rides the shared dynamic stack, every callee traces too:

```clojure
(binding [*lg-trace* true] (my-fn))   ; traces my-fn AND its call subtree
```

This is strictly better than the old frame-local special form (which never propagated to callees). The hot path stays free: a coarse atomic gate (`TraceArmed`, modeled on `ProfilingEnabled`) arms only on the first truthy bind/set of `*lg-trace*`, so there is zero per-call cost until tracing is actually used, and the precise per-frame var deref only runs once armed.

**Retire the now-dead opcodes with a bytecode format-version migration.** `OP_TRACE_ENABLE`/`OP_TRACE_DISABLE` are removed and the opcode enum renumbered. `FormatVersion` is bumped 2 → 3; the decoder still loads v1/v2 bundles via a **frozen, stride-aware opcode remap** (`18/19 → NOOP`, `≥20 → −2`) that walks instruction-by-instruction so argument words and packed high bits are never rewritten — instruction widths are preserved, so IPs, jump targets, and source maps stay valid. The new v3 binary boots by loading the prior-version committed bundle through the remap.

**Add the missing disassembler opcode-coverage guardrail:** a name+stride completeness test over every opcode, plus encode/decode round-trip and explicit v2→v3 remap tests — so future opcode enum changes are verifiable, not silent.

## Verification

- `make check-generated` — bundle + lowered tree in lockstep; regenerated bundle is v3.
- Behavior: `(defn trace …)` now works; `(binding [*lg-trace* true] (h 5))` traces `h` and propagates into its callees; outside the binding, normal output.
- Full `go test ./...` — green (18 packages), including the new coverage/remap tests and the AOT-diff gate.

## Heads-up

This branch is based on current `main`. The local `ir-stress-gate` pre-push hook currently fails on `main` itself due to a stale committed baseline (`:total 2347` vs the live corpus of 2366) — a pre-existing drift being addressed separately. This PR touches neither the baseline nor the corpus, so that failure is orthogonal to this change.

---
**Part of #257** (Epic: Clojure semantics parity) — removes the `trace` special-form lg-ism so the name resolves as an ordinary symbol.

## Tradeoff note: one-way `TraceArmed` latch

Once `*lg-trace*` has been bound or `set!` truthy anywhere in a process — even once, briefly — `TraceArmed` stays true for the process lifetime, and every subsequent frame entry pays one dynamic-var deref of `TraceVar` (now an `atomic.Pointer[Var]`). This is deliberate: it keeps the never-traced hot path at a single atomic load with zero per-frame deref, at the cost of a permanently-small overhead after first use. In a long-lived process (e.g. an nREPL server) that means using `*lg-trace*` once leaves the remaining session marginally slower; if that ever matters, the follow-up would be scoping the armed state rather than latching it.


nooga:
## Review: replace `trace` special form with `*lg-trace*` dynamic var

### Overview

Removes the `trace` compiler special form (which silently shadowed any user-defined `trace` function, since special forms are matched before var resolution) and replaces it with a dynamically-scoped `^:dynamic *lg-trace*` var. Tracing now propagates to callees via the shared binding stack, gated by a coarse `TraceArmed` atomic so the hot path costs nothing until tracing is used. The two retired opcodes (`OP_TRACE_ENABLE`/`OP_TRACE_DISABLE`) required a bytecode format bump (v2→v3) with a frozen, stride-aware remap table so old `.lgb` bundles still decode.

I checked out the branch, built it, ran the full suite (`go test ./...`, all green), ran `make check-generated` (passes), and manually verified both the bug fix and the callee-propagation claim:

```
$ go run . -e '(defn trace [s] (println s)) (trace "Hello")'
Hello
nil
```
and traced `g`→`h` correctly under `binding`, with no trace output once the binding scope exited.

### Code quality & style

- The opcode migration is careful and well-reasoned: the frozen `v2Stride`/`remapV2Opcode` table is decoupled from the live `vm.OP_*` enum by design (explicit comment: "do not update it when opcodes change again"), which is the right call for a wire-format migration. I hand-verified the v2 stride table against the pre-PR opcode numbering and it's correct (RECUR=4 words, TRY_PUSH=3, the 13 single-arg opcodes all line up).
- `TestRemapLegacyChunksV2ToV3` and `TestOpcodeDisassemblyCoverage` are exactly the right regression guardrails for this kind of change — they'll catch drift the next time an opcode is added/removed.
- `NewFrame` already resets `f.debug = false` on every acquire from the frame pool, so there's no cross-reuse trace leakage there — good, I checked this specifically since `f.debug` being pool-reused is an easy way to leak state.
- Comment density is high (multi-paragraph doc comments in `trace.go`/`opcode_migration.go`), but each one is explaining a genuinely non-obvious invariant (frozen table rationale, stride preservation, why disarming isn't needed) rather than restating code — appropriate given the subtlety of a bytecode migration.

### Potential issues & risks

**1. Data race on `TraceVar` (real bug, confirmed with `-race`).**
`pkg/vm/trace.go:41` declares `var TraceVar *Var` as a plain package-level pointer. It's written unsynchronized in `armTraceIfTruthy` (`TraceVar = v`) and read unsynchronized in `Frame.Run()` (`TraceVar != nil`) and again inside `armTraceIfTruthy` itself (`v == TraceVar`). Since `pushBinding`/`set!` can run concurrently across goroutines (the package's own comments elsewhere are explicit about this — `isDynamic` is `atomic.Bool` specifically "because push-binding/deref of a var can happen on different goroutines concurrently," and `bindingStack.head` is `atomic.Pointer[bindingFrame]` for the same reason), this is inconsistent with the rest of the package's concurrency discipline. I reproduced it directly:
```
WARNING: DATA RACE
Write at 0x... by goroutine 9: armTraceIfTruthy (trace.go:57, TraceVar = v)
Previous read at 0x... by goroutine 10: Frame.Run (vm.go:528, TraceVar != nil)
```
Fix is small: make it `atomic.Pointer[Var]` (mirrors `TraceArmed` and `Var.root` in the same file/package), e.g. `TraceVar.Load()`/`TraceVar.CompareAndSwap(nil, v)`.

**2. No `.lg` regression test for the actual behavior change.**
The project's testing policy (CLAUDE.md) calls for a `.lg` integration test for language-level features and fixes to core semantics. This PR fixes a real, user-visible bug (`(defn trace ...)` being unreachable) and adds a new dynamic var, but ships with only Go-level bytecode/disassembler tests. Nothing pins:
- `(defn trace [s] ...) (trace ...)` actually invoking the user function (the regression this PR exists to fix), or
- `(binding [*lg-trace* true] ...)` producing trace output and propagating into callees.

Both are easy to assert with `with-out-str` (I did so manually to verify the PR — it takes a few lines). Without this, a future refactor could reintroduce the shadowing bug silently.

**3. Permanent process-lifetime cost once tracing is used, with no way to disarm.**
`TraceArmed` is a one-way latch — once `*lg-trace*` is bound/set truthy anywhere in a process (even once, briefly, for debugging), `TraceArmed` stays `true` forever, and every subsequent `Frame.Run()` call in every goroutine pays an extra dynamic-var deref (walking the active binding chain) for the remainder of the process, even long after the triggering `binding` scope exits. This is intentional per the code comments ("disarming is unnecessary… an unbound/false `*lg-trace*` simply produces no trace") and is low-severity performance-wise, but it's a real, permanent throughput regression for the process's remaining lifetime, not scoped to the debugging session. Worth calling out more prominently (e.g., in a user-facing doc) given this is exactly the kind of thing that'd bite someone using `*lg-trace*` once in a long-lived nREPL server and then wondering why every call site got slightly slower for the rest of that session.

### Suggestions

- Fix #1 before merging — it's a genuine, confirmed race, not theoretical.
- Add the two `.lg` tests described in #2; they're cheap and directly pin the bug this PR fixes.
- Consider whether #3's blast radius (whole-process, permanent) is acceptable as-is, or whether scoping the "armed" state to something more granular is worth the complexity — a note in the PR description either way would help future readers know it's a deliberate tradeoff, not an oversight.

nooga (changes requested):
Great bytecode-migration engineering — the stride-aware remap and coverage tests are exactly right. Two things before merge: (1) `armTraceIfTruthy`'s var-wiring in `pkg/vm/trace.go` matches on bare `VarName()` only, not namespace-qualified — a same-named dynamic var in another ns could permanently hijack `TraceVar`; please qualify against `core`. (2) Please add a `.lg` integration test for `*lg-trace*` binding/propagation-to-callees per the testing policy.

nnunley:
## Reworking around #443: signature-keyed migration (no FormatVersion bump)

Per the discussion with @nooga and @mparrett: a newer VM should *tolerate* older bytecode (migrate) rather than force a recompile — important when the bundle vendor ≠ consumer. But that doesn't mean reverting #443; it means repurposing its opcode-set **signature** from a hard gate into the **key for a migration registry**.

Reframing this PR around two orthogonal axes:

- **`FormatVersion`** = *serialization structure* only. Removing the trace opcodes changes the opcode **set**, not the byte layout — so this PR **drops the v3 `FormatVersion` bump and the frozen stride table**.
- **Opcode-set signature** (#443's `OpcodeSetSignature()`) = the identity for the opcode dimension. v1/v2 predate the removal and share a frozen opcode set, so their signature is *implicit* (computable from the historical `opcodeNames`) — no legacy exception needed.

Decode becomes:

```
effective_sig(bundle) = stored signature (CapOpcodeSet)              if present
                        OpcodeSetSignature(frozen_pre_removal_names) otherwise  // implicit v1/v2

  effective_sig == current runtime          → run as-is
  effective_sig ∈ migration registry        → apply remap, then run   // tolerate old
  else                                       → reject ("recompile")    // backstop
```

The registry starts with one entry: `{ pre-removal signature → trace-era opcode remap }`, and grows by one per future opcode-set change — no per-version branching.

Concretely: #443 keeps all its encoder/decoder/signature plumbing; only its mismatch branch changes from `reject` → `migrate-if-known-else-reject`. This PR removes the two opcodes (→ new signature), registers the old signature's remap, and keeps the `*lg-trace*` dynvar feature. I'll push the reworked version shortly.

nnunley:
Pushed the reworked implementation. Summary of what changed from the original approach:

- **No `FormatVersion` bump** — stays at 2. Removing the trace opcodes changes the opcode *set*, not the serialization *format*.
- **`OP_TRACE_ENABLE`/`OP_TRACE_DISABLE` removed** from the const block and `opcodeNames`, so #443's `OpcodeSetSignature()` computes the new set's signature automatically.
- **Signature-keyed migration** (`pkg/bytecode/opcode_migration.go`): a frozen pre-removal opcode-name list yields `preRemovalSignature()`; a small `migrationRegistry` maps `(count, hash) → remap`. Seeded with `{ pre-removal signature → trace-era remap }` (TRACE→NOOP, opcodes ≥20 shifted down 2, instruction width preserved).
- **Decoder is now migrate-then-check** (softens #443's hard reject): effective signature = stored (`CapOpcodeSet`) or, for pre-capability bundles, the implicit pre-removal signature. Match current → run as-is; known-prior → apply remap after chunks decode → run; unknown → reject with the recompile error. So a newer VM tolerates every opcode set it knows how to migrate, and only rejects genuinely-unrecognized ones.
- **`*lg-trace*` dynvar feature preserved** (`pkg/vm/trace.go` + `(def ^:dynamic *lg-trace* false)`), trace special form removed from the compiler.

@mparrett — this keeps all of #443's signature plumbing; it only turns the mismatch branch from `reject` into `migrate-if-known-else-reject`, so the two mechanisms compose rather than conflict.

Gates: bytecode/vm/compiler tests, `TestLGBParity` (incl. `-tags bootstrap`), gogen-parity PARITY, ir-stress-gate, check-generated, and full `go test ./...` — all green. New tests cover the v2→current remap, the implicit-signature computation, registry lookup, a legacy bundle decoding + migrating, and unknown-set rejection.

nooga (changes requested):
Opcode-migration rework looks correct and reconciles cleanly with #443 (no competing versioning). But: (1) `armTraceIfTruthy` is still bare-name matched, not ns-qualified — unresolved from last review. (2) Still no test for `*lg-trace*`. (3) New bug: `armTraceIfTruthy` isn't called from `ExecContext.pushBinding`, so `(binding [*lg-trace* true] ...)` — the PR's own example — never arms tracing; only a bare top-level `set!` does. Please wire arming into `pushBinding` too and add the test.

nnunley:
All three findings from the detailed review are now addressed (commit fe5e0b1, rebased onto current main):

1. **Data race fixed**: `TraceVar` is now `atomic.Pointer[Var]` with `Load()`/`CompareAndSwap(nil, v)` for the first-wire, matching the package's discipline (`TraceArmed`, `isDynamic`, `bindingStack.head`). Verified clean under `-race` across 10 runs, including the frame-entry read path.
2. **`.lg`-level behavior tests added**: `TestLgTraceShadowingRegression` pins `(defn trace [s] (println s)) (trace "Hello")` invoking the user fn via `with-out-str`; `TestLgTraceOutputAndPropagation` pins that `(binding [*lg-trace* true] ...)` emits the frame trace into callees and goes silent when the scope exits — captured via `os.Pipe`, since the VM's trace dump writes to process stdout rather than `*out*` (which is also why a pure `.lg` `with-out-str` assertion can't see it).
3. **One-way latch documented as deliberate**: PR description now carries the tradeoff note — once `*lg-trace*` has been truthy anywhere in a process, every subsequent frame entry pays one dynamic-var deref for the process lifetime, trading a permanently-small cost after first use for a zero-cost hot path before it.

mparrett:
The race fix and latch documentation look good. The behavior-test finding is not fully closed yet: the shadowing test prints `FAIL` without failing, the propagation check can pass if only the outer frame traces, and resetting the global trace state before the final assertion does not test silence after the binding exits. Consider adding hard assertions, verify a callee-specific opcode/output, and test the post-binding call without resetting the armed state.

nooga (commented):
Functionally this looks correct and CI is green -- the opcode-migration mechanism (frozen v2Stride/remap table, signature registry composing cleanly with #443) and the *lg-trace* dynvar wiring (atomic TraceVar, arming from both pushBinding and set!, ns-qualified match) both address everything raised in the two earlier CHANGES_REQUESTED rounds. One thing is still open, matching what @mparrett already flagged and which no subsequent commit has addressed:

**test/lg_trace_test.go has three assertion bugs that make the new tests pass even when the feature they're pinning is broken:**

1. `TestLgTraceShadowingRegression` (~line 742-751): the compiled `.lg` body does `(println "FAIL:" output)` on mismatch instead of failing the test. Since `evalTrace` only checks the compile/eval error, this Go test can never fail based on the traced value -- it doesn't actually pin the shadowing regression it's named for.
2. `TestLgTraceOutputAndPropagation` (~line 792-796): it only checks the output contains generic `"run"`/`"LOAD_ARG"` substrings. Since `outer` itself takes an arg and would also emit those tokens, this passes even if propagation into `inner` were broken -- it never confirms the callee specifically got traced.
3. Same test, second half (~line 800): calling `resetTrace()` before the "no output after binding exits" check clears `TraceVar` to nil, so the `tv != nil` gate in `Frame.Run` trivially prevents tracing regardless of whether binding-pop actually restores the var correctly. It isn't testing the real permanently-armed-latch behavior described in the PR body.

Suggest: assert with `t.Fatalf` instead of printing FAIL; check for a callee-distinguishing marker (e.g. count occurrences of "run" == 2, or a token unique to `inner`'s frame) to confirm propagation; and for the post-binding silence check, don't call `resetTrace()` -- leave `TraceArmed`/`TraceVar` as they'd realistically be after a real `binding` scope exits, and confirm the *value* deref (not the coarse gate) is what suppresses tracing.

Nothing else blocking -- the race and ns-qualification issues from the earlier reviews are resolved and directly verifiable in the diff.

nnunley:
Head `116f32a` — rebased onto current main (`traceCompiler` removal reconciled with the typed-catch helpers; bundle/sums regenerated via `make generate`) and all three test-assertion findings addressed, plus a cleanup pass that found two real issues of its own.

**The three flagged assertion bugs:**
1. `TestLgTraceShadowingRegression` now THROWS on mismatch (surfacing through the eval error check) instead of printing `FAIL:` — the assertion is load-bearing.
2. `TestLgTraceOutputAndPropagation` calls `inner` with a sentinel argument only it receives and requires its own `run[42077]` frame-entry banner, so outer-only tracing cannot pass. Strengthening this exposed a subtlety worth knowing: a **tail call reuses the caller's frame** — tracing continues through the callee's instructions but no new banner prints — so the test uses a non-tail call and documents why.
3. The post-binding silence check no longer calls `resetTrace()`: it asserts the gate is still armed and `TraceVar` still wired (the realistic post-binding state), so what's proven to suppress tracing is the per-frame *value* deref, not the coarse gate. Output capture also drains the pipe concurrently via `io.ReadAll` — the old fixed 4096-byte single read could truncate before the callee's banner and deadlock on >64KB trace output.

**Found while polishing:**
- **`alter-var-root` never armed the gate** — the third root-mutation path (and what `with-redefs` is built on): `(alter-var-root #'*lg-trace* (constantly true))` silently didn't trace. Arming now also fires from `Var.AlterRootArgs`, with coverage in `TestLgTraceArming`.
- The frame-entry check loaded `TraceVar` unconditionally before consulting `TraceArmed`, paying two atomic loads per frame entry in the never-traced case — reordered so the armed gate is checked first, matching the documented single-load design and the `ProfilingEnabled` shape.
- The two duplicated opcode-signature/migration-lookup blocks in `decoder.go` collapsed into one `resolveOpcodeSet` helper, and `remapV2Opcode` now documents the append-only-enum constraint every registered migration depends on.

Gates: `make check-generated` OK, lint clean, full `go test ./...` green, gogen_ir `pkg/ir` + `./test` (incl. the compat suite) green.

mparrett:
Single-agent review against 116df32a surfaced a few issues

- [P1] Remap legacy opcodes before execution - commented inline on pkg/bytecode/decoder.go - important if v1 compat is desired since it causes silent misexecution. If dropping v1 compat, a contract/document decision is needed (drop/reject)
- [P2] Not strictly blocking, but `TestLegacyBundleDecodeWithMigration` does not actually exercise migration: its builder retains the current opcode-signature capability, and the assertion only checks code length.
- Verified the new trace tests pass under `-race`.
- Note: A full race run still hits an unrelated existing `ArrayVectorSeq.ChunkedFirst` race.

pkg/bytecode/decoder.go
mparrett: Version 1 returns through the frozen decoder without invoking the opcode remapper. I reproduced this with a v1 bundle containing legacy opcode 25 (`ADD`): it remained 25 and the current VM interpreted it as `MUL`. Because v1 predates the opcode removal and can contain these opcodes, both v1 decode paths need migration—or the compatibility guarantee must explicitly drop v1.

nnunley:
Good catch — confirmed and fixed in 055b0aa2. Both frozen v1 paths now resolve the implicit pre-removal signature at the version branch (same treatment as a capability-less v2 bundle) and remap after decode, so the frozen functions stay textually untouched. The module path also re-syncs the raw `ChunkData.Code` from the remapped live chunks so re-encoding a decoded v1 module can't carry unmigrated opcodes under a current signature. Added `TestV1DecodeAppliesOpcodeMigration` pinning both paths with a hand-crafted v1 fixture containing your exact repro (legacy 25/`ADD` → `OP_ADD`, not `OP_MUL`).

mparrett (commented):
All findings have been addressed. Let me know when you're ready and I can do a final check and approval after this is rebased.

nnunley:
> All findings have been addressed. Let me know when you're ready and I can do a final check and approval after this is rebased.

Ready as soon as the CI vets it. No further changes being generated on my side.

nnunley:
Ok. We're ready go go @mparrett

mparrett (approved):
Re-reviewed at 9862292 (rebased/squashed). All three prior rounds are resolved and verified in the diff: armTraceIfTruthy is ns-qualified to core/*lg-trace* with a CaS first-wire, pushBinding arms the binding path, and the three test assertions are now meaningful (throw-not-print, sentinel-arg callee distinction, no resetTrace() before the silence check). CI green, mergeable. LGTM.

nooga (approved):

nooga:
lgtm

mparrett:
I ran the repeat-A/B perf check (nooga/let-go#445 / #517) against this PR twice, on two different runners, and one benchmark regresses reproducibly:

- `SeqIteration/List/01000` — median **+18.95%** on [run 1](https://github.com/nooga/let-go/actions/runs/29449799608) (6/7 cycles positive) and **+11.25%** on [run 2](https://github.com/nooga/let-go/actions/runs/29463429360) (7/7, cycles +10 to +18%). It gates at every budget (6/8/10%) in both.

That's list-`seq` iteration getting ~11-19% slower. Nothing else holds up: run 1 also moved `VarDerefRoot`, `FrameDispatch`, and other `SeqIteration` variants, but that run's calibration anchor was noisy (13% drift) and none of those reproduced on the cleaner second run — they were measurement noise, not the change. The list-iteration regression is the one real signal.

Worth a look at whether moving `trace` to a dynamically-scoped `*lg-trace*` added per-step cost to the list-`seq` path specifically (`ArrayVector`/`PersistentVector` iteration stayed flat, so it's the list representation, not iteration in general). This is the informational shadow check, so it's not blocking — a specific, reproduced number, raised while it's cheap to act on.

PR #408, opened 2026-07-07
fix(vm): explicit :refer shadows the clojure.core auto-refer baseline

nnunley:
## Problem

An explicit `(:require [lib :refer :all])` (or `:refer [syms]`, or `use`) did **not** shadow the auto-refer'd `clojure.core`, diverging from JVM Clojure. A symbol defined by both `clojure.core` and a refer'd lib resolved **nondeterministically** — `Namespace.Lookup` iterated the `refers` **map** (unordered) with no precedence rule, and in practice `clojure.core` usually won.

Concretely, a program re-exporting a `clojure.core` name (e.g. `ys.std`/`ys.dwim` redefining `gt`/`le`, or any lib redefining `inc`) kept resolving to `clojure.core`:

```clojure
(ns ys.std) (defn gt [a b] {:ys-gt [a b]})
(ns app (:require [ys.std :refer :all]))
(gt 3 4)   ; => false  (clojure.core/gt)   ❌ expected {:ys-gt [3 4]}
```

This also affects compiled calls, since the compiler resolves unqualified call symbols through `CurrentNS().Lookup`.

## Fix

`pkg/vm/namespace.go` `Namespace.Lookup`: treat the `clojure.core` refer as a **lowest-priority baseline**. Explicit (non-core) refers are returned first, honoring `:refer :all` / `:refer [syms]`; the core mapping is used only if no explicit refer provides the symbol. The core refer stays a full-`:all` baseline (an explicit `:refer` of core is additive, matching Clojure — core is trimmed via `:refer-clojure :exclude`, not here).

```clojure
(gt 3 4)   ; => {:ys-gt [3 4]}   ✅
```

## Tests

`test/ns/refer_shadows_core_test.lg` — `:refer :all` and `:refer [sym]` shadow core; unlisted `:only` names don't leak; unshadowed names still resolve to core. Verified RED→GREEN (RED failed nondeterministically, exposing the underlying map-order bug).

Suites green on this branch: `pkg/vm` + `pkg/rt` + `pkg/compiler` (378), full `.lg` suite (171). Also validated against the external xsofy corpus (281 tests / 2733 assertions / 0 fail).

## Note

The local `ir-stress` pre-push ratchet currently fails on a **corpus-size drift** (2347→2366 forms) from other merged work — failure count is unchanged (11) and this change adds none. It's a stale-baseline / repo-health item, unrelated to this fix.

### DIFF
--- a/pkg/vm/namespace.go
+++ b/pkg/vm/namespace.go
@@ -277,13 +277,36 @@ func (n *Namespace) Lookup(symbol Symbol) Value {
 		}
 		// Unqualified miss: search refers. Snapshot first so we follow each
 		// refer's target via its own lock, never holding n's lock across.
+		//
+		// Precedence (Clojure semantics): an explicit refer — (:require [lib
+		// :refer :all]) or :refer [syms], and (use lib) — SHADOWS the
+		// clojure.core auto-refer that RegisterNS installs into every ns. So we
+		// treat the core refer as a lowest-priority baseline: return the first
+		// matching explicit refer, honoring :all / :only, and only fall back to
+		// core if no explicit refer provides the symbol.
+		var coreHit *Var
 		for _, ref := range n.refersSnapshot() {
-			if v := ref.ns.localVar(sym.(Symbol)); v != nil {
-				if v.isPrivate {
-					return NIL
-				}
-				return v
+			v := ref.ns.localVar(sym.(Symbol))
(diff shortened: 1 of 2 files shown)


mparrett:
Reviewed this and didn't find any issues. The targeted namespace/runtime tests passed in a clean temporary worktree.

nooga (approved):

PR #409, opened 2026-07-07
fix(cli): exit nonzero when script or -e fails

mparrett:
## Summary

- make `lg` exit nonzero when a script fails with an uncaught runtime error
- make `lg -e` exit nonzero when expression evaluation fails
- keep successful scripts, successful `-e`, and caught errors exiting zero

## Why

Runtime errors were printed but still returned process exit code 0. That makes shell scripts, pipelines, and CI unable to distinguish a failed `lg` run from a successful one.

This is split out from the related fix in #400 as a standalone CLI behavior change. Credit: @ingydotnet 

## Validation

- `go test -timeout 600s -count=1 -v ./test/e2e -run TestErrorExitCode`
- `go test -timeout 600s -count=1 -v ./test/e2e`

### DIFF
--- a/lg.go
+++ b/lg.go
@@ -625,15 +625,18 @@ func runMain() int {
 	// Script mode: treat only the first positional as the script to run.
 	// Any further positionals belong to the script (it reads os/args).
 	ranSomething := false
+	runFailed := false
 	if len(files) >= 1 {
 		script := files[0]
 		if filepath.Ext(script) == ".lgb" {
 			if err := runLGB(script); err != nil {
 				fmt.Print(vm.FormatError(err))
+				runFailed = true
 			}
 		} else {
 			if err := runFile(context, script); err != nil {
 				fmt.Print(vm.FormatError(err))
+				runFailed = true
 			}
 		}
 		ranSomething = true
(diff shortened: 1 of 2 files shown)


lg.go
```diff
@@ -625,15 +625,18 @@ func runMain() int {
 	// Script mode: treat only the first positional as the script to run.
 	// Any further positionals belong to the script (it reads os/args).
 	ranSomething := false
+	runFailed := false
```
nnunley, line 638: Can you fix this to use the guard pattern? I know it's pre-existing, but the increased number of error and special handling is making the nesting pretty deep.

nnunley (approved):
Loogs good, but another minor nit.

lg.go
nnunley: The comment region is supposed to be the nested if statements below the selected line.

mparrett: Done in d5be38f. Extracted the .lgb-vs-source dispatch into a `runScript` helper, so the block drops from three nested levels to one and the `FormatError`/`runFailed` handling isn't duplicated across the two branches anymore. Behavior unchanged; `TestErrorExitCode` still green and I smoke-tested the .lgb path through the new helper.

PR #410, opened 2026-07-07
refactor(rt,core): purify clojure.core — let-go.core / let-go.types for lg-isms

nnunley:
## Summary

Purify `clojure.core` so it reads as a faithful Clojure surface. `clojure.core`
carried **207 public names absent from both JVM Clojure and babashka** (the
`clojure ∩ bb` oracle); this relocates the lg-specific extras out of it while
keeping unqualified use working (auto-refer'd baselines) and shadowable by an
explicit `:refer`.

> **Stacked on #408** (`fix/refer-shadows-core`). This PR's parent commit is the
> refer-shadow fix from #408, so the diff shows both commits until #408 merges —
> review the `refactor(rt,core): purify clojure.core …` commit; the purge depends
> on that fix (an explicit `:refer` must shadow the auto-refer'd baselines).

## What moves

- **`let-go.core`** (auto-refer'd baseline): `gt lt ge le` (Clojure uses `> < >= <=`
  — the binops are now the native `clojure.core` primitives, were `(def > gt)`
  aliases), `now base64-* lines str-replace-first parse-int` (`parse-long` is the
  native name), `spy str-join range? contains-val?` (gogen-lowered).
- **`let-go.types`** (focused, visible): `array? bigint? big-int?` — none exist in
  `clojure.core`/bb.
- **`clojure.set`**: `difference`/`intersection` removed from `clojure.core`.

## Mechanism

- Baseline registry (`vm.AddBaselineNamespace`/`isBaselineNS`; `lgBaselineNSs`/
  `referBaselines`/`LgBaselineNSNames`): the lg-baseline namespaces are
  lowest-priority in `Namespace.Lookup`, so an explicit `:refer :all` shadows them.
- `eval.go` runs the baseline chunks eagerly at boot (they're auto-refer'd but
  never `require`d, so on-demand load never fires).
- **Fix the bootstrap Go `nsMacro`**: it silently dropped `:require` (an
  implementation accident) — so a namespace loaded before core.lg's own ns macro
  (i.e. core itself) couldn't import deps. It now emits require/refer/alias forms,
  so `clojure.core` imports its type predicates via a real
  `(ns core (:require [let-go.types :refer [array? bigint?]]))`.

## Testing

`pkg/vm` + `pkg/rt` + `pkg/compiler` (378), the `.lg` suite (171), the external
xsofy corpus (**2733 assertions, 0 fail**), and the vendored jank
`clojure-test-suite` (**233**, 0 compile-skip). `make check-generated` in lockstep.

## Deferred (design captured, follow-up change)

The remaining tail — re-home the `clojure.string` primitive veneer, IO isms
(`open`/`close!`/`write!`) → `io` ns, and privatize ~82 bucket-③ internals — is
entangled and gated on a new **annotation-driven primitive registrar**
(`//lg:ns` / `//lg:name` / `//lg:private`) plus a coercing Fn-adapter. Tracked as
a separate change/issue.

## Note

The local `ir-stress` pre-push ratchet trips on a **corpus-size drift** (2360 vs a
stale 2347 baseline; `main` is already ~2366 from prior merges) — failure count is
unchanged (11) and this change adds none. Stale-baseline / repo-health item,
unrelated to this change.


nnunley:
Design for the remaining namespace cleanup (string veneer re-home, IO isms → `io` ns, bucket-③ privatization) and the annotation-driven primitive registrar + Fn-adapter that unblocks it: #411

nooga:
Requesting changes — the PR description's jank-suite number doesn't match what I get running it.

I built this branch in an isolated worktree against the exact submodule commit pinned here (`41290a61`) and ran the full suite:

```
TOTALS: files=229 assertions: pass=5453 fail=0 | skipped: compile=3 panic=0 runtime=0
```

Three compile-skips (`mod`, `quot`, `rem`), all: `Can't resolve p/big-int? in this context`. That's a real regression, not env drift — contradicts the "233, 0 compile-skip" in the description.

Root cause: `portability.lg` dropped `(def big-int? clojure.core/big-int?)` in favor of just adding `[let-go.types :refer [big-int?]]` to its own `ns` form. That makes `big-int?` resolve *inside* `portability.lg`, but interns nothing there for other namespaces to pick up — `Namespace.Lookup` only ever sees a referred/aliased namespace's own local vars (`localVar`), never symbols it merely sees through its own refers. Confirmed with a minimal repro outside this codepath entirely:

```clojure
(ns a) (defn f [] :x)
(ns b (:require [a :refer [f]]))       ; b sees f via refer, doesn't intern it
(ns c (:require [b :refer [f]]))
(f)   ; => Can't resolve f in this context
```

`star.cljc`'s unqualified `:refer [big-int? …]` still passes, but only because `star.cljc` *also* gets `let-go.types` for free via the ordinary baseline auto-refer — that's masking the bug, not validating the re-export. `rem`/`quot`/`mod` use the qualified `p/big-int?` form (aliased to `clojure.core-test.portability`), which has no baseline fallback, so it fails outright.

Worth noting CI is all green here because the jank suite is submodule-gated and isn't wired into the standard CI jobs — this only ever surfaces on a manual run, which is presumably what happened between when the numbers in the description were captured and now.

Fix: put back a real local def in `portability.lg` — `(def big-int? let-go.types/big-int?)` — instead of relying on the ns-level `:refer` to do double duty as a re-export.

Also, minor cleanup while I'm here: `_ = lgTypes` at the end of the `let-go.types` registration block in `pkg/rt/lang.go` is dead — `lgTypes` is already used several times above it, so that line does nothing and can go.

Everything else in this PR checks out — the baseline-precedence `Lookup` change is correct and, as a good side effect, fixes a real pre-existing bug where `:refer [only-this]` wasn't actually enforced for unqualified resolution (nice catch, and `refer_shadows_core_test.lg`'s `hidden` assertion pins it down well). Once the `big-int?` regression is fixed, this looks mergeable.

nnunley:
Fixed the `p/big-int?` regression you flagged — pushed as a second commit (`fix(vm): qualified alias resolution follows target refers`).

**Root cause:** this PR moved `big-int?` into `let-go.types` and rewrote the portability shim from interning it (`(def big-int? clojure.core/big-int?)`) to `[let-go.types :refer [big-int?]]`. The `mod`/`quot`/`rem` suite files use it **qualified** (`p/big-int?`), and let-go's qualified-alias resolution only consulted the target ns's *interned* vars — it didn't follow the target's `:refer`s — so `p/big-int?` resolved to nil. (`let-go.types` was already a baseline; the gap was purely in the qualified path.)

**Fix:** extracted the refer-resolution logic into `Namespace.lookupViaRefers` and call it from the qualified path when `localVar` misses, preserving `:all`/`:only`/baseline/private semantics. This matches Clojure's `alias/sym` behavior (qualified refs see the target's refers), and fixes the whole class, not just `big-int?`.

**Verified:** `mod`/`quot`/`rem` load clean; full suite 1188 pass; jank 233 pass; `gogen-parity.sh` byte-identical (zero bytecode change — resolution-only). The three prior compile-skips are gone, so the description's jank count now holds.

mparrett:
I found no remaining code-level blocker on the latest head. The `big-int?` regression was fixed, and I reproduced the relevant suites successfully, including:

- normal + bootstrap package/language tests
- generated-artifact freshness
- jank compatibility suite: 232 files, 5,657 assertions, 0 failures/skips

We can do a final check once rebase/conflict resolution happens. If it's trivial, could consider that an optional step.

nnunley:
Addressed and ready for re-review, @nooga @mparrett:

- **`big-int?` regression** — fixed (qualified-alias resolution now follows the target's refers via `Namespace.lookupViaRefers`). The fix is preserved through the rebase.
- **Dead code** — removed the `_ = lgTypes` line in `pkg/rt/lang.go`; `lgTypes` is used above it, so it was a no-op.
- **Rebase** — squashed to a single commit and rebased conflict-free onto current `main`. The `Lookup` conflict against main's inline core-baseline logic resolved in favor of the generalized `lookupViaRefers` (handles clojure.core + let-go.core baselines, honors `:all`/`:only`/private).

Gates green: `check-generated` OK, `gogen-parity` PARITY (both backends), `ir-stress` 20/20, full `go test ./...` incl. e2e.

mparrett (approved):
Thanks for addressing the findings. Looks good.

nooga (approved):
Verified backward compat (unqualified resolution intact via baseline refers), regenerated artifacts in sync (`make check-generated` clean), and full test suite including `-tags bootstrap` green on the branch. No stray qualified references to moved names left in the tree. Nice, well-gated refactor — approve pending #408 landing/rebase as noted in the PR description.

PR #415, opened 2026-07-08
fix(core): scope record fields in defrecord protocol-method bodies

### LINKED ISSUE #414: defrecord doesn't lexically scope fields in protocol-method bodies
## Summary

`deftype` rewrites bare field references in inline protocol-method bodies so a
field name resolves to that field, but `defrecord` hands method bodies to
`extend-type` verbatim — so a bare field reference in a `defrecord` method fails
to compile.

## Repro

```clojure
(defprotocol P (get-a [this]))
(defrecord R [a] P (get-a [this] a))   ; bare field `a`
(get-a (->R 42))
```

```
CompileError: Can't resolve a in this context
```

`deftype` with the same shape works:

```clojure
(deftype T [a] P (get-a [this] a))     ; fine
```

## Cause / proposed fix

`deftype`'s macro rewrites method bodies (its `-dt-rewrite*` family in
`pkg/rt/core/core.lg`) because let-go has no compiler-level field scope;
`defrecord` never did the same. The fix generalizes that rewrite to take a
field-read emitter: `deftype` keeps interop `(.field recv)`, `defrecord` emits
keyword access `(:field recv)` (records are map-backed — canonical and
nil-safe). `case` test-constants that share a field name stay literal (only
field *reads* are rewritten, since `case` quotes its constants).

I have a fix with tests ready to open as a PR from a fork branch — this came up
porting weavejester/dependency + integrant to let-go.

abogoyavlensky:
deftype rewrites bare field references in inline protocol-method bodies (its -dt-rewrite* family), but defrecord handed method bodies verbatim to extend-type, so a bare field read failed to compile:

    CompileError: Can't resolve <field> in this context

Generalize the field-rewrite family to take a field-read emitter: deftype keeps interop .field access; defrecord emits keyword access (:field recv) (records are map-backed — canonical and nil-safe). case test-constants that share a field name stay literal (only field reads are rewritten), since case quotes its constants. Adds test/defrecord_field_scope_test.lg; regenerates the embedded core bundle.

Resolves: #414


nnunley (approved):
Hrm. You'll need to resolve merge conflicts,  but it looks good to me.

PR #422, opened 2026-07-08
feat(ir): liveness analysis pass + block-arg infrastructure

nnunley:
Adds a complete liveness analysis pass with block-arg-aware resolution, plus the supporting infrastructure that later optimization passes build on. All tests green (855 passes); generated artifacts in sync.

PR #423, opened 2026-07-09
feat(ir): pre-execution cleanup pass + lazy bundle source-map decode

nnunley:
**Stacked on #422** (liveness / block-arg infrastructure). The two net-new commits are the ones below; review the top two — the rest belong to #422 and merge with it.

## 1. Pre-execution cleanup pass (`ir.passes.cleanup`)
A pre-execution IR pass that compacts dead (DCE-tombstoned) block-params + their in-edge args, strips `:pop` markers, and DCE-sweeps orphaned insts to a fixed point. Runs on the **index-RPN pre-execution path only** (not the shared `optimize-fn` pipeline), so mainline bytecode/gogen lowering stays **byte-identical**. Promotes an earlier Go-side `compactDeadParams` spike into a real IR→IR pass — the spike now decodes already-clean IR and its `compactDeadParams` is deleted (`validateLiveInvariants` still holds).

- **Census:** 42→29 block-params (13 dead params + edge-arg positions stripped) over `pipelineCorpus`.
- **Un-gameable proof:** `TestCleanupCorpusInvariants` re-derives `validate.lg`'s arity law + tombstone-freedom over the corpus; `TestCleanupCompactsDeadBlockParams`; spike four-kernel parity intact.

## 2. perf(bytecode): lazy source-map materialization on bundle decode
Defers decoding each chunk's source map at `.lgb` load until the first `Lookup` (error/stack-trace) — source maps are a cold path. A new `DecodeToExecUnitBytes` entry captures each source-map section as a **zero-copy sub-slice** of the resident bundle (Reader gains an offset counter + `Slice`) and stores it **closure-free** in the `SourceMap`, decoded via `binary.Uvarint` on demand: **one allocation per source-map chunk instead of two**. No wire-format change.

- **`BenchmarkInitFromLGB`:** allocs/op 9148→**8355** (below the 8938 baseline), bytes/op **−24%**; `bench-ratchet` gate green (0 regressions).
- **Correctness:** lazy `LookupSource` is identical to eager across every chunk/IP of the real core bundle (`TestDecodeBytesLazySourceMapMatchesEager`) + lazy unit tests.

## Gates
`go test ./...` green (18 pkgs) · `make check-generated` byte-identical · `make parity-check` jank + lower-go PARITY.


nnunley:
Closing as superseded — already in `main`. The cleanup pass (`ir/passes/cleanup.lg`) rebases to an empty commit (main has it), and the lazy source-map optimization (`NewLazySourceMapRaw`/`decodeRawSourceMap` in `pkg/vm/source.go`) is likewise already upstream. Stale pre-merge history; no outstanding review comments.

PR #424, opened 2026-07-09
Runtime-only build: execute precompiled bytecode, no compiler linked

mparrett:
## Why

Embedding let-go, or shipping a small self-contained artifact, mostly means running precompiled code, not compiling source at runtime. Today every build links the full reader, compiler, and resolver even when the deployed program never reads a form.

A runtime-only binary drops them, which buys three things:

1. **No dynamic source execution.** A binary that carries `eval`, `load-string`, `read-string`, and a source-loading `require` can be driven into executing new source at runtime. `lg-runtime` removes that path by construction (the compiler is not in the binary), so the deployed artifact runs only bytecode compiled ahead of time by your toolchain. This is a deployment-mode property, not a sandbox; see notes.
2. **A smaller, more stable embedding surface**: `vm` + `bytecode`, without the compiler internals.
3. **Size trims**: modest on the stock build, real under TinyGo (below).

## What it does

`cmd/lg-runtime` is a dedicated main package that never imports `pkg/compiler` or `pkg/resolver`: no build tag, no gating in the root files. (The earlier draft tagged the root `main`; per review that duplicated the CLI lifecycle, so every future CLI fix would have needed two implementations.) `go list -deps ./cmd/lg-runtime` contains neither package, and `go tool nm` finds 251 `pkg/compiler` + 18 `pkg/resolver` symbols in the full build, 0 of each in `lg-runtime`. Fed a `.lg` source file, it rejects it (`invalid magic bytes`) rather than compiling it. Both properties are covered by e2e tests: a dep-graph guard (`go list -deps` must exclude compiler and resolver) and black-box runs.

- **`pkg/rt/run.go` + `pkg/rt/boot.go`** — the shared compiler-free run lifecycle: `LoadCore` boots from the embedded core `.lgb`; `DecodeExecUnit`/`RunExecUnit` decode and replay program bundles; `SetCommandLineArgs` and `InstallPersistentStorage` cover the CLI contract. The full `lg` now routes its `.lgb` and appended-bundle paths through the same helpers, so a CLI fix lands once.
- **`cmd/lg-runtime`** — boot, then run a precompiled `.lgb` with the full lifecycle: `*command-line-args*` from the positionals after the program, build metadata, persistent storage, `LG_RESOURCE_PATHS` resources, pod shutdown. No REPL, no `eval`/`load-string`/`read-string`, no `-c`/`-w`.
- **Bundle mode** — `lg -b app -bundle-base lg-runtime app.lg` produces a standalone, compiler-free executable: `lg-runtime` detects the appended payload at startup, including embedded resources and the baked storage id.
- A bytecode-only `NSLoader` replays bundled namespace chunks on `require` and rejects source loading: the enforcement point.

Closing the review findings surfaced two bugs, both fixed with regression tests:

- **Failed namespace replay was invisible.** The namespace is pre-registered as a placeholder during decoding, so when its chunk failed to replay, the registry found the placeholder, treated the load as a success, and consumed the retry marker, caching a half-initialized namespace. The loader now restores the marker on failure, so `require` reports it and a later require retries. Fixed at the loader rather than the registry because a loader returning nil against a pre-existing namespace is also the legitimate native+lazy pattern (`gogen`), where the pre-existing namespace is the result; the tests pin both behaviors.
- **lg baseline namespaces were nil stubs.** `let-go.core` and `let-go.types` are auto-referred into every namespace but never explicitly required, so on-demand loading never fires for them. The runtime-only boot never replayed their chunks, and calling `str-join`, `spy`, or `semver/*` died with "nil is not a function". `LoadCore` now replays them eagerly, mirroring `loadPrecompiledBundle`.

The default build's behavior is unchanged; it just calls the shared helpers.

## Validation

Same program, run source-at-runtime (full) vs precompiled (`lg-runtime`), release-shaped builds (`-ldflags "-s -w"`, what the Makefile and goreleaser pass):

| | full | lg-runtime | delta |
|---|---|---|---|
| stock size | 13.1 MB | 11.6 MB | −11% |
| stock cold start | ~6 ms | ~6 ms | ~0 |

Cold start is hyperfine on an M1 launching a trivial precompiled `.lgb`. The core-`.lgb` replay dominates boot and is unchanged by this, so on the stock build the payoff is size, not startup.

**Note re: TinyGo**
TinyGo is being casually investigated, locally. Under TinyGo the size delta compounds: dropping the compiler's package `init()` (it builds the reader/macro tables at every start) removes a sizable slice of an already-lean binary — −28% (4.6 MB → 3.3 MB). Caveat: vanilla `main` doesn't build under TinyGo today (its `reflect` is incomplete: the `reflect.Type.Method()` trap #466 notes), so that number comes from a separate enablement branch, not this PR as-is. Same for wasi: a `GOOS=wasip1` build needs `pkg/rt/term.go` gated off wasip1 first, so this PR makes no wasi claims; that's a separate change, in #466. Dedicated TinyGo work is tentatively coming.

## Try it

A dragon curve, compiled once with the full build, then run by the runtime-only binary:

```
go build -o lg .
go build -o lg-runtime ./cmd/lg-runtime
./lg -c dragon.lgb dragon.lg     # compile with the full build (2.5 KB .lgb)
./lg-runtime dragon.lgb          # run it with no compiler in the binary

# or as one standalone, compiler-free executable:
./lg -b dragon -bundle-base lg-runtime dragon.lg
./dragon
```

<details>
<summary><code>dragon.lg</code> (~30 lines)</summary>

```clojure
;; Dragon curve — the classic paper-folding fractal, rendered in the terminal.
;; Each iteration folds the path: s' = s + [:l] + reverse(flip s).

(def iterations 9)

(defn turns [n]
  (loop [s [] i 0]
    (if (= i n)
      s
      (recur (vec (concat s [:l] (map {:l :r, :r :l} (reverse s)))) (inc i)))))

(defn walk
  "Follow the turn sequence on a unit grid, returning the set of visited cells."
  [ts]
  (loop [visited #{[0 0]} pos [0 0] dir 0 ts (seq ts)]
    (let [step ([[1 0] [0 1] [-1 0] [0 -1]] dir)
          pos' [(+ (pos 0) (step 0)) (+ (pos 1) (step 1))]
          visited' (conj visited pos')]
      (if ts
        (recur visited' pos'
               (mod (+ dir (if (= (first ts) :l) 1 3)) 4)
               (next ts))
        visited'))))

(let [cells (walk (turns iterations))
      xs (map first cells)
      ys (map second cells)
      x0 (apply min xs) x1 (apply max xs)
      y0 (apply min ys) y1 (apply max ys)]
  (doseq [y (range y1 (dec y0) -1)]
    (println (apply str (map (fn [x] (if (cells [x y]) "##" "  "))
                             (range x0 (inc x1)))))))
```

</details>

```text
                                ####    ####
                              ######  ######
                              ##################
                              ##################
                        ####  ########      ####
                      ######  ########    ######
                      ############        ####
                      ##############
                      ################
                          ############
                          ########
                              ##############
                        ####  ################
                      ######  ################
                      ####################
                      ##############################
                      ################################
                          ############################
                          ########################
    ########                  ######################
    ##########          ####  ########      ##########
########  ####        ######  ########  ##############
########              ############      ##########
  ######              ##############      ########
    ########    ####  ################
############  ######  ################
##################################
  ##################################
            ##########      ##########
        ##############  ##############
        ##########      ##########
          ########        ########
```

The output matches a source run under the full build byte for byte. Fed the source instead, the runtime-only binary has no way to read it:

```
$ ./lg-runtime dragon.lg
error: invalid magic bytes: 3b3b2044
```

(`3b3b2044` is `;; D` — the first bytes of the source file's leading comment.)

## Notes for review

- **Guarantee scope.** "No dynamic source" is a deployment-mode property, not a sandbox: `lg-runtime` runs any `.lgb` it is handed (no signature or provenance check), and bytecode retains the full native `rt` surface (OS, network, processes). Untrusted code still needs OS- or wasm-level sandboxing around the process.
- **Follow-ups (called out, not done here).**
  1. The full build still boots via `pkg/compiler/eval.go`'s `loadPrecompiledBundle`, which mirrors `rt.LoadCore`; unifying them is left out to avoid churning the working boot path in the same change.
  2. The resolver's `execPrecompiled` has the same silent-failure shape the loader fix addresses; the same one-line treatment applies there if wanted.
  3. goreleaser/Makefile don't build `lg-runtime`; happy to add a target if you want it as a released artifact.


mparrett:
@nooga Somewhat experimental, and lower priority. I'd be curious to hear your thoughts on it when you have a chance.

mparrett:
Self-review of draft: The idea fits let-go, especially its “bytecode + VM are first-class,” portability, and embedding goals. The PR requires further development before consideration.

### Priority findings

1. **[P1] Runtime-only drops command-line arguments.**

   [`[lg_runtime.go](https://github.com/nooga/let-go/blob/e4fb02a4244fa118a28f21c96dd80161f990c355/lg_runtime.go#L41-L79)`](https://github.com/nooga/let-go/blob/e4fb02a4244fa118a28f21c96dd80161f990c355/lg_runtime.go#L41-L79) treats only `os.Args[1]` as meaningful and never publishes `os.Args[2:]` to `*command-line-args*`.

   I reproduced this with the same `.lgb`:

   ```text
   lg app.lgb one two          => (one two)
   lg-runtime app.lgb one two  => nil
   ```

   That violates documented CLI behavior. Move the argument initialization into shared, compiler-free code and add an e2e assertion.

2. **[P1] The advertised WASI build does not currently compile.**

   This fails:

   ```bash
   GOOS=wasip1 GOARCH=wasm go build -tags runtime_only .
   ```

   `pkg/rt/term.go` is selected for `wasip1` and references unavailable `x/sys/unix` facilities such as `SIGWINCH`, `PollFd`, and `Poll`. Either add a WASI terminal stub/build constraint plus CI coverage, or remove the WASI claim from this PR.

3. **[P1] The trust claim needs a narrower threat model.**

   The build proves “the compiler and source-loading path are absent.” It does not prove that executed code is trusted:

   - It accepts any supplied `.lgb`; there is no signature or provenance check.
   - Bytecode retains access to the broad native `pkg/rt` surface, including OS/network/process capabilities.
   - Compiling untrusted source with a trusted compiler does not make its behavior safe.

   Consider positioning this as a **compiler-free/no-dynamic-source deployment mode**, not as a safe environment for untrusted scripts. The PR’s “not a sandbox” note is good, but it undercuts some earlier security wording.

4. **[P2] Namespace replay errors can still be mistaken for success.**

   [`[bytecodeNSLoader.Load](https://github.com/nooga/let-go/blob/e4fb02a4244fa118a28f21c96dd80161f990c355/pkg/rt/boot.go#L65-L80)`](https://github.com/nooga/let-go/blob/e4fb02a4244fa118a28f21c96dd80161f990c355/pkg/rt/boot.go#L65-L80) returns `nil` on replay failure, expecting the registry to retain the retry marker. But decoding has already registered a placeholder namespace. `LookupOrRegisterNS` sees that placeholder after the loader returns and returns it before restoring `nsNeedsLoad`.

   The result can be a half-initialized namespace treated as successfully loaded. Add a regression test for a pre-existing, marked namespace whose loader fails—not only the current missing-namespace case.

### Cross-concerns

The alternate build-tagged `main` already duplicates bytecode execution and has drifted from the normal CLI lifecycle: arguments, build metadata, persistent storage, resource handling, and pod shutdown are absent.

A preferred alternative:

- a shared compiler-free bytecode runner/lifecycle package; and
- probably a dedicated `cmd/lg-runtime` entry point, unless this mode is intentionally going to become a fully supported root build variant.

That would make its product boundary clearer and avoid every future CLI fix needing two implementations. It also keeps this distinct from the project’s longer-term generated-Go path for authored minimal binaries, described in the [[contribution policy](https://github.com/nooga/let-go/blob/main/docs/contribution-policy.md)](https://github.com/nooga/let-go/blob/main/docs/contribution-policy.md).

Validation otherwise looks healthy: targeted runtime-only e2e tests and `go test -short ./...` pass, and current GitHub checks are green. The branch is presently 11 commits behind `main`, so I would rebase and rerun the cross-build checks after resolving the above.

nooga (approved):
Verified: `go build .` and `go build -tags runtime_only .` both succeed, default build/tests unaffected, and `test/e2e/TestRuntimeOnly*` pass. Traced the `LoadCore` vs `compiler.loadPrecompiledBundle` divergence (no eager hybrid-namespace loading) — confirmed safe today since `async`'s bundled content is macro-only and fully expanded before a `.lgb` exists, but worth a one-line comment in `boot.go` noting that assumption for future hybrid namespaces. LGTM.

nooga:
@mparrett despite the LGTM above (code-wise everything looks fine).

I think this makes sense in general as a separate binary or a bundle mode. Clojure is famously dynamic in nature so traditionally not having the compiler at runtime is unusual - compile time and runtime are quite fractal: macros compile and run while compiling the code while running the code... That being said it seems practical for many cases. We could even extend it to Go lowering as a mode i.e. "ship without the compiler".

Overall, I'm in favor if this does not increase maintenance burden significantly. Also curious about @nnunley 's thoughts on this.

One thing that stuck out in the description: measured cold start times of 27ms (both before and after) look way off - 1.11.1 boots in 8ms avg on my M1.

mparrett:
Thanks, and good catch on the cold start. The 27ms was a bad measurement (loaded machine); I re-benchmarked stripped release builds and stock boots in ~7.8ms, which matches your 8ms on 1.11.1. Updated the body. The same re-run also showed the TinyGo cold-start numbers I'd quoted were in-guest timing rather than wall-clock, so I pulled them and kept the size delta (~28%), which is the part that reproduces.

On shape: agreed, and I'd lean the same way. A dedicated `cmd/lg-runtime` (or a bundle mode) over the build-tagged `main`, with a shared compiler-free runner so the CLI lifecycle (args, build metadata, shutdown) lives in one place. That answers the maintenance-burden worry directly: no second entry point to keep in sync on every CLI fix. The "extend to Go lowering / ship without the compiler" angle fits the same shape.

And fair point on the dynamism — a Lisp with no compiler at runtime is an odd thing to say out loud. Maybe makes more sense framed as a "deploy mode for code that's done compiling" -  not as a general runtime.

Will see what @nnunley's has to say before any major rework.

mparrett:
Reworked per the review and the shape discussion above; branch rebased onto main and force-pushed. What changed:

- **Shape**: dedicated `cmd/lg-runtime` plus a shared compiler-free run lifecycle in `pkg/rt` (`DecodeExecUnit`/`RunExecUnit`/`SetCommandLineArgs`/`InstallPersistentStorage`). The `runtime_only` build tag is gone (`lg.go`/`wasm.go`/`lg_repl*.go` are back to main's shape), and the full build's `.lgb` and bundle paths now use the same helpers, so a CLI fix lands once. This is the maintenance-burden answer: there is no second CLI to keep in sync.
- **Args (P1)**: `*command-line-args*` is published from the positionals after the program; e2e asserts the set and empty cases.
- **wasi claim (P1)**: dropped from the body. `GOOS=wasip1` needs a `pkg/rt/term.go` capability split first; that composition is a separate change.
- **Trust wording (P1)**: reframed as a compiler-free deployment mode with an explicit not-a-sandbox scope note.
- **Replay failure (P2)**: fixed at the loader: it restores the needs-load marker before returning nil, so `require` reports the failure and retries later. Fixing it in the registry instead broke the legitimate native+lazy namespace pattern (`gogen`); the regression tests pin both behaviors.
- **Two bugs found en route**: the baseline namespaces (`let-go.core`, `let-go.types`) were nil stubs under the runtime-only boot (your `loadPrecompiledBundle` trace pointed straight at it), so `LoadCore` now replays them eagerly, and the hybrid/async assumption you asked about is documented in `boot.go`.
- **New**: bundle mode. `lg -b app -bundle-base lg-runtime app.lg` yields a standalone, compiler-free executable, e2e-covered.

Full suite and linux/js/plan9 cross-builds are green; body updated with re-measured numbers (sizes, cold start) and a lane caveat on the TinyGo figure.

Ready for another look; @nnunley's read on the overall shape is still welcome.

nooga (approved):
Verified independently on the current head (ce8382b4), which supersedes the one formal APPROVED review (anchored on the pre-rework commit e4fb02a424):

- `go build ./...` and `go test ./... -count=1` pass locally, including the new `pkg/rt/nsload_retry_test.go` and `test/e2e/runtime_only_test.go`.
- Traced the namespace-replay-failure fix in `pkg/rt/lang.go`'s `LookupOrRegisterNS`: when a placeholder namespace already exists (the bytecode-decode case), the function returns that placeholder as soon as the registry lookup after a failed `Load` succeeds — it never reaches the generic 'mark for retry' branch. So `bytecodeNSLoader.Load` restoring `MarkNSNeedsLoad` itself (pkg/rt/boot.go) is the correct fix point, not a workaround; the new tests (`TestBytecodeNSLoaderReplayFailure`, `TestRequireNSRetryAfterLoaderFailure`, `TestRequireNSNativeNamespaceLoaderNoop`) pin both the failure and the legitimate native+lazy (gogen) contract without regressing it.
- All P1 items from the self-review (dropped CLI args, broken WASI claim, overclaimed sandbox wording) and the P2 (silent replay failure) are addressed in the reworked diff, not just described in the PR body.
- `test/e2e/runtime_only_test.go` follows the existing e2e convention in this repo (same shape as `command_line_args_test.go`/`resources_test.go`) and covers the right surface: dep-graph guard (`go list -deps` excludes compiler/resolver), args, namespace require from bundle, baseline-namespace replay, bundle mode, embedded + env resource paths, and source rejection.
- Cold-start and TinyGo numbers in the PR body were corrected after the owner flagged the original 27ms figure as a bad (loaded-machine) measurement; the TinyGo number is honestly caveated as coming from a separate enablement branch.

No blocking issues. One process note for the record: the sole formal review is on a commit predating the force-push that changed the PR's whole shape (build-tag `main` -> dedicated `cmd/lg-runtime`), so it doesn't reflect the current diff — worth a fresh review pass from a human before merge even though this automated pass found nothing to block on.

mparrett:

pkg/bundle/bundle.go
```diff
@@ -161,6 +161,33 @@ func payloadFitsFile(lgbSize, resSize, idSize uint64, trailerLen, total int64) b
...
+// every entry point that can run as a standalone bundle (lg, lg-runtime).
+func ReadBundledSelf() (lgb []byte, res []byte, storeID string) {
+	candidates := make([]string, 0, 3)
+	if exe, err := os.Executable(); err == nil && exe != "" {
+		candidates = append(candidates, exe)
+	}
+	if len(os.Args) > 0 && os.Args[0] != "" {
+		candidates = append(candidates, os.Args[0])
+	}
+	candidates = append(candidates, "/proc/self/exe")
```
nnunley: Hrm. This needs a little more guarding. It's only valid on linunx systems. I imagine a windows user or macos user could have a malicious user add that directory, and hijack the system, potentialy.

mparrett: Good catch. Addressed in c3015ea by gating the `/proc/self/exe` behind `runtime.GOOS == "linux"` to mitigate the off-Linux risk reading a malicious/unintended file.

I considered dropping it entirely, but kept it since it covers the case where the binary is unlinked while running (self-contained bundles?) and `os.Executable()`'s path goes stale (should resolve to /proc/self/exe on linux).

> There is no guarantee that the path is still pointing to the correct executable. If a symlink was used to start the process, depending on the operating system, the result might be the symlink or the path it pointed to. If a stable result is needed, [path/filepath.EvalSymlinks](https://pkg.go.dev/path/filepath#EvalSymlinks) might help.

Ref: https://pkg.go.dev/os#Executable

nnunley:
Ok. I left a comment elsewhere. I agree that in general, while being able to eliminate eval is neat, that the way to get to what you're suggesting here is through tree-shaking and fine(r) grained runtime surfaces. I left a little more detail in #652 on what I'm thinking.

The idea I was having is that if we can move lang.go to to first class go packages, with appropriate metadata for lg, we can start loading subsets of the code based on functionality required. the AOT pass should be able to use DCE to avoid importing things that aren't required.

PR #428, opened 2026-07-09
feat(rt): read-line with no args reads stdin

ingydotnet:
(read-line) required an IOHandle arg, so Clojure code calling the 0-arity form failed with 'read-line expects 1 arg'. Default to the stdin handle when called with no args, like Clojure.


nnunley (commented):
Is the fix just adding the comment to suggest readers scroll backwards to the stdin implementation? Or is there something else pending?

ingydotnet:
I'm not sure what happened here. Closing this one.

PR #429, opened 2026-07-09
fix(compat): resolve clojure-test-suite incompatibilities

nnunley:
Advances `clojure-test-suite` from `41290a61` to accepted upstream `bd610c81` and uses the existing test-time patch-overlay mechanism for the targeted Unicode fixture correction. The overlay is the exact delta in jank-lang/clojure-test-suite#939, preserves let-go's full Unicode-scalar `char` behavior, applies idempotently, and is reverted after the test process so the submodule remains clean. This avoids pinning an unmerged fork commit.

## Compatibility gaps fixed

- preserve lazy sequence contracts for `concat`, `drop`, `map*`, and all `partition` arities;
- split keyword namespaces at the first slash, matching symbols;
- make `identical?` safe and identity-based for non-comparable collections;
- stop `run!` when its callback returns `Reduced`;
- preserve BigInt precision in `bigdec`, reject non-finite inputs, and canonicalize decimal signed zero;
- accept empty seqable inputs in typed-array constructors;
- match JVM `contains?` behavior for arrays, map entries, and unsupported collection/key shapes.

Each runtime gap has an in-repo regression test. Overlay tests cover apply, idempotency, cleanup, and the Unicode-scalar expectation.

## Verification

- current suite: **242 files, 6,311 assertions, 0 failures**;
- skipped: **0 compile, 0 panic, 0 runtime**;
- `make check-generated`;
- `go test -short -count=1 -timeout 600s ./...`;
- `go vet ./...`.

All generation and verification above ran serially with bounded Go heaps after the earlier high-memory incident.

Part of #257.


nooga (approved):

nooga (changes requested):
Approach and transparency here are good — clean gitlink-only bump with honest before/after numbers. Per your own notes and the failing CI `build` job, this needs the 18 new failures (remove 1-arity, realized? on lazy seqs, ArrayVector comparison, plus the 14 triaged-pending assertion fails) resolved or explicitly skip-listed with rationale before it comes out of draft.

nnunley:
Rebased onto current main (`8c31a809`) to refresh CI. The prior `build` failure was a stale-base WASM test issue (`term_wasm.go:91`, `syscall/js: call of Value.Int on undefined`), unrelated to the suite bump itself — it should clear on the fresh base. Keeping this in draft intentionally, since bumping the suite raises the failing-case count until those are worked down.

nnunley:
Re-triaged against current main (which now carries #447's update-keys/update-vals + remove 1-arity + regex nil-groups, and the typed-catch stack): **6 of the 18 failures are fixed on main** and no longer need skip-listing. The remaining 12 are skip-listed in `test/zz_compat_test.go` with per-test rationales: 4× multi-segment keyword parsing (`a/b/c`), 2× `realized?`-on-LazySeq contract, ArrayVector `identical?`, `drop` empty-result type, `map` 1-arity laziness, `(char 65895)` bounds, `run!` reduced handling, `*'` -0.0M canonicalization. None were honest quick-fixes — each needs a design decision, so they're documented rather than papered over. Suite: 242 files, 6012 assertions passing, 24 skip-listed. Leaving in draft for your call on whether the skip-list rationale bar is met.

nnunley:
Current failure classification from the `build` job (`86910036236`):

- **14 failed assertions + 3 errors across 11 fixtures** (17 total), not 18.
- **10 sequence contract failures**: `concat`, `drop`, `map`, and `partition` return eager or noncanonical empty values instead of the expected lazy-seq shapes.
- **4 keyword namespace-parsing failures**: `Keyword.Namespaced` accepts exactly one slash, so `(keyword "a/b/c")` is incorrectly treated as simple rather than splitting at the first slash.
- **1 identity failure**: `identical?` uses Go interface equality and panics on non-comparable `vm.ArrayVector`.
- **1 `run!` failure**: its reducer discards the procedure's `Reduced` result, so reduction does not terminate early.
- **1 compatibility decision**: let-go `char` accepts Unicode scalars through `0x10FFFF`; JVM Clojure rejects values above `0xFFFF`.

All other checks are green, including generated artifacts, gogen diff, lint, differential, and CodeQL. `remove` passes in this run, so the description's earlier failure list is stale.

Rebase assessment: this run used base `d80a5e5a`; current `nooga/main` is `a1d8bbbc`. The relevant implementations for all five categories are unchanged between those revisions, so rebasing is necessary to refresh CI but is expected to fix **zero of these current 17 failures**. Please rerun after rebase to confirm.

Also, current head `e5e05782` is a gitlink-only commit. `test/zz_compat_test.go` is unchanged from the base, so the skip list described in the prior comment is not present in the PR as currently published.

nnunley:
State update after replacing the original bump-only head:

- rebased the full compatibility stack onto current `nooga/main` (`362aa1e4`);
- regenerated each affected commit bottom-up under the bounded-memory workflow;
- pushed verified head `e9cb69cf`;
- GitHub reports the PR mergeable;
- current jank suite result: **242 files / 6,311 pass / 0 fail / 0 compile, panic, or runtime skips**;
- repository gates pass: generated-artifact check, short full Go suite, and `go vet ./...`.

The PR remains draft only because the Unicode scalar fixture correction is still open as jank-lang/clojure-test-suite#939. Its head is green, mergeable, and currently pinned here; after upstream merge, the gitlink should move to the accepted upstream commit.

nnunley:
Follow-up: switched the Unicode fixture from the unmerged #939 gitlink pin to let-go's established patch-overlay workflow.

- restored the accepted suite pin `41290a61`;
- added `test/patches/lg-char-unicode-scalar.patch` matching upstream #939;
- generalized the existing overlay harness to apply/revert multiple patches idempotently;
- added hermetic tests for apply, duplicate-apply no-op, cleanup, and the `:lg` Unicode-scalar expectation;
- pushed head `067d757c`.

Fresh local evidence under the bounded-memory policy: patched suite **242 files / 6,311 pass / 0 fail / 0 skips**, clean submodule after the run, generated-artifact check green, short full Go suite green, and `go vet ./...` green. The PR no longer depends on #939 merging; current CI and re-review remain.

nnunley:
Correction to the prior overlay update: `41290a61` was the pre-upgrade pin and caused CI to run only 232 files, exposing an obsolete `take-nth nil` expectation.

Current head `e217691a` now has the intended shape:

- suite gitlink → accepted upstream `bd610c81` (the 242-file #939 base);
- minimal local overlay → exact #939 Unicode-scalar delta only;
- external `/simplify` follow-up `420905ec` preserved;
- the old-pin `take-nth` experiment excluded from the PR and preserved locally for separate evaluation.

Fresh local result: **242 files / 6,311 pass / 0 fail / 0 skips**; `take-nth` passes, the suite checkout is clean after patch cleanup, generated artifacts are synchronized, the bounded short full Go suite passes, and `go vet ./...` passes.

nnunley:
Final rebase/simplification state:

- rebased onto current `nooga/main` at `0e05d00b`; head is `1918f6c7`;
- remote gitlink is exactly accepted jank-suite main `bd610c81`;
- retained broad Unicode-scalar behavior through the hermetic `:lg` overlay;
- removed the now-upstream map-order overlay, made the Unicode assertion exercise the actual pinned suite, simplified persistent-vector identity comparison, and folded the duplicate overlay commit;
- full suite: 242 files, 6,311 assertions, 0 failures, 0 compile/panic/runtime skips;
- `make check-generated`, `go test -short -count=1 -timeout 600s ./...`, and `go vet ./...` pass under the bounded-memory policy;
- all current GitHub checks pass and GitHub reports the head mergeable.

The remaining blocked state is the stale earlier CHANGES_REQUESTED review; `nooga` is re-requested on the current head.

nnunley:
Rebased onto current main, regenerated the bundled artifacts (`make generate`; `generated.sums` conflict resolved), and CI is fully green on `747df190` — the `build` job runs the suite at the pinned revision with **pass=6311 fail=0** across all 242 files (verified locally as well). The failures from the earlier review no longer exist on this branch.

nooga (approved):
Verified beyond the diff: initialized the clojure-test-suite submodule at the new pin and ran TestClojureTestSuite myself — 0 failures, 0 errors across all 241 fixture files, submodule left clean after the run (patch overlay applies and reverts correctly). Full go test ./... -count=1, make check-generated, go vet, and the js/wasm + plan9/amd64 cross-builds are all clean.

Traced correctness on the trickier pieces rather than just reading them: the PersistentVector identity token's non-zero-sized struct correctly avoids Go's zero-size-allocation pointer-aliasing pitfall; every vector-producing operation (including WithMeta, matching real Clojure's with-meta semantics) mints a fresh identity; BigDecimal signed-zero canonicalization funnels through every construction path including the arithmetic fallback in numbers.go, with no bypass; identical?'s reflect.Comparable()-based split is sound; and contains?'s new Keyed/String/Seq-reject/Indexed dispatch matches real JVM Clojure's throw-on-unsupported-collection behavior rather than being an invented divergence.

One thing worth a quick sanity check (not blocking): summing the per-fixture pass counts in my run gives ~12,622 assertions, not the ~6,311 in the PR description (close to exactly double) — 0 failures either way, but the exact number doesn't match what I reproduced.

nnunley:
Good catch on the arithmetic — reproduced it. Summing every `pass=N` in the `-v` log gives 12,622 because the `TOTALS:` line itself matches (`assertions: pass=6311`), so the aggregate gets added to its own components: 6,311 + 6,311. Excluding that line, the 242 fixture lines sum to exactly 6,311, matching the description. 0 failures either way, as you said.

Issue #425, opened 2026-07-09
AOT native-entry: generate the Go entry frame, self-contain `gogen`, and stop the silent perf-cliff


nnunley:
## Context: the LG-engine Step-0 spike

A by-hand spike proved the native-Go AOT pipeline (lower `.lg` → typed Go → standalone binary) end-to-end. fib(34), single-arity `fib`, variadic `-main`, same program and machine:

| Runner                  | Time   | Speedup |
|-------------------------|--------|---------|
| lg VM (bytecode)        | 1.111s | 1.0x    |
| Native binary (lowered) | 0.067s | 16.6x   |

Result correct in both (5702887); binary 13.3 MB with `-ldflags '-s -w'`. The lowered Go is high quality: `fib` came out as `func Fib(ec *vm.ExecContext, arg0 int) (vm.Value, error)` — argument **typed as `int`** with direct native self-recursion, which is where the 16x comes from. Variadic `-main` lowered as `func Main(ec *vm.ExecContext, args ...vm.Value)`.

Pipeline proven:

```
program.lg
  -> scripts/lg-compile out/ .../lowered src/main/core.lg   (LG_SOURCE_PATHS=src:ys/lg:$LET_GO_SRC/pkg/rt/gogen)
  -> out/main/core/core.go   (package core, typed native funcs)
  -> lg -c program.lgb driver.lg   (driver.lg = (require 'main.core))
  -> main.go + go.mod (replace github.com/nooga/let-go => checkout)
  -> go mod tidy && go build -> ./fib
```

Ingy owns the gloat-side wiring (`-ELG`/`-tLG` engine, `convert-file-LG-bin`, go.mod/build orchestration). This issue tracks the **let-go-owned** gaps the spike exposed — all in lgb / AOT-IR / interop. They're independent fixes sharing one root (the native-entry boundary), so they're filed together.

## Root cause: the two-path reachability split

A lowered fn has **two independent reachability paths**, gated by two predicates:

- **Dynamic** — `:override-eligible?` (`override-uniform-value?`, `pkg/rt/core/ir/lower_go.lg:2973`): only fns whose Go signature is *uniformly `vm.Value`* get a `__gogen_wrap` + `rt.RegisterGoOverrides` init() that clobbers the bytecode Var with a `NativeFn`. Reachable via var lookup / `ec.Invoke`.
[...]

PR #430, opened 2026-07-09
feat(gogen): math.NaN()/math.Inf() emission + go-ast read/COW import bindings

nnunley:
## Stacked on #387

**This is stacked on #387** (`feat/ir-lowering-strategy`). Until #387 merges, the commit list/diff here includes #387's commits — the substantive change in *this* PR is the single top commit `feat(gogen): emit math.NaN()/math.Inf()…`. Review/merge after #387.

## What

`gogen/float-lit` rendered NaN/±Inf as the bare tokens `strconv.FormatFloat` produces (`NaN`, `+Inf`, `-Inf`) — none of which are valid Go, so any lowered fn that inlined a non-finite float constant would emit uncompilable output. This PR:

1. **Emit `math.NaN()` / `math.Inf(1)` / `math.Inf(-1)`** for non-finite floats (finite floats byte-identical).
2. **go-ast read/COW bindings** (the reusable primitives, usable from lg):
   - `(gogen/references-pkg? node pkg)` — read: does the subtree reference `pkg.X`?
   - `(gogen/file-import-paths file)` — read: the file's import paths
   - `(gogen/add-file-import file path)` — **copy-on-write**: a new File with the import added; input never mutated; idempotent
   - `(gogen/ensure-gen-imports file)` — wires in the imports gogen itself introduced (currently just `math`)
3. **`ir.lower-go/file`** runs `ensure-gen-imports` over each assembled File. It calls the **native** gogen binding rather than taking a namespace `require` on the gogen macro layer, so the AOT-lowered lowering path stays self-contained (see #425).

## Why not fold the import scan into `cFile`

An ad-hoc `ast.Inspect` buried in the file constructor is invisible and hard to test. Proper read + COW go-ast bindings make the manipulation explicit, reusable, and unit-testable from lg.

## Tests

- `pkg/rt/gogen_test.go`: NaN/Inf emission, `references-pkg?`, `add-file-import` COW + idempotence + render.
- `test/gogen_math_import_test.lg`: end-to-end through the lg bindings (build File with a NaN literal → `ensure-gen-imports` → renders `import "math"` + `math.NaN()`; COW leaves the original unchanged).
- `make check-generated` green (lowered tree compiles + dispatches natively).

### DIFF
--- a/pkg/rt/core/ir/lower_go.lg
+++ b/pkg/rt/core/ir/lower_go.lg
@@ -3641,7 +3641,10 @@
         ;; carriers) with a nil :decl — skip those when collecting decls.
         decls   (filterv some? (mapv (fn [result] (:decl result)) results))
         import-nodes (mapv import-spec-node imports)]
-    (gogen/file pkg import-nodes decls)))
+    ;; gogen may emit math.NaN()/math.Inf() for non-finite float literals; those
+    ;; need the `math` import, which the caller can't know about. Let gogen wire
+    ;; it in (COW) from the assembled File.
+    (gogen/ensure-gen-imports (gogen/file pkg import-nodes decls))))
 
 (defn lower [f mode]
   (if (nil? f)
--- a/pkg/rt/generated.sums
+++ b/pkg/rt/generated.sums
@@ -2,4 +2,4 @@
 # Content digest of all .lg + lgbgen sources that feed the .lgb
 # bundle and the lowered Go tree. The genmanifest staleness test
 # fails if this no longer matches the sources on disk.
-7ce290e1dc2e4951b53603fee0782121c1ad77710a1f7f3253ec74a31eb1b75e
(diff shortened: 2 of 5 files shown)


nnunley:
#387 is now merged, so this is no longer stacked — rebased directly onto current main. It's a clean, self-contained change: `gogen/float-lit` emits `math.NaN()`/`math.Inf(±1)` for non-finite floats, plus the go-ast read/COW import bindings (`references-pkg?`, `file-import-paths`, `add-file-import`, `ensure-gen-imports`) that wire the `math` import in. `make check-generated` + gogen Go/`.lg` tests green.

nooga (approved):
Clean fix — confirmed the emitted `math.NaN()`/`math.Inf()` output actually compiles, and the COW import bindings don't alias the input (verified via the new Go + `.lg` tests, all passing). One non-blocking nit: `cAddFileImport`'s dedup/insert logic only looks at the first `import` GenDecl, fine today since `gogen/file` only ever emits one, but worth a comment for future-proofing. LGTM.

PR #432, opened 2026-07-10
feat(ir): reach typed natives from runtime values

ingydotnet:
Two gaps kept lowered typed fns unreachable from values that only exist at runtime (CLI args, collection elements):

1. Guarded coercion at typed direct-call boundaries. A lowered caller passing an unproven vm.Value into a typed sibling now emits a comma-ok type assert with the static native call on match and the identical trampoline fallback on miss, instead of abandoning the static call (which landed on the bytecode callee, since typed fns are not override-registered). Proven args, e.g. int-typed self-recursion, keep the flat emission and pay no guard. Also dedupe the function-scoped temp decls in the prologue: the structured attempt and the goto rebuild both register temps for a call lowered in both passes, which emitted duplicate var decls.

2. scripts/lg-compile now passes multi-arity defns to the pipeline, which already splits them into per-arity <Name>_<arity> funcs with static cross-arity calls and a len(args)-switch override adapter. The per-arity flatten also warns when an arity falls back instead of dropping it silently.

Together these let a lowered entry chain like -main -> main_1 -> fib(n int) run natively when fed CLI input: fib(34) via that chain drops from 1.18s (all bytecode) to 0.08s.

### DIFF
--- a/pkg/ir/lisp_crossns_directcall_test.go
+++ b/pkg/ir/lisp_crossns_directcall_test.go
@@ -61,24 +61,31 @@ func TestCrossNsLoweredDirectCallEmit(t *testing.T) {
 // leaked, unused import makes the emitted file fail to compile. Regression for
 // recording the import in lookup-direct-callable (before coercion) rather than
 // after emission succeeds.
+//
+// An UNPROVEN vm.Value arg into a scalar param no longer falls back: it takes
+// the guarded comma-ok path, which emits the static call and rightly records
+// the import (see TestCrossNsDirectCallGuardedRecordsImport). What still bails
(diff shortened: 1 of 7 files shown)


ingydotnet:
```
$ cat fib.ys
!yamlscript/v0

defn fib(n):
  if (< n 2):
    then: n
    else: (+ fib((- n 1)) fib((- n 2)))

defn main(n=34):
  println: fib(n)
$ gloat -rTElg fib.ys  # normal vm
5702887
> gloat run time: 1.059s
$ gloat -rTELG fib.ys  # with go lowering
5702887
> gloat run time: 0.068s
$ 
```

nnunley (approved):

PR #433, opened 2026-07-10
perf(ir): transient current-locals-flat + reduce-based every? (−270MB lowering)

nnunley:
Two hot-path allocation cuts in the IR lowering pipeline, verified against the full suite.

- **`current-locals-flat` (ir.build)** merged the whole locals-scope stack into a fresh persistent map per symbol resolution via `(reduce merge {} ...)` — ~950K calls / ~290MB in a full lowering pass. Now folds every scope into ONE transient map; nil-guard mirrors merge's skip of empty scopes; left→right fold preserves the inner-shadows-outer contract.
- **`every?` (clojure.core)** routed through `reduce` (zero-alloc ArrayVector fast path) instead of the recursive seq/first/next walk.

Verified: `make check-generated` (bundle + lowered tree in lockstep), full `go test ./...`, e2e `TestGogenAOTDiff`, gogen-parity. cloneAndSet 3924→3654 MB.


nooga (approved):
Logic checks out — transient in `current-locals-flat` is properly scoped (fresh per call, never escapes before `persistent!`), and fold order still gives inner-shadows-outer correctly given `:locals` is append-ordered. `every?`'s `reduce`+`reduced` rewrite is behaviorally identical, verified against the native `reduce` short-circuit handling. Only nit: consider adding a small `bench_test.go` case to pin the claimed allocation win as a regression guard, since none of the changed files are tests.

PR #434, opened 2026-07-10
perf(ir): transducer-fusion foundation — data-driven pass list + remove arity-1 + transduce≡lazy proof

nnunley:
Foundation for a (deferred) IR loop-fusion pass. Behavior-neutral + correctness; the fusion pass itself is not in this PR.

- **Data-driven `optimize-fn-tail`**: the hardcoded pass sequence becomes a declarative `[pass, validate-label]` list run by `run-pass-list`. Byte-identical generated artifacts (`make check-generated` green) — this is the seam a future fusion-placement harness plugs into.
- **`remove` arity-1 transducer fix**: `remove` had only a 2-arity form, so `(remove pred)` errored as an xform. Added the arity-1 mirroring `filter` with a negated predicate.
- **Equivalence + alloc tests**: `test/transduce-equiv.lg` (36-case `transduce ≡ lazy-seq` corpus over map/filter/remove/keep + comp chains) and `TestTransduceNoIntermediateSeqAlloc` (fused `(into [] (map f) v)` allocates < naive `(vec (map f v))`).

Gates green: full suite, gogen-parity, ir-stress (baseline synced to the true 2393/20 — verified coverage-neutral, the prior committed baseline was stale).


mparrett:
Nice one. No major findings. 

Validated the PR @ 196c20a with:

- `go test ./...`
- `make check-generated`
- The new allocation test repeated 10 times (stable: 178 vs. 193 allocations/run)

One minor test-quality note: [[the allocation test](https://github.com/nooga/let-go/blob/196c20a841bf7b7918f0990a567e746bce12734e/pkg/rt/corefns/transduce_alloc_test.go#L58)]

https://github.com/nooga/let-go/blob/196c20a841bf7b7918f0990a567e746bce12734e/pkg/rt/corefns/transduce_alloc_test.go#L58 compiles the source during every measured iteration, so it demonstrates an end-to-end allocation difference but does not isolate runtime/transducer allocations. A future benchmark could compile once and measure execution only.

mparrett (commented):
Looks good, just needs merge conflict resolution.

mparrett (approved):

PR #435, opened 2026-07-10
feat(lgbgen): self-hosting lowering — Go overrides + check-selfhost target

nnunley:
Self-hosting the IR lowering. Applies the native (gogen_ir) Go overrides inside `lgbgen` so the compiler lowers its own passes natively, and adds `make check-selfhost` — a 3b native re-lower + step-4 fixpoint check that verifies the native compiler reproduces its own committed lowered output byte-identically.

Consolidates the former #435 (lgbgen self-host) + #436 (check-selfhost target) into one commit, rebased onto current main. `make check-selfhost` passes (fixpoint byte-identical); check-generated, build, and tests green.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -156,6 +156,30 @@ perf-snapshot: lowered $(GO)
 lowered: $(GO)
 	@go run -tags bootstrap ./cmd/lgbgen --target=go >/dev/null
 
+# Self-hosting (3b + 4): re-lower the whole core with the NATIVE compiler and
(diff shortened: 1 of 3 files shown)


nooga (approved):
LGTM — clean, minimal wiring on top of already-tested override infra, CI green. Non-blocking suggestion: consider adding `check-selfhost` as a CI job (even non-blocking/scheduled) so native/bytecode fixpoint drift is caught automatically rather than relying on manual runs, and add a short docs mention of the new target.

PR #436, opened 2026-07-10
build(make): add check-selfhost — verify the native compiler reproduces its own lowered output

nnunley:
Adds a `make check-selfhost` target that verifies the self-hosting invariant: the compiler, built to run on its own natively-lowered passes, reproduces the committed lowered-Go tree byte-for-byte.

## What it does

1. Bootstrap the lowered-Go tree from source (`make lowered`, existing).
2. Build `lgbgen` with the `gogen_ir` tag so its lowering pipeline runs on **that tree's own native passes** (enabled by #435), then re-lower the whole core.
3. Byte-compare the re-lowering against the committed tree.

Byte-identical means the native compiler reproduces its own output — the self-hosting fixpoint holds. A difference means a pass changed and the tree is stale relative to source (run `make lowered` first). This makes the fixpoint a checkable invariant alongside `check-generated` and `check-lowered-fresh`.

## Why it matters

Running the compiler on its natively-lowered passes is measurably faster: the IR-compile benchmark runs ~16% faster under the native passes than under bytecode (18.5ms vs 22.0ms). This target proves that speedup is correctness-preserving, not just faster.

## Depends on

#435 — the fix that makes `lgbgen` apply its Go overrides so the passes actually run natively. Without it, step 2 runs interpreted and the check proves nothing.

## Test

```
$ make check-selfhost
OK: native self-hosting fixpoint holds (byte-identical).
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -156,6 +156,30 @@ perf-snapshot: lowered $(GO)
 lowered: $(GO)
 	@go run -tags bootstrap ./cmd/lgbgen --target=go >/dev/null
 
+# Self-hosting (3b + 4): re-lower the whole core with the NATIVE compiler and
+# verify the fixpoint. `lowered` (3a) bootstraps the tree from source with the
+# interpreted passes; here we build lgbgen -tags "bootstrap gogen_ir" so the
+# lowering pipeline runs on THAT tree's native passes, re-lower everything
+# (including the passes themselves), and byte-compare against the committed
+# tree. Byte-identical = the native compiler reproduces its own output, so the
+# passes are correct native code AND fast enough to self-host (bench-ratchet
+# IRCompile: native ~16% faster than bytecode). A diff means a pass changed and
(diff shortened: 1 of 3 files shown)


mparrett:
One potential blocker, please investigate.

- [[P1: The new check doesn’t prove native execution](https://github.com/nooga/let-go/blob/ba2af8a9cc215351cd84588a0004584cdf6ecf8d/Makefile#L173)](https://github.com/nooga/let-go/blob/ba2af8a9cc215351cd84588a0004584cdf6ecf8d/Makefile#L173). It only diffs the generated trees- passes if `gogen_ir` silently falls back to bytecode.

This is the same failure mode this PR says existed before `ApplyGoOverrides`. Suggestion: add an explicit assertion that native overrides were applied (for example, a gogen_ir-only applied-override counter / required namespace set) before accepting the byte-identical result.

Issue #411, opened 2026-07-07
Annotation-driven primitive registrar + Fn-adapter (finish namespace cleanup; improve lginterop model)


nnunley:
## Context

The namespace-purity work (#408 refer-shadow fix, #410 purify `clojure.core`)
moved the tractable lg-isms out of `clojure.core` into `let-go.core` /
`let-go.types` and made the surface assertable against `clojure ∩ bb`. The
**remaining** cleanup is uniformly blocked on the same missing infrastructure, so
this issue captures the design for it as a single follow-up.

### Remaining namespace cleanup (the "tail")

1. **Re-home the `clojure.string` primitives.** `clojure.string` is currently a
   *veneer* over Go primitives that live in `clojure.core`:
   `(defn upper-case [s] (core/upper-case (str s)))`, `(def trim core/trim)`,
   `index-of`/`last-index-of` wrap `core/…`, etc. To remove `trim`/`upper-case`/
   `split`/… from `clojure.core` the Go primitives must be registered into
   `clojure.string` (or `let-go.core` for the lg-renamed `str-replace`/`str-join`),
   and the veneer rewritten.
2. **IO isms → the `io` ns** (`pkg/rt/ions.go`, the `clojure.java.io` analog):
   `open`/`close!`/`write!`. Load-bearing — core.lg's `with-open` macro uses
   `close!`, and corpora (xsofy `debug.lg`) use `open` unqualified.
3. **Privatize bucket ③** — ~82 compiler/interop internals currently public in
   `clojure.core` (`-dt-*`, `make-*`, `destructure*`, `*emit*`, host-class
   registrations, …).

### Why it's blocked

Each primitive's **namespace, lg name, and visibility** are decided imperatively
at hundreds of `ns.Def("name", fn)` sites in `pkg/rt/lang.go`, where `fn` is an
anonymous `vm.NativeFnType.Wrap(func(vs []vm.Value) (vm.Value, error){…})` closure
that hand-asserts arg types and threads `ec` itself. There is **no** export/name/
privacy control other than the Def call site (the `;; lgbgen:skip`/`:skip-go`
markers only gate bundle/lowering; `lginterop` routes whole *external* Go packages
to a deps.edn alias). So re-homing/privatizing is manual and error-prone.

### 1. Annotation-driven primitive registrar (`//lg:` directives)

Named primitive functions declare their export via Go comment directives; a
generator (sibling to `cmd/lginterop`, for *internal* primitives) emits the
registration.

```go
[...]

PR #438, opened 2026-07-11
feat(native-bridge): //lg native-primitive generation — drain hot trampolines (IRCompile −13.6% allocs)

nnunley:
Generates native Go primitives via `//lg` annotations and converts hot compiler-path calls (`nth`/`deref`/`get`/`str`/`conj`, then `reduce`) to direct native calls, removing ~852 trampolines — IRCompile −13.6% allocs/op, −10.4% B/op (gogen_ir variant, benchstat n=10, p=0.000; wall time −3% point estimate, not significant). See the benchmark comment below for the full A/B vs the branch base.

**Base:** the branch parent is #410's commit, which has since merged into `main` — the diff now shows just the native-bridge delta.

**Squashed** from 6 commits to one, rebased onto current `main` (via #410). During the rebase, main's guard-based typed-call emission (`coerce-arg-to-type`) met native-bridge's typed-native constant args, which generated an illegal `vm.Int(0).(vm.Int)` (type-asserting a concrete constant) in 47 lowered functions. Fixed by widening a `:const` operand's boxed form to `vm.Value(...)` before the comma-ok assert, so it matches (native fast path) or falls to the trampoline (genuine mismatch); non-const guard sites are byte-identical.

**Gates:** `check-generated` OK (lowered tree compiles + dispatches natively), `gogen-parity` PARITY on both backends, `ir-stress` 2371/2391 (failures 20 ≤ 20), full `go test ./...` incl. `TestGogenAOTDiff` and e2e — all green.

---
Implements the annotation-driven native-primitive registrar infrastructure tracked in **#411**; addresses the silent AOT perf-cliff described in **#425**.


mparrett:
Luna Review:

- The primitive generator is nondeterministic. [`[emitFile](https://github.com/nooga/let-go/blob/b6f9eef2c3934a6c3e13ae3bdf209d5556203f31/cmd/lginterop/prims_emit.go#L93-L100)`](https://github.com/nooga/let-go/blob/b6f9eef2c3934a6c3e13ae3bdf209d5556203f31/cmd/lginterop/prims_emit.go#L93-L100) ranges over `moduleMap`, a Go map, when emitting registrations. With identical inputs, I observed two different outputs across 20 runs—17 in one ordering and 3 in another. This can intermittently dirty generated files or fail freshness checks. Sort the module keys and add a repeat-generation determinism test.

- The generation command passes `-out`, while primitive mode defines `-primitives-out`: (generator flags, invocation)

https://github.com/nooga/let-go/blob/b6f9eef2c3934a6c3e13ae3bdf209d5556203f31/cmd/lginterop/main.go#L41-L48

https://github.com/nooga/let-go/blob/b6f9eef2c3934a6c3e13ae3bdf209d5556203f31/scripts/generate.lg#L145-L148

It works only because the primitive output default happens to be correct; the explicit argument is ignored by primitive mode.

Validation:

- Targeted Go packages all passed: `cmd/lginterop`, `pkg/vm`, `pkg/rt`, `pkg/ir`, and `pkg/compiler`.
- The broader suite was green
- The generated native bindings survived core initialization in a direct probe.
- I found no obvious semantic regression in the converted primitives.

**Recommendation:**

Re-assess after branch is rebased and conflict-free. Verify the generator output is deterministically sorted with regression coverage, the flag mismatch is corrected, and CI, including `go test ./...`, generation freshness/parity, and preferably the performance benchmark passes.

nnunley:
Thanks @mparrett — both addressed, and the branch is now rebased conflict-free onto current main.

**1. Nondeterministic generator.** Fixed at the source: `RegisterGeneratedPrimitives()` now iterates the `(Ns|GoPkg)` module keys in sorted order (`prims_emit.go`) rather than ranging the raw map. gofmt (already applied in `generatePrimitives`) canonicalizes the import block, but it does not reorder function-body statements, so the registration order needed the explicit sort. Added `TestEmitFileDeterministic`, which asserts the gofmt'd output is byte-identical across 30 regenerations over a multi-package/multi-namespace spec set. Two back-to-back `-primitives` runs are now byte-stable, and `gofmt -l` on the checked-in file is clean.

**2. Flag mismatch.** Corrected `scripts/generate.lg` to pass `-primitives-out` (was `-out`, which primitive mode ignores — it only worked because the default matched).

**Gates:** `check-generated` OK (lowered tree compiles + dispatches natively), `gogen-parity` PARITY on both backends, `ir-stress` 20/20, full `go test ./...` incl. `TestGogenAOTDiff` and e2e — all green.

One heads-up on the diff: this branch is stacked on #410 (clojure.core purification); GitHub won't let me set a cross-fork base to the #410 branch, so until #410 lands the diff shows both. The rebase also surfaced an interaction between #410's guard-based typed-call emission and this branch's typed-native constant args (illegal `vm.Int(0).(vm.Int)`); fixed by widening a `:const` operand to `vm.Value(...)` before the comma-ok assert — details in the PR description.

mparrett:
@nnunley Thanks- the two review findings look correctly addressed and you might be rebasing now. Lint?

nooga (changes requested):
Solid mechanism, but please substantiate the -24% IRCompile claim with an actual benchmark, add parity tests for deref/get/conj/reduce/str (plus error-path cases) alongside the existing nth/subs/name coverage, and fix the genmanifest staleness tracking to include pkg/rt/native_prims.go so drift in the new annotated source is actually caught.

nnunley:
Fixed a source-stdlib-bootstrap regression that CI caught (the `Test with source stdlib bootstrap` job: `TestLGBParity/ns-require-string` → `Can't resolve s/join`).

**Root cause:** this branch added `LookupOrRegisterNS("string")` at rt-init (to ensure `clojure.string` exists before `RegisterGeneratedPrimitives` binds the native `upper-case` adapter into it). Under the precompiled bundle that's harmless — the ns is already compiled. But under source-stdlib bootstrap it forces a premature compile of `string.lg` during init, before the compiler is ready, leaving the `string` ns loaded-but-broken (`join` never interns) → `s/join` fails to resolve. That's why local `go test` (bundle) was green while the bootstrap CI lane failed.

**Fix:** dropped the premature `LookupOrRegisterNS("string")` — it's redundant, since the primitive binding creates the ns as needed — and switched the generated binding to the NoLoad variant (`LookupOrRegisterNSNoLoad`) so registration never triggers a source load at init either. `string.lg` now loads normally when user code requires it.

**Verified:** the full `-tags bootstrap` lane (bytecode/compiler/rt/vm/corefns/wasm) is green, `TestLGBParity` all-pass, native-bridge parity intact; plus `check-generated`, `gogen-parity` PARITY, and full `go test ./...` incl. e2e.

nooga (changes requested):
None of the three requested items are actually addressed in this push — it looks like it only carries forward fixes for a separate reviewer's earlier findings (generator determinism, CLI flag, bootstrap regression). Specifically: (1) still no benchmark substantiating -24% IRCompile; (2) native-bridge-parity.lg still only covers nth/subs/name, no deref/get/conj/reduce/str or error-path cases; (3) genmanifest now tracks `pkg/rt/builtins/` (an unrelated existing file) instead of `pkg/rt/native_prims.go`, so edits to the actual `//lg:native` annotations still won't trip staleness. Please address the original comment directly.

nnunley:
All three items addressed in 74f84b12, plus an honest correction on the headline number.

**1. Benchmark.** Ran `BenchmarkIRCompile` A/B: base = this branch's parent 3a3c030 (#410, now on main) vs head, `-count=10` total per side, Apple M3, `benchstat`:

`-tags gogen_ir` (native passes — where the drained trampolines live):

| metric | base | head | delta |
|---|---|---|---|
| sec/op | 18.22m ± 2% | 17.67m ± 5% | ~ (−3.0%, p=0.089, n.s.) |
| B/op | 6.535Mi ± 0% | 5.859Mi ± 0% | **−10.35%** (p=0.000) |
| allocs/op | 169.8k ± 0% | 146.7k ± 0% | **−13.59%** (p=0.000) |

Bytecode variant: flat across all three metrics, as expected (the direct calls only exist in the lowered tree).

The **−24% wall-time claim does not reproduce** post-rebase — it came from a pre-rebase measurement against a different base and I can't substantiate it, so I've corrected the PR title/description to the measured numbers: the durable, statistically solid win is **−13.6% allocs / −10.4% bytes per IR compile**; wall time is a ~3% point estimate that doesn't clear significance on this corpus.

Repro:
```
git worktree add /tmp/base 3a3c03039 && (cd /tmp/base && make generate)
(cd /tmp/base && go test -tags gogen_ir ./pkg/ir/ -run '^$' -bench '^BenchmarkIRCompile$' -count=10) > base.txt
go test -tags gogen_ir ./pkg/ir/ -run '^$' -bench '^BenchmarkIRCompile$' -count=10 > head.txt
benchstat base.txt head.txt
```

**2. Parity coverage.** `test/native-bridge-parity.lg` now covers all 14 annotated prims: `deref` (1-arity over atom/delay, 3-arity blocking form with timeout hit and delivered-value cases), `get` (2/3-arity over map/vector/set/nil), `conj` (vector/list/set/map/nil-coll, multi-arg), `reduce` (init/no-init, empty-coll `(f)` and init-passthrough-without-calling-f, `reduced` early-exit incl. over an infinite lazy source), `str` (0-arg, nil elision, variadic value rendering), plus error paths: `nth` OOB on vector and list, `subs` OOB both arities, `deref` on a non-IDeref, `name` on a number. Verified green under both `TestRunner` (bytecode) and `-tags gogen_ir` (native dispatch).

**3. genmanifest.** `pkg/rt/native_prims.go` is now a tracked single-file generator input (`sourceFiles` in `pkg/genmanifest/manifest.go`) — an `//lg:native` annotation edit there changes the committed digest and trips `TestGeneratedArtifactsAreFresh` / `make check-generated`. Covered by a new assertion in `TestSourceFilesCoversCoreAndLgbgen`. (`pkg/rt/builtins/` stays tracked as well — it's a genuine lowering input for direct-call targets; the miss was that it wasn't *the* annotated source.)

nnunley:
@mparrett Lint: ran `make lint` (golangci-lint v2.12.2, config in `.golangci.yml`) plus the pre-commit formatters (gofmt -s, goimports) on the rebased branch — no findings. The branch is now squashed to one commit and rebased onto current main (the typed-catch stack and #447 landed under it; bundle/sums conflicts resolved via `make generate`, `check-generated` in lockstep). Parity file green under both backends (`TestRunner` and `-tags gogen_ir`), genmanifest tests green, ir-stress ratchet green.

mparrett (approved):
I looked at be352a516 and -- earlier findings have been addressed, parity coverage in place, and benchstat run looks solid. Thanks!

mparrett (approved):

mparrett:
@nnunley Here's some context excluded from my previous review due to verbosity. Maybe it's helpful to consider while a final blessing from @nooga is pending:

- Benchmark: A/B `benchstat` supplied; run with `n=10`; the unsupported −24% wall-time claim was removed. The PR now claims the reproducible −13.6% allocations and −10.4% bytes improvements, while accurately describing the wall time result as statistically insignificant.
- Parity coverage: now covers `deref`, `get`, `conj`, `reduce`, and `str`, including blocking deref, empty/infinite reduction, early termination, and relevant error paths.
- Staleness tracking: `pkg/rt/native_prims.go` is explicitly tracked by genmanifest and asserted in its tests.
- Earlier findings remain fixed: deterministic generation, correct CLI flag, bootstrap-safe namespace registration, and lint cleanup.
- Branch status: mergeable, focused to the native-bridge delta, and all active CI checks pass.

Independently ran the generator/genmanifest tests and the native-bridge parity test successfully. GitHub’s `generated-artifacts`, `gogen-diff`, build, lint, bootstrap, CodeQL, and differential checks are green. The only administrative wrinkle is that GitHub still shows `CHANGES_REQUESTED`.

mparrett:
Took another look and tried harder to poke holes. Confirmed, the original checklist is mechanically addressed.

Deeper probing found two concrete bugs and one test blind spot that could matter from a compat/maintenance standpoint:

### Findings (P1)

**1. `with-redefs` semantics break across lowered function boundaries.**

I reproduced this program:

```clojure
(let [calls (atom 0)
      original reduce]
  (with-redefs [reduce
                (fn [& args]
                  (swap! calls inc)
                  (apply original args))]
    (group-by identity [:a :b :a])
    @calls))
```

Results:

- bytecode backend: `1`
- `gogen_ir`: `0`

Generated `GroupBy` calls `rt.Reduce3`, `rt.Get3`, and `rt.Conj` directly, bypassing their Vars. The existing safeguard only disables direct calls inside a function that itself contains `with-redefs`/`alter-var-root`; it cannot protect a separately lowered helper invoked within that dynamic extent.

That contradicts the promise in [`[lower_go.lg](https://github.com/nooga/let-go/blob/8fa59c0665cfcb16e0dd13b4411f0ae41baf978a/pkg/rt/core/ir/lower_go.lg#L1810-L1817)`](https://github.com/nooga/let-go/blob/8fa59c0665cfcb16e0dd13b4411f0ae41baf978a/pkg/rt/core/ir/lower_go.lg#L1810-L1817) that these overrides remain observable. Because the new primitives produce over a thousand direct call sites, this is materially broader than a corner case.

**Acceptable resolutions:**

- a cheap root-identity/version guard that takes the direct path only while the Var still points at the registered primitive, falling back otherwise; or
- an explicit project decision that `gogen_ir` has Clojure-style “direct linking” semantics where redefinitions are unsupported, with tests and documentation changed accordingly.

Given the existing comments and tests, the first option appears consistent with intended behavior.

**2. Incorrect behavior in the new import deduper**

At [`[lower_go.lg:3660](https://github.com/nooga/let-go/blob/8fa59c0665cfcb16e0dd13b4411f0ae41baf978a/pkg/rt/core/ir/lower_go.lg#L3639-L3662)`](https://github.com/nooga/let-go/blob/8fa59c0665cfcb16e0dd13b4411f0ae41baf978a/pkg/rt/core/ir/lower_go.lg#L3639-L3662), replacing an earlier unaliased import uses `(pop out)`. That removes whichever import happens to be last, not necessarily the duplicate.

I reproduced:

```clojure
[{:path "example/a"}
 {:path "example/b"}
 {:path "example/a" :alias "a"}]
```

The result drops `example/b` and retains both `example/a` entries. Current generation happens not to hit this ordering, but future native modules can yield missing and duplicate imports. This should be fixed with indexed replacement or a two-pass path selection and covered by a focused regression test.

**3. `clojure.string/upper-case` is not actually left native.**

After:

```clojure
(require '[clojure.string])
(type clojure.string/upper-case)
```

both backends report `let-go.lang.Fn`, whereas `str` reports `let-go.lang.NativeFn`.

`RegisterGeneratedPrimitives` runs before the string namespace source is loaded; the later [`[defn upper-case](https://github.com/nooga/let-go/blob/8fa59c0665cfcb16e0dd13b4411f0ae41baf978a/pkg/rt/core/string.lg#L19-L25)`](https://github.com/nooga/let-go/blob/8fa59c0665cfcb16e0dd13b4411f0ae41baf978a/pkg/rt/core/string.lg#L19-L25) overwrites the generated native adapter. The “native bridge parity” test verifies output but never verifies that this native bridge is active.

Registration should survive namespace loading, and a test should assert the resulting Var root is native after `require`.

### Performance and size

Less critical: the revised performance story is credible but should be refreshed.

- My current-head run reproduces approximately `146.7k allocs/op` and `5.86 MiB/op`, matching the author’s reported head.
- The unsupported wall-time claim has correctly been removed.
- However, the posted A/B used `3a3c030`; the live parent is now `49ff263`, ten commits later, including compiler, VM, core, and `ir/build` changes. A final benchmark against the actual merge base would remove that ambiguity.
- The new `reduce` implementation dropped dedicated ArrayVector/Range loops, but my focused benchmarks found no meaningful regression—the chunked path and Go escape analysis mostly recover it.
- Untagged binary growth is modest:
  - normal: `+53,424 bytes` (`+0.286%`)
  - stripped: `+33,408 bytes` (`+0.255%`)

From a compactness standpoint, that looks acceptable for the measured allocation reduction.

### Additional hardening

Not necessarily blocking, but worth follow-up before expanding annotation use:

- The generator scans every `.go` file, including `_test.go` and files excluded by build tags. A platform-specific annotation could generate unconditional references and break other targets.
- Go parse failures are silently treated as “no primitives,” potentially replacing valid generated output with an empty registrar.
- The parity test still lacks representative `conj` and `reduce` error cases and does not exercise transitive direct-call/redefinition behavior.

**Summary:** Would like to see changes for the redefinition bug, import deduper, and `upper-case` registration lifecycle. Consider moving forward once those have regression tests, and (preferably) the benchmark is rerun against `49ff263`. The underlying mechanism and size/performance tradeoff otherwise look reasonable.

nooga (dismissed):

nooga (commented):
All three items from the two prior CHANGES_REQUESTED reviews look genuinely addressed now (verified independently, not just taking the PR comments at face value):
- Benchmark: real benchstat A/B posted, and the headline number was honestly corrected from an unreproducible -24% wall-time claim to the statistically solid -13.6% allocs/-10.4% bytes (wall time ~-3%, n.s.). Good.
- Parity coverage: test/native-bridge-parity.lg now covers deref/get/conj/reduce/str plus error paths (nth/subs OOB, deref-on-non-IDeref, name-on-number), in addition to the pre-existing nth/subs/name coverage.
- genmanifest: pkg/genmanifest/manifest.go now has a `sourceFiles` entry for pkg/rt/native_prims.go (the actual `//lg:native` annotation source) with a dedicated assertion in TestSourceFilesCoversCoreAndLgbgen, so annotation drift will now trip staleness.

Two remaining issues found while reviewing the fix itself:

1. **Latent bug in `distinct-imports`** (pkg/rt/core/ir/lower_go.lg, the `file` fn's import-dedup rewrite). When a later duplicate of an import path carries an alias and the prior occurrence didn't, the code does `(conj (pop out) entry)` to replace it in place — but `pop` removes the *last* element of `out`, not necessarily the prior same-path entry. If another path's import is interleaved between the two same-path entries, `pop` silently drops that unrelated import instead. Repro (isolated the function and ran it standalone):
```clojure
(distinct-imports [{:path "rt"} {:path "other"} {:path "rt" :alias "rt"}])
;=> [{:path rt} {:path rt, :alias rt}]   ; "other" is gone, and "rt" is duplicated
```
This doesn't currently manifest in the generated pkg/rt output because the native-module import collector (pipeline.lg, `native-import-results`) is always appended as a single batch at the very end of the results list, after any per-function aliased `rt` import — so the aliased entry always wins the "seen" slot before the unaliased duplicate is ever considered. But that's an invariant of caller structure, not something `distinct-imports` itself enforces, so it's one refactor away from silently corrupting a generated file's import block. Suggest tracking replacement by index (e.g. `seen` mapping path -> index into `out`) rather than relying on `pop`.

2. **Test-file convention deviation.** test/native-bridge-parity.lg has no `(ns ...)` form and uses bare `(assert ...)` throughout instead of the `deftest`/`testing`/`is` pattern CLAUDE.md documents and all 152 other `.lg` test files use. It does get exercised (TestRunner's `runFile` compiles it and fails the subtest if an assert throws), but it loses per-assertion failure names, doesn't participate in the `test` namespace's isolation/reset machinery, and stands out from every other test file in the suite. Worth converting to the standard shape for consistency, ideally before merge since it's new code in this PR, not a pre-existing pattern being extended.

Neither of these blocks the core mechanism, which is solid and well-covered by the parity suite plus the full CI matrix (build/lint/gogen-diff/generated-artifacts/CodeQL all green). Also: the squashed commit's body text still opens with "IRCompile −24%)" even though the PR title/description were corrected to the honest number — worth a quick edit so `git log` doesn't repeat the retracted claim.

nnunley:
All three P1s addressed, rebased onto current main (`536b212`), squashed head `f9fed21`.

**1. with-redefs across lowered boundaries** — took the root-identity-guard route. Every registered native-primitive var's canonical root is tracked in the vm (`Var.GuardRoot`; re-defs carry the guard across the Var replacement, `alter-var-root`/`SetRoot` re-evaluate it against the canonical value by pointer identity). Lowered native-module call sites are emitted as `if rt.NativePrimsIntact() { direct } else { trampoline }` — one atomic counter load on the fast path; any active override steers all native sites onto var-dispatch until the root is restored. The posted repro now yields `calls: 1` on both backends with `group-by` still the lowered NativeFn, and the `redefinition-guard` deftest in `test/native-bridge-parity.lg` covers the flip → observe → restore cycle plus steady-state intactness (`rt/native-prims-intact?`). Lowered-sibling (non-native-module) direct calls keep their existing semantics — that seam predates this PR.

**2. import deduper** — replaced `(pop out)` with indexed replacement (`seen` maps path → index into `out`). `TestLowerGoFileDedupesInterleavedAliasedImports` reproduces the exact failure first (interleaved `example.com/b` dropped, duplicate imports) and is green after.

**3. upper-case registration lifecycle** — generated adapter bindings are recorded at init and reapplied right after a namespace's on-demand load completes, so the bootstrap `(defn upper-case ...)` no longer wins. `(type clojure.string/upper-case)` is `NativeFn` after require on both backends (asserted in the parity test); user `with-redefs` still overrides and restores correctly (it lands after the load).

**Benchmark vs the actual merge base** (`536b212`, includes the typed-catch stack), gogen_ir, benchstat n=10, guard included:
allocs/op **−13.27%** (p=0.000), B/op **−10.24%** (p=0.000), sec/op **−7.39%** (p=0.001) — wall time is now statistically significant.

**Hardening follow-ups from the review also landed:** the lginterop scan skips `_test.go`/`zz_`/build-constrained files (deterministically — any constraint skips, so committed output can't depend on the scanning host), and a Go parse failure is a hard error instead of "no primitives"; both have unit tests. The parity file is converted to the standard `ns` + `deftest`/`testing`/`is` shape with the missing `conj`/`reduce` error paths. The squashed commit message no longer mentions the retracted −24%.

**Gates:** `make check-generated` OK, `gogen-parity` PARITY (jank + lower-go suites), `ir-stress` improved to 2378/2397 (19 ≤ 20 baseline), full `go test ./...` + gogen_ir + lint green.

**One pre-existing finding, out of scope here:** `TestUnrollBytecodePipelineAll` under `-tags gogen_ir` fails on **clean main** with the identical numbers (ops loop=212 unroll=113, 1.88x vs the 2x bar) once the lowered tree is generated and wired — it only surfaces because the wire test file exists post-`make generate`, so CI never runs it. Same result on this branch; worth its own issue.

nnunley:
Cleanup round pushed (`5cc3a18`). Post-review polish plus one real find:

- **Eager-hybrid load path now also reapplies generated adapters**: `loadPrecompiledBundle`'s eager chunk pass runs bootstrap defs outside the on-demand loader, so it now calls `rt.ReapplyGeneratedPrimitives` too — the same gap I'd flagged for BootCore on #499, present in-tree.
- Covering that path let the jank suite's `string_test/upper_case` see the native adapter — which exposed that `UpperCase(s string)` was stricter than JVM Clojure (`(upper-case :a)` must coerce via toString to `":A"`). The primitive now takes `vm.Value` and coerces via the `str` rendering, nil still throws; suite green on both backends.
- Minor: deduped `toKebabCase` against the existing `kebabCase`, removed a dead conditional and a hand-rolled `strings.Contains`, and `hasBuildConstraint` scans lines without allocating the whole file.

Gates rerun: check-generated OK, lint clean, full default suite + gogen_ir `./test` (incl. the 232-file compat suite) green. The one gogen_ir `pkg/ir` failure is the pre-existing main issue addressed in #505.

nnunley:
One more fix from probing the cleanup round's var-lifecycle question: the reapply path bound adapters via `Namespace.Def`, which creates a fresh Var — splitting var identity for code compiled during the ns load. Observable: `(with-redefs [str/upper-case f] (str/capitalize "ab"))` didn't see the override (capitalize was stranded on the pre-reapply var object), where JVM Clojure does. The primitives lifecycle now mutates the EXISTING var's root in place (`SetRoot`, matching the language-level def path, which interns via `LookupOrAdd` and already has stable identity), falling back to `Def` only for a first bind. Regression assertion added to the parity test (redef observed through `capitalize`, restore verified). Gates green on both backends; head `f6141b4`.

mparrett:
Filed #506 to track the `loadPrecompiledBundle`/`BootCore` core-load unify from #499. Your coordination note there — calling `reapplyGeneratedPrimitives` after the eager chunk loop so a BootCore'd binary doesn't keep the bootstrap closure and trip the root-deviation counter — is captured as the single-point item on #506, so whichever of these lands second has one place to wire it.

PR #440, opened 2026-07-11
feat(ir): loop-fusion (deforestation) — pass, enabled, comp-chains + into

nnunley:
Consolidates the full IR loop-fusion (deforestation) effort into one PR against current `main`. Supersedes and folds in #441.

## What it does
Adds an IR-level deforestation pass (`ir.passes.fusion`) that rewrites producer→consumer seq chains into transducer form, eliminating the intermediate `LazySeq` churn. Because it runs in the shared optimizer, it benefits **both** backends (bytecode + native `gogen_ir`).

- **The pass** (`pkg/rt/core/ir/passes/fusion.lg`): matches `(reduce g init (map|filter|remove|keep f coll))` and rewrites to `(transduce (map f) g init coll)`.
- **Enabled by default** — measured ~20% fewer allocations on ClojureTestSuite (aot_native 14.16M → 11.33M; bytecode 8.53M → 6.79M) and −14..22% on IRCompile.
- **v2 — comp-chains + `into`:**
  - Nested producers fuse into a single `comp` transducer, innermost-first: `(reduce g init (filter p (map f coll)))` → `(transduce (comp (map f) (filter p)) g init coll)`.
  - `into` as a fusion consumer: `(into to (map f coll))` → `(into to (map f) coll)`.
  - **Maximal-suffix partial fusion:** if an inner stage escapes or has an impure callback, the qualifying outer stages still fuse while the failing producer stays a lazy leaf.

## Safety
Each fused stage is independently gated: stable producer var (not `with-redefs`'d), single-use ⇒ non-escape, same-block, and callback purity (required for chunk/short-circuit correctness). Length-1 chains emit a bare xform (no `comp`), keeping single-stage output byte-identical.

## Tests & gates
- 17 IR-shape unit tests (`pkg/ir/lisp_fusion_test.go`) + an 84-assertion `transduce`≡lazy-seq property test (`test/transduce-equiv.lg`), including a non-commuting `(filter even? (map inc coll))` case that guards against `comp`-order inversion.
- `scripts/gogen-parity.sh` PARITY across both backends; `make ir-stress-gate` (2378/2398, failures 20 ≤ baseline 20); `make check-generated`; full `go test ./...` incl. e2e — all green.

---
**Part of #258** (Epic: Lowering & cross-target codegen) — IR-level loop-fusion optimization on the lowering pipeline (both backends).


nooga (changes requested):
Solid design and unusually rigorous equivalence/order-inversion testing (comp-order test, partial-suffix, control on/off test are all good signs). One correctness-relevant gap before merge: please add a test that runs the *fused* path (via `optimize-fn`) against an infinite/lazy source with a `reduced`-based early exit, proving it terminates and matches the unfused result — none of the current equivalence tests exercise short-circuiting through an actual fused chain, which is the highest-risk untested combination for a deforestation pass. Also fix the stale "fusion OFF (default)" comment in `test/fusion-semantics.lg` now that `*enable-fusion*` defaults true.

nooga (changes requested):
Neither requested change from the prior review appears in this commit — `test/fusion-semantics.lg` still has the stale "fusion OFF (default)" comment, and no test (in transduce-equiv.lg, fusion-semantics.lg, or lisp_fusion_test.go) exercises `reduced`/early-exit through a fused chain over an infinite or lazy source. This looks like a rebase, not a fix-up. Please add the termination/early-exit test and correct the comment before merge.

nooga (approved):
Verified both previously-requested changes are now in place on HEAD (9d1ef6c4):

- `TestFusionEarlyExitReduced` (pkg/ir/lisp_fusion_test.go) plus the three new cases in test/fusion-semantics.lg exercise `reduced`/early-exit through an actual fused `(transduce xform (completing g) init coll)` chain over `(range)` (infinite) and lazy `filter`/`remove` sources — this was the highest-risk untested combination flagged in review and is now covered and passing.
- The stale "fusion OFF (default)" comment in test/fusion-semantics.lg is fixed to reflect that `*enable-fusion*` defaults true.

Ran the suite locally at this commit in a clean worktree: `go build ./...` clean, all 17 fusion Go tests pass, `test/fusion-semantics.lg` (8/8) and `test/transduce-equiv.lg` (84/84, including the non-commuting `filter∘map` order-inversion guard and multi-stage `comp` chains) pass, full `go test ./...` green. CI (`gh pr checks`) is all green — build, lint, gogen-diff, gold-differential, generated-artifacts, CodeQL; the bench/comment jobs are skipped as expected for PR runs.

Test coverage follows the repo's convention: IR/compiler-pass logic gets Go unit tests (pkg/ir/lisp_fusion_test.go), transducer-equivalence semantics get .lg integration tests. The safety gating (stable producer var, single-use/non-escape, same-block, callback purity) and the maximal-suffix partial-fusion fallback are sound and specifically tested (TestFusionPartialSuffix, TestFusionGateBlocksImpureChainStage). No further blocking issues found.

One non-blocking nit for a follow-up: `docs/perf/ir-stress-baseline.edn`'s note about the deferred placement sweep (spec Section A.1) is fine to leave as future work, but it'd be worth a tracking issue reference so it isn't lost.

PR #441, opened 2026-07-11
feat(ir): enable loop-fusion by default (~20% fewer allocations)

nnunley:
Turns the loop-fusion pass ON by default (`*enable-fusion*` true) at its current placement (in `default-pass-list` right after `cse`).

> **Stacked on #440** (the fusion pass itself). Until #440 merges, its commits appear in this diff; the delta here is the one-line default flip + regenerated artifacts.

## Measured win (full corpus, 2026-07-11)
Allocation counts (deterministic, unlike wall-clock):

| Benchmark | allocs/op OFF | ON | Δ |
|---|---:|---:|---:|
| `ClojureTestSuite` [aot_native] | 14,155,276 | 11,334,065 | **−20%** |
| `ClojureTestSuite` [bytecode] | 8,534,727 | 6,788,794 | −20% |
| `ClojureTestSuite` [ir_bytecode] | 14,187,067 | 11,338,846 | −20% |
| `IRCompile` [gogen_ir] | 203,268 | 174,292 | −14% |
| `IRCompile` [bytecode] | 130,608 | 101,328 | −22% |

The intermediate `LazySeq`s between `map`/`filter`/`remove`/`keep` and their `reduce`/`into` consumers are gone; the corpus has many such chains, so the win is larger than the narrow v1 pattern suggested.

## Correctness (corpus-wide, fusion active on both backends)
- `gogen-parity.sh`: **PARITY** — bytecode ≡ gogen_ir; jank 5657/5657 assertions pass.
- `go test ./...`: green.
- `ir-stress-gate`: no new lowering failures (`:failed` unchanged at 20).
- `make check-generated`: bundle + lowered tree in sync.

## Placement note
The exhaustive permutation-harness sweep (spec §A.1) across 8 placements × 3 workloads is deferred as an optimization — this after-`cse` placement is the spec's starting hypothesis and already banks a large, verified win. A follow-up can build the sweep to squeeze a few more percent.

## Known follow-up
The `with-redefs`-of-`map` case is guarded by `safe-to-fuse?`'s `stable-load-var?` clause (and the corpus run shows no breakage), but lacks a dedicated unit test (IR fixture complexity).


nnunley:
Folded into #440, which now carries the full loop-fusion effort (pass + enabled by default + v2 comp-chains/into) as a single commit rebased onto current main. Closing to avoid a redundant stacked PR.

PR #442, opened 2026-07-11
perf(rt): close timeout channels from one shared timer daemon

mparrett:
Every `(timeout ms)` call currently spawns its own goroutine: it holds a `time.Timer`, selects on it against scope cancellation, and closes the channel. That's a goroutine spawn, a stack, and a timer per call.

Timeout-heavy patterns pay for it. The canonical one is a frame-paced loop, `(<!! (timeout n))` per frame, which becomes a goroutine-per-frame treadmill. On stock Go that's measurable churn; on runtimes with heap-allocated fixed-size goroutine stacks (TinyGo), each spawn is a large allocation plus the GC cycle it drives, and the loop's cost is dominated by spawning rather than by anything the program does.

This PR replaces the per-call goroutine with one long-lived daemon that owns a deadline queue and closes due timeout channels. A `timeout` call now costs the channel plus a queue entry.

Semantics are unchanged:

- Timeout channels still close after their deadline.
- On scope cancellation a timeout closes early only when its own context generation was cancelled — each entry carries its creation-time context, matching the old per-call `select` on `ctx.Done()`. Timeouts created after a `CancelAll` installed a fresh generation keep their deadlines under a respawned daemon.
- The daemon starts lazily on the first `timeout` call; programs that never use `timeout` never run it.

Benchmark (darwin/arm64, Apple M2 — creation cost per `(timeout 1)` call, old shape kept in the bench file as the baseline):

```
BenchmarkTimeoutDaemon-8             3690286    338.8 ns/op    137 B/op    1 allocs/op
BenchmarkTimeoutGoroutinePerCall-8   1327947   2316 ns/op      438 B/op    6 allocs/op
```

Two implementation notes:

- The daemon rescans its queue linearly on each wake. Realistic timeout counts are small (the pathological pattern is many calls over time, not many pending at once), so a deadline heap felt like premature structure; easy to add later if a workload proves otherwise.
- `(timeout ms)` with `ms <= 0` now returns an already-closed channel synchronously. The old code spawned a goroutine whose 0-duration timer fired immediately, so the close landed asynchronously shortly after return; "closes after 0ms" is preserved, just without the detour.

The existing async suite covers close-after-ms and cancellation; the generation regression test and the benchmark pair are new. Full suite passes.

### DIFF
--- a/pkg/rt/async.go
+++ b/pkg/rt/async.go
@@ -37,6 +37,119 @@ type Pub struct {
 
 func init() { RegisterInstaller(installAsyncNS) }
 
+// --- Shared timeout daemon ---
+//
+// All (timeout ms) channels are closed by ONE long-lived goroutine instead of
+// a goroutine per call. A per-call goroutine is cheap on stock Go but not
+// free: each (timeout ms) paid a goroutine spawn, its stack, and a timer. On
+// runtimes with heap-allocated fixed-size goroutine stacks (TinyGo), a
+// timeout-per-frame animation loop turns that into a large allocation plus a
+// GC cycle per frame. The daemon makes timeout calls allocation-free apart
+// from the channel and a queue entry.
+//
+// The daemon runs in the root scope and exits on scope cancellation, closing
+// every pending timeout channel (same observable behavior as the old per-call
+// select on ctx.Done). timerRunning lets the next timeout call restart it in
+// the scope's next context generation after a CancelAll/Drain.
+
+type timerEntry struct {
+	deadline time.Time
+	ch       vm.Chan
+}
+
+var (
+	timerMu      sync.Mutex
+	timerQueue   []timerEntry
+	timerKick    = make(chan struct{}, 1)
+	timerRunning bool
+)
+
+// timeoutChan returns a channel that the timer daemon closes after ms.
+func timeoutChan(ms int) vm.Chan {
+	ch := make(vm.Chan)
+	if ms <= 0 {
+		close(ch)
+		return ch
+	}
+	deadline := time.Now().Add(time.Duration(ms) * time.Millisecond)
+	timerMu.Lock()
+	timerQueue = append(timerQueue, timerEntry{deadline: deadline, ch: ch})
+	start := !timerRunning
+	if start {
+		timerRunning = true
+	}
+	timerMu.Unlock()
+	if start {
+		vm.Goroutines.Go(timerDaemon)
+	} else {
+		select {
+		case timerKick <- struct{}{}:
+		default:
+		}
+	}
+	return ch
+}
+
+func timerDaemon(ctx context.Context) {
+	for {
+		timerMu.Lock()
+		now := time.Now()
+		var due []vm.Chan
+		keep := timerQueue[:0]
+		var next time.Time
+		for _, e := range timerQueue {
+			if !e.deadline.After(now) {
+				due = append(due, e.ch)
+			} else {
+				keep = append(keep, e)
+				if next.IsZero() || e.deadline.Before(next) {
+					next = e.deadline
+				}
+			}
+		}
+		timerQueue = keep
+		timerMu.Unlock()
+		for _, ch := range due {
+			close(ch)
+		}
+
+		if next.IsZero() {
+			// Idle: park until a new timeout arrives or the scope cancels.
+			select {
+			case <-timerKick:
+				continue
+			case <-ctx.Done():
+				timerDaemonStop()
+				return
+			}
+		}
+		t := time.NewTimer(time.Until(next))
+		select {
+		case <-t.C:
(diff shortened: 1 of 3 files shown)


mparrett:
Found one blocking correctness issue in the shared timeout daemon:

**[P1] Queue entries can cross scope context generations.** After `CancelAll` installs a fresh root context, the previous daemon may still have `timerRunning == true`. A new timeout is therefore appended to the old daemon's global queue; when that canceled daemon runs `timerDaemonStop`, it closes the newly created timeout immediately, even though the timeout belongs to the fresh context generation.

I reproduced this 10/10 times with a focused regression using `GOMAXPROCS(1)`, an existing long timeout, `vm.Goroutines.CancelAll()`, then `timeoutChan(200)`. The post-`CancelAll` channel closed immediately instead of waiting for its deadline.

The daemon/entries need to be associated with a context generation, or the canceled daemon must be fully retired before fresh-generation entries can enter its queue. Please also add a regression test covering `CancelAll()` immediately followed by a new timeout.

mparrett:
Confirmed and fixed in 954758a. The old per-call goroutine captured its
context at call time, so a post-`CancelAll` timeout survived the old
generation's cancellation — the daemon version lost that.

Each queue entry now carries the `Done` channel of the scope context current
at its creation. A stopping daemon closes only entries whose own generation is
cancelled and respawns a fresh daemon for any survivors; if another
`CancelAll` races the respawn, the new daemon's first pass closes the newly
cancelled entries via the same per-entry check.

Added the regression you asked for (`TestTimeoutSurvivesCancelAll`: long
timeout → `CancelAll` → `timeout(200ms)` under `GOMAXPROCS(1)`): it fails
10/10 on the previous commit with the exact symptom and passes with the fix.
Full suite green.

nooga (changes requested):
Concurrency design looks correct — verified with `-race` (x3) and the full suite, no issues found in locking/close-once/generation handling. Before merging, please add a Go benchmark (`bench_test.go`) showing the perf win per project policy for perf-sensitive changes, since that's the PR's whole premise. Optional: note the linear per-tick queue scan and the sync-close-on-`ms<=0` behavior change in the description.

mparrett:
Benchmark added in bb899ac, with the old per-call shape kept in the bench file
as the baseline so the pair runs comparably:

```
BenchmarkTimeoutDaemon-8             3690286    338.8 ns/op    137 B/op    1 allocs/op
BenchmarkTimeoutGoroutinePerCall-8   1327947   2316 ns/op      438 B/op    6 allocs/op
```

(darwin/arm64, Apple M2, `go test -bench Timeout -benchmem ./pkg/rt/`.)
~6.8× faster and 6→1 allocs per call on stock Go — and stock Go is the
favorable case for the old shape; the motivating case (fixed-size
heap-allocated goroutine stacks) is qualitatively worse.

Both description notes added as well: the linear per-wake queue scan (small
realistic queue sizes; a deadline heap is an easy later upgrade if a workload
proves otherwise) and the `ms <= 0` synchronous-close change.

mparrett:
Methodology and a robustness check on those numbers, since a single run can
mislead:

**Setup.** Both shapes run in one file: `timeoutChan` (daemon) against
`oldTimeoutChan`, the pre-PR implementation preserved as an in-file baseline.
Each op is one create with a 1ms deadline; entries drain about as fast as they
are created, so the daemon holds a steady queue of roughly 3k entries and its
per-wake linear rescan is exercised, not idle. That makes the bench regime
adversarial for the daemon: the real frame-pacing pattern is create-then-wait
with a queue near 1.

**Results** (`go test -bench Timeout -benchmem -cpu=1,8 -count=3`,
darwin/arm64, Apple M2):

```
                                   -cpu=1                    -cpu=8
TimeoutDaemon             119–156 ns/op, 1 alloc    370–502 ns/op, 1 alloc
TimeoutGoroutinePerCall  2074–2352 ns/op, 5 allocs  2133–2430 ns/op, 5–6 allocs
```

Roughly 15–19× single-core and 4.3–6.6× at 8 cores. The `-cpu=1` rows are the
robustness anchor: there the daemon's queue rescan must share the only core
with the caller, so its cost is fully included — no work hiding on another
core — and the gap widens rather than narrows. Run-to-run variance stays
within about ±15% per configuration; the allocation counts (1 vs 5–6) are
stable across every run, and allocation pressure is what matters most on the
constrained runtimes the PR description calls out.

mparrett:
Ready for a final look.

nooga (approved):
Benchmark resolves the prior ask — real, reproducible 3x time / 4x alloc win (`BenchmarkTimeoutDaemon` vs `BenchmarkTimeoutGoroutinePerCall`), clean pair-comparison, no correctness regressions (race-clean, full suite green, diff scope unchanged since last review). LGTM.

PR #443, opened 2026-07-11
feat(bytecode): reject .lgb bundles compiled against a different opcode set

mparrett:
If the opcode enum changes between the tree that compiled a `.lgb` and the tree
running it — an opcode inserted, removed, or reordered — every later opcode
shifts meaning. Today that fails as an "unknown instruction" panic deep in
execution (or worse, runs silently wrong), and nothing in the error points at
the cause. I lost most of a day to this after #379 inserted the bitwise
opcodes mid-enum: a bundle from a post-#379 tree run on a pre-#379 runtime
died with `unknown instruction op=37`, which reads like a VM bug, not a
version skew.

This PR makes the skew fail fast at decode with a message that names the cause:

```
opcode set mismatch: bundle compiled with 52 opcodes (signature 8c6f19e2a4b07d31), runtime has 45 (7350b02386b0793d) — recompile the bundle with a matching lg
```

Mechanism:

- `vm.OpcodeSetSignature()` returns the opcode count and an FNV-64a hash of the
  mnemonics in enum order. The mnemonics move from a literal inside
  `OpcodeToString` to a package-level table so the signature and the
  disassembler share one inventory.
- The signature rides the v2 header's existing capability mechanism as the
  first defined bit, `CapOpcodeSet`: mask, then varint count + uint64 hash.
  `ModuleBuilder.Build` sets it on every new bundle.
- The decoder (both the module and exec-unit paths, now sharing one
  `readCapabilities` helper) compares against the running VM and rejects on
  mismatch.

Compatibility:

- Pre-guard bundles (no `FlagCapabilities`) decode unchanged.
- A new bundle on an older runtime hits the existing
  `unsupported capability mask` check: a clean refusal instead of a misdecode.
  Anyone distributing bundles to older runtimes needs to keep producing them
  with a matching older `lg`.
- One existing test (`TestUnsupportedCapabilityMask`) hand-crafted bit 0 as an
  unsupported capability; it now uses bit 1, preserving its intent.

Tests: roundtrip (same-VM bundle passes), tampered hash and tampered count both
rejected with the mismatch message, legacy bundle without a capability mask
still decodes.


mparrett:
Found two issues:

**[P1] The committed core bytecode bundle is stale.** `ModuleBuilder.Build` now emits `FlagCapabilities | CapOpcodeSet`, which changes every newly generated bundle, but `pkg/rt/core_compiled.lgb` was not regenerated. I ran `make check-generated`; it failed with:

```
ERROR: pkg/rt/core_compiled.lgb is stale — the regenerated bytes differ.
       Run 'make generate' and commit the regenerated bundle.
```

Please run `make generate` and commit the updated artifact.

**[P2] The signature inventory can silently drift from the actual opcode enum.** `OpcodeSetSignature()` hashes only the manually maintained `opcodeNames` slice. If a future opcode is added or reordered without updating that slice, emitted bytecode changes while the signature remains unchanged, defeating the compatibility guard. Please add an `OP_COUNT` sentinel plus an enforced `len(opcodeNames) == int(OP_COUNT)` check, or generate both definitions from one source.

The affected package tests (`go test ./pkg/bytecode ./pkg/vm`) otherwise pass.

mparrett:
Both addressed:

**P1** — regenerated in d81998c via `make generate`; only `core_compiled.lgb`
changed (the Go-gen artifacts came out byte-identical) and
`make check-generated` passes again. I had missed that the builder change
invalidates the committed bundle.

**P2** — ff91636 adds the `OP_COUNT` sentinel at the end of the enum and an
`init` that panics unless `len(opcodeNames) == OP_COUNT`, so an opcode added
or removed without updating the inventory fails every build immediately. One
honest limit: a pure reorder that keeps the count can still desynchronize the
slice — but that has always broken `OpcodeToString`'s disassembly too, since
both read the same table; the sentinel closes the add/remove case the guard
actually created.

Full suite green after both.

nooga (approved):
Solid, well-tested fix — verified independently (full `go test ./... -count=1` passes on the PR branch). The signature mechanism (count + FNV hash, shared `opcodeNames` table with an init-time consistency panic) correctly closes the exact failure mode described, and backward/forward compatibility is handled cleanly. LGTM.

PR #444, opened 2026-07-11
fix(core): format time macro elapsed duration as msecs

nooga:
## Summary
- `(time expr)` printed the raw nanosecond count (e.g. `Elapsed: 3611916`) instead of a formatted duration, because `time.Time.Sub` returns a `time.Duration` whose Go kind is `int64`, and `vm.BoxValue`'s integer fallback unboxes it to a bare `Int` rather than preserving `Duration.String()`.
- Reformats the `time` macro to compute milliseconds directly (`(/ (double elapsed) 1000000.0)`) and print via `prn`, matching Clojure/babashka's `"Elapsed time: X.XXXXXX msecs"` output.
- Regenerated `pkg/rt/core_compiled.lgb` via `go generate ./pkg/rt/`.

## Test plan
- [x] `go test ./... -count=1`
- [x] Added `test/time_test.lg` covering the return value and the printed message format
- [x] Manually verified: `go run . -e '(time (reduce + (map inc (range 9999))))'` → `"Elapsed time: 1.907959 msecs"`


nooga (commented):
LGTM — the msecs formatting matches Clojure's own `time` macro semantics, `double`/`Int` boxing checked out, and the new `test/time_test.lg` covers both return value and print format. Full test suite (including genmanifest staleness check) passes.

(Posting as a comment rather than an approval since GitHub doesn't allow self-approval on your own PR.)

nnunley (approved):

PR #446, opened 2026-07-12
feat(ir): deterministic lowered tree — densify + work-unit typeinfer guard

nnunley:
Makes the `-tags gogen_ir` lowered Go tree **byte-deterministic** across machines and runs, so a byte-identical generate-twice gate is meaningful and the tree can eventually become a committed artifact with near-zero regeneration churn.

## Two timing-dependent nondeterminism sources removed

**A. Concurrent position allocation → formatting jitter.** Generated `ast.Node`s got synthetic `token.Pos` from a global `allocPos()` counter. Under concurrent lowering, goroutines' allocations interleave, leaving position gaps that `go/format` renders as spurious blank lines — identical ASTs rendered to different bytes run-to-run. Fixed by `renderFset` + `eachPosField` (`pkg/rt/gogen.go`): at render time, every `token.Pos` reachable from the node is renumbered to a dense, sorted, gap-free sequence, making rendered text a pure function of the AST.

**B. Wall-clock typeinfer budget → code divergence.** `typeinfer.lg` bailed its fixpoint drain on a 2000ms wall-clock budget (`System/nanoTime`), so a slow run inferred fewer types → different lowered Go. Replaced with a deterministic **work-unit cap** `*typeinfer-max-drains*` (drain count, not milliseconds). The bail point is now machine-independent; the cap is a defensive backstop set far above real core (`TestTypeinferNeverBailsOnCore` verifies zero-bail on the full stdlib). The codegen path binds it to `nil` (unbounded).

## The gate

`test/e2e/lowering_determinism_test.go`'s `TestLoweringDeterminism` is repurposed from two-cold-runs into **cold** (`-tags bootstrap`) vs **hot** (`-tags "bootstrap gogen_ir"`) `--target=go` **byte-identical** comparison — validating the self-hosting fixpoint — and is now default-on (env gate removed; `testing.Short()` skip kept).

## Verification

- Full suite (12 packages) green; `make check-generated` in lockstep; bundle byte-stable.
- Gate passes: cold and hot lowering byte-identical; `make check-selfhost` OK.
- **No perf regression.** An isolated branch-vs-main A/B (same machine, anchor-normalized) shows all compile/runtime benchmarks flat within ±3.6% (single-sample noise): IRCompile `[gogen_ir]` +1.1%, ClojureTestSuite `[ir_bytecode]` −0.1%, `[aot_native]` −0.5%. On fast hardware the old budget never fired, so `nil` does identical work; the benefit is guaranteeing that on slower hardware too.

## Ordering note

The work-unit bail point is deterministic because the drain/enqueue order already is (`graph/toposort` is DFS deps-first with a sorted-root tie-break; `sccs` sorts its seed; `initial-work-queue` is structural block/inst order). No change needed there.

## Follow-ups (deliberately out of scope)

- **Spans for provenance:** `vm.SourceInfo` already carries full spans and the reader captures them, but IR build attaches name-only info. Threading real spans → `//line` directives would give debuggable generated Go. Separate change (does not replace densify — `go/format` needs positions monotonic in emission order, which source spans are not).
- `eachPosField` could grow a warning-logging `default` case for defense-in-depth against future AST node types.
- `bench-ratchet` `baseline.json` is stale vs current main and should be re-anchored in its own change on main.

---
**Part of #258** (Epic: Lowering & cross-target codegen) — lowered-tree determinism, the theme #175 opened.


nooga:
## Review

Solid, well-motivated fix for the lowering-nondeterminism problem — pushing position assignment out of the concurrent lowering path into a pure post-processing step at render time is the right approach, and swapping the wall-clock typeinfer budget for a drain-count cap is the correct way to make the bail point reproducible. Two issues found, one blocking.

### Blocking: `TestTypeinferNeverBailsOnCore` hardcodes a personal machine path

`pkg/ir/lisp_typeinfer_test.go` (new test, no `t.Skip`, runs under plain `go test ./...`):

```go
cmd.Dir = filepath.Join(os.Getenv("HOME"), "development", "let-go")
if cmd.Dir == "/development/let-go" {
    // Fallback if HOME is not set
    cmd.Dir = "/Users/ndn/development/let-go"
}
```

This only resolves correctly if the repo is checked out at exactly `$HOME/development/let-go`. GitHub Actions checks out to `$GITHUB_WORKSPACE` (e.g. `/home/runner/work/let-go/let-go`), so this will fail in CI. The "fallback" branch is effectively dead code (`cmd.Dir` won't equal that literal string in any real environment), and even if triggered it falls back to `/Users/ndn/development/let-go` — an apparent personal path.

The repo already has the right pattern for this: `test/e2e/helpers_test.go`'s `repoRoot(t)` walks up from `runtime.Caller(0)` to find `go.mod`. This test should use the equivalent instead of guessing from `$HOME`.

### Correctness: `scripts/ti-baseline.lg`'s `CAP` constant is now the wrong unit

```clojure
(def CAP 50)   ; ms per-fn typeinfer budget          ; <- untouched by this PR
...
(binding [ir.passes.typeinfer/*typeinfer-max-drains* CAP   ; <- renamed by this PR
```

`CAP` was tuned as "50ms" of wall-clock budget. The PR renames the binding site to the new drain-count var but leaves `CAP`/its comment as digital "ms". Given the new default cap elsewhere is 2,000,000 drains, and `TestTypeinferNeverBailsOnCore` asserts real core functions never hit that, 50 drains is almost certainly far below what any nontrivial function needs to reach fixpoint. This script will now bail near-instantly on most/all functions, silently corrupting the wide-join-fraction/timing data it's meant to produce. Not CI-breaking, but worth fixing alongside the rename since this PR is what changed the semantics.

### Minor / already acknowledged

`eachPosField` doesn't cover `ast.TypeSpec.Assign` or `ast.ImportSpec.EndPos` — not a live bug today (those fields are never populated by `cTypeDecl`/`cImportSpec`), and the PR's own follow-ups already flag a defensive `default` case for `eachPosField` that would also catch this.

### Test coverage

`TestRenderFsetIsPositionIndependent` / `TestRenderFilePositionIndependent` directly exercise the fix's core claim, and the cold-vs-hot rework of `TestLoweringDeterminism` is a stronger gate than the old two-cold-runs version. The hardcoded-path issue above undermines the new typeinfer test's value since it can't run anywhere but the author's machine as written.

nooga:
## Re-review (after fixes)

**Scope:** diff between `main` (d6c09d24) and `feat/lowered-tree-determinism` (35e6a28), 12 files.

### Previous blocking issue: RESOLVED ✅

The earlier review flagged `pkg/ir/lisp_typeinfer_test.go`'s `TestTypeinferNeverBailsOnCore` for hardcoding `$HOME/development/let-go` (and a dead `/Users/ndn/...` fallback). That test no longer exists anywhere in the tree — its assertion ("typeinfer never bails on real core") has been folded into `TestLoweringDeterminism` in `test/e2e/lowering_determinism_test.go`, which now captures **stdout** from both the cold and hot `lgbgen` runs and fails if `"typeinfer: hit *typeinfer-max-drains*"` appears. I verified `println`/`prn` route through `*out*` (defaults to `os.Stdout`, `pkg/rt/iort.go:392`), so the capture is correctly wired — a real, portable replacement, not just a deletion of coverage.

### Previous correctness issue: NOT RESOLVED ⚠️

`scripts/ti-baseline.lg` still has:
```clojure
(def CAP 50)   ; ms per-fn typeinfer budget
...
(binding [ir.passes.typeinfer/*typeinfer-max-drains* CAP ...
```
The binding site was renamed to the new drain-count var, but `CAP`/its comment are untouched. `50` was tuned as *milliseconds*; against a drain-count var whose new production default is `2,000,000`, `50` will make the script bail almost immediately on essentially every function, silently corrupting whatever timing/join-fraction data `ti-baseline.lg` is meant to produce. Not CI-breaking (standalone script), but this PR is exactly what changed the semantics of that binding, so worth fixing alongside it.

### New finding: stale comment in `Makefile` (not touched by this diff, but now factually wrong)

`Makefile:290-291` (in `check-generated`, describing why `core_go_lowered/` isn't committed):
```
#     trips the wall-clock *typeinfer-budget-ms*, so the bytes are not
#     reproducible and committing it would churn.
```
This describes precisely the nondeterminism this PR fixes, using the variable name this PR renamed — it now contradicts the PR's own premise ("the tree can eventually become a committed artifact"). Worth a follow-up edit.

### Verified correct (no new issues found)

- **`renderFset`/`eachPosField` (`pkg/rt/gogen.go`)**: built a standalone repro of the densify logic with two `FuncDecl`s separated by an artificial position gap (simulating concurrent `allocPos` interleaving) — `go/printer` still inserts the blank line between top-level funcs regardless of gap size, so densifying doesn't collapse deliberate inter-declaration spacing; it only removes *spurious* gaps.
- `eachPosField` is still missing `ast.TypeSpec.Assign` / `ast.ImportSpec.EndPos` — confirmed still dead (not populated by any `c*` constructor), and already flagged in the PR's own follow-up notes. Not blocking.
- CI wiring (`go.yml`) is consistent: main `build`/`build-bootstrap` jobs already pass `-short`, so the standalone "Expensive lowering e2e" job now runs `TestLoweringDeterminism`/`TestDeftypeSkeletonNativeLowering` unconditionally, matching intent.
- No remaining references to the old `*typeinfer-budget-ms*` name in any file this PR touches.

### Bottom line

The blocker from the last round is fixed well. One previously-flagged correctness nit (`ti-baseline.lg` CAP) is still outstanding, plus one new minor doc-drift item (`Makefile:290`). Neither blocks merge, but both are quick fixes caused directly by this PR's rename.

nnunley:
Both outstanding nits from the re-review are fixed (branch also squashed to one commit and rebased onto current main): (1) `scripts/ti-baseline.lg` CAP is now a drain-count — 50,000 with the comment naming the production default (2,000,000); 50 was the stale ms-era value corrupting every measurement. (2) The `Makefile` check-generated comment now describes the post-PR reality (deterministic `*typeinfer-max-drains*`, not wall-clock `*typeinfer-budget-ms*`), and the parallel-lowering design doc got the same correction. `grep -rn 'typeinfer-budget-ms'` is zero hits tree-wide; full suite + ir-stress ratchet green.

mparrett:
Agent review found two blockers and one smaller correctness issue.

1. **[P1] The “hot” determinism test is actually cold.**

   At `test/e2e/lowering_determinism_test.go:48`, `-tags gogen_ir` is built before generating the ignored lowered tree and `cmd/lgbgen/main_gogen_ir.go`. CI only runs `go generate ./pkg/rt`, which regenerates the bytecode bundle. In a clean checkout, `go list` shows only `[generate.go main.go]` and no `core_go_lowered` dependency, so the gate compares the interpreted compiler with itself and can pass without testing self-hosting. Generate the lowered tree/wireup before building hot, and assert the native dependency is present.

2. **[P1] Position allocation remains data-racy.**

   The new render-time remapping at `pkg/rt/gogen.go:2059-2067` assumes concurrent `allocPos` calls merely interleave, but `goPosNext++` is unsynchronized. A focused `go test -race` reproduction reports races at `gogen.go:1886-1888`. Lost updates can create duplicate positions, which the later unique-position map cannot reconstruct. Synchronize or replace the shared allocator and retain a concurrent race test.

3. **[P2] Drain cap permits one extra unit.**

   `pkg/rt/core/ir/passes/typeinfer.lg:533` checks `tick > max`; a cap of zero still performs one drain. Check queue exhaustion first, then use `>=` to enforce the documented cap without reporting a bail when convergence occurs exactly at the limit.

Existing PR checks are green. Locally, the changed renderer tests, `pkg/rt`, `test`, and short e2e tests passed. The temporary race reproduction was removed and the checkout was clean after review.

nooga (commented):
Two earlier review rounds (in-thread) already got the wall-clock-budget rename and the HOME-hardcoded test fixed — those are resolved in the current HEAD (38d3594). But @mparrett's most recent review (posted after this commit) raises three issues that remain unaddressed; I re-verified all three directly against the diff/tree and they're real:

**P1 — `pkg/rt/gogen.go:1884-1889` (`allocPos`): unsynchronized shared counter is a genuine data race.**
```go
func allocPos() token.Pos {
	p := token.Pos(goPosNext)
	goSynthFile.AddLine(goPosNext)
	goPosNext++
	return p
}
```
`goPosNext` is a package-level `int` incremented from concurrent lowering goroutines with a plain `++` — no mutex, no atomic. This is exactly the mechanism the PR's own description says interleaves under concurrent lowering ("goroutines' allocations interleave, leaving position gaps"). A `go test -race` repro confirms actual races here (per mparrett), and this repo doesn't run `-race` anywhere in CI or the Makefile, so nothing catches it. Beyond gaps, a lost update can hand two unrelated AST nodes the same raw position, which `renderFset`'s collect/remap can't disambiguate — undermining the byte-determinism guarantee this PR is supposed to establish. Please synchronize `goPosNext` (`atomic.Int64`, or a mutex around the read-increment) rather than relying on the render-time remap to paper over a racy source.

**P1 — `test/e2e/lowering_determinism_test.go` (`buildLgbgenTags`/hot build): the "hot" build never generates the gogen_ir wireup, so the gate may compare cold against itself.**
`hot := buildLgbgenTags(t, root, "bootstrap gogen_ir")` just does `go build -tags "bootstrap gogen_ir"`. `cmd/lgbgen/main_gogen_ir.go` (gated `//go:build bootstrap && gogen_ir`) and the native `core_go_lowered` tree it wires in are gitignored build artifacts per this PR's own Makefile comment ("NOT committed... regenerated on demand"). Nothing in `.github/workflows/go.yml`'s "Expensive lowering e2e" step (which just runs `go test ... -run 'TestLoweringDeterminism|...'` after `go generate ./pkg/rt/`, which only regenerates the bytecode bundle) produces that wireup first. In a clean checkout, `-tags gogen_ir` therefore selects no additional file, and the 'hot' binary is compiled identically to 'cold' — the comparison can trivially pass without ever exercising the native self-hosting path this PR's gate claims to validate. Please have the harness generate the wireup (e.g. run `lgbgen --target=go`/`--target=both` before the hot build) and assert the native dependency is actually present (e.g. via `go list -tags ...`) before comparing.

**P2 — `pkg/rt/core/ir/passes/typeinfer.lg` (drain-loop `cond`): off-by-one lets the cap admit one extra drain.**
The bail check `(and *typeinfer-max-drains* (> tick *typeinfer-max-drains*))` runs before the empty-queue check and uses `>` on a counter that starts at 0, so a cap of `N` performs `N+1` drains before bailing (cap `0` still does one). Minor given the production default is 2,000,000, but worth tightening (`>=`, and/or check queue exhaustion first) since this PR is what's redefining the cap's semantics.

None of the three are architecture calls — all are mechanical fixes well within scope for this PR, and the author has moved fast on the prior two rounds. But #1 and #2 both go to whether this PR's central claim (byte-identical, race-free, self-hosting-validated lowering) actually holds, so I'd hold off merging until they're addressed rather than relying on CI green (CI doesn't run `-race` and doesn't currently prove the hot build is exercising the native path).

nnunley:
All three round-3 items verified against the code and addressed in d4e3d411 (rebased onto post-#407 main along the way):

**P1 — `allocPos` race.** Confirmed: plain read-increment on `goPosNext` from concurrent lowering goroutines; a lost update can mint duplicate positions that `renderFset`'s densify cannot disambiguate. Fixed with a mutex around the whole body rather than an atomic on the counter alone — that also keeps `AddLine` calls monotonic, which the function's own line-table comment relies on. `TestAllocPosConcurrentUniqueness` hammers it from 16 goroutines under `-race` and asserts every position unique; verified red (race detector fires) with the mutex removed.

**P1 — vacuous hot build.** Confirmed: in a clean checkout `-tags gogen_ir` selected no files and hot compiled identical to cold. The harness now materializes the wireup + native lowered tree with the cold binary before the hot build (the same default-location artifacts `make lowered` produces — gitignored, so no tracked state is touched), and `assertNativeLoweringLinked` fails the gate unless `go list -deps -tags "bootstrap gogen_ir" ./cmd/lgbgen` actually contains `core_go_lowered`. The full gate (materialize → assert → cold/hot byte-compare) passes end-to-end, now demonstrably exercising the native path.

**P2 — drain-cap off-by-one.** Confirmed: `>` on a 0-started tick gave N+1 drains and cap 0 drained once. The `cond` now checks queue exhaustion first (a queue that empties at the cap is a completed pass, not a bail) and uses `>=`, so cap N performs exactly N drains and cap 0 none. `TestTypeinferDrainCapSemantics` pins exact counts via `lat/*ti-counters*`: cap 0 → 0 drains, cap 2 → 2, uncapped > 2, ample cap == uncapped.

One note for the record: the byte-determinism gate is now ~10 minutes (two full lowerings plus the materialize), which exceeds the pre-push hook's 60s `go test` budget by design — it runs standalone and in the CI e2e step.

nooga (approved):
Verified: all three blockers fixed. goPosNext is now guarded by a mutex with a 16-goroutine race test; the determinism test now materializes the native core_go_lowered tree before comparing cold vs. hot; and the typeinfer drain-count check is now >= with correct ordering, pinned by a new test. Also appreciate the incidental fix to the bail-message format mismatch with the ratchet script.

mparrett (approved):
All three round-3 findings verified fixed at b63a6b2a — approving.

Verification:

1. **`allocPos` race**: the mutex covers the whole read/`AddLine`/increment body, and every `goPosNext`/`goSynthFile` access in the tree goes through `allocPos`. `TestAllocPosConcurrentUniqueness` passes under `-race`; I re-verified red independently (with the lock removed the race detector fires and the test fails; restored, it passes).
2. **Vacuous hot build**: `materializeNativeLowering` produces the wireup + lowered tree with the cold binary before the hot build, and `assertNativeLoweringLinked` fails unless `go list -deps` on the hot tags contains `core_go_lowered` — the gate can no longer compare cold with itself. I ran the full gate locally: materialize, link assertion, and cold/hot byte-compare all pass (~180s here).
3. **Drain cap**: queue exhaustion is checked before the cap and the comparison is `>=`; `TestTypeinferDrainCapSemantics` pins exact counts (cap 0 → 0 drains, cap 2 → 2, ample cap == uncapped).

The self-review follow-ups hold up too: the bail line's trailing "on <name>" round-trips through turnaround-ratchet's `parse-bailed-fns`, and the design-doc caveat on `renderFset`'s rendered-once contract is the right warning to leave for the future `pmapv` step. The env-var → `testing.Short()` gating swap is consistent across the three heavyweight e2e tests, CI runs them full, and `make test` (which now passes `-short`) keeps the local loop fast.

Locally at b63a6b2a: full suite green (e2e determinism gate run standalone, as designed). CI is green on the head.

PR #447, opened 2026-07-12
feat(core): add update-keys/update-vals

### LINKED ISSUE #426: stdlib compat: `update-keys`/`update-vals`, transducer-arity `remove`, and `re-find` empty-group vs nil
## Context

Found while getting [lambdaisland/cli](https://github.com/lambdaisland/cli/tree/let-go-compat) to run under let-go (the `let-go-compat` experiment branch). These are small `clojure.core` / regex-semantics gaps, verified against current `main`. The `clojure.java.io` shim from the same report is filed separately.

## 1. Missing `update-keys` / `update-vals`

Clojure 1.11 core fns, not present in let-go (`grep -n 'update-keys\|update-vals' pkg/rt/core/*.lg` → nothing). The medley-compat work (#157) merged separately but did not add these. Straightforward `.lg` additions:

```clojure
(defn update-keys [m f] (into {} (map (fn [[k v]] [(f k) v])) m))
(defn update-vals [m f] (into {} (map (fn [[k v]] [k (f v)])) m))
```

## 2. `remove` missing the 1-arity transducer form

`pkg/rt/core/core.lg:1497` defines only `(defn remove [pred coll] ...)`. Clojure's `remove` also has a **1-arity transducer** form `([pred] (filter (complement pred)))`, which cli relies on. Add the `[pred]` arity mirroring `filter`.

## 3. (resolved — noted for completeness)

The same report hit an arity-detection bug: a single-body multi-arity form with a destructured param and `& args` —

```clojure
(defn assoc-flag ([opts {:keys [middleware handler] :as flagspec} & args] ,,,))
```

— failed with `function ... doesn't have a 2-arity variant`. **This no longer reproduces on current `main`** (verified: the exact form runs correctly). No action needed; recording it so the report is fully accounted for.

## 4. Regex `\R` (expected — noted, no action)

Go's RE2 doesn't support `\R`. cli worked around it with `\n`; a faithful replacement is `(\r\n|\r|\n)`. This is an RE2 limitation, not a let-go bug — noting only in case we want to document known regex-dialect differences.

## 5. `re-find` yields `""` for a non-participating optional group (should be `nil`)

When an optional capturing group doesn't participate — e.g. `(no-)?` matching nothing — Clojure's `re-find` returns `nil` for that group; let-go returns `""`.

Root cause: `pkg/rt/lang.go:5145` (`re-find`) uses `re.FindStringSubmatch` and boxes **every** element as `vm.String(m)`. Go returns `""` for a non-participating group, indistinguishable from an empty match via that API. Fix: use `FindStringSubmatchIndex` and emit `nil` where a group's `[start,end]` pair is `-1,-1`, boxing the rest as `String`. (`re-matches`, `re-groups`, and any other `FindStringSubmatch` callers likely share this and should be fixed together.)

Repro:
```clojure
(re-find #"a(b)?" "a")   ; Clojure => ["a" nil] ; let-go => ["a" ""]
```

## Code references
- `remove`: `pkg/rt/core/core.lg:1497`
- `re-find`: `pkg/rt/lang.go:5145`

nooga:
## Summary
- Adds `update-keys`/`update-vals` to `clojure.core` (`pkg/rt/core/core.lg`), matching Clojure 1.11 semantics (metadata preserved, later key wins on collision).

Fixes #426 (the update-keys/update-vals portion; #426's third item — transducer-arity `remove` — is untouched and stays open).

## Scope note
Rebased and narrowed per review: the `re-find`/`re-matches`/`re-seq` nil-vs-`""` capture group fix originally included here is dropped — #456 already landed the equivalent fix on `main` independently (verified: `pkg/rt/lang.go`'s `re-find`/`re-matches`/`re-seq`/`re-groups` all already distinguish absent vs empty-string groups via index-based submatches). That overlap was the source of the merge conflict.

## Test plan
- [x] `test/update_keys_vals_test.lg` — deftest covering basic behavior, empty maps, key-collision, and metadata preservation
- [x] `go build ./...`
- [x] `go test ./... -count=1`
- [x] `go generate ./pkg/rt/` (regenerated precompiled core bytecode)


nnunley (commented):
Heads-up before this lands: the `re-find`/`re-matches`/`re-seq` half here is now **superseded by #456** (merged), which rewrote the same `pkg/rt/lang.go` handlers + `pkg/vm/regex.go` to the identical nil-vs-`""` semantics — that overlap is the source of the current conflict. Verified on current `main`: `(re-find #"(a)?b" "b")` ⇒ `[b nil]`, `(re-matches #"(a)?b" "b")` ⇒ `[b nil]`, `(re-seq #"(x)" "abc")` ⇒ `nil`, all already correct.

`update-keys`/`update-vals` are still needed (verified missing on `main`) and look right — `reduce-kv` + transient, `meta` preserved. Suggest rebasing and reducing this PR to just the two map fns.

One more: #426's third item — transducer-arity `remove` — isn't covered by either PR, so #426 shouldn't auto-close on this one.

nnunley:
One gap: `re-groups` (pkg/rt/lang.go:6282) still boxes every group via `FindAllStringSubmatch`, so it keeps returning `""` for a non-participating optional group — worth routing through the new `FindAllStringSubmatchPresent` in the same PR so all four regex builtins agree.

nooga:
Rebased onto current \`main\` and narrowed scope per review — dropped the \`re-find\`/\`re-matches\`/\`re-seq\` changes since #456 already landed the equivalent nil-vs-\`""\` fix (confirmed \`re-groups\` on main already routes through index-based submatches too, so there's no remaining gap there). This PR is now just \`update-keys\`/\`update-vals\`. Updated the PR description accordingly; #426 stays open for the transducer-arity \`remove\` item.

nnunley (approved):

PR #449, opened 2026-07-12
fix(compiler): parse empty catch bodies with simple class names

### LINKED ISSUE #448: Empty catch body with a simple class name fails to compile
Found while loading [`clojure.tools.cli` v1.4.256](https://github.com/clojure/tools.cli/tree/v1.4.256), which contains a typed catch clause with no body.

## Repro

```clojure
(try (throw "boom") (catch Throwable _))
```

Actual:

```
error: Can't resolve _ in this context
```

Expected: the clause catches the thrown value and returns nil, as in Clojure.

## Cause

`tryCompiler` decides whether the first catch token names a class. A qualified name (fixed in #413) or a token count of three or more marks it as one. `(catch Throwable _)` has two tokens and an unqualified name, so the compiler reads `Throwable` as the exception binding and `_` as a body form. The IR builder (`pkg/rt/core/ir/build.lg`) applies the same rule and fails the same way.

## Fix

Treat an uppercase simple symbol as class-shaped in both the compiler and the IR builder.

Relation to the other tools.cli issues: independent, mergeable in any order. The branch regenerates `pkg/rt/generated.sums` because `ir/build.lg` feeds the digest.

abogoyavlensky:
`(catch Throwable _)` with no body parsed the class symbol as the binding and the binding as a body form, so the clause failed to compile. #413 fixed the same ambiguity for qualified class names; simple names still fell back to a three-token minimum. Treat an uppercase simple symbol as class-shaped in both the compiler and the IR builder, which mirrors the same catch parsing.

Found while loading `clojure.tools.cli`, which contains a valid typed catch clause with no body.

Merge note: independent of the other `tools.cli` compat branches. It regenerates `pkg/rt/generated.sums` (`ir/build.lg` feeds the digest), so re-run the generator after rebasing past any branch that also touches it.

Resolves: #448

### DIFF
--- a/pkg/compiler/compiler.go
+++ b/pkg/compiler/compiler.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"sort"
 	"strings"
+	"unicode"
 
 	"github.com/nooga/let-go/pkg/rt"
 	"github.com/nooga/let-go/pkg/vm"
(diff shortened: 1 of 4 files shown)


nooga (approved):

nooga:
@abogoyavlensky thx, LGTM

PR #451, opened 2026-07-12
fix(rt): add compile-only Exception constructor stub

### LINKED ISSUE #450: (Exception. msg) on a never-executed path keeps a namespace from loading
Found while loading [`clojure.tools.cli` v1.4.256](https://github.com/clojure/tools.cli/tree/v1.4.256). The library calls `(Exception. msg)` in a fallback branch that a successful parse never runs, but the unresolved constructor keeps the whole namespace from loading.

## Repro

```clojure
(defn f [] (Exception. "x"))
:loaded
```

Actual:

```
error: Can't resolve ->Exception in this context
```

Expected: the namespace loads. Calling `(f)` may fail, since let-go has no JVM exceptions to construct.

## Fix

Resolve `Exception.` and `->Exception` to a compile-only stub that fails loudly when called (`"Exception. is unavailable under let-go"`), instead of fabricating a JVM exception. This follows the compile-only classpath-scan stubs from #421 and the existing `java.util.ArrayList.` load-only constructor.

Relation to the other tools.cli issues: independent, mergeable in any order; touches only `pkg/rt/lang.go` and one test file.

abogoyavlensky:
(Exception. message) appears in JVM-oriented libraries on branches that never run under let-go. `clojure.tools.cli` reaches it only in a fallback that a successful parse skips, yet the unresolved constructor kept the whole namespace from loading. Resolve `Exception.` and `->Exception` to a stub that fails loudly when called, instead of fabricating a JVM exception. This follows the compile-only stub approach from #421.

Merge note: independent of the other tools.cli compat branches; touches only `pkg/rt/lang.go` and one test file.

Resolves: #450


nooga:
Thanks for the contribution @abogoyavlensky ... I'm not sure about this though. I think much better approach would be to just implement exceptions properly at this point. Would you be interested in expanding this and #449 into Clojure-compatible extensions?

abogoyavlensky:
@nooga Yes, sure, I will take a look into this.

abogoyavlensky:
@nooga Can we make it in several stages:
- first we can merge this one #449  as-is (it's parsing, needed regardless)
- then I prepare hierarchy + constructors as the reworked ctor-stub PR
- then typed dispatch

Also I have a question: do you want the hierarchy under the `java.lang.*` names (max library compat) or as `let-go.lang.*` with aliases?

fyi And one caveat: runtime errors currently all surface as `ExceptionInfo`, so finer-grained classification (e.g. `ArithmeticException` for divide-by-zero) can come later without breaking anything.

nooga:
@abogoyavlensky sounds good! 

Let's go with `java.lang.*` (e.g. `java.lang.Exception`, `java.lang.RuntimeException`, `java.lang.ArithmeticException`, `java.lang.IllegalArgumentException`...), matching the `java.lang.Throwable` root and the existing `java.util.*`/`clojure.lang.*` namespacing already in `pkg/rt/lang.go`.

nnunley (approved):
Verified on the branch: `(defn f [] (Exception. "x")) :loaded` now **loads** (the dead-path constructor no longer blocks the namespace), and calling `(f)` fails loudly with a clear message. The compile-only stub pattern matches the existing `enumeration-seq`/classpath-scan stubs.

I pushed two small commits via maintainer-edit to unblock CI: the stub's error string was reworded to lead lowercase (staticcheck `ST1005` — golangci-lint now `0 issues`), and `test/clojure_compat_aliases_test.lg` updated to assert the new message. All checks green. LGTM.

abogoyavlensky:
@nooga fyi, In continuation of the discussion on exceptions improvement, I've created 3 separate PRs, with recommended merging order: https://github.com/nooga/let-go/pull/472 -> https://github.com/nooga/let-go/pull/474 -> https://github.com/nooga/let-go/pull/476

PR #453, opened 2026-07-12
fix(core): support the partition padding arity

### LINKED ISSUE #452: partition lacks Clojure's padding arity (partition n step pad coll)
Found while running [`clojure.tools.cli` v1.4.256](https://github.com/clojure/tools.cli/tree/v1.4.256), which pairs validation predicates with messages via `(partition 2 2 (repeat nil) validate)`.

## Repro

```clojure
(partition 2 2 (repeat nil) [pos? "must be positive" int?])
```

Actual:

```
error: function <mfn partition ...> doesn't have a 4-arity variant
```

Expected (Clojure): `((pos? "must be positive") (int? nil))`.

## Semantics

Clojure's `(partition n step pad coll)` emits complete groups as usual. When input remains but fewer than `n` items are available, it tops up that final group from `pad`, emits it, and stops. A short `pad` may still leave the final group short, and empty input yields no padded group.

abogoyavlensky:
partition accepted only two and three arguments. Clojure also defines (partition n step pad coll): emit complete groups as usual, then pad the final short group from pad, emit it, and stop; empty input yields no padded group, and a short pad may still leave the final group short. Add that arity to the existing loop, matching Clojure's semantics for each case above.

`clojure.tools.cli` relies on it to pair validation predicates with messages via (partition 2 2 (repeat nil) validate).

Merge note: regenerates `pkg/rt/core_compiled.lgb` and `pkg/rt/generated.sums`, as does the fn `pre/postconditions` branch. Either merge order works; re-run the generator on whichever branch rebases second, since the binary bundle will not merge.

Resolve: #452


nooga (approved):
thanks @abogoyavlensky , LGTM

PR #454, opened 2026-07-12
test(vm): stabilize BenchmarkVectorAccess — sink result + hoist key boxing

nnunley:
`BenchmarkVectorAccess` measured a fiction, in both directions — which is why the "Are we fast yet?" page shows an enormous ArrayVector "regression" that isn't real.

Two defects in the old loop:

1. **Results discarded.** `arrayVec.ValueAt(...)` return values were dropped, so once the compiler devirtualizes + inlines the `Lookup` call it can dead-code-eliminate the whole loop → a fictional sub-nanosecond time. When an unrelated change perturbs devirtualization, the loop starts running for real and the number leaps ~50×, reading as a giant regression.

2. **Per-iteration key boxing.** It passed `Int(size/2)` / `Int(size-1)` at the call site. Boxing an `Int > 255` into the `Value` interface argument escapes and allocates, so `/1000` and `/10000` measured **2 allocs/op of Int boxing**, not vector access.

Fix: sink each result to a package-level var (an observable store the compiler must keep), and hoist the key boxing out of the timed loop. Access is now a stable **~9ns/iter (3 accesses), 0 allocs, size-independent** — which is what an O(1) array index should be.

Confirmed no real runtime regression underneath: the honest number is flat across `/10`…`/10000`. The perf page's ArrayVector "+several-thousand-percent" was entirely this benchmark artifact.

Pure test change; no runtime code touched.

### DIFF
--- a/pkg/vm/vector_benchmark_test.go
+++ b/pkg/vm/vector_benchmark_test.go
@@ -30,6 +30,13 @@ func BenchmarkVectorCreation(b *testing.B) {
 	}
 }
 
+// benchVecSink prevents the compiler from dead-code-eliminating the
+// benchmarked ValueAt calls: storing the result to a package-level var is an
+// observable side effect it must keep. Without a sink the whole access loop
+// can be optimized away (giving a fictional sub-nanosecond time that then
+// reads as a huge "regression" the moment devirtualization stops firing).
+var benchVecSink Value
+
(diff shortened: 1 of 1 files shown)


nooga (approved):

nooga:
no notes, LGTM

PR #456, opened 2026-07-12
fix(rt): return Clojure-compatible regex match results

### LINKED ISSUE #455: Regex results diverge from Clojure: truthy no-match re-seq, unmatched captures as empty strings
Found while running [`clojure.tools.cli` v1.4.256](https://github.com/clojure/tools.cli/tree/v1.4.256).

## Repro

```clojure
(re-seq #"(x)" "abc")     ;; let-go: ()        Clojure: nil
(re-find #"(a)?b" "b")    ;; let-go: ["b" ""]  Clojure: ["b" nil]
(re-matches #"(a)?b" "b") ;; let-go: ["b" ""]  Clojure: ["b" nil]
```

The empty list is truthy, so `(condp re-seq token ...)` picks a wrong branch. Returning `""` for an absent optional group hides it; tools.cli's long-option pattern needs that to tell `--verbose` from `--port=`. The cause is Go's `FindStringSubmatch` API, which flattens absent and empty groups both to `""`.

## Fix

Read matches through Go's submatch index API: an absent group becomes `nil`, a participating empty group stays `""`, and `re-seq` returns `nil` on no match. `re-find`, `re-seq`, `re-matches`, and `re-groups` share one conversion helper; result shapes are unchanged.

abogoyavlensky:
Three result conversions differed from Clojure and broke predicate-driven code such as `clojure.tools.cli`'s option parsing:

- `re-seq` returned an empty list when nothing matched. The empty list is truthy, so (condp `re-seq` ...) picked a wrong branch. It now returns nil, like Clojure.
- `re-find` and `re-seq` returned `""` both for an optional group that did not participate and for a group that matched zero characters. They now read Go's submatch indices and return nil for an absent group and `""` for a participating empty one.
- `re-matches` and re-groups had the same absent-vs-empty gap. They now share the index-based conversion. `re-groups` keeps its existing shape: always a vector per match, nil no-match, and its (`re-groups` regex str) all-matches signature.

Result shapes are otherwise unchanged: a groupless match is still a string and a grouped match is still `[full g1 g2 ...]`.

Merge note: independent of the other tools.cli compat branches; touches `pkg/rt/lang.go` and `pkg/vm/regex.go`, no generated artifacts.

Resolves: #455

### DIFF
--- a/pkg/rt/lang.go
+++ b/pkg/rt/lang.go
@@ -5381,6 +5381,26 @@ func installLangNS() {
 		return vm.String(string(s.(vm.String)) + "\n"), nil
 	})
 
+	regexSubmatchVector := func(s string, indices []int) vm.ArrayVector {
+		result := make(vm.ArrayVector, len(indices)/2)
+		for i := range result {
+			start, end := indices[2*i], indices[2*i+1]
+			if start < 0 {
+				result[i] = vm.NIL
+				continue
+			}
+			result[i] = vm.String(s[start:end])
+		}
+		return result
+	}
+
+	regexSubmatchValue := func(s string, indices []int) vm.Value {
+		if len(indices) == 2 {
+			return vm.String(s[indices[0]:indices[1]])
+		}
+		return regexSubmatchVector(s, indices)
+	}
+
 	// re-find: find first match of regex in string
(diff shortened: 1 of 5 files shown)


nooga:
thanks @abogoyavlensky, LGTM

nooga (approved):

PR #458, opened 2026-07-12
fix(core): support fn pre and post conditions

### LINKED ISSUE #457: fn/defn ignore :pre, and :post with % fails to compile
Found while loading [`clojure.tools.cli` v1.4.256](https://github.com/clojure/tools.cli/tree/v1.4.256), which declares `defn` postconditions with `%`.

## Repro

A postcondition fails to compile, because the reader rewrites every `%` token to `%1`, even outside `#(...)`:

```clojure
(defn g [x] {:post [(map? %)]} {:x x})
```

```
error: Can't resolve %1 in this context
```

A precondition compiles but never runs; the condition map parses as an ordinary first body form and its value is discarded:

```clojure
(defn h [x] {:pre [(pos? x)]} x)
(h -1) ;; let-go: -1   Clojure: AssertionError
```

## Fix

Two parts:

- `fn-expand` in `pkg/rt/core/core.lg` strips the condition map when more body forms follow, asserts each `:pre` condition before the body, evaluates the body once, and binds its result so `:post` conditions can read `%`. Works for single- and multi-arity functions.
- The reader rewrites `%` arguments only inside `#(...)`, and nested short-fn reads restore the enclosing state.

Regression tests live in `test/fn_prepost_test.lg` and `pkg/compiler/reader_test.go`.

Relation to the other tools.cli issues: independent, mergeable in any order. Both this branch and the partition padding branch regenerate `pkg/rt/core_compiled.lgb` and `pkg/rt/generated.sums`; whichever lands second needs one rerun of the generator, since the binary bundle will not merge.

abogoyavlensky:
fn and defn treated a Clojure `:pre`/`:post` condition map as an ordinary first body form, so conditions never ran. A postcondition could not even compile: the reader rewrote every `%` token to `%1`, including outside `#(...)`, so `{:post [(map? %)]}` failed to resolve at namespace load. `clojure.tools.cli` uses exactly that form.

`fn-expand` now strips the condition map when more body forms follow, asserts each `:pre` condition before the body, evaluates the body once, and binds its result so `:post` conditions can read `%`. The expansion works for single- and multi-arity functions. The reader now rewrites `%` arguments only inside `#(...)`, and nested short-fn reads restore the enclosing state.

Merge note: regenerates `pkg/rt/core_compiled.lgb` and `pkg/rt/generated.sums`, as does the partition padding branch. Either merge order works; re-run the generator on whichever branch rebases second, since the binary bundle will not merge.

Resolves: #457


nnunley (approved):
Verified locally on the branch:

- `:post` with `%` → works (`(defn g [x] {:post [(map? %)]} {:x x})` ⇒ `{:x 1}`), was `Can't resolve %1`.
- `:pre` now fires (`(h -1)` ⇒ assert), was silently returning `-1`.
- `#(...)` short-fns untouched — `#(* % 2)`, `#(+ %1 %2)` fine. Gating the `%`→`%1` rewrite behind `inShortFn` with defer-restore is the right shape, and the tests cover the tricky bits (hygiene vs. locals named `let`/`assert`, explicit `%1` staying a user local, multi-arity).

LGTM on substance. One non-blocking note: the branch now conflicts on the regenerated bundle (`core_compiled.lgb`/`generated.sums`) after the recent parity merges — it needs a `main` merge + `make generate` before it can land. Happy to push that via maintainer-edit if useful.

PR #459, opened 2026-07-12
perf(vm): per-var root-binding gate for lock-free var deref (Phase 1)

nnunley:
**Phase 1** of recovering the `Var.Deref` fast path after the #203/#210 concurrency rework, which routed every `^:dynamic` var deref through a walk of the single shared `globalBindingStack` — regressing deref ~8× on the perf timeline (worst on the `*Parallel` variants, where all threads read the one shared head).

## Change

Add a per-var `rootBindDepth atomic.Int32` counting the var's live **root-context** bindings. Deref's root path reads only that per-var field (its own cache line, no shared-structure access) and walks the shared stack **only** when the counter is non-zero:

```go
func (v *Var) derefRoot() Value {
    if v.rootBindDepth.Load() == 0 { return v.Root() } // non-dynamic OR declared-but-unbound → no shared read
    if val, ok := globalBindingStack.current(v); ok { return val }
    return v.Root()
}
```

- `Var.Deref()` and `ExecContext.deref()`'s root branch share this one helper (can't diverge). The **child** branch is unchanged — isolated per-goroutine stacks, `isDynamic`-gated (#210 isolation preserved).
- The counter is maintained by every root-stack mutation with race-safe ordering: `pushBinding` increments **before** the push, `popBinding` pops **before** the decrement, and `RunWithBindings` bumps gaining vars before the snapshot swap and drops losing vars after — so a reader never sees `depth==0` while a settled binding is still walkable. `set!`/`setCurrent` doesn't change bound-status and correctly leaves the counter alone.

## What this fixes / doesn't

| Case | This PR |
|---|---|
| non-dynamic (fns, defs) | ✅ single per-var atomic load, contention-free |
| `^:dynamic` declared-but-unbound (`*warn-on-reflection*`, `*print-length*`, …) | ✅ `rootBindDepth==0` → skip the shared walk |
| `^:dynamic` currently root-**bound** (`Bound`/`BoundParallel`) | ⛔ still walks the shared stack — **deferred to Phase 2**, gated on measuring a real bound-parallel workload (e.g. malli) |

Bench (M3): `VarDerefPreviouslyBound` — the common declared-but-unbound read — drops from ~2.5ns (walking the shared stack) to **1.6ns serial / 0.31ns parallel**, and every `*Parallel` variant is now *faster* than its serial counterpart (contention gone). Success criterion is contention-flat + near the atomic-load floor, not restoring the pre-atomics 0.25ns.

## Tests

`-race` gate (the acceptance bar): child-context isolation, 16-goroutine concurrent bind/deref on the same var, and 16 distinct vars concurrent — all pass with zero data races. Full non-race suite green (18 packages).

**Note (not this PR):** `go test ./pkg/vm/ -race` also surfaces a **pre-existing** data race in `TestArrayVectorSeqChunkedFirstRace` (`vector.go:268`) — confirmed identical on `main`; this branch doesn't touch `vector.go`. Worth its own fix.

Spec/plan: `docs/superpowers/specs/2026-07-12-var-deref-fast-path-design.md`.

---
**Part of #464** (Epic: Runtime performance & concurrency) — var-deref regression recovery, Phase 1.

### DIFF
--- a/pkg/vm/exec_context.go
+++ b/pkg/vm/exec_context.go
@@ -126,10 +126,14 @@ func (ec *ExecContext) orRoot() *ExecContext {
 
 // --- dynamic-var resolution (ec owns the binding state) ---------------------
 
-// deref returns v's current value in this context: the top dynamic binding if
-// any, else v's root.
 func (ec *ExecContext) deref(v *Var) Value {
 	ec = ec.orRoot()
+	if ec == RootExecContext {
+		return v.derefRoot()
+	}
+	// Child context: an isolated, per-goroutine stack — uncontended, so walk it
+	// directly. rootBindDepth tracks only ROOT bindings, so child bindings are
+	// gated by isDynamic (set at declaration or on any bind) as before.
 	if v.isDynamic.Load() {
 		if val, ok := ec.bindings.current(v); ok {
 			return val
@@ -140,11 +144,25 @@ func (ec *ExecContext) deref(v *Var) Value {
 
 func (ec *ExecContext) pushBinding(v *Var, val Value) {
 	v.isDynamic.Store(true)
-	ec.orRoot().bindings.push(v, val)
+	root := ec.orRoot()
+	if root == RootExecContext {
+		// Increment BEFORE the push: a reader that sees depth>0 then walks will
+		// either find the binding or (racing this in-progress push) fall through
+		// to Root(). It never sees depth==0 while a settled root binding exists.
+		v.rootBindDepth.Add(1)
+	}
+	root.bindings.push(v, val)
 }
 
 func (ec *ExecContext) popBinding(v *Var) {
-	ec.orRoot().bindings.pop(v)
+	root := ec.orRoot()
+	root.bindings.pop(v)
+	if root == RootExecContext {
+		// Decrement AFTER the pop: depth stays >0 while the binding is still
+		// walkable, so a concurrent reader never sees depth==0 with the binding
+		// still present.
+		v.rootBindDepth.Add(-1)
+	}
 }
 
 func (ec *ExecContext) hasBinding(v *Var) bool {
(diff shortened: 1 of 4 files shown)


nooga (commented):
Nice piece of work — the increment-before-push / decrement-after-pop ordering (and the mirrored bump-before-swap / drop-after-swap in `rootInstallSnapshot`) is exactly the right shape to guarantee `rootBindDepth` can only be transiently over-counted, never under-counted while a binding is live, which is what makes the `== 0` fast path safe. The new `var_binding_race_test.go` coverage (unbound/bound/nested, child isolation, concurrent same-var and distinct-var bind/deref, `RunWithBindings` consistency) is thorough and matches the repo's Go-unit-test convention for VM-internal perf paths. Phase 1 also looks safe to land on its own: the still-root-bound path is untouched, so this is a pure additive fast path with no behavior change for the case Phase 2 will optimize.

Two non-blocking things:

1. This repo's CI doesn't run `go test -race` anywhere (`.github/workflows/*.yml` has no `-race` flag), so the acceptance bar this PR cites ("`-race` gate ... zero data races") is only ever checked locally by whoever runs it, not automatically. Given this PR is explicitly about lock-free concurrent access, it might be worth adding a `-race` CI job (could be its own follow-up PR/issue rather than blocking this one).
2. The diff removes `ExecContext.deref`'s summary doc comment (`// deref returns v's current value in this context...`) and doesn't replace it with an equivalent — the explanation is now split between the inline child-branch comment and `derefRoot`'s doc on `var.go`. Worth restoring a one-line summary on `deref` itself for anyone who lands there without already knowing the split.

Also flagging (not a bug, just worth a comment somewhere): `rootInstallSnapshot`'s counter bookkeeping assumes no two `RunWithBindings` calls race the same underlying stack — true today, but if `RunWithBindings` is ever resurrected for concurrent spawn sites, a bad interleaving could leave `rootBindDepth` permanently wrong. Worth a one-line warning in the `RunWithBindings` doc comment for future readers.

nnunley:
Superseded: #462 (the stack tip) carried these four commits and has merged — the Phase-1 rootBindDepth work landed with it in cfd2025's successor main. Closing per the stack-tip merge convention.

PR #460, opened 2026-07-12
fix(rt): expose key-pending? on plan9

mparrett:
## What changed

- define `term/key-pending?` in the Plan 9 terminal namespace
- return `false`, consistent with Plan 9's existing stubbed `term/read-key`
- include `key-pending?` in the terminal namespace surface test

## Why

Plan 9 builds create the `term` namespace but did not register `key-pending?`. Code using `(term/key-pending?)` therefore failed during compilation with `Can't resolve term/key-pending? in this context`, even though the operation is present on native and WASM targets.

This restores a consistent namespace surface without claiming unsupported Plan 9 input polling behavior.

## Validation

- `go test -count=1 -run 'TestRunner/term_test.lg' ./test`
- `GOOS=plan9 GOARCH=amd64 go test -c -o /tmp/let-go-plan9-term.test ./test`
- `GOOS=js GOARCH=wasm go build ./...`
- `GOOS=linux GOARCH=amd64 go build ./...`
- `gofmt -d pkg/rt/term_plan9.go`
- `git diff --check`

### DIFF
--- a/pkg/rt/term_plan9.go
+++ b/pkg/rt/term_plan9.go
@@ -35,6 +35,9 @@ func installTermNS() {
 	stubTrue, _ := vm.NativeFnType.Wrap(func(vs []vm.Value) (vm.Value, error) {
 		return vm.TRUE, nil
 	})
+	stubFalse, _ := vm.NativeFnType.Wrap(func(vs []vm.Value) (vm.Value, error) {
+		return vm.FALSE, nil
+	})
 	stubNil, _ := vm.NativeFnType.Wrap(func(vs []vm.Value) (vm.Value, error) {
 		return vm.NIL, nil
 	})
@@ -42,6 +45,7 @@ func installTermNS() {
 	ns.Def("raw-mode!", stubTrue)
 	ns.Def("restore-mode!", stubTrue)
 	ns.Def("read-key", stubNil)
+	ns.Def("key-pending?", stubFalse)
 
 	sizeFn, _ := vm.NativeFnType.Wrap(func(vs []vm.Value) (vm.Value, error) {
 		return vm.NewPersistentVector([]vm.Value{vm.MakeInt(80), vm.MakeInt(24)}), nil
--- a/test/term_test.lg
+++ b/test/term_test.lg
@@ -10,6 +10,7 @@
     (is (not (nil? term/write)))
     (is (not (nil? term/write-at)))
     (is (not (nil? term/read-key)))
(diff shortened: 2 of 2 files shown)


mparrett:
PR #460 is correct for its narrow goal. No code issues. Registering an always-false `key-pending?` is consistent with Plan 9’s existing nil `read-key` stub, and #461 replaces both with real behavior.

### Testing gap

**[P3] The added test does not actually exercise the Plan 9 namespace.**

The assertion in [[test/term_test.lg](https://github.com/nooga/let-go/blob/73a093ee2f59442c905271a8575205d319082714/test/term_test.lg#L12-L14)](https://github.com/nooga/let-go/blob/73a093ee2f59442c905271a8575205d319082714/test/term_test.lg#L12-L14) runs against the host implementation, where `key-pending?` already existed. Cross-compiling the Plan 9 test binary confirms that the Go code compiles, but it does not execute this `.lg` test or resolve the Plan 9 namespace.

Consequently, removing the Plan 9 registration again would not fail the normal host test. Recommend running the compiled test under Plan 9 via qemu, or adding a platform-surface test that directly checks the selected namespace implementation.

**Conclusion:**
- Safe to merge #460.
- Do not release or leave main at #460 alone for long: xsofy will compile but its poll loop will never observe input because [`key-pending?` always returns false](https://github.com/nooga/let-go/blob/73a093ee2f59442c905271a8575205d319082714/pkg/rt/term_plan9.go#L38-L48).
- Merge #460 first, then retarget/rebase #461 onto `main` and verify #461’s diff no longer includes #460’s commit.

The targeted host test passes, and the Plan 9/amd64 test binary cross-compiles successfully.

nnunley (approved):

nooga (approved):
LGTM. key-pending? was defined on native (term.go) and wasm (term_wasm.go) but missing from plan9 — this closes that gap with a stubFalse fn matching the existing stubTrue/stubNil pattern in the same file. Consistent with the already-stubbed read-key (returns nil) — an always-false key-pending? is the honest counterpart.

One note (already surfaced in the PR thread): the added assertion in test/term_test.lg runs under the host build, so it verifies the symbol resolves but doesn't specifically exercise the plan9 build-tagged file — no CI job builds/tests with GOOS=plan9. That's a pre-existing gap for this whole file rather than something introduced here, so it shouldn't block this PR, but worth a follow-up (e.g. a plan9 cross-compile+run-under-qemu job) if this namespace keeps drifting.

PR #461, opened 2026-07-12
feat(rt): real Plan 9 terminal input (raw mode, read-key, key-pending?)

mparrett:
Stacked on #460. That PR made the Plan 9 build compile again by registering `term/key-pending?`, but as a `false` stub. Because xsofy's input loop only calls `read-key` once `key-pending?` reports a key is ready, an always-`false` `key-pending?` means the game animates but never reads input. This PR replaces the input stubs with a working path, so xsofy takes keystrokes and is playable on ANSI-capable Plan 9 terminals.

## The problem specific to Plan 9

The native and WASM key paths both rely on a non-blocking "is a byte ready?" check — `FIONREAD` on native, the SharedArrayBuffer fill level in the browser. Plan 9 has no `poll(2)`/`FIONREAD` equivalent for the console, so there is no way to answer `key-pending?` by peeking the OS directly.

The fix is to keep the bytes on our side of that boundary. A background goroutine does the blocking read and appends to a buffer; `key-pending?` then answers from the buffer (non-blocking), and `read-key` tokenizes off it.

## What's here

A new `KeySource` implementation, `NewQueuedKeySource(io.Reader)` (`pkg/rt/keysource_queued.go`):

- a reader goroutine blocks on the `io.Reader` and appends chunks to a shared queue under a mutex; it never holds the lock across the blocking read
- `ReadKey` blocks until a whole key token is available, tokenizing one key per call off the queue via the existing `scanKey`
- `KeyPending` returns whether bytes are buffered — non-blocking, and false at end-of-input, matching native's `FIONREAD`-based behavior so a per-tick poll loop doesn't busy-spin once input is exhausted

It's platform-neutral and exported: Plan 9 needs it, but any embedder whose input is a plain blocking reader (a pipe, a socket) can bind it via `api.WithKeySource`. State lives on the instance rather than a package global, so the reader goroutine starts only when this source is consulted — an `api.WithKeySource` override reads nothing.

The Plan 9 wiring (`pkg/rt/term_plan9.go`):

- `read-key`/`key-pending?` route through the bound `*keys*` source, so `api.WithKeySource` and `(binding [*keys* …])` still override them
- `raw-mode!`/`restore-mode!` open `/dev/consctl` and write `rawon`/`rawoff`. Plan 9 reverts to cooked mode when the consctl fd closes, so the fd is held open until `restore-mode!`. Raw mode is gated on `fd 0` being the console (`Fd2path`), so a piped stdin returns nil rather than raw-moding a console it isn't reading — mirroring native's not-a-TTY path
- `size` reads `/env/COLS` and `/env/LINES` as live files (80x24 fallback), so it tracks alacritty9/drawterm resizes rather than a value cached at startup
- `flush` is a no-op on the console/pipe: Plan 9 has no `fsync` there (sync returns "bad arg in system call"), so a naive sync crashed on exit — the Plan 9 twin of #308

## Multi-byte keys can split across reads

`read-key` hands out one key per call, and a single key can be multiple bytes (an arrow is `ESC [ A`). With a chunked background reader, a multi-byte sequence can straddle two reads — a short read mid-sequence is normal when the console is a network transport (drawterm/9P). If `read-key` emitted the partial bytes immediately, it would hand out a broken `ESC [` and desync every following key.

So on an incomplete front token, `read-key` waits up to a short inter-byte grace — reset by each byte that arrives — and re-scans, stitching a sequence that dribbles in across reads. A genuinely bare `ESC` (nothing more within one grace window) resolves to the Escape key; a complete key pays no latency. This is the classic terminal ESC delay: with no `FIONREAD` peek, a bare `ESC` and the head of a split arrow are indistinguishable synchronously — emit eagerly and arrows break, wait unconditionally and `ESC` hangs.

## Validation

- `go test -race` over the queued source: held-key bursts, single-chunk arrows, an SS3 key, an `ESC`/`[`/`A` arrow and a UTF-8 rune each split byte-by-byte across reads, a bare `ESC` resolving after the grace, a truncated sequence, a non-EOF reader error surfacing after its bytes, and the `key-pending?` transitions (false → true → false-at-EOF).
- Cross-compiles: `GOOS=plan9`, `GOOS=js/wasm`, `GOOS=linux` build; `go vet` clean on native and plan9; the Plan 9 test binary compiles.

**Tested on real Plan 9 (9front amd64 under QEMU, including agent9).** The bare-console smoke accepts keystrokes, enters and exits raw mode, resolves a bare Escape through the grace window, and restores the terminal on quit. The full xsofy game then runs under **alacritty9** — title screen, movement, inventory, combat, potions, and a clean quit, which exercises the `flush` shutdown path (that clean exit is what the flush fix buys — before it, quitting crashed on `sync /dev/stdout`). `term/size` tracks live window resizes there (measured 84×27 → 98×28 on a manual resize). A second pass on real hardware is still welcome — especially the ANSI arrow path and rendering throughput under alacritty9/drawterm. Thanks @dharmatech / @Alino (nooga/xsofy#141).

**Smoke test (qemu):**
image

## Out of scope

- **Output throughput.** The slowness reported in nooga/xsofy#141 is most likely the renderer's per-cell write amplification, not input; batching terminal output is a separate change.
- Mouse reports and per-handle raw mode stay native-only.
- Real per-window pixel geometry via `/dev/wctl` is a follow-up; the `/env`-based `size` above covers terminals that publish `$COLS`/`$LINES` (alacritty9, drawterm), with the 80x24 fallback where they're unset.

Refs nooga/xsofy#141.

### DIFF
--- a/pkg/rt/keysource_queued.go
+++ b/pkg/rt/keysource_queued.go
@@ -1,0 +1,132 @@
+/*
+ * Copyright (c) 2026 Matt Parrett
+ * SPDX-License-Identifier: MIT
+ */
+
+package rt
+
+import (
+	"io"
+	"sync"
+)
+
+// queuedKeySource is a KeySource for hosts without a non-blocking input peek: a
+// background goroutine blocks on an io.Reader and appends to a shared queue,
+// ReadKey tokenizes one key per call off it (via scanKey), and KeyPending peeks
+// it. This is the shape Plan 9 needs — it has no poll(2)/FIONREAD, so a
+// non-blocking key-pending? can only be answered from a buffer a concurrent
+// reader fills, where native (term.go) reads stdin synchronously inside ReadKey.
+// It is platform-neutral, so it is also available to embedders whose input is a
+// plain blocking reader (a pipe, a socket) — see NewQueuedKeySource.
+//
+// State lives on the instance, not a package global like native's keyBuf: the
+// goroutine consumes the reader, so it must start only when THIS source is the
+// one being read. If an embedder rebinds *keys* via api.WithKeySource, this
+// source's ReadKey/KeyPending are never called and the reader never starts — no
+// bytes stolen from the source that replaced it.
+type queuedKeySource struct {
+	mu      sync.Mutex
+	cond    *sync.Cond
+	r       io.Reader
+	buf     []byte // read but not yet tokenized out
+	eof     bool
+	started bool
+}
+
+// NewQueuedKeySource returns a KeySource that reads keys from r via a background
+// goroutine and a buffered queue, so KeyPending is non-blocking on platforms
+// that lack a poll/FIONREAD peek. The reader goroutine starts lazily on the
+// first ReadKey/KeyPending, so binding this at *keys* without ever consulting it
+// (e.g. when api.WithKeySource overrides it) reads nothing.
+func NewQueuedKeySource(r io.Reader) KeySource {
+	s := &queuedKeySource{r: r}
+	s.cond = sync.NewCond(&s.mu)
+	return s
+}
+
+// queuedReadChunkSize matches native's readChunkSize: a blocking Read returns as
+// soon as any data is available, so a larger buffer adds no latency to a single
+// keystroke — it just collects held-key bursts in fewer reads.
+const queuedReadChunkSize = 256
+
+// start launches the background reader once. Called with s.mu held.
+func (s *queuedKeySource) start() {
+	if s.started {
+		return
+	}
+	s.started = true
+	go s.readLoop()
+}
+
+// readLoop blocks on r.Read WITHOUT holding the lock, then locks only to append
+// the chunk and wake a waiter. Runs until EOF or a read error, which it records
+// as eof so a parked ReadKey unblocks with the nil contract.
+func (s *queuedKeySource) readLoop() {
+	chunk := make([]byte, queuedReadChunkSize)
+	for {
+		n, err := s.r.Read(chunk)
+		s.mu.Lock()
+		if n > 0 {
+			s.buf = append(s.buf, chunk[:n]...)
+		}
+		if err != nil || n == 0 {
+			s.eof = true
+			s.cond.Broadcast()
+			s.mu.Unlock()
+			return
+		}
+		s.cond.Broadcast()
+		s.mu.Unlock()
+	}
+}
+
+// ReadKey blocks until a whole key token is available (or EOF), tokenizing one
+// key per call off the shared buffer via scanKey. "" signals end-of-input (the
+// read-key nil contract). Blocking here matches native (which blocks in poll);
+// callers stay responsive by gating on KeyPending first.
+func (s *queuedKeySource) ReadKey() (string, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.start()
+	for len(s.buf) == 0 && !s.eof {
+		s.cond.Wait()
+	}
+	if len(s.buf) == 0 {
+		return "", nil // EOF
+	}
+	for {
+		status, n := scanKey(s.buf)
+		if status == keyNeedMore && n > 1 && !s.eof {
+			// A multi-byte token (CSI/SS3/UTF-8) split across reader chunks —
+			// short reads are normal over a network transport like 9P/drawterm.
+			// Wait for one more append, then re-scan to stitch it, rather than
+			// emitting a broken partial that would desync every following key.
+			// Bounded: each wait adds bytes (progress) or hits eof (then emit
+			// best-effort below). A lone ESC (scanKey → keyNeedMore, 1) is
+			// deliberately NOT waited on — it is the Escape key, emitted now
+			// rather than parked until the next keypress.
+			before := len(s.buf)
+			for len(s.buf) == before && !s.eof {
+				s.cond.Wait()
+			}
+			continue
+		}
+		tok := string(s.buf[:n])
+		s.buf = s.buf[n:]
+		if len(s.buf) == 0 {
+			s.buf = nil // release the backing array once drained
+		}
+		return tok, nil
+	}
+}
+
+// KeyPending reports whether a key is buffered and ready, without consuming it.
+// Non-blocking and eof-blind (false once the queue drains at end-of-input),
+// mirroring native's FIONREAD-based rawPending so a per-tick input poll doesn't
+// busy-drain end-of-input forever.
+func (s *queuedKeySource) KeyPending() bool {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.start()
+	return len(s.buf) > 0
+}
--- a/pkg/rt/keysource_queued.go
+++ b/pkg/rt/keysource_queued.go
@@ -8,6 +8,7 @@ package rt
 import (
 	"io"
 	"sync"
+	"time"
 )
 
 // queuedKeySource is a KeySource for hosts without a non-blocking input peek: a
@@ -26,12 +27,24 @@ import (
 // bytes stolen from the source that replaced it.
 type queuedKeySource struct {
 	mu      sync.Mutex
-	cond    *sync.Cond
 	r       io.Reader
-	buf     []byte // read but not yet tokenized out
-	eof     bool
-	started bool
+	buf     []byte        // read but not yet tokenized out
+	err     error         // sticky reader error; nil for a clean EOF
+	done    bool          // reader finished (EOF or error) — no more bytes coming
+	started bool          // reader goroutine launched
+	notify  chan struct{} // buffered(1) wakeup; reader signals on append/done
+	grace   time.Duration // inter-byte wait for stitching a split escape sequence
 }
+
+// escGrace bounds how long ReadKey waits for the rest of an incomplete escape
+// (or UTF-8) sequence before emitting what it has. It's the classic terminal
+// "ESC delay": long enough to stitch a sequence split across reads on a slow
+// transport (drawterm/9P), short enough that a bare Escape key still registers
+// promptly. Only an incomplete front token pays it — a complete key returns with
+// zero added latency. Without it, distinguishing a bare ESC from the start of a
+// split arrow is impossible on a host with no FIONREAD peek (a lone ESC would
+// either be emitted too eagerly, breaking split arrows, or waited on forever).
+const escGrace = 40 * time.Millisecond
 
 // NewQueuedKeySource returns a KeySource that reads keys from r via a background
 // goroutine and a buffered queue, so KeyPending is non-blocking on platforms
@@ -39,9 +52,7 @@ type queuedKeySource struct {
 // first ReadKey/KeyPending, so binding this at *keys* without ever consulting it
 // (e.g. when api.WithKeySource overrides it) reads nothing.
 func NewQueuedKeySource(r io.Reader) KeySource {
-	s := &queuedKeySource{r: r}
-	s.cond = sync.NewCond(&s.mu)
-	return s
+	return &queuedKeySource{r: r, notify: make(chan struct{}, 1), grace: escGrace}
 }
 
 // queuedReadChunkSize matches native's readChunkSize: a blocking Read returns as
@@ -58,9 +69,21 @@ func (s *queuedKeySource) start() {
 	go s.readLoop()
 }
 
+// signal wakes a waiting ReadKey without blocking; the buffered channel
+// coalesces bursts, and ReadKey re-checks the buffer under the lock after each
+// wake, so a dropped (already-pending) signal is never a missed update.
+func (s *queuedKeySource) signal() {
+	select {
+	case s.notify <- struct{}{}:
+	default:
+	}
+}
+
 // readLoop blocks on r.Read WITHOUT holding the lock, then locks only to append
-// the chunk and wake a waiter. Runs until EOF or a read error, which it records
-// as eof so a parked ReadKey unblocks with the nil contract.
+// the chunk and wake a waiter. On a read error it retains the error (draining
+// any bytes that came with it) and stops; io.EOF is a clean end, recorded as a
+// nil error so ReadKey returns the "" contract. A (0, nil) read is not EOF (the
+// io.Reader contract) — it loops and reads again.
 func (s *queuedKeySource) readLoop() {
 	chunk := make([]byte, queuedReadChunkSize)
 	for {
@@ -69,55 +92,95 @@ func (s *queuedKeySource) readLoop() {
 		if n > 0 {
 			s.buf = append(s.buf, chunk[:n]...)
 		}
-		if err != nil || n == 0 {
-			s.eof = true
-			s.cond.Broadcast()
+		if err != nil {
+			if err != io.EOF {
+				s.err = err
+			}
+			s.done = true
 			s.mu.Unlock()
+			s.signal()
 			return
 		}
-		s.cond.Broadcast()
 		s.mu.Unlock()
+		if n > 0 {
+			s.signal()
+		}
 	}
 }
 
-// ReadKey blocks until a whole key token is available (or EOF), tokenizing one
-// key per call off the shared buffer via scanKey. "" signals end-of-input (the
-// read-key nil contract). Blocking here matches native (which blocks in poll);
-// callers stay responsive by gating on KeyPending first.
+// ReadKey blocks until a whole key token is available (or the reader ends),
+// tokenizing one key per call off the shared buffer via scanKey. It returns ""
+// with a nil error at a clean EOF (the read-key nil contract), or "" with the
+// retained error if the reader failed. Blocking here matches native (which
+// blocks in poll); callers stay responsive by gating on KeyPending first.
 func (s *queuedKeySource) ReadKey() (string, error) {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 	s.start()
-	for len(s.buf) == 0 && !s.eof {
-		s.cond.Wait()
+
+	for len(s.buf) == 0 && !s.done {
+		s.waitLocked(0) // no deadline: block until data or reader end
 	}
 	if len(s.buf) == 0 {
-		return "", nil // EOF
+		return "", s.err // clean EOF (err nil) or a real reader failure
 	}
+
+	var deadline time.Time // set on the first incomplete front token
 	for {
 		status, n := scanKey(s.buf)
-		if status == keyNeedMore && n > 1 && !s.eof {
-			// A multi-byte token (CSI/SS3/UTF-8) split across reader chunks —
-			// short reads are normal over a network transport like 9P/drawterm.
-			// Wait for one more append, then re-scan to stitch it, rather than
-			// emitting a broken partial that would desync every following key.
-			// Bounded: each wait adds bytes (progress) or hits eof (then emit
-			// best-effort below). A lone ESC (scanKey → keyNeedMore, 1) is
-			// deliberately NOT waited on — it is the Escape key, emitted now
-			// rather than parked until the next keypress.
-			before := len(s.buf)
-			for len(s.buf) == before && !s.eof {
-				s.cond.Wait()
-			}
-			continue
+		if status == keyReady || s.done {
+			// Complete token, or the reader ended so no completion is coming —
+			// emit best-effort (a lone ESC at EOF is the Escape key; a truncated
+			// sequence surfaces rather than hangs).
+			return s.take(n), nil
 		}
-		tok := string(s.buf[:n])
-		s.buf = s.buf[n:]
-		if len(s.buf) == 0 {
-			s.buf = nil // release the backing array once drained
+		// keyNeedMore and the reader is still live: a front token split across
+		// reads. Wait for more bytes up to the grace period, then re-scan; if the
+		// grace expires with no completion, emit what we have. This is what makes
+		// a bare ESC vs. a split arrow decidable without a FIONREAD peek. The
+		// grace is inter-byte — each byte that arrives resets it — so a sequence
+		// dribbling in over a slow transport keeps stitching, while a truly bare
+		// ESC (nothing more within one grace window) emits as the Escape key.
+		if deadline.IsZero() {
+			deadline = time.Now().Add(s.grace)
 		}
-		return tok, nil
+		remaining := time.Until(deadline)
+		if remaining <= 0 {
+			return s.take(n), nil
+		}
+		before := len(s.buf)
+		s.waitLocked(remaining)
+		if len(s.buf) > before {
+			deadline = time.Time{} // new bytes — restart the inter-byte window
+		}
 	}
+}
+
+// take slices the front n bytes off the buffer as a token, releasing the backing
+// array once drained. Called with s.mu held.
+func (s *queuedKeySource) take(n int) string {
+	tok := string(s.buf[:n])
+	s.buf = s.buf[n:]
+	if len(s.buf) == 0 {
+		s.buf = nil
+	}
+	return tok
+}
+
+// waitLocked releases the lock, waits for a reader wake (or the timeout, if
+// nonzero), and re-acquires it — the condition-variable wait, done over a
+// channel so it can be time-bounded. Called with s.mu held; returns holding it.
+func (s *queuedKeySource) waitLocked(timeout time.Duration) {
+	s.mu.Unlock()
+	if timeout > 0 {
+		select {
+		case <-s.notify:
+		case <-time.After(timeout):
+		}
+	} else {
+		<-s.notify
+	}
+	s.mu.Lock()
 }
 
 // KeyPending reports whether a key is buffered and ready, without consuming it.
--- a/pkg/rt/keysource_queued.go
+++ b/pkg/rt/keysource_queued.go
@@ -7,6 +7,8 @@ package rt
 
 import (
 	"io"
+	"os"
+	"strconv"
 	"sync"
 	"time"
 )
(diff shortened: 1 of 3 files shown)


mparrett:
Three issues to address before moving to review:

1. **[P1] Arrow sequences can still split after the first `ESC`.**  
   [`[ReadKey](https://github.com/nooga/let-go/blob/5c8ac0bfcc549e78a3701e2049b47c182b798f4a/pkg/rt/keysource_queued.go#L97-L113)`](https://github.com/nooga/let-go/blob/5c8ac0bfcc549e78a3701e2049b47c182b798f4a/pkg/rt/keysource_queued.go#L97-L113) waits for more data only when `n > 1`. If the reader returns `"\x1b"` followed by `"[A"`, the first byte is immediately emitted as Escape and the remainder becomes two ordinary keys. In xsofy, that Escape may cancel a menu or quit. Conversely, `ESC [` with no completing byte waits forever. Both need an inter-byte timeout/grace period; the tests currently split only after `ESC[`, missing the important boundary.

2. **[P2] Reader errors are converted to clean EOF.**  
   [`[readLoop](https://github.com/nooga/let-go/blob/5c8ac0bfcc549e78a3701e2049b47c182b798f4a/pkg/rt/keysource_queued.go#L67-L76)`](https://github.com/nooga/let-go/blob/5c8ac0bfcc549e78a3701e2049b47c182b798f4a/pkg/rt/keysource_queued.go#L67-L76) discards every non-nil error, so `ReadKey() (string, error)` can never report an underlying console/socket failure. It should retain the error, drain any returned bytes, then return that error. The same block incorrectly treats `(0, nil)` as EOF; Go’s `io.Reader` contract says that means nothing happened, not end-of-input. [Go `io.Reader` documentation](https://pkg.go.dev/io#Reader).

3. **[P2] Opening `/dev/consctl` does not detect redirected stdin.**  
   [`[raw-mode!](https://github.com/nooga/let-go/blob/5c8ac0bfcc549e78a3701e2049b47c182b798f4a/pkg/rt/term_plan9.go#L70-L90)`](https://github.com/nooga/let-go/blob/5c8ac0bfcc549e78a3701e2049b47c182b798f4a/pkg/rt/term_plan9.go#L70-L90) assumes opening `/dev/consctl` fails when stdin is piped. Normally that path remains present in the process namespace after fd 0 is redirected, so this can put the user’s window console into raw mode while the key source reads an unrelated pipe. The Plan 9 manual describes `/dev/cons` and `/dev/consctl` as separate files and confirms that holding `consctl` open controls raw mode. [Plan 9 `cons(3)`](https://9p.io/magic/man2html/3/cons).

Otherwise, the locking structure looks good: the blocking read occurs outside the mutex, buffered bytes are protected, and holding `consctl` open until restoration matches Plan 9’s documented behavior.

I ran `go test -race ./pkg/rt` successfully and confirmed the Plan 9/amd64 test binary cross-compiles. For the QEMU run, I’d specifically test an arrow delivered as three separate reads—`ESC`, `[`, `A`—plus a deliberately truncated `ESC [` sequence.

mparrett:
All three fixed in 3dc689d and 45b0e2f.

**1 — ESC split (P1).** You're right that gating the wait on `n > 1` broke the ESC boundary: a lone ESC was emitted immediately, so `ESC` then `[A` in separate reads became Escape plus two stray keys, and a truncated `ESC [` waited forever. With no FIONREAD peek there's no synchronous way to tell a bare ESC from the head of a split sequence, so `ReadKey` now uses an inter-byte grace: on an incomplete front token it waits up to a grace window — reset by each arriving byte — and re-scans, so a sequence dribbling in over a slow transport stitches, while a genuinely bare ESC resolves to Escape once the window passes. A complete key adds no latency. Tests now split `ESC`/`[`/`A` (and a UTF-8 rune) byte-by-byte across reads, plus a bare ESC resolving after the grace and a truncated `ESC [`.

**2 — reader errors (P2).** `readLoop` retains a non-EOF error, drains any bytes that arrived with it, and `ReadKey` returns the error once the buffer empties; `io.EOF` stays the clean `""` contract, and `(0, nil)` loops instead of ending, per the `io.Reader` contract. Test covers a non-EOF error surfacing after its bytes.

**3 — redirected stdin (P2).** `raw-mode!` now checks `syscall.Fd2path(0) == "/dev/cons"` before touching `/dev/consctl` and returns nil when stdin isn't the console — the plan9 analogue of native's not-a-TTY gate, so `lg <somepipe` no longer raw-modes the window. `tty?` uses the same check.

`go test -race ./pkg/rt` and the plan9/amd64 cross-compile are green. Still not run on real Plan 9 — I'm bringing up a 9front amd64 VM to exercise it, and the three-separate-reads arrow plus truncated `ESC [` you flagged are now in the unit tests, so the device run is about confirming the `/dev/consctl` handshake and console reads against the real thing.

mparrett:
## Trying this on Plan 9

Validated so far on stock 9front amd64 (QEMU): raw mode engages on the console, keys echo unbuffered one token each, a bare Escape resolves through the grace window, and quitting restores cooked mode. The paths I *couldn't* reach here and would value a second pair of eyes on: **alacritty9 / drawterm** (the ANSI-terminal path, where arrows arrive as `ESC [ A` rather than Plan 9's native key runes) and the full xsofy game.

### Build

Pure Go, so either cross-build and copy the binary over, or build on Plan 9 directly if you have Go:

```
git fetch origin agent/plan9-terminal-input && git checkout agent/plan9-terminal-input
GOOS=plan9 GOARCH=amd64 go build -o lg .    # cross-build; scp/hget lg to Plan 9
# — or, on Plan 9 with go installed —
go build -o lg .
```

### Smoke test

Save as `smoke.lg` and run `lg smoke.lg` in a cons window (a rio window or the bare text console — both own `/dev/cons` + `/dev/consctl`):

```
(term/write (str "key-pending? before input: " (term/key-pending?) "\r\n"))
(term/raw-mode!)
(term/write "raw mode on. type keys ('q' or EOF quits):\r\n")
(loop []
  (let [k (term/read-key)]
    (cond
      (nil? k)  (term/write "EOF - bye\r\n")
      (= k "q") (term/write "q - bye\r\n")
      :else     (do (term/write (str "  read-key => " (pr-str k) "\r\n"))
                    (recur)))))
(term/restore-mode!)
```

Expected: each key echoes the instant you press it (no Enter). `q` quits and the shell returns to normal line editing — if it doesn't, raw mode wasn't restored.

### Notes

- **Arrow keys differ by terminal.** On the bare rio console they come through as Plan 9's native key runes — single tokens that may render as tofu because the font lacks a glyph, which is fine. Under alacritty9 or drawterm they arrive as ANSI `ESC [ A` sequences instead. `read-key` handles both as one token per key; the ANSI path is the one I'd most like confirmed on real hardware.
- **raw-mode! needs stdin to be the console.** It gates on `fd 0` being  `/dev/cons`, so `lg <somefile` deliberately won't enter raw mode (it would otherwise raw-mode a console it isn't reading).

### Useful to report

Whether keys echo unbuffered, arrows come through as single tokens (rune *or* `ESC[` sequence depending on your terminal), a bare Escape registers, and the shell is left in cooked mode after `q`. An arrow delivered as three separate reads (`ESC`, `[`, `A`) over a slow drawterm link is the interesting edge — it should still arrive as one token.

nooga:
## Potential issues / points worth a second look

1. **`stdinIsConsole()` heuristic (`term_plan9.go`)** checks `syscall.Fd2path(0) == "/dev/cons"` exactly. On some Plan 9 variants (9front's `#c/cons`, or a console reached through a different mount) this could false-negative and silently skip raw-mode entry. Not a correctness bug given the PR's own admission it hasn't been validated on real hardware — but worth flagging in the follow-up hardware validation the author is already requesting.
2. **`escGrace = 40ms` is a fixed constant**, not configurable. For a slow/high-latency transport (drawterm over a WAN link, vs. local 9P) 40ms might be too short and occasionally split an arrow key into a stray literal `ESC` + `[A`. This is a tuning question rather than a bug — the PR frames it correctly as the classic terminal ESC-delay trade-off — but it's the one number here that can't be validated without real Plan 9 hardware.
3. **Background reader goroutines are never explicitly stopped.** Fine for the single, process-lifetime `*keys*` binding, but worth noting for the `api.WithKeySource(NewQueuedKeySource(...))` embedder path: each `Run` that installs a fresh queued source over a reader that never returns EOF leaks a goroutine per call. Likely out of scope for this PR, just flagging for future embedder use.
4. Minor: `rawMode`'s early `!stdinIsConsole()` return happens before the `consctlMu` lock and the "already raw" check — harmless since nothing is mutated on that path, just noting the ordering isn't symmetric with the rest of the function.

mparrett:
## Device testing on agent9

Second terminal type tested. Ran the input smoke on agent9 (9front amd64, QEMU), which gives a second real terminal to compare against the bare console. Result confirms the raw-mode gate behaves correctly on both.

| Terminal | `fd 0` | `raw-mode!` | Notes |
|---|---|---|---|
| bare 9front console | `/dev/cons` | engages | consctl `rawon` works; keys arrive unbuffered |
| agent9 vtwin | `#\|/data` (a pipe) | no-ops (returns nil) | stdin is vt's pipe, not the console |

`read-key` and `key-pending?` tokenized correctly on both — letters, `\n`, etc. all came through right. In vtwin the keys are line-buffered because vt cooks its own pipe, and that's exactly why `raw-mode!` no-ops there: `fd 0` is `#|/data`, so the `fd2path(0) == "/dev/cons"` gate refuses to touch consctl. That's the pipe case from the earlier review working as intended, now confirmed against a real pipe-backed terminal rather than argued from the manual.

For anyone testing the full game later: vt-family terminals don't take raw input through consctl at all — they switch on a terminal escape. alacritty9's notes describe raw mode "via the alt-screen toggle," and xsofy enters the alternate screen when it starts, so on those terminals the raw ANSI input path runs through the alt-screen escape while `raw-mode!` correctly no-ops on the pipe stdin. That path is xsofy-side and I haven't confirmed it on hardware yet; the consctl path (bare console) and the gate (both) are confirmed.

dharmatech:
> ### Smoke test
> Save as `smoke.lg` and run `lg smoke.lg` in a cons window (a rio window or the bare text console — both own `/dev/cons` + `/dev/consctl`):
> 
> ```
> (term/write (str "key-pending? before input: " (term/key-pending?) "\r\n"))
> (term/raw-mode!)
> (term/write "raw mode on. type keys ('q' or EOF quits):\r\n")
> (loop []
>   (let [k (term/read-key)]
>     (cond
>       (nil? k)  (term/write "EOF - bye\r\n")
>       (= k "q") (term/write "q - bye\r\n")
>       :else     (do (term/write (str "  read-key => " (pr-str k) "\r\n"))
>                     (recur)))))
> (term/restore-mode!)
> ```
> 
> Expected: each key echoes the instant you press it (no Enter). `q` quits and the shell returns to normal line editing — if it doesn't, raw mode wasn't restored.

Results of the `smoke.lg` (qemu, agent9, drawterm, rio, term):

image

I pressed `q` and it exited as expected. 👍

dharmatech:
Here's what I get when I run xsofy (qemu, agent9, alacritty9):

image

I'm using these branches of let-go and xsofy:

```
cpu% cd ../let-go/
cpu% git/branch
heads/agent/plan9-terminal-input
cpu% cd ../xsofy/
cpu% git/branch
heads/main
```

dharmatech:
When I try to run xsofy from a Windows terminal (connect to agent9 via drawterm):

image

```
> $env:PASS = 'agent9agent9'; C:\Users\dharm\src\drawterm\build\msvc\drawterm.exe -h 127.0.0.1 -a 127.0.0.1 -u glenda -G -c "cd /usr/glenda/src/xsofy; /usr/glenda/src/let-go/lg main.lg"
error: sync /dev/stdout: permission denied
  --> main.lg:39:3
    |
 39 |   (term/flush)
    |   ^^^

stack trace:
  at flush (main.lg:39:3)
  at shutdown-terminal (main.lg:600:7)
  at -main (main.lg:608:3)
```

mparrett:
@dharmatech Thanks for testing! The flush issue should be addressed by 6323eef707ee63b4fed9d80b4767d68c59194b95.

I got the game to load and read input in alacritty9. It's slow, and needs to read the terminal size, as 80x24 isn't workable at the moment. Let me know if you get another chance to test with the latest changes.

dharmatech:
So I'm running some tests in a fresh 9front QEMU image.

I ran into this during `go build -o lg .`:

image

I didn't run into that when I build on agent9.
Not sure if something changed or if it's something to do with this particular image.
Just thought I'd share the issue.

Full terminal log shown below:

```
cpu% echo ipv6 >/net/cs
cpu% webfs
cpu% 
cpu% hget https://raw.githubusercontent.com/Alino/agent9/main/pac9/pac9 >/rc/bin/pac9
cpu% chmod +x /rc/bin/pac9
cpu% mkdir -p /sys/lib/pac9
cpu% hget https://raw.githubusercontent.com/Alino/agent9/main/pac9/registry >/sys/lib/pac9/registry
cpu% 
cpu% pac9 install go
pac9: fetching https://github.com/Alino/agent9/releases/download/go/go-amd64.tar.gz
pac9: installed go
cpu% go
Go is a tool for managing Go source code.

Usage:

	go <command> [arguments]

The commands are:

	bug         start a bug report
	build       compile packages and dependencies
	clean       remove object files and cached files
	doc         show documentation for package or symbol
	env         print Go environment information
	fix         apply fixes suggested by static checkers
	fmt         gofmt (reformat) package sources
	generate    generate Go files by processing source
	get         add dependencies to current module and install them
	install     compile and install packages and dependencies
	list        list packages or modules
	mod         module maintenance
	work        workspace maintenance
	run         compile and run Go program
	telemetry   manage telemetry data and settings
	test        test packages
	tool        run specified go tool
	version     print Go version
	vet         report likely mistakes in packages

Use "go help <command>" for more information about a command.

Additional help topics:

	buildconstraint build constraints
	buildjson       build -json encoding
	buildmode       build modes
	c               calling between Go and C
	cache           build and test caching
	environment     environment variables
	filetype        file types
	goauth          GOAUTH environment variable
	go.mod          the go.mod file
	gopath          GOPATH environment variable
	goproxy         module proxy protocol
	importpath      import path syntax
	modules         modules, module versions, and more
	module-auth     module authentication using go.sum
	packages        package lists and patterns
	private         configuration for downloading non-public code
	testflag        testing flags
	testfunc        testing functions
	vcs             controlling version control with GOVCS

Use "go help <topic>" for more information about that topic.

cpu% cd
cpu% mkdir src
cpu% pwd
/usr/glenda
cpu% cd src
cpu% pwd
/usr/glenda/src
cpu% lc
cpu% git/clone https://github.com/nooga/let-go.git
fetching... /
indexing 9268 objects: 100%
checking out repository...
cpu% cd let-go
cpu% lc
.agents/				.goreleaser.yml			README.md				lg_ansi.go				pkg/
.claude/				.pre-commit-config.yaml	benchmark/				lg_ansi_plan9.go		scripts/
.git/					.vscode/				cmd/					lg_profile.go			tap_migrations.json
.gitattributes			CONTRIBUTORS			deps.edn				lg_profile_default.go	test/
.github/				Casks/					docs/					lg_repl.go				wasm/
.gitignore				Dockerfile				examples/				lg_repl_stub.go			wasm.go
.gitmodules				HomebrewFormula/		go.mod					meta/
.golangci.yml			LICENSE					go.sum					mise.toml
.gomodcache/			Makefile				lg.go					openspec/
cpu% git/branch -a
heads/main
remotes/origin/agent/fix-plan9-key-pending
remotes/origin/agent/plan9-terminal-input
remotes/origin/ci/dedicated-determinism-step
remotes/origin/codex/key-tokenizer-fsm
remotes/origin/delint-golangci-v2
remotes/origin/docs-status-skill
remotes/origin/docs/docs-status-tool
remotes/origin/feat/ir-substrate-passes
remotes/origin/feat/js-url-param
remotes/origin/feat/set-rand-seed
remotes/origin/fix-http-header-raw-strings
remotes/origin/fix-var-doc-metadata
remotes/origin/fix/426-update-keys-vals-re-find
remotes/origin/fix/ns-loader-require-semantics
remotes/origin/fix/pod-shutdown-without-op
remotes/origin/fix/source-bootstrap-lgbgen
remotes/origin/homebrew-tap-let-go
remotes/origin/main
remotes/origin/migrate-homebrew-cask
remotes/origin/perf-data
remotes/origin/perf-timeline-page
remotes/origin/perf/lower-go-registry-exact-resolve
remotes/origin/perf/pr-fast-tune
remotes/origin/pr/wasm-host-eval
remotes/origin/release-brews-arm-linux
remotes/origin/release-workflow-tag-branch-fix
remotes/origin/release/v1.9.0
remotes/origin/runtime-only-build
remotes/origin/spike/libyamlstar-cshared
remotes/origin/sweep/feature/http-request-function
remotes/origin/sweep/reader-unit-tests
remotes/origin/tco
remotes/origin/wasip1-build
cpu% 
cpu% git/branch agent/plan9-terminal-input
refs/heads/agent/plan9-terminal-input: ccff61420c1415507c10747d13ca729f5fbacda1
cpu% go build -o lg .
go: downloading github.com/hashicorp/go-uuid v1.0.3
go: downloading github.com/zeebo/bencode v1.0.0
go: downloading github.com/zeebo/xxh3 v1.1.0
pkg/rt/pods.go:21:2: unable to resolve git version: failed to execute git version: exec: "git": executable file not found in $path
pkg/rt/interop_xxh3.go:6:2: unable to resolve git version: failed to execute git version: exec: "git": executable file not found in $path
pkg/nrepl/server.go:17:2: unable to resolve git version: failed to execute git version: exec: "git": executable file not found in $path
```

mparrett:
The `git` in that error is Go's own build-time VCS stamping, not let-go. When you `go build` inside a checkout, the toolchain shells out to a `git` binary to stamp the version into `runtime/debug.BuildInfo`, and a stock 9front only has the `git/clone`-style commands, not a single `git` on `$path`. agent9 ships a `git` shim, which is why the build worked there.

Two ways past it:

- `go build -buildvcs=false -o lg .` — skips the stamp and builds clean.
- Or don't build on 9front at all: cross-build on another box with `GOOS=plan9 GOARCH=amd64 go build -o lg .` and copy the binary over (`hget`/`scp`). That's how I've been producing the test binaries.

Thanks for putting it through its paces. The flush crash you hit earlier is fixed in 6323eef, so a retest with the latest branch should get past that too.

mparrett:
Thanks for the review. Addressed 1 and 2 in b98950b; 3 and 4 below.

**1 — `stdinIsConsole()` exact match.** Widened past the literal `/dev/cons`: it now also accepts the raw device form `#c/cons` and a console reached through a different mount (`…/dev/cons` suffix), while still rejecting the pipe (`#|/data`) and redirected-file stdins the gate exists to catch. I couldn't reproduce the `#c/cons` path on the hardware I have, so it's coded from the device semantics rather than observed — worth a second look if you've hit it in the wild.

**2 — `escGrace` fixed at 40ms.** Now overridable per source via `LETGO_ESC_GRACE_MS`; the default is unchanged. A test covers the override and the invalid-value fallback. That gives slow-link users (drawterm over a WAN) a knob without retuning the default for local 9P.

**3 — reader goroutine never stopped.** Agreed this is a real leak on the `WithKeySource(NewQueuedKeySource(...))` per-`Run` path. Leaving it for a follow-up: a clean stop means adding a `Close` to the `KeySource` seam and wiring teardown to call it, and no source implements one today, so it's a wider change than this PR. Happy to do it as a separate one.

**4 — early `!stdinIsConsole()` return before the lock.** Kept as is — the gate is a cheap path read that lets `raw-mode!` bail before taking `consctlMu` when stdin isn't the console, and nothing is mutated on that path. Can reorder it for symmetry if you'd rather.

mparrett:
## Device result: arrow tokenization on alacritty9

Ran the arrow smoke on real 9front amd64 + alacritty9 against the branch tip (b98950b).

The path matters here. Probes that call `raw-mode!` stay line-buffered under alacritty9, because `raw-mode!` correctly no-ops there — fd 0 is a pipe, not `/dev/cons`. alacritty9 takes raw input through the alternate-screen toggle instead, which is what xsofy enters on startup. A probe that enters the alt-screen first (like the game) gets unbuffered keys, one `read-key` token per press.

On that path the ANSI arrows tokenize cleanly:

```
↑  codes=(27 91 65)  ESC [ A
←  codes=(27 91 68)  ESC [ D
→  codes=(27 91 67)  ESC [ C
```

One quirk worth recording: ↓ arrives as `U+F800` (Plan 9's Kdown rune), not `ESC [ B` — consistently, every press. That's a keymap gap on this alacritty9 build, not a tokenizer issue; `read-key` returns exactly what the terminal delivers, and xsofy binds the rune alongside the escape in a separate change.

So the arrow-token-on-hardware check is covered: split-safe tokenization, one token per key, on the real ANSI path. With the earlier bare-console (consctl raw mode) and vtwin (pipe-stdin gate) results, the input path is confirmed across all three terminal types.

mparrett:
@dharmatech A correction on the flush fix — I was too quick earlier. `6323eef` only handled the local console's `EINVAL`, not the `EPERM` your drawterm run hit on `/dev/stdout`, so my "should be fixed, retest" was premature.

I reproduced your exact crash (`sync /dev/stdout: permission denied` at `term/flush`) on a 9front amd64 VM and traced the root cause: the flush no-op was gated on `os.Stdout.Stat().IsRegular()`, but on Plan 9 a console Stats as regular through Go's `FileMode` (no device-type bits), so flush misclassified the console as a real file and actually `Sync`'d it.

Fixed properly in `0a8d77d`: syncability is now decided by the fd's namespace path (`/dev/cons`, `/dev/stdout`, and `#`-devices/pipes are never fsync'd), and `EPERM` is swallowed alongside `EINVAL` as a backstop. Verified on the VM — a bare `(term/flush)` with stdout `= /dev/stdout` returns cleanly instead of crashing.

If you get another chance, a drawterm retest on the latest branch would confirm it on your setup — that's the environment that surfaced it.

dharmatech:
Here's how I got past the git messages in the let-go build.

I installed certs and the build ran fine:

```
cpu% hget https://curl.haxx.se/ca/cacert.pem >/sys/lib/tls/ca.pem
cpu% 
cpu% go build -o lg .
go: downloading github.com/hashicorp/go-uuid v1.0.3
go: downloading github.com/zeebo/bencode v1.0.0
go: downloading github.com/zeebo/xxh3 v1.1.0
go: downloading github.com/klauspost/cpuid/v2 v2.2.10
```

dharmatech:
Dude!!! It's working! I'd say it's at a speed where it would be playable.

image

So that's in a Windows terminal, connected to plan9 via this drawterm command:

```
$env:PASS = 'PASSWORD'; C:\Users\dharm\src\drawterm\build\msvc\drawterm.exe -h 127.0.0.1 -a 127.0.0.1 -u glenda -G -c "cd /usr/glenda/src/xsofy; /usr/glenda/src/let-go/lg main.lg"
```

dharmatech:
`legmacs` also running in the same way 😅

image

dharmatech:
When I run `alacritty9` on stock 9front (not agent9) connected via drawterm, `alacritty9` is totally unresponsive.
Doesn't respond to key presses.
CC: @alino

dharmatech:
Running alacritty9 in the qemu console works:

image

xsofy seems slower this way than from Windows Terminal.

pkg/rt/keysource_queued.go
```diff
@@ -0,0 +1,215 @@
...
+func NewQueuedKeySource(r io.Reader) KeySource {
+	return &queuedKeySource{r: r, notify: make(chan struct{}, 1), grace: resolveGrace()}
+}
+
+// queuedReadChunkSize matches native's readChunkSize: a blocking Read returns as
+// soon as any data is available, so a larger buffer adds no latency to a single
+// keystroke — it just collects held-key bursts in fewer reads.
+const queuedReadChunkSize = 256
+
+// start launches the background reader once. Called with s.mu held.
```
nnunley: This does not hold the s.mu lock, per your comment. It looks like s.readLoop has finer grained locking, so you probably should remove this comment.

mparrett:
@dharmatech Followed my own suggestion and confirmed the flush fix on real drawterm, so no need to wait on me — though a sanity check from your setup is still welcome.

Built drawterm on macOS (rcarmo's fork, `CONF=osx-cocoa`) and connected `-G` to a local 9front cpu server running the `0a8d77d` binary:

```
; ./lg flushtest.lg          # (term/write) (term/flush) (term/write)
FLUSH_TEST_START
FLUSH_TEST_OK_NO_CRASH
```

`(term/flush)` over drawterm's `/dev/stdout` returns cleanly now, instead of the `sync /dev/stdout: permission denied` you hit. The game itself also renders over `drawterm -G` — truecolor SGR and cursor positioning came through — with no errors during the run.

So the shutdown crash is fixed at the source, not papered over. A run on your own drawterm would still confirm it end to end, but this should unblock you.

pkg/rt/keysource_queued.go
```diff
@@ -0,0 +1,215 @@
...
+// with a nil error at a clean EOF (the read-key nil contract), or "" with the
+// retained error if the reader failed. Blocking here matches native (which
+// blocks in poll); callers stay responsive by gating on KeyPending first.
+func (s *queuedKeySource) ReadKey() (string, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.start()
+
+	for len(s.buf) == 0 && !s.done {
+		s.waitLocked(0) // no deadline: block until data or reader end
```
nnunley: Hrm. for single user, this is probably fine. Having a deadline might be a longer term consideration. But this should only block the channel monitoring input, right?

pkg/rt/keysource_queued.go
```diff
@@ -0,0 +1,215 @@
...
+	if len(s.buf) == 0 {
+		s.buf = nil
+	}
+	return tok
+}
+
+// waitLocked releases the lock, waits for a reader wake (or the timeout, if
+// nonzero), and re-acquires it — the condition-variable wait, done over a
+// channel so it can be time-bounded. Called with s.mu held; returns holding it.
+func (s *queuedKeySource) waitLocked(timeout time.Duration) {
```
nnunley: The method name doesn't match the intent.

nnunley (commented):
Some concerns on comments and naming. High level, it looks good.

mparrett:
Game loads and is interactive on emulator via drawterm. Pretty slow, but it works.

**xsofy over drawterm  →  agent9 (9front amd64)**

image

mparrett:
@dharmatech Appreciate the helpful testing, and glad to hear it's working! I have some open PRs for render optimization that should help somewhat.

mparrett:
Glad it's playable for you.

On alacritty9 over drawterm — you can skip it entirely. `drawterm -G` already gives you a real ANSI terminal (your Windows Terminal does the rendering), which is exactly why the plain `lg main.lg` run worked. alacritty9 is for when you're at the physical 9front display — the qemu console, where you saw it work. Running it *inside* a drawterm session just stacks a second terminal on top, and routing input down through drawterm into it is an alacritty9 question (@alino's area), not a let-go one. For the drawterm path, `-G` straight to `lg` is the whole story.

Related, since it came up: over drawterm all four arrow keys arrive as standard ANSI and work. The Kdown-as-`U+F800` quirk is specific to alacritty9's own keymap, so it doesn't affect the drawterm path.

pkg/rt/keysource_queued.go
mparrett: Good prompt — reworded rather than removed, since the precondition is real: `start()` flips `s.started`, so it has to run under `s.mu` or two callers could both launch the reader. The confusion was on my end; the note now says the lock requirement is on `start()` itself, and that the `readLoop` goroutine it spawns does its own finer-grained locking (no lock held across the blocking `Read`). cee4746.

pkg/rt/keysource_queued.go
mparrett: Right — it blocks only the caller (in xsofy, the input-poll loop), never the reader goroutine or anything else; the reader keeps filling the buffer independently. Callers stay responsive by gating on `KeyPending` before `ReadKey`, matching native's poll-then-read. And agreed the deadline is worth a look longer term (an embedder wanting a bounded read) — left it out here to keep the single-user path simple.

pkg/rt/keysource_queued.go
mparrett: Agreed — renamed to `condWait` in cee4746. It unlocks, waits, then re-locks (the `sync.Cond.Wait` shape), so `waitLocked` read backwards.

nooga (approved):
Verified locally on the PR branch: native build/vet/gofmt clean, `GOOS=plan9 GOARCH=amd64 go build ./...` clean, and `go test -race ./pkg/rt/...` green including the new `keysource_queued_test.go` suite (split ESC/CSI/UTF-8 stitching, bare-ESC grace timeout, reader-error propagation, EOF-blind `KeyPending`, `LETGO_ESC_GRACE_MS` override). Test coverage is the right shape per CLAUDE.md — `queuedKeySource` is a new runtime data structure, so a Go unit test in `pkg/rt/` is correct; a `.lg` integration test isn't practical here since the code is plan9-build-tag-gated.

The two items from @nooga's earlier review that remain open are appropriately non-blocking:
- Reader goroutine never stopped (point 3): real leak on the `api.WithKeySource(NewQueuedKeySource(...))` embedder path per `Run`, correctly scoped out of this PR pending a `Close()` seam on `KeySource`. Worth opening a tracking issue so it isn't forgotten, but it doesn't affect the process-lifetime `*keys*` binding this PR wires up on Plan 9.
- Lock-ordering nit on `raw-mode!`'s early return (point 4): cosmetic, no mutation happens on that path either way.

Nice validation trail: 9front console (consctl rawon/rawoff), agent9 vtwin (pipe stdin correctly no-ops raw-mode), and alacritty9 (ANSI arrow tokenization) all confirmed on real hardware, plus an independent tester catching and getting a `term/flush` EINVAL crash fixed along the way. Good to merge once #460 lands.

mparrett:
Thanks for the review and the approval. Filed the goroutine-stop follow-up (your point 3) as #498 so it doesn't get lost — the `KeySource.Close()` seam for the `WithKeySource` embedder path. #461 is stacked on #460, so it'll rebase onto `main` once #460 lands.

PR #462, opened 2026-07-13
perf(vm): per-var root binding chains — O(1) bound deref (Phase 2, stacks on #459)

nnunley:
**Phase 2**, stacked on #459. Replaces the shared-binding-stack walk for bound dynamic vars with **per-var binding chains + a membership registry**, making root-var deref **O(1)** regardless of how many other vars are bound.

## The problem Phase 1 left

Phase 1 fixed the *unbound* dynamic case. A genuinely *bound* dynamic var still resolved via `globalBindingStack.current(v)`, which walks the shared chain **from the head until it finds `v`** — so bound-var deref was **O(number of dynamic vars bound after it)**. Deref of `*a*` in `(binding [*a* *b* *c*] …)` walked past `*c*` and `*b*`. Measured (M3):

| others bound | Phase 1 | **Phase 2** |
|---|---|---|
| 0 | 1.10 ns | **0.55 ns** |
| 4 | 3.08 ns | **0.58 ns** |
| 16 | 7.54 ns | **0.58 ns** |
| 64 | **46.9 ns** | **0.58 ns** |

This is a **serial** cost — it hits single-threaded code, which is most Clojure code — whenever several dynamics are bound (nested `binding`, config-heavy frameworks). Deref of a bound var shouldn't depend on *other* vars' bindings; now it doesn't.

## Change

Root bindings move onto each `Var` as its own immutable-frame chain behind an atomic head (`rootBind`):

```go
func (v *Var) derefRoot() Value {
    if val, ok := rootDerefHead(v); ok { return val } // this var's own head — O(1), per-var cache line
    return v.Root()
}
```

- **O(1), contention-free:** the chain is per-var, so the head *is* the current binding — no walk, no shared read. `rootBindDepth` (Phase 1) is retired; the head pointer subsumes it.
- **Registry** (`map[*Var]struct{}` under one mutex) lets `Child()`/`RunWithBindings` enumerate live root bindings. Written **only** on first-bind/last-unbind transitions; **deref never touches it**.
- **Single source of truth, no cache:** per-var chain authoritative for values, registry for membership — updated together, nothing to keep coherent.
- **Lock ordering** is acyclic: push/pop take `v.mu`, release, then `rootRegistryMu` (un-nested); `rootInstall` nests `rootRegistryMu`→`v.mu`; never both held in push/pop.
- **Child contexts unchanged** — isolated per-goroutine stacks, seeded via `rootSnapshot()` at spawn. #210 isolation preserved.

## Tests

`-race` gate: 16-goroutine concurrent bind/deref/**snapshot** on the same var (registry read-while-write), distinct vars, child isolation, `RunWithBindings` bracketing, full-unbind→registry-empty. Zero data races. Full non-race suite green.

`BenchmarkVarDerefBoundDepth` proves flat-in-N (the O(1) payoff); `BoundParallel` 0.14 ns (contention gone).

**Note (not this PR):** `go test ./pkg/vm/ -race` surfaces a **pre-existing** `TestArrayVectorSeqChunkedFirstRace` race (`vector.go:268`), confirmed on main; unrelated.

Spec/plan under `docs/superpowers/`. Shows the combined diff with #459 until that merges.

---
**Part of #464** (Epic: Runtime performance & concurrency) — var-deref regression recovery, Phase 2 (stacks on #459).

### DIFF
--- a/pkg/vm/exec_context.go
+++ b/pkg/vm/exec_context.go
@@ -49,12 +49,15 @@ func NewExecContext() *ExecContext {
 // A nil receiver seeds from the root context.
 func (ec *ExecContext) Child() *ExecContext {
 	src := ec
-	if src == nil {
-		src = RootExecContext
-	}
 	c := NewExecContext()
-	c.bindings.installSnapshot(src.bindings.snapshot())
-	c.scope = src.scope
+	if src == nil || src == RootExecContext {
+		c.bindings.installSnapshot(rootSnapshot())
+	} else {
+		c.bindings.installSnapshot(src.bindings.snapshot())
+	}
+	if src != nil {
+		c.scope = src.scope
+	}
 	return c
 }
 
@@ -98,11 +101,10 @@ func OpenChildEC(ec *ExecContext) *Scope {
 // suitable for passing to NewExecContextFrom to isolate a child context.
 // A nil receiver uses the root context.
 func (ec *ExecContext) BindingSnapshot() BindingSnapshot {
-	src := ec
-	if src == nil {
-		src = RootExecContext
+	if ec == nil || ec == RootExecContext {
+		return rootSnapshot()
 	}
-	return src.bindings.snapshot()
+	return ec.bindings.snapshot()
 }
 
 // NewExecContextFrom returns a fresh context whose binding stack is seeded from
@@ -126,10 +128,14 @@ func (ec *ExecContext) orRoot() *ExecContext {
 
 // --- dynamic-var resolution (ec owns the binding state) ---------------------
 
-// deref returns v's current value in this context: the top dynamic binding if
-// any, else v's root.
 func (ec *ExecContext) deref(v *Var) Value {
 	ec = ec.orRoot()
+	if ec == RootExecContext {
+		return v.derefRoot()
+	}
+	// Child context: an isolated, per-goroutine stack — uncontended, so walk it
+	// directly. Child bindings are gated by isDynamic (set at declaration or on
+	// any bind) as before.
 	if v.isDynamic.Load() {
 		if val, ok := ec.bindings.current(v); ok {
 			return val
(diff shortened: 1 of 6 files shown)


nooga (approved):
Verified locally on the PR branch: clean build, full `go test ./... -count=1` green, and all new `-race` tests pass (`root_bindings_test.go`, `var_binding_race_test.go`). Re-ran `BenchmarkVarDerefBoundDepth` myself — deref cost is flat (~1ns) across others=0/1/4/16/64, confirming the O(1) claim vs. Phase 1's O(N) walk. Lock ordering (v.mu released before rootRegistryMu in push/pop, nested the other way only in rootInstall) is acyclic as documented. Test coverage is the correct kind per CLAUDE.md (Go unit tests in pkg/vm/ for VM-internal binding infrastructure).

Two non-blocking notes for a follow-up:

1. **`root_bindings.go`**: concurrent `rootPush`/`rootPop` on the *same* var from different goroutines at the root context (not through `ec.Child()`) can race the registry update relative to the chain update — e.g. G1's `rootPush` observes `old == nil`, gets preempted before `registryAdd`, and G2's `rootPop` runs the full push→pop cycle in between (its `registryRemove` is a no-op since the add hasn't landed yet); G1 then adds `v` to `rootRegistry` even though `v.rootBind` is already nil again. This produces a permanent phantom registry entry for that var. It's harmless in practice — `rootSnapshot` already drops zero-length chains (the `continue // raced an unbind` case) so it never surfaces a wrong deref/snapshot value, just an unbounded-but-tiny amount of registry bloat. Worth a comment acknowledging it (or a CAS-based registry transition if you want to close it), but not worth blocking on.

2. **Vestigial state**: `RootExecContext.bindings` / `globalBindingStack` are no longer touched by any root-context operation (push/pop/deref/snapshot/install all route through the new `rootXxx` helpers). Fine to leave for now but flag for cleanup in a later pass — right now it's dead weight that could confuse a future reader into thinking it's still load-bearing.

3. **Stacking note**: since #459 is still open, this diff (per `gh pr diff`) is the combined #459+#462 changeset against `main`. `mergeable` shows clean today, but merge #459 first (or be deliberate about squashing) to avoid the two PRs' commit histories colliding later.

None of the above changes my recommendation — approving.

PR #465, opened 2026-07-13
chore(hooks): reject committing compiled executables (pre-commit/prek)

nnunley:
## What

Adds a fail-fast `pre-commit` hook that refuses staged **compiled executables** (Mach-O / ELF / PE) — e.g. an accidentally-`git add`-ed `go build` output — before they bloat history.

- `scripts/reject_compiled_binaries.py` — stdlib-only Python (consistent with the repo's other hook scripts), magic-byte detection.
- `.pre-commit-config.yaml` — wired as the first local `pre-commit` hook so it fails cheaply before build/lint.

## Why magic-byte, not size

Detection is by executable magic prefix, so it **does not** flag the repo's legitimate binary assets: the core bundle starts with `LGB\x01` and images (`meta/logo.png`) aren't executable formats. A blanket size cap would false-positive on `core_compiled.lgb`.

## Tested

- Rejects a real Mach-O `go build` output (exit 1, names the file).
- `prek run forbid-compiled-binaries --all-files` **passes** on the entire repo — no false positives on `core_compiled.lgb` / `logo.png`.

## Caveat

Git-only — `jj` doesn't run git hooks, same as the repo's existing `scripts/pre-commit` / ratchet hooks. Bypass an intentional commit with `git commit --no-verify`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -27,6 +27,17 @@ repos:
 
   - repo: local
     hooks:
+      # Fail fast on an accidentally-staged compiled executable (e.g. a stray
+      # `go build` output). Magic-byte based (Mach-O / ELF / PE), so it ignores
+      # the repo's legitimate binary assets — the LGB core bundle and images.
+      # No type filter: it must inspect every staged file's leading bytes.
+      - id: forbid-compiled-binaries
+        name: forbid committing compiled executables
+        description: Reject staged Mach-O / ELF / PE binaries before they bloat history
+        entry: python3 scripts/reject_compiled_binaries.py
+        language: system
+        stages: [pre-commit]
+
       # Auto-fix formatting before linters run. gofmt ships with Go.
       # goimports is fetched on demand via `go run ...@latest` so contributors
       # don't need a separate install; the module cache makes subsequent runs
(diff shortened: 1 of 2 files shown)


nooga (approved):
Verified functionally by checking out the branch: `reject_compiled_binaries.py` correctly catches a real `go build` Mach-O binary and synthetic ELF/PE inputs, ignores an `LGB\x01`-prefixed file and an empty file, and handles directories/missing paths (deleted-file-in-commit case) without raising. The hook's lack of `types:`/`files:` filters is correct and intentional (it must sniff every staged file's magic bytes, not just certain extensions) — good call-out in the inline comment.

Two optional, non-blocking nits:
- `classify()` silently returns `None` on any `OSError`/`IsADirectoryError`, which is the right call for deleted-file/dir entries, but it also means a permissions error on a real binary would silently pass through undetected. Not worth complicating for a fail-fast local hook, just flagging for awareness.
- No automated test for the script itself, but this matches the existing convention for sibling hook scripts (`check_test_location.py`, `docs_frontmatter_hook.py`), which also ship untested — consistent with repo practice, not a regression.

No blocking issues. CI is green across the board (lint, build, CodeQL, gogen-diff, etc.)

PR #466, opened 2026-07-13
Build let-go as a WASI (`GOOS=wasip1`) module

mparrett:
# Build let-go as a WASI (`GOOS=wasip1`) module

let-go already targets `GOOS=js` for the browser. Extending that to `wasip1` turns it into a standalone WASI module any wasm host can load: wasmtime from a CLI, or wazero from a Go program embedding let-go as a sandboxed, capability-gated compute runtime. That embedding shape is the payoff: a Clojure dialect you can drop into a Go host, with the host controlling what it can touch.

Like the existing browser build, this uses `GOARCH=wasm`. That's a 32-bit-memory target, but standard Go keeps `int` 64-bit on it, so arithmetic matches the native runtime — `10^18` and `2^62` exact (verified below).

## The change

Three build-tag flips, extending the existing `js` platform gates to also cover `wasip1`:

```
lg_repl.go       !plan9 && !js          → !plan9 && !js && !wasip1
lg_repl_stub.go  plan9 || js            → plan9 || js || wasip1
pkg/rt/term.go   !js && !plan9          → !js && !plan9 && !wasip1
```

The REPL (`lg_repl.go`) and interactive terminal (`pkg/rt/term.go`) pull in `chzyer/readline` and `x/sys/unix` poll/ioctl, which have no wasip1 backing. The gates stopped at `!js`, so a `GOOS=wasip1` build dragged them in and failed to link. Routing wasip1 to the existing `lg_repl_stub.go` fixes it; the stub seam was already there for `js`/`plan9`, so no new stub was needed, and nothing in the default entry point force-links a terminal namespace.

Also in the PR:

- **Fix:** requiring an unavailable native namespace under wasip1 now reports it cleanly instead of recursing. Gating out `term.go` leaves no `term` installer, and `loadEmbedded`'s `term` special case called `rt.NS("term")`, which re-registered a placeholder and re-entered the loader until the wasm stack was exhausted (`(require 'term)` → `call stack exhausted`). It now uses a non-registering lookup and returns `unable to load namespace term`. Guarded by a resolver regression test.
- **CI:** a `wasip1-build` job that cross-builds the target on every PR, so a change that pulls readline or unix terminal code back into the wasip1 link fails the PR instead of surfacing downstream.
- **Docs:** `docs/guide/wasi.md` (build, run, and the host-controlled capability model), plus a short `docs/guide/tinygo.md` noting that TinyGo isn't currently supported and the blocker it hits — surfaced while adding this target.

## Validation

Standard Go `GOOS=wasip1 GOARCH=wasm go build`, run under wasmtime:

| check | result |
|---|---|
| build | clean, 24 MB `.wasm` |
| `-e '(println (str "wasi says " (* 6 7)))'` | `wasi says 42` — stdio works |
| `-e '(* 1000000000 1000000000)'` | `1000000000000000000` (10^18) — 64-bit exact |
| `-e '(bit-shift-left 1 62)'` | `4611686018427387904` (2^62) — full width |

Scope: this is the standard-Go target. TinyGo's `-target=wasi` is also `GOOS=wasip1`, but it doesn't build let-go as-is — TinyGo 0.41.1 traps at init on `reflect.Type.Method()` — so it's out of scope here.

## Notes for review

- **Capabilities.** The module has no ambient authority. No `term` namespace (the interactive terminal is gated out; requiring it reports `unable to load namespace term` rather than recursing), no sockets, no threads. Filesystem access is host-controlled: nothing by default, but a wasip1 host can preopen directories (e.g. wasmtime `--dir`), which `slurp` and friends then see. Fine for headless eval and embedding; a TUI-over-wasi would need a headless `term` stub (future work).
- **Independent of the runtime-only build (#424).** These compose but don't couple: different files (platform gates in `pkg/rt`/repl vs. compiler linkage in `eval.go`), different risk (mechanical vs. design), different guarantees (capability vs. trust surface). Layering runtime-only on top gives the "no `eval`, run untrusted precompiled bytecode under a wasm sandbox" shape, but that's a separate change.
- **Follow-up, not here: the wazero embedding demo.** wasmtime already proves it runs as wasip1; the embedding pitch is the same module loaded from a Go host via wazero, fed a form, capturing output. Happy to add it as a follow-up example if that's useful.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -95,6 +95,27 @@ jobs:
           LETGO_RUN_LOWERING_DETERMINISM: "1"
         run: go test -v -timeout 600s ./test/e2e/ -run 'TestLoweringDeterminism|TestDeftypeSkeletonNativeLowering'
 
+  # Cross-build guard for the wasip1 (WASI) target. The REPL and interactive
+  # terminal are build-tagged off wasip1 (routed to lg_repl_stub.go); without
+  # this gate, a change that pulls chzyer/readline or x/sys/unix terminal code
+  # back into the wasip1 link would break the module and only surface
+  # downstream. Keeps the "let-go builds as a GOOS=wasip1 module" guarantee
+  # honest. Build-only: the committed bytecode bundle is enough to link, no
+  # regeneration or wasm runtime needed.
+  wasip1-build:
+    name: wasip1-build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
+      - name: Build for GOOS=wasip1
+        run: GOOS=wasip1 GOARCH=wasm go build -v -o /tmp/lg-wasip1.wasm .
+
   # Differential gold check against a LIVE Clojure. The everyday `build` job is
   # clojure-free (it compares let-go to the committed test/gold/*.out goldens);
   # this job re-derives those goldens from real Clojure and fails if they drift.
--- a/docs/README.md
+++ b/docs/README.md
@@ -57,6 +57,8 @@ A subdir is earned when a cluster of related docs justifies one; one-off cross-c
 | Version requirements, range matching (`let-go.semver`) | `guide/semver.md` |
 | `io/resource`, `-resource-paths` / `-source-paths` resolution | `guide/resources-and-source-paths.md` |
 | Embedding let-go in a Go program | `guide/embedding-in-go.md` |
+| Running as a WASI (`GOOS=wasip1`) module | `guide/wasi.md` |
+| let-go under TinyGo (status: not currently supported) | `guide/tinygo.md` |
 | nREPL server + editor setup | `guide/nrepl.md` |
 | Clojure compatibility: namespace table + differences | `guide/clojure-compatibility.md` |
 | Running, compiling, WASM, project mgmt (lgx) | `guide/usage.md` |
--- a/docs/guide/tinygo.md
+++ b/docs/guide/tinygo.md
@@ -1,0 +1,21 @@
+---
+status: active
+last-verified: 2026-07-13
+human-verified:
+---
+
+# let-go under TinyGo
+
+**Status: not currently supported.** This note records what's known, surfaced
+while adding the standard-Go wasip1 target (which shares `GOARCH=wasm`).
+
+TinyGo's `-target=wasi` is also `GOOS=wasip1`, so the wasip1 build gating applies
+to it too — but let-go does not build and run under TinyGo as-is:
+
+- **Traps at initialization.** With stock TinyGo 0.41.1 the module builds, but
+  traps before executing any code: `unimplemented: (reflect.Type).Method()`.
+  let-go's runtime boxes Go values through `reflect`, which TinyGo does not fully
+  implement, so namespace installation fails at startup.
+
+Until that's resolved, use the standard-Go build — see
+[Running let-go as a WASI module](wasi.md).
--- a/docs/guide/wasi.md
+++ b/docs/guide/wasi.md
@@ -42,12 +42,9 @@ $ wasmtime lg.wasm -e '(bit-shift-left 1 62)'
 
 ## TinyGo
 
-This guide covers the standard Go toolchain, which is what builds let-go as a
-wasip1 module. TinyGo's `-target=wasi` is also `GOOS=wasip1`, but it does not
-build let-go as-is: stock TinyGo 0.41.1 traps at initialization with
-`unimplemented: (reflect.Type).Method()` before running any code, because the
-runtime's reflect-based boxing isn't available there. So the build here is the
-standard-Go one.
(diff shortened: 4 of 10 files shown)


mparrett:
I tested the PR head locally. The standard-Go WASI build and documented arithmetic examples pass, but I found two runtime concerns that the build-only CI job does not catch.

1. **[P1] Requiring `term` crashes the WASI module.**

   Excluding `pkg/rt/term.go` leaves no `term` namespace installer. `NSResolver.loadEmbedded("term")` assumes it was already registered and calls `rt.NS("term")`, which re-enters the loader indefinitely.

   Repro:

   ```sh
   wasmtime lg.wasm -e "(require 'term)"
   ```

   This traps with `wasm trap: call stack exhausted` rather than reporting an unavailable namespace. A WASI `term` stub or non-recursive missing-namespace handling would fix it; a runtime smoke test would catch regressions.

2. **[P2] The TinyGo compatibility claim is currently false with TinyGo 0.41.1.**

   `tinygo build -target=wasi` succeeds, but the resulting module traps during initialization—even for `-e '42'`:

   ```text
   panic: unimplemented: (reflect.Type).Method()
   github.com/nooga/let-go/pkg/vm.NewBoxed
   github.com/nooga/let-go/pkg/rt.installLangNS
   ```

   It never reaches the expression, so 32-bit integer width is not the only tradeoff. The documentation should remove or version-qualify the TinyGo claim unless a compatible path is added and exercised.

mparrett:
Both fixed in the latest push.

**P1 (term recursion).** Good catch. Root cause was `loadEmbedded`'s `term` special case calling `rt.NS("term")` on the assumption term was installed at init — which holds on platforms with an `installTermNS`, but not wasip1, where the terminal is build-tagged out. `rt.NS` re-registered a placeholder and re-entered the loader while `nsNeedsLoad` stayed set, so it recursed until the stack was exhausted. It now uses the non-registering `rt.LookupNS`, so requiring an uninstalled native ns returns `unable to load namespace term` cleanly. Added a resolver regression test that drops the natively-installed term ns to mimic wasip1 and asserts a clean error rather than recursion. I went with a clean error over a headless `term` stub for this PR — the stub (working degraded term under wasi) is worth doing but is a separate change.

**P2 (TinyGo claim).** Correct, and thanks for running it. Those numbers came from a branch that also carries the reflect shims that let let-go build under TinyGo; this PR has only the gating, so stock TinyGo 0.41.1 traps at init on `reflect.Type.Method()` as you saw. I pulled the claim — the body and `docs/guide/wasi.md` now scope to the standard-Go build and note TinyGo support is separate and currently traps without that enablement.

mparrett:
Fresh pass: the two earlier findings are fixed, but I found one remaining documentation/capability issue.

**[P2] Document host-preopened filesystem access.**

`docs/guide/wasi.md` says the module is "compute + stdio only," but the standard-Go WASI artifact can access directories that the host preopens:

```sh
wasmtime --dir=/tmp lg.wasm \
  -e '(println (slurp "/tmp/wasi-capability-test.txt"))'
```

I verified this successfully reads and prints the file. The module has no ambient filesystem access by default, but a wasip1 host can explicitly grant/mount directories. Since this PR presents the module as a capability-gated sandbox, the guide should describe filesystem access as host-controlled and opt-in rather than implying that it is unavailable.

mparrett:
Fixed the guide and PR body which overstated "compute + stdio only"; both now describe filesystem access as host-controlled and opt-in: nothing by default, but a wasip1 host can preopen directories (e.g. wasmtime `--dir`), which `slurp` and other file ops then see. That's a better fit for the capability-gated framing anyway — the host decides what the module can reach.

nooga (approved):

nooga (approved):
Verified locally against the actual PR branch: native build and `CGO_ENABLED=0 GOOS=wasip1 GOARCH=wasm go build .` both succeed, full `go test ./... -count=1` passes, and the new `TestRequireUnregisteredTermReportsUnavailable` regression test passes and correctly exercises the fixed recursion path (`rt.LookupNS` instead of `rt.NS` in `loadEmbedded`'s term case).

The three issues raised in earlier review rounds are all addressed in the current diff:
- Term-namespace infinite recursion under wasip1 (P1) — fixed with a non-registering lookup, backed by a regression test.
- False TinyGo compatibility claim (P2) — removed from PR body/docs, replaced with an accurate `docs/guide/tinygo.md`.
- Overstated "compute + stdio only" capability claim (P2) — corrected to describe filesystem access as host-controlled/opt-in.

Nit (non-blocking): the regression test sits in `pkg/bytecode/lgb_integration_test.go`, which is a bit of an odd home for resolver-namespace-loading behavior — `pkg/resolver/resolver_test.go` would be more discoverable — but it's not wrong given it reuses that file's existing compiler+resolver+rt setup.

Change is small, mechanical (build-tag gates + a non-recursive lookup fix), well-tested per the project's Go-unit-test convention for compiler/resolver edge cases, and CI is fully green including the new `wasip1-build` cross-build guard. LGTM.

mparrett:

nnunley (approved):

mparrett:
> Nit (non-blocking): the regression test

Moved to pkg/resolver/resolver_test.go in 6db32e3 - agreed it's the discoverable home for resolver namespace-loading behavior. No cycle since resolver already imports `compiler/rt/vm`, so the internal test package can too.

PR #467, opened 2026-07-13
chore(ir-stress): exclude scripts/lginterop.lg from corpus — stale gogen/parse-stmts

### LINKED ISSUE #293: scripts/lginterop.lg references undefined gogen/parse-stmts (stale build-wrapper-body)
## Summary

`scripts/lginterop.lg` references a `gogen` fn that does not exist:

```
scripts/lginterop.lg:130  ;; "Returns a vector of gogen statement nodes (via parse-stmts)."
scripts/lginterop.lg:159  (gogen/parse-stmts (str arity-check unboxes "\n" ret))))
```

`parse-stmts` is **not defined** anywhere in `pkg/rt/gogen/gogen.lg` (0 matches). So `build-wrapper-body` surfaces as `ir/build: unresolved symbol gogen/parse-stmts` in the `ir-stress` corpus — a stale-script reference, not a lowering gap.

## Scope

Minor / low-priority. `lginterop.lg` is a build-time script (not loaded by the runtime or tests), so this doesn't break anything in use — it's dead/stale code that pollutes the corpus failure set. Either:

- point it at the real gogen statement-parsing entry point (if one exists under a different name), or
- remove the dead `build-wrapper-body` path, or
- add `scripts/lginterop.lg` to the `:exclude` list in `scripts/ir-stress-corpus.edn` with this rationale.

Surfaced during the #258 self-AOT corpus close-out. Relates to the prior "lginterop blocker is a phantom" finding.

nnunley:
## Summary
`scripts/lginterop.lg`'s `build-wrapper-body` calls `gogen/parse-stmts`, which is not defined anywhere in `pkg/rt/gogen/gogen.lg` (gogen builds AST from s-expression forms; it has no Go-source statement parser). The smart-wrapper path is therefore dead-on-arrival today and only pollutes the ir-stress corpus failure set.

Per the options in #293, this excludes the file from the corpus with a rationale in `:exclude` (the map is consumed via `keys` in `scripts/ir-stress.lg:116`, so the rationale string rides along as the value). The wrapper model itself is being superseded by the annotation-driven registrar (#411), so repairing `build-wrapper-body` against the s-expression gogen API would be effort spent on a path already slated for replacement.

Fixes #293

## Test plan
- [x] corpus EDN still parses; `scripts/lginterop.lg` remains listed in `:stress` and is now subtracted via `:exclude`
- [x] `go test ./...` green

### DIFF
--- a/scripts/ir-stress-corpus.edn
+++ b/scripts/ir-stress-corpus.edn
@@ -1,8 +1,9 @@
 {:doc
  "IR-stress lower-go corpus: every repo .lg the AOT lowering should handle. Failures against :stress are real lowering gaps - forward/mutual-ref resolution, unsupported arg patterns, gogen emission, lowering-cost timeouts. Widest net: core via go:embed all:core plus tests, examples, scripts; external deps are themselves lowered .lg so they stay IN. Consumer: make ir-stress runs lower-go on :stress minus :exclude. Regenerate the file list via scripts/ir-stress.lg."
- ;; Nothing excluded: every listed file must read and lower.
+ ;; Every non-excluded file must read and lower. Excludes carry a rationale.
  :exclude
- {}
+ {"scripts/lginterop.lg"
+  "stale build-wrapper-body references undefined gogen/parse-stmts (#293); the smart-wrapper path is superseded by the annotation-driven registrar (#411)"}
  :stress
  ["examples/async.lg"
   "examples/concurrent-primes.lg"


nooga (approved):
Confirmed `gogen/parse-stmts` has zero definitions anywhere in the repo — `build-wrapper-body` is genuinely dead code, matching the #293 finding. The chosen fix (exclude with rationale, per one of the options #293 explicitly sanctioned) is reasonable given the wrapper model is being superseded by #411's annotation-driven registrar; repairing it would be wasted effort. Verified the `:exclude` consumption path in `scripts/ir-stress.lg:116-117` (`(set (keys (:exclude corpus-edn)))` then `remove` against `:stress`) matches what the PR describes — the rationale string is just carried as the map value and correctly ignored by `keys`. EDN is well-formed, CI is fully green (including gogen-diff/gold-differential, which would catch any corpus wiring regression). No concerns.

PR #469, opened 2026-07-13
fix(ir): accept defn attr-map position in AOT lowering

### LINKED ISSUE #289: AOT lowering: defn attr-map (e.g. {:added "1.0"}) reported as 'unsupported arg pattern'
Part of the self-AOT lowering close-out (#258). Surfaced by the `ir-stress` corpus.

## Symptom

`test/metadata_test.lg` defn `meta-doc-fn` fails to lower:

```
unsupported arg pattern: :added
```

## Cause

The defn in question uses the Clojure-standard **attr-map** position:

```clojure
(defn meta-doc-fn "defn doc" {:added "1.0"} [x] x)
```

Clojure's `defn` arglist grammar is `(defn name docstring? attr-map? [params*] body)`. The AOT lowering's defn/arglist parser does not recognize the `attr-map?` form between the docstring and the parameter vector, so it treats `{:added "1.0"}` as an arg pattern and rejects it.

Same construct appears on `meta-doc-macro` / `meta-doc-multi` in the same file (`{:added "3.0"}`, `{:added "2.0"}`).

## Expected

The attr-map is metadata; lowering should skip it (as the bytecode compiler does — the test asserts `(:added (meta (var meta-doc-fn)))` returns `"1.0"`) and lower the body normally.

## Repro

```
LG_STRESS_TIMEOUT_MS=15000 LG_STRESS_PASSES=1 ./lg scripts/ir-stress.lg corpus
# row: test/metadata_test.lg  meta-doc-fn  unsupported arg pattern: :added
```

nnunley:
## Summary
Clojure's defn grammar is `(defn name docstring? attr-map? [params*] body)`. The IR builder's `build-fn` did its own header parse that only understood the optional docstring, so a defn using the attr-map position — `(defn meta-doc-fn "defn doc" {:added "1.0"} [x] x)` — treated `{:added "1.0"}` as the arglist and failed to lower with `unsupported destructuring pattern: :added` (surfaced by the ir-stress corpus on `test/metadata_test.lg`).

The fix wires in `parse-defn-args-body`, which already implemented the full grammar but had **zero callers**. The attr-map is var metadata the bytecode compiler attaches; lowering skips it, as the issue specifies. Covers all four header shapes (doc+attr / attr-only × single / multi-arity).

Fixes #289

## Test plan
- [x] new `TestBuildFnDefnAttrMap` (pkg/ir): doc+attr, attr-only, multi-arity variants + doc-only/bare regression guards
- [x] runtime repro gone: `(ir.build/build-fn '(defn meta-doc-fn "defn doc" {:added "1.0"} [x] x))` builds
- [x] `make generate` + `make check-generated` (build.lg is a core source)
- [x] pre-push gate: `go test ./...`, Clojure compat suite (both backends), ir-stress coverage ratchet — all passed

### DIFF
--- a/pkg/ir/lisp_build_defn_test.go
+++ b/pkg/ir/lisp_build_defn_test.go
@@ -1,0 +1,40 @@
+/*
+ * Copyright (c) 2026 Norman Nunley, Jr <nnunley@gmail.com>
+ * Part of the let-go project; see CONTRIBUTORS for full list of authors.
+ * SPDX-License-Identifier: MIT
+ */
+
+package ir_test
+
+import (
+	"testing"
+
+	"github.com/nooga/let-go/pkg/vm"
+)
+
+// TestBuildFnDefnAttrMap covers the defn header grammar
+// (defn name docstring? attr-map? [params] body) in the IR builder (#289):
+// the attr-map position ({:added "1.0"} etc.) is var metadata the lowering
+// must skip, not an arg pattern. Before the fix, build-fn treated the map as
+// the arglist and threw "unsupported destructuring pattern: :added".
+func TestBuildFnDefnAttrMap(t *testing.T) {
+	ensureLoader()
+
+	cases := []struct {
+		label string
+		src   string
+	}{
+		{"doc+attr single-arity", `(defn meta-doc-fn "defn doc" {:added "1.0"} [x] x)`},
+		{"attr only single-arity", `(defn attr-fn {:added "1.0"} [x] x)`},
+		{"doc+attr multi-arity", `(defn meta-doc-multi "doc" {:added "2.0"} ([x] x) ([x y] y))`},
+		{"attr only multi-arity", `(defn attr-multi {:added "2.0"} ([x] x) ([x y] y))`},
+		{"doc only (regression guard)", `(defn doc-fn "doc" [x] x)`},
+		{"bare (regression guard)", `(defn bare-fn [x] x)`},
+	}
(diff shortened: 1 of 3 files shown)


nooga (approved):
Verified the fix logically and empirically:

- `parse-defn-args-body`'s cond correctly disambiguates all four defn header shapes (bare / doc-only / attr-only / doc+attr) for both single- and multi-arity forms. For multi-arity, `build-fn`'s `(cons args-vec body-forms)` correctly reconstitutes the full arity-clause list regardless of which branch fired — traced this by hand, it holds.
- The regular (non-AOT) compiler already supports defn attr-map (see `test/metadata_test.lg:103`), so this bug was isolated to the self-hosted IR lowering path in `pkg/rt/core/ir/build.lg`; the new Go test in `pkg/ir/lisp_build_defn_test.go` is the right test kind for that (matches the existing `lisp_*_test.go` convention in that package, reuses `buildLispIR`/`ensureLoader` rather than reinventing a harness) — no need for a separate `.lg` integration test since the language-level behavior was already covered.
- `pkg/rt/core_compiled.lgb` correctly did not change — `build.lg` feeds the gitignored `core_go_lowered` tree, not the core bytecode bundle — so the `generated.sums`-only diff is expected, and the generated-artifacts/gogen-diff/gold-differential CI checks confirm consistency.
- Ran the new test plus full `go test ./... -count=1` on the PR branch: all pass. CI is green.

No issues found. Good fix — small, targeted, well-tested.

PR #470, opened 2026-07-13
fix(compiler): evaluate top-level (do ...) subform-by-subform like Clojure

### LINKED ISSUE #195: eval: top-level (do ...) isn't evaluated subform-by-subform (diverges from Clojure) — breaks (do (require 'x) (x/f))
## Summary

let-go does not evaluate a **top-level `(do ...)`** subform-by-subform the way Clojure does. A top-level `do` is compiled as a single unit, so a namespace-qualified symbol in a later subform is resolved at **compile time — before** an earlier subform's `require` has run. This breaks the common `(do (require 'x) (x/f))` idiom (and any top-level `do` where a later form depends on the compile-time effect of an earlier one), and it diverges from Clojure.

## Repro (zero dependencies, two files)

`a.lg`:
```clojure
(ns a)
(defn hello [] :hi)
```

```console
# a) do-wrapped: FAILS
$ lg -e "(do (require 'a) (a/hello))"
error: Can't resolve a/hello in this context
  --> EXPR:1:18
   |
 1 | (do (require 'a) (a/hello))
   |                  ^^^

# b) the SAME two forms, un-wrapped (separate top-level forms): WORKS
$ printf "(require 'a)\n(println (a/hello))\n" > run.lg && lg run.lg
:hi
```

So it is specifically the `do` wrapper that forces single-unit compilation; let-go already evaluates *separate* top-level forms sequentially (run.lg works), it just doesn't extend that to the subforms of a top-level `do`.

> Note: `clojure.*` stdlib namespaces are embedded and resolve at compile time *without* `require`, so they **mask** this bug — `(do (require 'clojure.set) (clojure.set/union #{1} #{2}))` returns `#{1 2}` even though `(find-ns 'clojure.set)` is still `false`. You must use a **user namespace loaded from disk** (like `a` above) to reproduce.

## Expected (Clojure behavior)

Clojure's `eval` special-cases a top-level `do`: `clojure.lang.Compiler.eval` checks whether the form is a seq beginning with `do` and, if so, **recursively evaluates each subform in turn** (compile + eval one, then the next), returning the last. So earlier subforms' effects (a `require` loading a namespace, an `in-ns`, an `intern`) are visible when later subforms are compiled. This is also why a macro may expand into `(do (defines-X ...) (uses X ...))` at the top level.

```clojure
;; Clojure (and babashka) — works:
(do (require '[clojure.string :as s]) (s/upper-case "hi"))  ;=> "HI"
```

let-go gives `Can't resolve s/upper-case in this context` for the user-namespace equivalent.

## Impact

- **REPL / nREPL ergonomics:** require-then-use in one submission must be split into separate top-level forms; `(do (require ...) (use ...))` — the natural thing to type/send — fails. (Hit this driving the let-go nREPL from an MCP tool.)
- **Macros:** a macro that expands to a top-level `(do ...)` whose later forms depend on an earlier form's compile-time effect won't work the way it does in Clojure.

## Workaround

Submit the forms as **separate top-level forms** (not wrapped in `do`, `let`, `when`, or a fn body — those fail in both let-go and Clojure since they're genuinely one compiled unit). The only thing missing in let-go is Clojure's top-level-`do` special case.

## Version

`lg 1.9.0 (9cb2167)`, macOS arm64.

nnunley:
## Summary
Clojure's `Compiler.eval` special-cases a top-level `(do ...)`: each subform is compiled **and evaluated** in turn, so an earlier subform's compile-time effect (a `require` loading a namespace, an `in-ns`, an `intern`) is visible when a later subform compiles. let-go compiled a top-level `do` as one unit, so the common REPL idiom broke:

```clojure
(do (require 'a) (a/hello))   ; before: Can't resolve a/hello in this context
                              ; after:  :hi
```

`CompileMultiple` already compiles+runs each *separate* top-level form sequentially; this change splices a top-level `(do ...)` into that same sequence, recursively (nested top-level `do`s splice too). Non-top-level `do` is untouched — `(println (do (require 'a) (a/hello)))` still fails, exactly as it does in Clojure. Empty `(do)` falls through to the normal path and evaluates to nil; value semantics (`(do 1 2 3)` → 3) are unchanged.

Note: like Clojure, the check is on the literal form. Clojure also macroexpands before the check (a macro expanding to a top-level `do` gets spliced); that extension is left for a follow-up if wanted — it requires hoisting the compiler's inline macro-invocation out of `compileForm`.

Fixes #195

## Test plan
- [x] new `test/top_level_do_test.lg` + `test/topdo_helper.lg` (disk-loaded ns, so embedded-clojure.* masking doesn't hide the bug): require-then-use in one top-level do, nested do splice, value/nil semantics
- [x] `make generate` → bundle byte-identical (compiler change does not perturb codegen); `make check-generated` OK
- [x] pre-push gate: `go test ./...`, Clojure compat suite (both backends), ir-stress ratchet — all passed

### DIFF
--- a/pkg/compiler/compiler.go
+++ b/pkg/compiler/compiler.go
@@ -112,13 +112,23 @@ func (c *Context) CompileMultiple(reader io.Reader) (*vm.CodeChunk, vm.Value, er
 	chunk := vm.NewCodeChunk(c.consts)
 	var result vm.Value = vm.NIL
 	compiledForms := 0
-	for {
-		o, err := r.Read()
-		if err != nil {
-			if isErrorEOF(err) {
-				break
+	var evalTopForm func(o vm.Value) error
+	evalTopForm = func(o vm.Value) error {
+		// Clojure's Compiler.eval special-cases a top-level (do ...): each
+		// subform is compiled AND evaluated in turn, so an earlier subform's
+		// compile-time effect (a require loading a namespace, an in-ns, an
+		// intern) is visible when a later subform compiles. This is what
+		// makes (do (require 'x) (x/f)) work at the REPL (#195). An empty
+		// (do) falls through to compileForm and evaluates to nil.
+		if lst, ok := o.(*vm.List); ok && lst.Next() != nil {
+			if first, isSym := lst.First().(vm.Symbol); isSym && first == "do" {
+				for s := lst.Next(); s != nil; s = s.Next() {
+					if err := evalTopForm(s.First()); err != nil {
+						return err
+					}
+				}
+				return nil
 			}
-			return nil, result, err
 		}
 		if compiledForms > 0 {
 			chunk.Append(vm.OP_POP)
@@ -126,10 +136,10 @@ func (c *Context) CompileMultiple(reader io.Reader) (*vm.CodeChunk, vm.Value, er
 		formchunk := vm.NewCodeChunk(c.consts)
 		c.chunk = formchunk
 		c.resetSP()
-		err = c.compileForm(o)
+		err := c.compileForm(o)
 		c.chunk.SetMaxStack(c.spMax)
 		if err != nil {
-			return nil, result, err
+			return err
 		}
 		chunk.AppendChunk(formchunk)
 
@@ -143,9 +153,22 @@ func (c *Context) CompileMultiple(reader io.Reader) (*vm.CodeChunk, vm.Value, er
 		result, err = f.RunProtected()
 		vm.ReleaseFrame(f)
 		if err != nil {
-			return nil, result, err
+			return err
 		}
 		compiledForms++
+		return nil
+	}
+	for {
+		o, err := r.Read()
+		if err != nil {
+			if isErrorEOF(err) {
+				break
+			}
+			return nil, result, err
+		}
+		if err := evalTopForm(o); err != nil {
+			return nil, result, err
+		}
 	}
 
 	c.chunk = chunk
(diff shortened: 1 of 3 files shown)


nooga (approved):
Reviewed the diff against pkg/compiler/compiler.go and the two new test/*.lg files.

Correctness: the recursive `evalTopForm` closure splices a literal top-level `(do ...)` into the same compile-one-subform/eval-it/append-to-chunk loop `CompileMultiple` already uses for separate top-level forms, which is exactly how Clojure's `Compiler.eval` treats it. Empty `(do)` (no `Next()`) correctly falls through to the normal `compileForm` path, so `(do)` still evaluates to nil. The shared `compiledForms` counter and `OP_POP` insertion are threaded correctly through the recursion, so intermediate values are still discarded and only the final top-level value survives — verified by the `(do 1 2 3)` → 3 and `(assert (nil? (do)))` cases in the new test. Non-top-level `do` (e.g. inside a fn body) is untouched, matching Clojure's behavior and the PR's stated scope.

Test coverage: correctly uses a `.lg` integration test (test/top_level_do_test.lg) per CLAUDE.md's guidance for new language-level semantics, and the disk-loaded `topdo_helper.lg` namespace is the right way to avoid the embedded-clojure.* masking the author called out. This follows the existing convention of assertion-free helper `.lg` files living alongside `_test.lg` files in test/ (e.g. fib.lg, hello.lg already do this), so it being picked up as its own vacuous subtest by test/language_test.go's walk is expected, not a regression.

Minor/non-blocking: a Go-level unit test in pkg/compiler directly exercising `CompileMultiple` on a `(do (def a 1) (do 2 3))`-shaped input would pin down the splicing/stack-discipline logic independent of runtime `require` behavior, but the existing `.lg` test already covers the observable behavior end-to-end, so this is a nice-to-have rather than a gap.

No correctness, convention, or CI issues found. CI is green across build/lint/gold-differential/generated-artifacts/gogen-diff; bench jobs are skipped by design for this PR, not failing.

PR #472, opened 2026-07-13
feat(rt): java.lang exception class hierarchy and constructors

### LINKED ISSUE #471: java.lang exception classes with real constructors (expands #451 per review)
Follow-up to the #451 review discussion: implement exceptions properly instead of the compile-only stub, with java.lang.* naming as agreed.

## Gap

let-go has one exception value (`ex-info`) and no exception classes. `(Exception. msg)` resolves through #451's stub but fails when called, `(instance? Exception x)` cannot be expressed, and a failed `assert` throws a bare string.

## Change

- `vm.ExceptionClass`, a ValueType with a parent pointer, and a 15-class java.lang.* hierarchy rooted at `java.lang.Throwable` (Error, AssertionError, Exception, IOException, InterruptedException, RuntimeException and its common subclasses).
- Real constructors for every class: `(Exception. msg)`, `(RuntimeException. msg cause)`, bare and qualified spellings. Replaces and removes the #451 stub.
- `instance?` walks registered type ancestry, so subclass values answer true for parent classes. `ex-info` values are `ExceptionInfo` under `RuntimeException`; `(type (ex-info ...))` output is unchanged.
- VM runtime errors are tagged `java.lang.Exception` (previously they printed as ExceptionInfo).
- `assert` throws `(AssertionError. msg)`.

Ready on `abogoyavlensky/let-go` branch `clojure-exception-classes` with regression tests in `test/exception_classes_test.lg`. PR to follow.

## Merge order

First of three: `fix-finally-abnormal-exit` is independent, and `typed-catch-dispatch` stacks on both. Nothing dispatches on class yet in this PR; typed catch dispatch is the third one.

abogoyavlensky:
Add `vm.ExceptionClass`, a `ValueType` with a parent pointer, and canonical `java.lang.*` instances rooted at `java.lang.Throwable`. ExInfo carries an optional class tag; tagged values print their class and report it from `Type()`, while plain ex-info values keep `let-go.lang.ExceptionInfo`.

Register each class under qualified and bare names with X., ->X constructor spellings accepting `[]`, `[msg]`, and `[msg cause]`. This replaces the compile-only Exception. stub (#451) with real constructors, per maintainer feedback on that PR; the stub and its fails-loudly test are removed.

`instance?` now walks registered type ancestry when the exact match fails, so subclass values answer true for parent `classes. ExInfoType` chains through `RuntimeException`. Wrapped Go runtime errors are tagged `java.lang.Exception`: their type name changes from
`let-go.lang.ExceptionInfo`, and `(instance? ExceptionInfo e)` is now false for them. Catch clauses do not dispatch on class yet; that lands in the follow-up typed-catch-dispatch change, which this classification exists to serve.

`assert` now throws `(AssertionError. msg)` instead of a bare string, so failed assertions are catchable by class. The macro expands to the `->AssertionError` factory spelling, which syntax-quote qualifies to `core/->AssertionError`: hygienic against caller shadowing, while the dotted spelling would qualify into an unresolvable form.

Resolves: #471


nnunley (approved):
Reviewed as part of the typed-catch stack (#472→#474→#476). Class descriptors as data (name+parent, single ExInfo value representation) is the right shape; hierarchy integration via directTypeParents keeps instance?/ancestry honest.

PR #474, opened 2026-07-13
fix(vm): run finally on abnormal exits and propagate the exception

### LINKED ISSUE #473: finally is skipped or the exception swallowed on abnormal exits
Two try/finally bugs in the bytecode VM, found while designing typed catch dispatch (the no-match rethrow path steps on both).

## Repro

```clojure
;; 1. finally-only try SWALLOWS the exception: prints "outer boom"? No:
(try
  (try (throw "boom") (finally (println "F")))
  (catch e (str "outer " e)))
;; actual: F is printed, then the try RETURNS "boom" as its result;
;; the outer catch never runs.

;; 2. a throw from a catch body SKIPS the finally:
(try
  (try (throw "x") (catch e (throw "y")) (finally (println "F")))
  (catch e e))
;; actual: returns "y" but F never prints.
```

Clojure guarantees finally runs on every exit. The IR/lowered engine already gets both cases right (its `:try` orchestration always invokes finally and propagates afterward); only the bytecode VM diverges.

## Change

The handler survives catch entry when a finally exists, a pending-throw slot parks the in-flight error while the (single, shared) finally block runs, and a new `OP_FINALLY_END` opcode resumes unwinding. `handleError` becomes the one dispatch path (the INVOKE/TAIL_CALL error paths had four inline copies). The bundle is regenerated because the opcode-set signature changes (#443).

Behavior change: code relying on a finally-only try swallowing exceptions now sees them propagate.

Ready on `abogoyavlensky/let-go` branch `fix-finally-abnormal-exit`, tests in `test/try_finally_test.lg`, parity-check agrees on both engines. PR to follow.

## Merge order

Independent of the exception-classes PR; merge in any order. `typed-catch-dispatch` stacks on both and should land last.

abogoyavlensky:
Two abnormal-exit bugs in the bytecode VM's try/finally:

- `(try (throw x) (finally f))` with no catch ran the finally but then returned the thrown value as the try's result, swallowing the exception.
- A throw from inside a catch body skipped the finally entirely.

The compiled layout keeps a single finally block shared by both paths, ending in the new `OP_FINALLY_END`. `TRY_PUSH` encodes "no catch" as a zero catch offset, and the handler survives catch entry whenever a finally follows (a `TRY_POP` retires it when the catch completes normally). handleError walks the handler stack with three rules: an armed catch consumes the error; otherwise a present finally parks the error in the handler's pending slot, normalizes the stack with a placeholder, and enters the shared finally; a throw escaping such a finally replaces the in-flight exception. `OP_FINALLY_END` carries the finally's own address so it resumes unwinding only for its own abnormally-entered handler, not for an outer try mid-finally.

The `INVOKE/TAIL_CALL` error paths carried four inline copies of the old single-shot dispatch; they now share `handleError`, which also fixes them jumping to a consumed catch.

The IR/lowered engine already had these semantics (its :try orchestration always runs finally and propagates afterward), so this aligns the bytecode VM with it; make parity-check agrees on both suites. The bundle is regenerated because the opcode-set signature changed with `OP_FINALLY_END`.

Behavior change: code that relied on a finally-only try swallowing exceptions now sees them propagate, matching Clojure.

Resolves: #473


nnunley (approved):
Approving with the stack — finally-on-abnormal-exit is a prerequisite for #476's no-match rethrow.

nnunley:
Superseded by #478, which carries this commit rebased past the #472 squash (authorship preserved) — this branch went conflicting on the regenerated bundle artifacts after #472 landed as a squash. Will close when #478 merges.

nnunley (approved):

PR #476, opened 2026-07-13
Typed catch dispatch

### LINKED ISSUE #475: Catch clauses ignore the exception class; multiple clauses concatenate bodies
Catch clauses ignore the exception class, and multiple clauses silently concatenate their bodies.

## Repro

```clojure
(try (throw (ex-info "a" {}))
     (catch ExceptionInfo e :first)
     (catch e :second))
;; actual: :second - the class symbols are discarded and every clause
;; body runs concatenated, so the last value wins.
```

## Change

Both front ends (bytecode `tryCompiler`, IR `parse-try`) desugar typed and multiple clauses into one handler with class dispatch in source order and a rethrow on no match. Semantics live in one `core/catch-matches?` helper:

- classes resolve at dispatch time, so a JVM-only class let-go does not model (integrant's `(catch java.io.FileNotFoundException _)`) compiles into a never-matching clause instead of failing the namespace;
- `Throwable` catches ANY thrown value, plain strings included, so the conventional `(catch Throwable e ...)` stays a catch-all; `instance?` itself stays honest;
- other classes match by type identity or registered ancestry.

Also tightens clause parsing: a clause is typed only when the first token is qualified or uppercase-simple (the former 3-token fallback would misread the bare `(catch e foo bar)` under dispatch).

Behavior change: typed catches now filter. A known limitation is called out in the commit: class symbols resolve against core, so a same-ns deftype or aliased class in a catch never matches - follow-up material.

Ready on `abogoyavlensky/let-go` branch `typed-catch-dispatch`, tests in `test/catch_dispatch_test.lg` (plus `test/catch_class_test.lg` rewritten to the new semantics), parity-check agrees on both engines. Verified against clojure.tools.cli, medley, dependency, bond, and integrant under lgx.

## Merge order

Last of three: stacked on `clojure-exception-classes` (classes to dispatch on) and `fix-finally-abnormal-exit` (the no-match rethrow must run finally).

abogoyavlensky:
Resolves: #475


nnunley:
Superseded by #478, which carries this commit rebased past the #472 squash (authorship preserved) — this branch went conflicting on the regenerated bundle artifacts after #472 landed as a squash. Will close when #478 merges.

nnunley (approved):

PR #479, opened 2026-07-13
feat(ir): legalize cross-block refs and run lambda-lifted fns natively in gogen_ir

nnunley:
Two coupled changes that let lambda-lift templates execute natively instead of self-trampolining through `rt.CachedVarFn` back to bytecode.

## 1. `ir.passes.legalize` — cross-block ref repair

`build.lg` emits cross-block direct refs for shapes its construction-time threading misses (e.g. the `:try` lowering), which the build validator rejects — forcing legal-semantics fns onto the bytecode trampoline. The new pass threads those values through block args (respecting the `:branch-if` symmetric-args invariant) before validation, on demand: fns with no cross-block refs are untouched.

## 2. Lambda-lift native activation

- **Correct ns on sibling refs**: `rewrite-site!` now qualifies the lifted-sibling `:load-var` with the current ns — bare symbols default to `"core"` in lower-go, which dangles for any other ns, and lifted vars have **no bytecode fallback** (lambda-lift runs only on the `:go` path).
- **Registration**: siblings and the main carry their full lowering result through the multi-fn-template flatten, so each registers as a native override and enters the direct-call registry. Intra-ns callers now emit direct Go calls to the main — **73 fns across the lowered tree go native** (e.g. `AnalyzeNsForms` callers call it directly instead of trampolining).
- **Uniform-ABI siblings**: siblings lower through a typeinfer-free pass list, giving a uniform `vm.Value` ABI by construction. They are only ever invoked boxed through their interned var, and a typed signature (e.g. a `bool`-returning predicate) would silently fail override registration and leave the main body dereferencing an uninterned var — a guaranteed runtime nil-deref.
- **Guard**: if any sibling still is not registrable, the whole form falls back to bytecode; a native main with a dangling sibling `LookupVar` is never emitted.
- **Variadic inners are not lifted**: overrides cannot adapt variadic signatures, so variadic capture-free inners stay inline fn-templates lowered as plain Go func-lits (the pre-lift path).

## Invariant

Every `rt.LookupVar(ns, "<name>__liftedN")` in the generated tree has a matching `"<name>__liftedN": __gogen_wrap` registration in that package's init — verified tree-wide (134 references, 0 dangling).

## Verification

- `make parity-full` green: jank suite + ir-stress lower-go + ir-stress ir-compile, untagged vs `-tags gogen_ir` all parity-identical (738/738, matching failure buckets)
- `pkg/ir` (367) and `pkg/rt` (196) suites pass under both tags
- New tests: `lisp_legalize_test.go`, `lisp_lambda_lift_ns_test.go`, `lisp_lambda_lift_register_test.go` (incl. typed-sibling case), `lisp_lambda_lift_main_test.go`

Note: `TestUnrollBytecodePipelineAll` fails under `gogen_ir` **pre-existing** (A/B-verified at clean main with identical op counts) — native core-fn dispatch deflates the VM-op ratio the test asserts; not introduced here.

### DIFF
--- a/pkg/ir/lisp_constfold_test.go
+++ b/pkg/ir/lisp_constfold_test.go
@@ -50,7 +50,7 @@ func ensureLoader() {
 			"ir.passes.mutability", "ir.passes.cse",
 			"ir.passes.typeinfer", "ir.passes.infer-arg-types",
 			"ir.passes.licm", "ir.passes.lambda-lift", "ir.passes.inline", "ir.passes.fusion",
-			"ir.passes.liveness", "ir.passes.blockarg", "ir.passes.cleanup", "ir.passes.pipeline", "ir.dump", "ir.dominance", "ir.lower-go"} {
+			"ir.passes.liveness", "ir.passes.blockarg", "ir.passes.cleanup", "ir.passes.legalize", "ir.passes.pipeline", "ir.dump", "ir.dominance", "ir.lower-go"} {
 			if res.Load(ns) == nil {
 				panic("namespace failed to load: " + ns)
 			}
--- a/pkg/ir/lisp_lambda_lift_main_test.go
+++ b/pkg/ir/lisp_lambda_lift_main_test.go
@@ -1,0 +1,64 @@
+/*
+ * Copyright (c) 2026 Norman Nunley, Jr <nnunley@gmail.com>
+ * Part of the let-go project; see CONTRIBUTORS for full list of authors.
+ * SPDX-License-Identifier: MIT
+ */
+
+package ir_test
+
+import (
+	"regexp"
+	"strings"
+	"testing"
+
+	"github.com/nooga/let-go/pkg/vm"
+)
+
+// Part 3 of the lambda-lift native-safety work: the MAIN fn of a lambda-lift
+// multi-fn-template must carry its full lowering result through the flatten,
+// exactly like its lifted siblings (Part 2). Otherwise the main is stripped to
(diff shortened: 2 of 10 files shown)


nooga (approved):
Traced both passes by hand (ir.passes.legalize's block-arg threading + branch-if symmetric-args handling, and the lambda-lift multi-fn-template flatten's :lifted? vs :multi-arity? tagging) and didn't find a correctness issue. The override-eligibility guard (whole form falls back to bytecode if any sibling fails to register) correctly closes the dangling-LookupVar failure mode described in the PR body, and the new Go unit tests in pkg/ir (lisp_legalize_test.go, lisp_lambda_lift_{ns,register,main}_test.go) are the right test kind for a compiler/IR-internal change per CLAUDE.md and cover the interesting edge cases (typed-sibling-still-registers, ns-qualification of lifted refs, main registered + direct-callable).

One non-blocking note: the PR frames this as enabling native (vs. trampolined) execution for 73 fns, but no before/after benchmark numbers are included and the bench CI job was skipped for this PR. Given CLAUDE.md's guidance to cite bench numbers for performance-sensitive changes, it'd strengthen the record to add a quick hyperfine/internal-bench comparison in a follow-up, even though the change is primarily a correctness fix (previously-silently-bytecode-trampolined fns now actually go native) rather than a tuning change.

No blocking feedback from other reviewers to reconcile (reviews/comments are both empty). CI is fully green including the gogen_ir-tag-specific checks (gogen-diff, gold-differential, generated-artifacts) that would catch a stale lowered-tree regen.

Issue #482, opened 2026-07-13
VM: a throw during top-level lazy-seq realization escapes as a Go panic (thrownPanic) instead of a rendered error


nnunley:
Companion to #481 (which is how it was found, via @seancorfield's HoneySQL experiment). Any let-go-level throw that first surfaces during lazy-seq realization at the top level crashes the process with a raw Go panic dump:

```
user=> (defprotocol Y :extend-via-metadata true (foo [x]))
panic: (*vm.thrownPanic) 0x...
github.com/nooga/let-go/pkg/vm.(*LazySeq).sval  pkg/vm/lazy_seq.go:50
github.com/nooga/let-go/pkg/vm.(*LazySeq).seq   pkg/vm/lazy_seq.go:67
...
```

Generic repro without defprotocol — any lazily-deferred throw realized outside a protected frame:

```clojure
(first (map (fn [x] (throw "boom")) [1]))   ; fine inside eval (protected)
;; but when realization happens in printer/require/macroexpansion plumbing
;; outside RunProtected, the thrownPanic escapes to Go's runtime.
```

`LazySeq.sval` (pkg/vm/lazy_seq.go:50) panics with `thrownPanic` expecting an enclosing recover; the REPL/exec top level doesn't recover around every place a lazy seq can be forced (printing results, macroexpansion products, require machinery). Whatever the fix shape — recover-at-boundary in the REPL/driver, or making sval return errors on the paths that can't panic-safely — a user error must never produce a Go stack dump.

The user-visible bar: `(defprotocol Y :extend-via-metadata true (foo [x]))` should print a normal `error: ...` line (with the defprotocol form's location once #481's eager parsing lands) and leave the REPL alive.

PR #488, opened 2026-07-14
fix(core): parse defprotocol options eagerly

### LINKED ISSUE #481: defprotocol: option keywords (:extend-via-metadata) crash — lazy sig-parsing lets the error escape as a Go panic / masked require error
Reported by @seancorfield trying HoneySQL on lgx ([let-go branch](https://github.com/seancorfield/honeysql/tree/let-go)). Two symptoms, one root cause.

## Repro

```clojure
user=> (defprotocol Y :extend-via-metadata true (foo [x]))
panic: (*vm.thrownPanic) 0x...
github.com/nooga/let-go/pkg/vm.(*LazySeq).sval  pkg/vm/lazy_seq.go:50
```

And the same form inside a required namespace produces the famously unhelpful:

```
user=> (require 'badproto)
error: first expected Seq
  --> <embedded:core>:2219:41
  at first (<embedded:core>:2219:41)
  at seq (<embedded:core>:369:24)
  at require (REPL:1:1)
```

## Root cause

`defprotocol` (pkg/rt/core/core.lg:2210-2225) handles an optional docstring but not Clojure's **option keyword pairs** (`:extend-via-metadata true`, also `:on`/`:on-interface`). The keyword flows into `sigs`, and `(map first sigs)` calls `(first :extend-via-metadata)` → throw. Because `method-names`/`method-defs` are **lazy** `map`s, the throw escapes the defprotocol expansion and detonates wherever the seq is first realized:
- at the REPL top level → un-recovered **Go panic** through `LazySeq.sval` (filed separately — that's a VM robustness bug for any error inside lazy realization),
- inside `require`'s machinery → an error attributed to `<embedded:core>`, masking the user's file entirely (this is what made Sean ask "is that the error you get whenever a namespace cannot be required for any reason at all?").

Note: require error reporting is otherwise fine — a plain CompileError in a required file reports `failed to load <file>: ...` correctly. The masking is specific to this lazy escape.

## Fix shape

1. Parse option keyword pairs off the front of `sigs` (accept + record `:extend-via-metadata`; actually honoring it in protocol dispatch can be a follow-up, but parsing must not crash).
2. Make sig parsing **eager** (`mapv`) and validate each sig is a seq whose head is a symbol, so a malformed defprotocol errors AT the defprotocol form with its own source location.

Verified against main (79b96e5-era behavior still present on current HEAD).

nnunley:
## Summary
- parses defprotocol option pairs before signature realization
- preserves ordinary signatures and reports malformed signatures as user errors
- adds focused eager-error regressions

## Verification
- `make check-generated`
- `go test ./...`

Fixes #481


nooga (commented):
Logic looks correct and matches the #481 repro/root-cause exactly — options are now parsed eagerly (mapv) before method-name extraction, malformed option pairs and non-symbol sig heads throw at the defprotocol form instead of escaping through a lazy `map` later. Test coverage is the right kind (test/defprotocol_options_test.lg, language-level macro fix) and covers the three cases called out in the issue. CI is green including the full test suite and generated-artifacts checks.

One blocker before merge: GitHub reports this PR as CONFLICTING against main. Locally `git merge-tree --write-tree origin/main <branch>` resolves cleanly (exit 0) thanks to this repo's `merge=lgb` / `merge=sums` custom drivers, which regenerate pkg/rt/core_compiled.lgb and generated.sums automatically — but GitHub's merge UI doesn't have those drivers, so it sees a raw binary/digest diff against main (churned by the already-merged #476) and flags it. Please rebase onto current main and push (`make generate` will resolve the regenerated artifacts via the same drivers used locally) so the PR shows mergeable again.

No other correctness or convention issues found.

mparrett (approved):
Verified against the diff — eager parse + sig validation match the root cause, options land in var metadata, artifact conflict resolved, CI green. Crash/masking fully addressed here; #482/#487 track the rest. LGTM.

nooga (approved):
Verified: the pushed commit ("Merge main; make generate") directly resolves the artifact conflict I flagged — mergeable/mergeStateStatus now shows CLEAN. Full test suite plus the generated.sums staleness check pass on the PR head. No other concerns.

PR #489, opened 2026-07-14
fix(core): accept return hints on arity vectors

### LINKED ISSUE #484: Return-type hints on defn arity vectors break compilation (two distinct failures)
From @seancorfield's HoneySQL run ([let-go branch](https://github.com/seancorfield/honeysql/tree/let-go)) — honey.sql.util's optimized `str` had to be `#_`'d out entirely. Clojure attaches `^String` before an arglist as `:tag` metadata on the params vector; let-go's defn/fn parsing chokes on it two different ways (both verified on current main):

**Multi-arity** — the metadata map's contents leak into a let-binding position:
```clojure
user=> (defn g (^String [] "") (^String [^Object a] "x"))
error: let binding name must be a symbol: :tag
```
This is the exact error that blocked `honey.sql.util` (`failed to load src/honey/sql/util.cljc: ... let binding name must be a symbol: :tag`).

**Single-arity** — a different explosion:
```clojure
user=> (defn h ^String [] "y")
error: ExecutionError: calling apply
        caused by ExecutionError: calling apply*
        caused by ExecutionError: calling seq
        caused by seq expected Seqable
```

Param-position hints (`[^Object a]`) and `^{:tag String}` map-form metadata on a binding both already work — it's specifically metadata attached to the ARGLIST VECTOR (return-type position) that breaks, in both the single-arity (`defn name ^T [args]`) and per-clause (`(^T [args] body)`) spots.

Expected: tolerate (preserve if cheap, else drop) `:tag` metadata on arity vectors. Every perf-conscious Clojure library type-hints returns, so this blocks a lot of unmodified code.

Note the desugaring-consolidation angle: defn header parsing is duplicated per frontend (#477 §generalization, and #289 was the same disease in the attr-map position) — whatever fix lands should cover the IR builder's `build-fn`/`expand-fn-args` path too, not just the bytecode compiler.

nnunley:
## Summary
- accepts return-type metadata on single and multi-arity vectors
- fixes both bytecode and IR/Go-lowering paths
- preserves source spans through lowering

## Verification
- `make check-generated`
- `go test ./...`

Fixes #484


nooga (approved):
Verified locally (worktree at PR head): `go build ./...`, full `go test ./... -count=1`, and `go generate ./pkg/rt/` all clean/green (generated.sums matches the diff, so nothing stale). The fix — stripping the reader's `(with-meta [args] {:tag T})` wrapper before the vector?/list? shape checks in `core.lg`'s defn/fn macro, `pkg/rt/core/ir/build.lg`'s IR builder, and the multi-arity discrimination logic — correctly matches the root cause in #484: metadata was leaking into let-binding position and breaking single-arity parsing. Manually reproduced both original repros plus mixed-arity/nested-arity variants; all now parse correctly with no metadata leakage.

Test coverage is right per CLAUDE.md: `.lg` integration tests (test/return_type_hint_test.lg, test/return_type_hint_ir_test.lg) for the language-level fix, plus a Go unit test in pkg/ir/lisp_lower_go_test.go for the IR-lowering source-span-preservation detail.

LGTM.

PR #490, opened 2026-07-14
fix(core): preserve dotdot member chains

### LINKED ISSUE #485: (.. obj (method)) compiles to unresolvable ->-> — constructor-shorthand rewrite cascades on the .. symbol
From @seancorfield's HoneySQL run. He hit `caused by CompileError: Can't resolve ->-> in this context` with no such symbol anywhere in his code (he suspected his `or-fn`, but that compiles fine — the or-fn errors in his transcript were a cascade from honey.sql.util failing to load).

## Repro + root cause (verified on current main)

```clojure
user=> (.. "abc" (length))
error: Can't resolve ->-> in this context
```

`compileForm`'s record-constructor shorthand — any head symbol ending in `.` rewrites `(Name. args...)` → `(->Name args...)` — doesn't exclude `..`:

1. `(.. x (m))` → head `..` ends in `.` → strip the dot, prepend `->` → `(->. x (m))`
2. `->.` ALSO ends in `.` → `(->-> x (m))` → "Can't resolve ->->".

## Fix shape

- Exclude `..` (and generally any head that is only dots / starts with a dot or `->`) from the ctor rewrite — a real ctor shorthand is `Name.` with a non-empty class-shaped stem.
- Then decide what `..` should DO: if the `..` threading macro doesn't exist yet, add it (it's core Clojure, a trivial macro over `.`) or at minimum produce "Can't resolve .." so the error names the actual form.

The mangled-symbol error is a nasty debugging experience — the reported name appears nowhere in user code, which is exactly what burned Sean.

nnunley:
## Summary
- prevents constructor-shorthand rewriting from cascading through `..`
- preserves source metadata through the macro, compiler, and IR pipeline
- tests Lisp and Go-lowering shapes

## Verification
- `make check-generated`
- `go test ./...`

Fixes #485


nooga (commented):
Verified locally against the PR head (be969bd3) in a worktree: `go build ./...`, `go test ./... -count=1`, and `make check-generated` all pass, and manual exercise of `(.. "abc" (replace "b" "x") (replace "x" "z"))`, `macroexpand` of a `..` form, and record ctor shorthand (`Foo.`) all behave as expected. The fix (excluding `.`/`->`-prefixed symbols from the ctor-shorthand rewrite, and excluding `..` from the single-dot member-call branch) matches the root cause in #485, and the new `..` macro in core.lg is a faithful port of Clojure's own implementation. Test coverage is appropriate per CLAUDE.md: `test/dotdot_test.lg` for the new macro (chaining, macroexpand shape, and a regression test that real `Name.` ctor shorthand still works), plus a Go unit test in `pkg/ir/lisp_lower_go_test.go` for source-span preservation through the IR lowering pipeline. The `-copy-form-source!` plumbing through compiler.go/lang.go/pipeline.lg is a clean, minimal, no-op-safe way to keep FormSource spans alive across macro/IR rewrites.

One thing to check before merging: `gh pr checks 490` shows only the Perf-PR/Perf-WASM bench workflows completed (both skipped) — the actual `Go` workflow (lint, build, full test matrix, gogen-diff, generated-artifacts) and `CodeQL` never ran on this commit at all, unlike neighboring PRs (#495, #496) which show full green runs. Worth confirming those actually fire and pass on GitHub (a re-push or manual re-run may be needed) rather than relying only on local verification, since the current PR checks list gives no real signal either way.

nooga (approved):
Verified: diff is unchanged since the earlier review (SHA only moved from a rebase onto fix/484-arity-return-hints), builds and full test suite pass cleanly on the worktree head, including the new dotdot regression tests. One non-blocking note: Go/CodeQL haven't run yet because this PR's base isn't main — please confirm they're green (or retarget) before merging into main.

PR #491, opened 2026-07-14
fix(vm): recover top-level lazy realization errors

### LINKED ISSUE #482: VM: a throw during top-level lazy-seq realization escapes as a Go panic (thrownPanic) instead of a rendered error

nnunley:
## Summary
- recovers user throws during lazy top-level realization and rendering
- prevents raw Go panic and stack leakage in REPL and `-e`
- tests REPL and expression boundaries

## Verification
- `make check-generated`
- `go test ./...`

Fixes #482


nooga (commented):
Nice, targeted fix — the two-part recovery (RecoverPanic around Compile/CompileMultiple for macroexpansion-time panics, SafeString around the print-time String() calls) matches the two escape hatches issue #482 identified, and the new tests (pkg/compiler/lazy_macro_error_test.go, pkg/vm/safe_string_test.go, test/e2e/repl_lazy_error_test.go) exercise the real bug rather than just the plumbing.

Two things before merge:

1. `pkg/nrepl/server.go:293` (`valStr = val.String()`) has the identical unrecovered-panic hazard this PR fixes for the CLI REPL/-e path — a lazy throw forced while rendering the eval result will still crash the whole nREPL server process (there's no recover() anywhere in that file). Worth swapping in `vm.SafeString` there too, since an nREPL server crash is more consequential than a one-shot CLI process dying.

2. Minor: in `pkg/vm/vm.go` around the OP_ADD/OP_SUB block (~line 1110), the edited comment line (`// These avoid NativeFn.Invoke, interface boxing, and RecoverPanic.`) lost its leading tabs relative to the sibling comment lines directly above/below it — cosmetic but easy to fix.

3. Process check: looking at the check-runs for this PR's head SHA (c88b87f), only `Perf PR`/`Perf WASM` ran and both are `skipped` (they're path-filtered and unrelated to this change's paths). The main `Go` workflow (build + `go test ./...` + clojure-test-suite) doesn't appear to have run on this commit at all, even though `go.yml` has no path filter and fires unconditionally on `pull_request` → `main`. Can you confirm the full suite actually ran green somewhere, or trigger it, before merging?

mparrett:
@nnunley 

The change itself looks correct and well-tested. The two-part recovery (RecoverPanic around Compile/CompileMultiple for macroexpansion-time realization, SafeString around the print-time String() calls) maps cleanly onto the two escape hatches #482 identified, and the new tests exercise the real bug rather than the plumbing.

Could we please address the nrepl result rendering hazard in this PR?

**`pkg/nrepl/server.go:293` — `valStr = val.String()` is unguarded, and there's no `recover()` anywhere in that file.** A lazy throw forced while rendering an eval result will crash the whole nREPL server process, killing every session — worse than the one-shot CLI exit this PR fixes. Result rendering is squarely in this PR's scope, so fixing CLI/`-e` but not nREPL ships an inconsistent partial fix, and the remedy is the same one-liner (`vm.SafeString(val)`, routing the error into the existing error-response branch). I'd rather see it here than as a follow-up.

Two smaller notes:

- `pkg/vm/errfmt.go:126` still filters on the `recoverThrownPanic` string, but after the rename that identifier only survives in the `errfmt_origin_test.go` fixture. Real stacks now show `RecoverPanic`. Worth updating the fixture so the test reflects actual output.
- Process: this is stacked on `fix/485-dotdot-member-chains` (#490, still open), so it can't land on `main` until that merges or this rebases. And per the check-runs, confirm the full `Go` workflow ran green on the head SHA — only the path-filtered perf jobs show up, both skipped.

nnunley:
Both addressed in 75388535:

1. `pkg/nrepl/server.go` eval-result rendering now goes through `vm.SafeString`; a render-time lazy throw becomes an nREPL `err` response (+ `done`) instead of killing the server process.
2. The vm.go comment indent was already re-aligned by an interim gofmt pass during the stack rebase — all three lines sit at two tabs and `gofmt -l` is clean on the file.
3. On the CI question: the head you reviewed (c88b87f) predated today's rebase; this push (and the rebase before it) trigger fresh `Go` workflow runs on the current head, so the full suite result will be attached to the PR checks now. Locally: build + pkg/nrepl + pkg/vm green, `make check-generated` clean.

mparrett:
The nREPL follow-up (7538853) closes the last substantive gap, and cleanly — it routes rendering through `vm.SafeString`, emits a proper `err` response, and sends `status: done` before returning so the client isn't left hanging. Code looks ready to me.

The only thing left is procedural, and both gates collapse to one root cause: this is stacked on #490 (still open), so its base isn't `main` — which means the `Go` workflow (build + `go test ./...` + clojure-test-suite) has never fired here. I confirmed zero workflow runs on head `7538853`. Once #490 lands, the base auto-retargets to `main`, the full suite finally runs, and if green this is good to merge.

Non-blocking: `errfmt_origin_test.go` still hardcodes the old `recoverThrownPanic` name in its fixture (and `errfmt.go` filters both) — cosmetic, fine to clean up in passing or leave.

mparrett (approved):
Approving. The nREPL rendering follow-up (7538853) closed the last substantive gap, and the procedural gate is cleared: #490 merged, this PR now bases on main, and all checks are green at 5101563.

nooga (approved):

PR #492, opened 2026-07-14
fix(errors): preserve namespace load diagnostics

### LINKED ISSUE #486: REPL renders a bare/old-format error (no file/line) for ns-load failures — the formatted error path exists but isn't used
From the same HoneySQL session. Two error-presentation defects that made every other bug harder to find — per @nooga on Slack: *"this is old error format, might be a repl quirk … the repl is for some reason showing a bare error value instead of a formatted error with line numbers, which we also have — this seems to be a bug."*

## Symptom 1 — bare error value, wrong location

```
user=> (require 'honey.sql.protocols)
error: first expected Seq
  --> <embedded:core>:2163:41
  at first (<embedded:core>:2163:41)
  at seq (<embedded:core>:362:24)
  at require (REPL:1:1)
```

The user's file never appears. (This particular masking has a second contributing cause — #481's lazy defprotocol parsing detonating inside require's machinery — but the rendering path is what showed the bare value.)

## Symptom 2 — CompileError chains carry no positions

```
user=> (require 'honey.sql)
error: failed to load src/honey/sql/util.cljc: CompileError: compiling def value
        caused by CompileError: compiling fn body
        caused by CompileError: let binding name must be a symbol: :tag
```

Sean: *"(is there a reason there are no line numbers?)"* — the file is named but no line/column for any link of the chain, even though the compiler tracks FormSource info and the REPL demonstrably CAN render `--> file:line:col` framing (symptom 1 shows the formatter working for a different error kind).

## Ask

- Route ns-load/require failures through the same formatted-error renderer the REPL uses for direct eval errors (file, line, column, source excerpt).
- Give each `caused by` link its position where the compiler has one (it usually does — `compileForm` records source info per form).

Debuggability is the multiplier on every other compat gap: with positions, Sean would have self-served the root cause of #484 in seconds.

nnunley:
## Summary
- propagates namespace loader errors without pre-printing
- renders compile chains with precise file:line:column spans
- keeps the REPL alive after load failure

## Verification
- `make check-generated`
- `go test ./...`

Fixes #486


nooga (commented):
Solid fix for the two symptoms in #486 — I verified locally (worktree at c4706355) since go.yml never ran on this branch: `go build ./...`, `go test ./... -count=1`, and `make check-generated` are all green. Test coverage is appropriate per the four kinds in CLAUDE.md (Go unit tests in pkg/compiler and pkg/resolver, plus a black-box test/e2e subprocess test asserting the REPL survives a load failure).

Two things worth fixing before merge:

1. **pkg/resolver/resolver.go, `execPrecompiled`** (around the new `return nil, fmt.Errorf("failed to load precompiled namespace %s: %w", name, err)`): this loses structure. `FormatError` (pkg/vm/errfmt.go) dispatches on concrete Go type via a type-switch, not `errors.As`/`errors.Unwrap`, so wrapping an `*ExecutionError`/`*ThrownError` from `RunProtected()` in a plain `fmt.Errorf("...: %w", err)` means a runtime failure while executing a *precompiled/embedded* namespace's top-level code (e.g. core) will render as a single flat line — no source snippet, no stack trace — exactly the diagnostic loss this PR is meant to fix, just on the embedded-bundle path instead of the source-file path. Since `CompileError`/`ExecutionError`/`TypeError` already implement a `Wrap(err) errors.Error` + `GetCause()` pattern used everywhere else in this file, this spot should use that instead of `fmt.Errorf(%w)` (or `FormatError` needs to start unwrapping through `errors.Unwrap`).

2. **pkg/vm/errfmt.go, `FormatError`**: the early-return `if ce, ok := err.(compileErrorLike); ok { return formatCompileError(ce) }` at the top means the new `case compileErrorLike:`/`frame.compile` branch added to the frame-collection loop below is unreachable whenever `err` is itself a `*CompileError` (the common case exercised by the new tests) — it only matters if a `CompileError` shows up nested as some other error type's cause, which I don't see happening anywhere today. `formatCompileError` and the loop's compile-frame branch now implement near-duplicate 'root + caused-by chain, snippet per frame' logic; worth consolidating into one so the two don't drift.

Minor context for the maintainer: this PR sits on top of an unmerged stack (#491 <- #490 <- #489 <- #488), so `go.yml` (build/lint/test) hasn't actually run against this diff yet — only the Perf PR/WASM checks fired, and both report skipped.

nnunley:
Both addressed in 2d5cbd7b:

1. `execPrecompiled` now wraps `RunProtected` failures in `vm.NewExecutionError(...).Wrap(err)` — `FormatError` walks the cause chain via `GetCause`, so an embedded-bundle load failure renders with its full snippet/stack chain instead of the flat `%w` line.
2. `FormatError` consolidated: the early-return `formatCompileError` is gone; one frame-chain renderer handles compile errors top-level or nested (frames render in chain order — root, then causes, snippets where present). Output is byte-compatible for existing chains: pkg/vm + pkg/compiler + pkg/resolver + test/ + the e2e REPL-survival test all pass unchanged.

On the process note: this head is freshly pushed post-rebase, so the full `Go` workflow is running against the current diff now.

nooga (approved):
Verified: both blockers from the earlier review are genuinely fixed — execPrecompiled now wraps failures with a real vm.ExecutionError (Wrap/GetCause), threaded through LoadWithError/NSLoaderWithError with test coverage, and the duplicate FormatError path was removed entirely in favor of one unified renderer.

mparrett:
Reviewed at 84f3308f (post-#493 merge). Both #486 symptoms are fixed and everything is green locally. Same procedural gates as the rest of the stack before merge.

Verification:

1. **Symptom 1 (bare error, wrong location)**: a runtime failure during ns load now renders formatted with the user's file — my probe (`(def boom (first 5))` in a required ns) shows `--> src/broken/rt.lg:3:11` with a snippet and a stack ending `at require (REPL:1:10)`, instead of the old bare value anchored in `<embedded:core>`.
2. **Symptom 2 (chains without positions)**: every `caused by` link now carries `file:line:col` plus a snippet — my nested def/fn/let probe renders all three links with their own positions, matching the issue's ask. The e2e test pins this (per-link positions, single render, REPL evaluates the next form).
3. **Threading design**: `NSLoaderWithError` as an optional interface keeps the `NSLoader` API intact for embedder loaders, and the `nsLoadErrors` side channel is `nsMu`-guarded with retry-clean deletion in `RequireNS`.
4. **Earlier review items**: `execPrecompiled` wraps in `vm.NewExecutionError(...).Wrap(err)` so `FormatError`'s concrete-type chain walk sees it, and the duplicate `formatCompileError` path is gone — one renderer for compile chains top-level or nested.
5. **Dispatch warnings (the merged #493)**: smoke-tested — `(set! *warn-on-reflection* true)` plus a dynamic `(. x member)` emits once with source position; the once-per-site dedup is mutex-guarded and the docs are honest that this is not JVM reflection.

Locally at 84f3308f: full `go test ./...` green, Clojure compat suite 5657/5657 assertions across 232 files.

One non-blocking observation: `NSResolver.Load` (the old API) now returns nil on failure without the stderr print it used to emit. Nothing in-repo calls it anymore — `RequireNS` prefers `LoadWithError` — but an embedder calling the resolver's `Load` directly goes from a stderr line to silence. Fine to leave if that surface is considered internal.

Procedural gates (shared with #494):

- Conflicting with its base (`fix/482-lazy-top-level-errors`) — generated artifacts only (`core_compiled.lgb`, `generated.sums`), so rebase plus `make generate` clears it.
- No CI test job ran on this combined head; the run you kicked off landed on 2d5cbd7b, before the #493 merge. This looks structural: the Go workflow doesn't trigger for PRs based on other PR branches, so CI should attach once this retargets main after #491 merges.

I'll add my approval once it retargets main with CI green — with nooga's approval already in, treat this as a second confirmation.

PR #493, opened 2026-07-14
feat(compiler): warn on dynamic dispatch fallbacks

### LINKED ISSUE #483: Implement *warn-on-reflection* with real let-go semantics (dynamic-dispatch fallback warnings), well documented
From @seancorfield's HoneySQL-on-lgx run; nooga on Slack: "I think it's a good idea to add it, we just never got around to it." Revised after discussion (nnunley): implement the compatibility vars and give `*warn-on-reflection*` real, documented let-go semantics.

## Today

```clojure
user=> (set! *warn-on-reflection* true)
error: SET_VAR invalid Var
;; and inside a required .cljc file:
error: failed to load src/honey/sql/protocols.cljc: ExecutionError: SET_VAR invalid Var
```

Serious Clojure libraries commonly set this var in their `:clj` branch. let-go currently takes that branch unless an `:lg` branch exists, so the missing var forces source edits in otherwise-portable code.

## Approved semantics

`*warn-on-reflection*` is a dynamic core var, default `false`, and top-level `set!` is accepted. When truthy, let-go emits a compile/lowering-time warning with the precise user file:line:column, at most once per source site, for these actionable fallback classes:

1. **Dynamic host interop**: member/method access whose target host type cannot be resolved statically and therefore dispatches per call at runtime.
2. **Missed known direct/native lowering**: the callee is a known var/direct/native candidate, but the IR/Go lowering cannot select that path and emits the generic `InvokeValueEC` trampoline.

The warning does **not** cover ordinary intended IFn dispatch: higher-order calls, protocol dispatch, or unrelated generic calls with no known direct/native candidate remain silent. This keeps the signal analogous to Clojure's warning: a statically optimizable operation fell back to a slower dynamic mechanism.

Direct/static host calls and successfully direct/native-lowered calls remain silent. Warning state is consulted dynamically at compilation/lowering time.

## Other compatibility vars

Define `*unchecked-math*` as a dynamic, settable compatibility var. let-go's numeric tower has no corresponding unchecked primitive-arithmetic mode, so its value is currently accepted but has no code-generation effect; document that explicitly.

## Documentation requirement

The docstring and user docs must state plainly that this is not JVM reflection. Warnings may fire at compile/lowering time rather than call time. A host-class type hint may resolve host interop; direct/native fallback remediation instead requires a typeable/canonical call shape or native registration.

## Repro

```console
$ lg -e '(set! *warn-on-reflection* true)'
# expected: succeeds; var exists and its root is set
```

nnunley:
## Summary
- adds dynamic `*warn-on-reflection*` with real let-go dispatch-warning semantics
- warns once per precise source site for dynamic host dispatch and known native/direct fallback
- keeps ordinary IFn and protocol dispatch silent
- adds documented no-op `*unchecked-math*` compatibility var
- tests compiler and IR/Go-lowering source spans

## Verification
- `make check-generated`
- `go test ./...`

Fixes #483


nooga (commented):
Feature works and is tested (Go unit tests in `pkg/compiler/reflection_warning_test.go` + two new cases in `pkg/ir/lisp_lower_go_test.go`, plus a `.lg` test for the new dynamic var's surface behavior — right test kinds per CLAUDE.md), verified locally with `go build ./...`/`make check-generated` clean.

One doc-accuracy issue worth fixing: the docstring and `docs/guide/clojure-compatibility.md` say "Host-class hints ... can resolve host warnings." I tested this directly — `(defn g [^String s] (.length s))` with `*warn-on-reflection*` true still warns at the `.length` call site. `hostTargetStaticallyKnown` (compiler.go) never inspects hint metadata, and per existing comments in compiler.go (~lines 200, 1199) the compiler doesn't attach `:tag` metadata to fn params/locals at all yet. Net effect: the warning fires for essentially every `.` call whose receiver is a plain symbol — i.e. most real interop code — and the documented remediation doesn't actually work. Either fix the docs to not promise hint-based suppression, or note this as a known phase-1 limitation, so users chasing the warning per the docs aren't misled.

Minor: `call-assign-stmts` in `lower_go.lg` calls `lookup-direct-candidate` twice per call site (once directly, once again inside `lookup-direct-callable`) — duplicate registry lookups in a path the surrounding comments flag as perf-sensitive. Worth folding `lookup-direct-callable` to accept the already-computed candidate instead of recomputing it.

mparrett:
Built and ran this at `d6b8d2e`; `go test ./pkg/...` is green and I exercised the feature end-to-end with the built binary. No correctness blockers. Two minor nits and one merge-order note below.

## Verified

- Default-off is respected: a `.missing` form compiles silently with no var set.
- The idiomatic activation path works. This was my main worry: the compiler path derefs `RootExecContext` unconditionally (compiler.go carries no `ec`) while the lowering path threads the real `ec`, so I wanted to confirm the two didn't diverge for real usage. `(set! *warn-on-reflection* true)` at the top of a file does reach the root and fire for the rest of that compilation, so the split is fine in practice.
- Per-site dedup works; two calls on different lines stay distinct sites.
- The `lookup-direct-candidate` extraction out of `lookup-direct-callable` reads cleanly.

## Findings

### 1. Redundant candidate lookup on the lowering path (minor perf)

In `invoke-stmts` (lower_go.lg ~2157), `candidate` is computed unconditionally via `lookup-direct-candidate`, and then `lookup-direct-callable` calls `lookup-direct-candidate` again inside. On a direct-call hit (the common case) `candidate` is computed and thrown away. This runs per call site being lowered, which is the cost the neighboring `pos? (count *lowered-registry*)` "O(1) emptiness check" comment is guarding against, so it cuts against the grain of the surrounding code. Computing it lazily in the fallback branch keeps the hit path free and confines the `ir/source-infos` work to the path that actually warns:

```clojure
(or direct
    (do
      (when-let [candidate (lookup-direct-candidate f (first refs) arg-count)]
        (-warn-reflection! (ir/source-infos nid f) :direct-native-fallback
                           "known direct/native candidate could not be lowered; using runtime dispatch"))
      (trampoline-call-stmts f closed-exprs nid target refs)))
```

### 2. Stringly-typed args in `-warn-reflection!` (minor)

`strings.TrimPrefix(vs[1].String(), ":")` and `strings.Trim(vs[2].String(), "\"")` rebuild the kind and reason by munging the printed form. It works for the current callers (all fixed literals), but a type-assert to `vm.Keyword` / `vm.String` on the raw value is less fragile than trimming quote characters off `.String()`.

### 3. No lg-level test for the activation idiom (nice-to-have)

The Go tests enable warnings via `RootExecContext.PushBinding` / `warnVar.SetRoot`, which proves the mechanism but not the user-facing `(set! *warn-on-reflection* true)` path. I confirmed that path by hand; a small `.lg` test would lock in the contract that actually ships.

One non-blocking observation: `hostTargetStaticallyKnown` warns on every bare-symbol target, so a class-style static target like `(. Foo bar)` warns too. Fine given it's opt-in and default-off, but a candidate for a follow-up refinement or a comment.

## Merge order

This is stacked three deep: #493 → #492 (`fix/486-repl-error-rendering`) → `fix/482-lazy-top-level-errors`. `CLEAN` here only means clean against #492's head — it can't land until the ancestors do, and if any of them squash-merges the base needs re-pointing to `main`.

mparrett:
@nnunley To be clear regarding the last review comments, they are optional; this PR cleared my review. ✅

There are some unaddressed asks from previous reviews.

nnunley:
Both addressed in f47f5c32:

1. Confirmed the doc promise was wrong — `hostTargetStaticallyKnown` never consults hint metadata and the compiler attaches no `:tag` to params/locals, so hinted receivers warn like any plain symbol. `clojure-compatibility.md` now documents this as an explicit phase-1 limitation (hints NOT consulted; expect the host warning on most plain-symbol interop) instead of promising hint-based suppression.
2. The double registry resolution is gone: `lookup-direct-callable` is now `direct-callable-entry`, a pure filter over the candidate `call-assign-stmts` already resolved — one `resolve-call-entry` per call site on the lowering path.

nooga (approved):
Verified: the *warn-on-reflection* docs/docstrings now honestly disclose the phase-1 tag-hint limitation instead of the previous incorrect claim, and the duplicate lookup-direct-candidate call in call-assign-stmts was consolidated into a single lookup passed through to direct-callable-entry. New warning-dedup code and tests look solid.

PR #494, opened 2026-07-14
feat(compat): run stock HoneySQL examples

### LINKED ISSUE #487: Tracking: run stock HoneySQL on let-go (seancorfield's report, 2026-07-13)
Umbrella for @seancorfield's first-contact report trying HoneySQL under `lgx repl` in the HoneySQL tree ([his `let-go` branch with minimal edits](https://github.com/seancorfield/honeysql/tree/let-go), `lgx.edn` = `{}`). Same pattern as #134 (malli). Every child issue below carries an exact one-line REPL repro verified against current main.

## Blocking bugs (each filed with repro + root cause)

- [ ] #481 — `(defprotocol Y :extend-via-metadata true (foo [x]))` crashes: option keywords unhandled, lazy sig-parsing lets the throw escape
- [ ] #482 — that escape at the REPL top level is a raw **Go panic** (`thrownPanic` via `LazySeq.sval`) — VM must never dump a Go stack for a user error
- [ ] #483 — `(set! *warn-on-reflection* true)` → `SET_VAR invalid Var`; wanted as a no-op compat var (nooga on Slack: "good idea to add it")
- [ ] #484 — return-type hints on defn arity vectors: `(defn g (^String [] ""))` → `let binding name must be a symbol: :tag`; `(defn h ^String [] "y")` → seq error. Blocked `honey.sql.util`.
- [ ] #485 — `(.. "abc" (length))` → `Can't resolve ->->`: ctor-shorthand rewrite cascades on `..` (the mystery symbol that appeared nowhere in Sean's code)
- [ ] #486 — REPL renders bare/old-format errors without file:line for ns-load failures (the debuggability multiplier — made all of the above harder to find)

## Also observed, not yet filed as bugs

- `(StringBuilder.)` → `->StringBuilder` unresolvable — `java.lang.StringBuilder` isn't a modeled host class. That's the interop-registry design space (#411/#427/#97, and #480 proposes the shared class-descriptor registry); noting here rather than filing a one-off.
- `:extend-via-metadata` SEMANTICS (dispatch via metadata) is follow-up work beyond #481's parse-and-don't-crash.

## Definition of done

`(require 'honey.sql)` loads and basic `(sql/format {:select [:*] :from [:table]})` works on Sean's branch with, ideally, zero let-go-specific edits (the `#?(:lg ...)` workarounds reverted).

nnunley:
## Summary
- adds remaining Clojure compatibility seams found by untouched HoneySQL
- includes reusable `MetaValue[T]`, `MetaFn`, and `TypeUnion` descriptors
- adds `partition-all` and `remove` transducer forms, keyword reader compatibility, and terminal positive lookahead
- includes the HoneySQL compatibility design and implementation plan
- validates untouched HoneySQL ee0403d and 10 official examples externally

## Verification
- `make check-generated`
- `make test`
- `go test ./...`
- untouched HoneySQL official-example corpus: 10/10

Fixes #487


mparrett:
Review findings — I would not merge this as-is. I found three correctness blockers and two lower-priority issues.

1. **[P1] Lookahead regexes cannot survive bytecode serialization.**

   [`NewRegex`](https://github.com/nooga/let-go/blob/bb7e4e8c5ec05e9c3e78d1caf0e8c618d9d2feea/pkg/vm/regex.go#L223-L242) retains the original lookahead pattern, which the encoder writes verbatim. Both bytecode decoders still call Go's `regexp.Compile`, which rejects `(?=...)`. A regex accepted by this PR therefore works from source but fails when loaded from `.lgb`:

   ```text
   recompiling regex "(\\w)-(?=\\w)": invalid or unsupported Perl syntax: `(?=`
   ```

   The v2 decoder should reconstruct through the compatibility-aware path, with a bytecode round-trip regression test.

2. **[P1] Repeated lookahead matching changes `^` and boundary semantics.**

   [`lookaheadMatches`](https://github.com/nooga/let-go/blob/bb7e4e8c5ec05e9c3e78d1caf0e8c618d9d2feea/pkg/vm/regex.go#L261-L293) repeatedly searches `s[offset:]`. This makes `^` anchor to every sliced suffix rather than the original string and similarly loses preceding-character context for boundaries.

   Differential reproduction:

   ```clojure
   (re-seq #"^(?=a)" "aa")
   ```

   Clojure 1.12.5 returns `("")`; this branch returns `("" "")`.

3. **[P1] `TypeUnion` breaks exact-type-over-interface protocol precedence.**

   [`extend-type*`](https://github.com/nooga/let-go/blob/bb7e4e8c5ec05e9c3e78d1caf0e8c618d9d2feea/pkg/rt/lang.go#L5251-L5254) implements interface extension by registering the implementation directly on every concrete type. Consequently, a later interface extension overwrites an existing exact-type implementation. In Clojure, the exact type remains more specific regardless of registration order.

   Reproduction: extend `(type [])` first and `clojure.lang.IPersistentVector` second. Clojure dispatches the concrete implementation; let-go dispatches the interface implementation. This can silently route compatibility libraries to the wrong method.

4. **[P2] The synthesized lookahead capture leaks into replacement expansion.**

   [`replaceLookahead`](https://github.com/nooga/let-go/blob/bb7e4e8c5ec05e9c3e78d1caf0e8c618d9d2feea/pkg/vm/regex.go#L307-L326) passes the raw match—including `__let_go_terminal_lookahead`—to `ExpandString`. With `#"(a)(?=b)"`, replacement `$2` should fail because the source pattern has no group 2; let-go instead substitutes the hidden assertion and produces `"bb"`.

5. **[P3] Remove the committed `.cpcache` files.**

   [`.basis`](https://github.com/nooga/let-go/blob/bb7e4e8c5ec05e9c3e78d1caf0e8c618d9d2feea/.cpcache/959655937.basis) and [`.cp`](https://github.com/nooga/let-go/blob/bb7e4e8c5ec05e9c3e78d1caf0e8c618d9d2feea/.cpcache/959655937.cp) are already ignored by `.gitignore` and contain author-specific `/Users/ndn/.m2/...` paths. They are nonportable accidental artifacts.

Verification and size notes:

- Generated manifest, bundle comparison, gogen build, and native-dispatch smoke check pass.
- GitHub currently reports only four skipped perf jobs; there is no passing CI test job attached to the PR.
- Stripped binary: **+33,776 bytes (+0.257%)** versus the PR base.
- Embedded core bundle: **+1,388 bytes (+0.536%)**.
- Normal regex performance is unchanged. The lookahead path adds per-match allocations and eagerly collects matches, but I found no clear superlinear case.
- This review is relative to the declared stacked base, `feat/483-dispatch-warnings`.

nooga (commented):
Confirming and reinforcing mparrett's review — I reproduced all five findings against the branch and none have been addressed (this PR is still a single commit, predating that review):

1. **Bytecode round-trip breaks lookahead regexes (P1, confirmed).** Both `pkg/bytecode/decoder.go` TagRegex sites (~L1004, ~L1369) call raw `regexp.Compile(pattern)` instead of routing through `vm.NewRegex`, which has the lookahead-compatibility fallback. Extending the existing `TestRegexRoundtrip` in `pkg/bytecode/bytecode_test.go` with a lookahead pattern reproduces the failure immediately: `recompiling regex "(\w)-(?=\w)": invalid or unsupported Perl syntax: '(?='`. Needs a regression test added to that file and a decoder fix that reconstructs via the compat path.

2. **`^` anchors to every match offset, not string start (P1, confirmed).** `(re-seq #"^(?=a)" "aa")` returns `("" "")` here vs Clojure's `("")`, because `lookaheadMatches` (pkg/vm/regex.go) searches `s[offset:]` repeatedly rather than matching against the full string with proper start-position handling.

3. **TypeUnion breaks exact-type-over-interface precedence (P1, confirmed).** `protocol.Extend` (pkg/vm/protocol.go:51-53) is a plain `p.impls[vt] = implMap` write. The new TypeUnion branch in `extend-type*` (pkg/rt/lang.go) fans an interface extension out to every concrete member type, so whichever of {exact-type extension, interface extension} runs last wins — order-dependent, and divergent from Clojure where the exact type always wins regardless of registration order.

4. **Lookahead capture leaks into replacement expansion (P2, confirmed).** `(str/replace "ab" #"(a)(?=b)" "$2")` returns `"bb"` here (substituting the synthesized `__let_go_terminal_lookahead` group) instead of failing on a nonexistent group 2.

5. **Remove `.cpcache/959655937.{basis,cp}`.** Already gitignored; contain the author's local `/Users/ndn/.m2/...` paths.

Additionally: `docs/superpowers/plans/2026-07-13-honeysql-compat.md` and `docs/superpowers/specs/2026-07-13-honeysql-compat-design.md` read as internal agent-workflow planning artifacts (references to jj workspaces, Slack, `memory_reflect.py`, "agentic workers") rather than intended project documentation — worth dropping unless there's a deliberate reason to keep them versioned.

On the positive side: `partition-all`/`remove` transducers and the keyword-colon reader fix have proper `.lg` integration tests, and `MetaValue`/`MetaFn` has a Go unit test, consistent with CLAUDE.md's test-kind conventions.

Separately: statusCheckRollup shows only two perf workflows (both skipped) — no actual `go.yml` test run appears attached to this PR, so there's no CI green signal beyond local runs. This also stacks on open PR #493, so it's blocked on that regardless of the above.

nnunley:
All four in-scope findings addressed in 202ec92d — fixed at the subsystem seams rather than piecemeal, since they share one root (the lookahead compat layer and the TypeUnion fan-out both shipped with unfinished integration seams):

1. **Bytecode round-trip**: both `TagRegex` decode sites reconstruct via `vm.NewRegex`, so lookahead literals survive a bundle round-trip. `TestRegexLookaheadRoundtrip` pins pattern AND matching semantics post-decode.
2. **`^` anchor context**: `NewRegex` classifies the fallback pattern's leading anchor (none / `^`/`\A` / `(?m)^`); resumed searches constrain resume positions accordingly — text-anchored patterns match only at position 0, `(?m)` only at real line starts. `(re-seq #"^(?=a)" "aa")` now yields `("")`. One honest caveat: `\b` at a resume boundary still sees a fresh text start — Go's regexp has no match-at-offset-with-context API, so that sub-case is documented in-code as a known limitation rather than silently wrong.
3. **Synthetic group leak**: replacements expand against the visible indices, so `$n` addresses the original pattern's groups and an out-of-range reference expands empty (consistent with non-lookahead patterns) instead of leaking the assertion's text. `(str/replace "ab" #"(a)(?=b)" "$2")` → `"b"`.
4. **Dispatch precedence**: exact `extend-type` now beats a TypeUnion interface fan-out in **both** registration orders (`Extend` marks the type direct; `ExtendViaUnion` never clobbers a direct impl). The both-orders `.lg` test was verified red against the unfixed code — concrete-first genuinely failed before.

Finding 5 (`.cpcache`) isn't in this diff. Full suite (1412) + Clojure compat suite (233) green; branch rebased onto the current chain tip.

nooga (changes requested):
Solid progress — the bytecode round-trip, ^ anchor semantics, and TypeUnion precedence fixes are all verified working. One item from the original review is still open:

str/replace with a genuinely nonexistent capture group reference doesn't error. E.g. (str/replace "ab" #"(a)(?=b)" "$2") — the pattern has one capturing group, (?=b) is a lookahead not a group, so $2 doesn't exist. Real Clojure throws IndexOutOfBoundsException: No group 2 here; let-go silently substitutes empty string and returns "b".

The lookahead-text-leaking-into-$2 part is fixed (no longer returns "bb"), but referencing a nonexistent group should still be a hard error rather than silently producing a plausible-looking wrong answer. Please make an out-of-range $n reference throw, matching Clojure's fail-loud semantics.

mparrett:
Verified all four in-scope fixes at 202ec92d — the code is ready. Three procedural items below before this can merge.

Verification:

1. **Round-trip**: both `TagRegex` decode sites (v1 and v2) reconstruct through `vm.NewRegex`, and `TestRegexLookaheadRoundtrip` pins pattern and post-decode matching. I also grepped for remaining raw `regexp.Compile` on user patterns — none left outside `NewRegex` itself.
2. **Anchor semantics**: `(re-seq #"^(?=a)" "aa")` → `("")`, `(re-seq #"(?m)^(?=a)" "aa\nab")` → `("" "")`, and the `\A` variant all match Java now. The `\b` resume limitation reproduces exactly as your in-code comment describes (`(re-seq #"\b(?=\w)" "ab")` gives `("" "")` where Java gives `("")`) — disclosed and scoped, fine for the compat fallback.
3. **Group leak**: `(str/replace "ab" #"(a)(?=b)" "$2")` → `"b"` (empty expansion; Java throws here, but empty matches let-go's existing out-of-range `$n` behavior, the right consistency for this layer), and `$1` addresses the real group in both `replace` and `replace-first`.
4. **Dispatch precedence**: both registration orders dispatch the concrete implementation in my probe; the `direct` marking in `Extend`/`ExtendViaUnion` reads correctly, and unions still overwrite only other unions.

`.cpcache` is gone from the diff. Locally at 202ec92d: full `go test ./...` green, Clojure compat suite 5657/5657 assertions across 232 files.

Procedural gates:

- GitHub reports the branch conflicting with its base (`fix/486-repl-error-rendering`). The conflict is confined to generated artifacts (`core_compiled.lgb`, `generated.sums`) — source merges clean — so a rebase plus `make generate` clears it.
- No CI test job ran on this head (only the skipped bench/comment jobs), same as round 1. The next push should get a real CI run attached before merge.
- This is stacked on #492, which is still open; merge order applies.

I'll approve once the rebase lands with CI green.

nnunley:
Out-of-range group references now throw — fixed in 7eb3b9f.

`validateReplacement` runs before expansion in both `ReplaceAll` and `ReplaceFirst`, so it covers plain patterns too, not just the lookahead path: `(str/replace "ab" #"(a)b" "$2")` was also silently expanding empty (Go's `Expand` semantics), and Clojure throws for both. Named references to nonexistent groups (`${y}`) throw as well, and the synthetic lookahead group is excluded from the addressable set.

One deliberate scope note: references are parsed exactly as Go's `Expand` parses them, so `$2x` reads as the name `2x` (Java would read `$2` then literal `x`). That divergence predates this PR; the change here makes it fail loud instead of silently emptying, which seemed strictly better than leaving it silent while chasing byte-for-byte Java template parsing.

Also reconciled the branch with main after the #488–#492/#496/#446 squashes (merge commit 31fd7b08, append-only) — conflicts were the usual generated artifacts plus keeping main's final versions of `resolver.go`/`errfmt.go` (the branch carried stale pre-#492-fix copies from when it sat on that stack) and keeping both additions in `regex.go` (`MustRegexFromReadable` from #496 + the lookahead compat layer). Full suite green: 1592 tests across 55 packages, `make check-generated` clean. PR now bases on `main` directly.

mparrett:
Verified 7eb3b9f5 end to end — the lint failure is the only thing between this and my approval.

- **Merge reconciliation (31fd7b08)**: I diffed the rebased diff against the round-2 diff I reviewed; outside the regenerated artifacts, the only residue is context drift from keeping both `regex.go` additions (this branch's plus #496's `MustRegexFromReadable`). The stale pre-#492 `resolver.go`/`errfmt.go` copies are gone in favor of main's finals, as described.
- **Fail-loud group references**: probed plain, lookahead, named, and `replace-first` paths — nonexistent `$n`/`${name}` all throw with the pattern in the message; `$$`, a bare `$`, and a malformed `${1x` pass through literally, matching Go `Expand`; `$0` and braced refs work. Widening the fix to plain patterns (not just lookahead) was the right call — `(str/replace "ab" #"(a)b" "$2")` silently emptying was the same trap.
- Full `go test ./...` and the compat suite (5657/5657) green locally at 7eb3b9f5.

One suggestion on the disclosed template-parsing divergence: the common Java shape `"$1_$2"` (or `"$1x"`) now throws, because Go's name parsing eats the trailing word characters, and the error reads as if the group is missing (`no group named "1_"`) when the actual fix is `"${1}_${2}"`. When an unknown name starts with a digit, appending a hint (group references followed by letters, digits, or underscores need braces: `${1}_`) would let someone hitting this from a stock Clojure library self-serve. Take it or leave it; the fail-loud behavior itself is what matters and it's in place.

I'll approve as soon as the QF1001 lint fix lands and CI is green.

nooga (approved):
Verified: the last outstanding item (`str/replace` silently substituting empty string for a nonexistent `$n` group reference) is fixed — `validateReplacement` now rejects it, matching Clojure's fail-loud semantics. Confirmed live (`$2` against a lookahead-only pattern now throws `"no group 2"`), plus edge cases (`$$`, `${name}`, `$0`, named groups) all behave correctly. Full test suite passes clean on PR head. All four originally-flagged issues are now resolved.

mparrett (approved):
Approving at 180c4aec. The force-push delta from the verified 7eb3b9f5 is the QF1001 fix alone — the `gi != lookaheadGroup` simplification is equivalent, since slot 0's name is always empty and a validated name never is. CI is green and my probes reproduce identically at the new head; full verification detail in the comment above.

Issue #352, opened 2026-06-28
Epic: Combinator inlining & higher-order specialization (EPIC-013)


nnunley:
Part of #258 (Lowering & cross-target codegen).

## Why
EPIC-007 (native dispatch, `MakeNativeMultiArity`) and EPIC-012/#345 (cross-package direct calls) made the YamlStar combinators *dispatch* natively and *direct-call* across packages — but an AOT-lowered parse stays **dispatch-bound** (native lowering of the 212 grammar `def`-closures measured ~55 ms/op vs 36 ms interpreted — slower). Root cause: every PEG combinator (`chr`/`rng`/`all`/`any`/`rep`) allocates a fresh `name*`-wrapped closure per call and the `call` trampoline dynamically dispatches into it. This epic collapses the combinator tree into **inlined straight-line recursive descent** via general optimizer passes (lambda lifting + higher-order specialization).

Discharges the deferred `AC-PO-MA.1b` (YamlStar parse-profile proof) and is the prerequisite for the EPIC-011 native binding being *fast*, not just native.

Designs (in-repo, gitignored superpowers dir): `docs/superpowers/specs/2026-06-28-lambda-lift-and-higher-order-specialization-design.md` and `…-multiblock-inliner-and-loop-unroll-design.md`.

## Iterations
- [x] **ITER-0031 — lambda-lift (capture-free).** `ir.passes.lambda-lift` hoists inner capture-free `fn` literals (the grammar rule-closures) to top-level direct-call-eligible siblings via the `:multi-fn-template` flatten; bytecode path isolated. *(done, parity-identical)*
- [x] **ITER-0032 — single-block inliner + `name*` transparency + bounded iteration.** `ir.passes.inline`: automatic eligibility (mutability/size), `name*`-transparent callee resolution, single-block body splice, bounded iterative driver. *(done)*
- [x] **ITER-0033 — tail-position multi-block inliner.** Inline multi-block callees when the call is in tail position (no continuation/live-across threading). *(done — `splice-multiblock-tail!` in `ir.passes.inline`, gated on `tail-call-of-block?`; shipped in #387)*
- [x] **ITER-0034 — fold-over-rest specialize + unroll (`any`/`all`).** Recognize the fold-over-rest-arg loop idiom; specialize to fixed arity for N known operands and unroll into straight-line direct calls. The research-grade core. *(done — `fold-over-rest?`/`specialize-fold` in `ir.passes.inline`; shipped in #387)*
[...]

PR #495, opened 2026-07-14
bench: YamlStar real-corpus parse benchmark + ITER-0035 gate (#352)

nnunley:
Part of #352 (EPIC-013) — the last open iteration, ITER-0035.

Adds `test/benches/ysbench.{lg,sh}`: times `(yamlstar.parser/parse "foo: 42")` against the real 212-rule yamlstar grammar in two modes — interpreted VM bytecode vs `*ir-compile*` with `ir.passes.inline` enabled — and asserts the epic's closing criterion (optimized parse ≥1.10× faster than the interpreted baseline). Skips cleanly when no yamlstar checkout is present; knobs: `LG_YAMLSTAR_SRC`, `LG_YSBENCH_MIN_SPEEDUP`.

One call-out: the gate currently **fails** — recorded numbers (Apple Silicon, 2026-07-13): interp 36.4 ms/call (matching the epic's 36 ms baseline), ir+inline 34.1 ms/call, speedup 1.07×. The ITER-0031..0034 passes don't reach this corpus yet: the runtime inline registry seeds only from same-namespace `defn` siblings, while yamlstar's grammar rules are top-level `def` closures invoking combinators from another namespace (`prelude`), so the parse stays dispatch-bound in the combinator tree walk. This PR is the measurement instrument the epic's acceptance is judged by; #352 stays open on making it pass.

### DIFF
--- a/test/benches/README.md
+++ b/test/benches/README.md
@@ -31,6 +31,32 @@ The IR-on number depends on the loop/recur lowering fix in this branch
 (see the parent commit); before it, `pegbench-ir.lg` crashed. The remaining
 VM→JVM gap is the lowered-Go target (tracked separately).
 
+## ysbench — YamlStar parse benchmark (ITER-0035 gate)
+
+The real-corpus counterpart to pegbench: times `(yamlstar.parser/parse
+"foo: 42")` against yamlstar's actual grammar (212 `def`-closure rules).
+Needs a [yamlstar](https://github.com/yaml/yamlstar) checkout; skips
+cleanly without one.
+
+| file | what | run |
+|------|------|-----|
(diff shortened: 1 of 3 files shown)


nooga (approved):
Clean, additive benchmark harness consistent with the existing pegbench.lg / pegbench-ir.lg pattern in test/benches/. Verified: (1) ir.passes.inline/*enable-inline* is reachable via the required ir.passes.pipeline namespace, so the alter-var-root call is valid; (2) the skip path with no yamlstar checkout exits 0 with a clean message, confirmed by running it directly; (3) the gate script is not referenced by any GitHub Actions workflow, so the documented current failing result (1.07x vs 1.10x required) cannot break CI - it is a local/manual instrument only. No functional or core code is touched, so no additional .lg or Go tests are needed per CLAUDE.md benchmarking guidance. Nothing blocking.

PR #496, opened 2026-07-14
perf(compiler): compile regex literals at read time (Clojure parity)

nnunley:
Reworked after review feedback (twice — see thread): **no caching anywhere, no permanent rooting anywhere.**

What Clojure actually does: `re-pattern` never caches (two calls with the same string return distinct Patterns); regex **literals** are cheap because the reader compiles `#"..."` into a Pattern constant at read time. This PR matches that exactly:

- `readRegex` returns the compiled `vm.Regex`; invalid patterns now error at read time (as in Clojure). The compiler treats `RegexType` as a literal constant; the `.lgb` const pool already round-trips Regex.
- Dynamic `(re-pattern s)` still compiles per call — memoization there stays the caller's choice.
- IR pipeline: `ir.build` carries the pre-compiled literal as a `:const` (it only recognized the old `(re-pattern "...")` call form), and lower-go reconstructs it via a new `vm.MustRegexFromReadable`.

**Constant lifetimes (from the review round):** transient compilation units — each `compiler.Eval` call and each REPL input — now compile through `Context.ChildForEval()`, a child constant pool layered on the global one. Constants an eval introduces are reachable only through the chunks/functions that use them, so they're garbage-collected with their chunk instead of rooting the process-global pool (previously +1 pool entry and ~650B retained per re-eval of a regex literal). Regex constants use pure pointer identity, so Clojure's regex equality semantics hold exactly: `(= #"x" #"x")` is false, distinct occurrences and distinct evaluations yield distinct objects. Two layering fixes make child pools safe unconditionally (child `Intern` refuses parent hits minted after the child's base; `Consts.get` recurses through layered parents). Regression tests pin zero global-pool growth across 1000 evals, finalizer-proven collectibility, and identity equality.

Microbench: 200K evaluations of `(re-matches #"chr\((.)\)" s)` drop from 268 ms to 54 ms.

One call-out: this **improved native lowering coverage** as a side effect — core fns containing regex literals previously hit `:build/unrecognized-form` and fell back to bytecode; the ir-stress ratchet improves accordingly.

Found while profiling the yamlstar parse for #352 (~4,850 regex-literal evaluations per parse).


mparrett:
Blocking performance concern: this change makes each read of a regex literal produce a fresh `*vm.Regex`, then `compileForm` interns it into the process-global constant pool. `constHash`/`constEqual` intentionally use pointer identity for `*Regex`, so repeated compilation of the same source can never deduplicate it.

I reproduced this with 10,000 calls to `compiler.Eval` on the source `#"x"` and forced a GC afterward:

- `main`: constant-pool delta `0`, live-heap delta ~17 KB
- this PR: constant-pool delta `+10,000`, live-heap delta ~6.48 MB

The interactive REPL and nREPL both reuse a compiler context backed by this same global pool, so a long-lived session that evaluates regex literals grows without bound (roughly 650 bytes retained per evaluation in this probe). This is distinct from the intended per-literal-site identity: constants needed by live compiled functions must remain stable, but constants from transient top-level eval chunks should become collectible.

Please give transient compilation units/chunks an appropriate constant-pool lifetime (for example, a child pool retained only by the generated chunk/functions), or otherwise preserve per-site identity without permanently rooting every transient regex. A regression test that checks pool/live-object growth across repeated evals would be valuable.

nnunley:
Reworked to your preferred design — head `fdf3f04`: transient compilation units now get a **child constant pool whose lifetime is the chunk's**, so transient constants are genuinely garbage-collected instead of being rooted anywhere.

- `compiler.Eval` and each REPL input compile through `Context.ChildForEval()` — a fresh context over `vm.NewChildConsts(pool)`. Constants an eval introduces are reachable only through the chunks/functions that reference them; a `defn` keeps its child pool alive exactly as long as the fn lives, and a discarded eval's constants are collected with its chunk. The global pool stays global — shared constants still dedupe against the parent; children only hold what an eval adds.
- Two layering fixes make child pools safe unconditionally: a child's `Intern` only accepts parent dedup hits minted before the child's `base` (a later parent index would collide with the child's index space — refused hits re-intern in the child, duplication but never aliasing), and `Consts.get` recurses so a child-over-child (eval over a bundle-decoded pool) resolves correctly.
- Regex constants are back to **pure pointer identity** — the per-site tagging from the previous push is reverted; with per-eval pools, plain identity gives exact Clojure semantics with no leak.

Regression tests, all red/green verified:
- `TestRegexLiteralReEvalDoesNotGrowConstPool`: 1000 evals → global pool growth **exactly 0** (was +1 per eval).
- `TestTransientRegexConstantIsCollectible`: a dropped eval's regex is proven collected via finalizer under `runtime.GC` (with filler evals to flush recycled VM frames' stack-slot residue — pooled frames pin the last eval's values until reuse; noting that as a possible micro-follow-up in `ReleaseFrame`).
- `TestRegexLiteralIdentityEquality`: `(= #"x" #"x")` false, self-equality true; the jank suite's `eq`/`not_eq` also pass on both backends.

Gates: lint clean, full `go test ./...` (incl. genmanifest after regen), gogen_ir `pkg/ir` + `./test` with the compat suite, `check-generated` — all green.

nnunley:
@mparrett re-review requested — the blocking pool-growth concern is reworked per your first suggestion (per-eval child const pools with chunk-lifetime collection, global pool unchanged as the shared layer; regex constants back to plain pointer identity). Details in the comment above; head `fdf3f04`, all gates green.

mparrett (changes requested):
@nnunley Thanks for the rework. The new targeted tests pass, and the broader local suite was healthy. One finding to resolve:

> The constant-pool leak still exists in several transient evaluation paths: nREPL compiles directly into its long-lived session context; `load-string`, native `eval`, and namespace-aware eval also use the process-global pool. On PR head `fdf3f04`, temporary regression probes confirmed that 10 regex evaluations through each path add 10 permanently retained constants. These should also use per-evaluation child pools and receive regression tests.

nnunley:
Addressed in bf6edfb, and widened. Your four paths each get a per-evaluation child pool, plus three more the same audit turned up: `api.LetGo.Run` (embedder session pool) and wasmhost's `<host-request>`/`<host-inspect>` evals (long-lived host pool).

Rather than patching each site with a bare `vm.NewChildConsts`, the mechanism is pulled up a level: `compiler.NewTransientCompiler(parent, ns)` is the single constructor for transient evaluations (`ChildForEval` delegates to it), and every transient site routes through it — so a grep for bare `NewCompiler(` is now an audit of the intentionally-persistent pools (boot, require loading, build tools), which is documented on the constructor.

Regression tests probe each of your four paths for pool growth — the load-string probe was verified red against the unfixed code — and the nREPL test drives `handleEval` over a drained pipe against its own session pool. The pods path is extracted to a named `evalInNSChild` so its test hits the exact registered function. Full suite green (1396 tests).

nooga (approved):
Verified: all four paths mparrett flagged are genuinely fixed via NewTransientCompiler routing (nREPL handleEval, load-string, native eval, EvalInNS), each with regression tests (before/after const-pool-size checks over 100 evals). Full compiler/nrepl/api/wasmhost suites pass on head.

One non-blocking note for a follow-up: pkg/rt/wasm/rendermain.go's legacy _lgEval/LetGoHost.eval() compat hook still calls ctx.Compile(code) directly instead of going through a child pool, so a long-lived browser session using that older API path still leaks one constant per eval — same bug class, just not one of the four paths originally flagged. The newer _lgRequest path is already fine.

mparrett (approved):
Re-review of the child-pool rework at 397424df — approving.

Verified:

- The original blocker is gone. Re-running my 10,000-eval probe on `#"x"` gives a constant-pool delta of 0 and no live-heap growth (previously +10,000 entries, ~6.48 MB retained).
- All four paths from my last round (nREPL session, `load-string`, native `eval`, `EvalInNS`) now compile through `NewTransientCompiler`, each with a pool-growth regression test; the nREPL test drives `handleEval` itself over a drained pipe. I ran the bare-`NewCompiler(` audit at head: the remaining sites (boot core and baseline-NS load, the `initCompiler`/`NewLetGo` session pools, resolver ns-loading, lgbgen, wasm boot) are all legitimately persistent.
- Mechanism: the `idx < c.base` guard in `Intern` refuses parent hits minted after child creation (duplication, never aliasing), and the recursive `get` resolves child-over-bundle layering.
- Semantics: `(= #"x" #"x")` stays false, and regex literals inside quoted collections compile and stay distinct — the structural `constEqual` path compares `*Regex` by identity, so cross-form merging can't happen.
- gogen: `MustRegexFromReadable` round-trips `String()`'s Go-quoted `#%q` form.
- Full `go test ./...` green locally at 397424df.

One note, no action needed: a grandchild pool (`ChildForEval` over a session pool that is itself a child) dedupes only against its immediate parent's own layer, so it can re-intern grandparent constants. That is duplication only, and transient — harmless at current depths.

PR #499, opened 2026-07-14
feat(rt): BootCore() — compiler-free clojure.core load for AOT/embedded binaries

mparrett:
## Why

`scripts/lg-compile` lowers a set of `.lg` namespaces to native Go packages that import only `pkg/rt` and `pkg/vm`. By design there's no compiler: an AOT-lowered binary shouldn't carry the reader and compiler.

Such a binary can only reach part of `clojure.core`. The native core fns (`str`, `+`, `vector`, …) are registered by installers at package init, so a lowered call to one resolves from a bare `vm.NewExecContext()`. The `.lg`-defined core fns (`frequencies`, `partition`, …) bind only when the embedded core bundle runs, which nothing triggers unless the compiler's eval path runs. Until then they're unbound nil stubs, and the generated call — `ec.Invoke(rt.CachedVarFn(&v, "clojure.core", "frequencies"), …)` — nil-panics.

There's no public "load core, hand me an `ExecContext`" entry point. `lg`'s own boot runs through `compiler.Context`, which the AOT binary deliberately doesn't link. So a lowered binary that touches `.lg`-defined core has no way to load it short of reproducing the core boot by hand.

## What it does

`rt.BootCore() (*vm.ExecContext, error)` decodes and runs the embedded precompiled core (`CoreCompiledLGB`) and returns a ready `ExecContext`:

- runs core's main chunk (replays every `def`/`defn`/`defmacro`),
- runs the lg baseline namespaces (`let-go.core`, …), which are auto-refer'd everywhere but never explicitly required, so nothing else runs them,
- runs the chunks of hybrid namespaces (native fns plus bundled lg source, e.g. `async`), whose lg-defined vars are reachable via qualified symbols without a `require`.

It links only `pkg/rt`, `pkg/vm`, and `pkg/bytecode`, so an AOT binary can call it and then invoke the full `clojure.core` surface. Callers that touch only native core still pass `nil` or a fresh `vm.NewExecContext()`; `BootCore` is what additionally binds the `.lg`-defined core.

## Notes for reviewers

`BootCore` currently parallels the runtime half of `pkg/compiler/eval.go`'s `loadPrecompiledBundle`: same decode, same eager chunk-runs, minus the compiler-side lazy-loader wiring. I kept it separate to keep the change small; the follow-up is to have `loadPrecompiledBundle` call `BootCore` so there's one core-load path (noted in the doc comment and below).

The hybrid-namespace eager-run mirrors `loadPrecompiledBundle`: a namespace that exists before decode is native-backed, and a bundle chunk for one of those is a hybrid whose chunk must run eagerly, since qualified resolution finds the already-registered ns and its decoded stubs without ever triggering a loader.

## Testing

- New `TestBootCoreBindsLgDefinedCore` (`pkg/rt`): after `BootCore`, `clojure.core/frequencies` is bound and `(frequencies ())` runs without error, the case a fresh `ExecContext` leaves as a nil stub.
- End to end against an AOT-lowered binary: `(count (frequencies (range 5)))` nil-panics on a fresh `ExecContext` and returns `5` after `rt.BootCore()`.
- `go test ./pkg/rt/` green; `gofmt` and `go vet` clean; builds under `GOOS=linux`, `js/wasm`, and `plan9`.

## Follow-up

Unify `loadPrecompiledBundle` onto `BootCore` so the compiler path and the compiler-free path share one core-load implementation. Kept out of this PR to stay focused on the new entry point.

### DIFF
--- a/pkg/rt/bootcore.go
+++ b/pkg/rt/bootcore.go
@@ -1,0 +1,103 @@
+//go:build !bootstrap
+
+/*
+ * Copyright (c) 2026 let-go contributors
+ * SPDX-License-Identifier: MIT
+ */
+
+package rt
+
+import (
+	"fmt"
+
+	"github.com/nooga/let-go/pkg/bytecode"
+	"github.com/nooga/let-go/pkg/vm"
+)
+
+// BootCore decodes and runs the embedded precompiled clojure.core bundle
+// (CoreCompiledLGB) so a program that links only pkg/rt + pkg/vm — with no
+// compiler — can resolve and invoke the full clojure.core surface plus the lg
+// baseline namespaces. It returns a fresh ExecContext ready to Invoke.
+//
+// The motivating case is an AOT-lowered binary (scripts/lg-compile output): its
+// generated funcs call core via ec.Invoke(rt.CachedVarFn(..., "clojure.core",
+// name), ...), which needs those vars bound. Programs that touch only NATIVE
+// core fns (str, +, vector, …) already work from a bare vm.NewExecContext(),
+// because installers register those at package init; BootCore is what
+// additionally binds the .lg-DEFINED core (and let-go.core / let-go.types).
+//
+// Call once at startup. This mirrors the runtime half of pkg/compiler/eval.go's
+// loadPrecompiledBundle without linking the compiler; a future refactor should
+// unify the two behind this function so there is a single core-load path.
+func BootCore() (*vm.ExecContext, error) {
+	if len(CoreCompiledLGB) == 0 {
+		return nil, fmt.Errorf("BootCore: embedded core is empty (built -tags bootstrap?)")
+	}
+
+	// Namespaces present before decode are native-backed (installers ran at
+	// package init); a bundle chunk for one of them is a hybrid ns whose chunk
+	// must run eagerly (below), since qualified refs bypass the on-demand loader.
+	preexisting := map[string]bool{}
+	for name := range AllNSes() {
+		preexisting[name] = true
+	}
+
+	// DefNSBare gives a var a home ns without triggering a loader; DefStub avoids
+	// warn-on-shadow while the chunk Defs the real value later.
+	resolve := func(nsName, name string) *vm.Var {
+		n := DefNSBare(nsName)
+		if v := n.LookupLocal(vm.Symbol(name)); v != nil {
+			return v
+		}
+		return n.DefStub(name)
+	}
+
+	unit, err := bytecode.DecodeToExecUnitBytes(CoreCompiledLGB, resolve)
+	if err != nil {
+		return nil, fmt.Errorf("BootCore: decode core: %w", err)
+	}
+
+	// Replay core's def/defn/defmacro definitions.
+	if err := runChunk(unit.MainChunk); err != nil {
+		return nil, fmt.Errorf("BootCore: run core chunk: %w", err)
+	}
+
+	// The lg baseline namespaces (let-go.core, …) are auto-refer'd everywhere but
+	// never explicitly required, so nothing else runs them — do it eagerly, after
+	// core (whose defs they depend on).
+	baseline := map[string]bool{}
+	for _, name := range LgBaselineNSNames() {
+		baseline[name] = true
+		if unit.NSChunks != nil {
+			if err := runChunk(unit.NSChunks[name]); err != nil {
+				return nil, fmt.Errorf("BootCore: run baseline %s: %w", name, err)
+			}
+		}
+	}
+
+	// Hybrid namespaces (native fns + bundled lg source, e.g. async) are reachable
+	// via qualified symbols without a require, so their chunks must run eagerly
+	// too or the lg-defined vars stay nil stubs.
+	if unit.NSChunks != nil {
+		for name, ch := range unit.NSChunks {
+			if name == "core" || baseline[name] || !preexisting[name] {
+				continue
+			}
+			if err := runChunk(ch); err != nil {
+				return nil, fmt.Errorf("BootCore: run hybrid %s: %w", name, err)
+			}
+		}
+	}
+
+	return vm.NewExecContext(), nil
+}
+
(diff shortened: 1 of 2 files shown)


mparrett:
Focus on maintainability, compatibility, artifact size, performance, and project architecture. Changes requested before merge (2):

1. **[P1] Preserve `*ns*` across core boot.** Each eager namespace chunk executes its `(ns ...)` form. Because `runChunk` creates a frame without a local `ExecContext`, `in-ns` mutates root `CurrentNS`; the hybrid chunks are then executed in randomized map order. I reproduced identical runs starting at `core` and ending at either `async` or `io`. The returned fresh context has no `*ns*` binding, so lowered code that dereferences `*ns*` can observe that process-random namespace. Save and restore `CurrentNS` on every exit (or execute under a boot-local binding), and use `unit.NSOrder` where eager chunk dependency order matters. See [bootcore.go lines 82–86](https://github.com/nooga/let-go/blob/57020ed2816db736b5923b7e95be01e24487b498/pkg/rt/bootcore.go#L82-L86).

2. **[P2] Keep `BootCore` available under `bootstrap`.** The `//go:build !bootstrap` constraint removes the public API entirely in the project's standard bootstrap configuration. A downstream program calling `rt.BootCore()` fails with `undefined: rt.BootCore`, so the function's `len(CoreCompiledLGB) == 0` error path is unreachable in exactly the configuration it describes. Remove the constraint or provide a bootstrap-tagged stub returning the documented error. See [bootcore.go line 1](https://github.com/nooga/let-go/blob/57020ed2816db736b5923b7e95be01e24487b498/pkg/rt/bootcore.go#L1).

Validation otherwise looks healthy:

- Full `go test ./... -count=1` and targeted race test passed.
- GitHub build, lint, CodeQL, generated-artifact, and differential checks are green.
- wasm and Plan 9 cross-compiles passed.
- An unused `BootCore` caused zero stripped-binary growth versus `main`.
- Calling it added approximately 242 KiB, matching the 240 KiB embedded bundle.
- Boot measured roughly 1.7–4.1 ms and 0.83 MB allocated locally.

Architecturally, consider unifying the shared decode/chunk-execution portion with `compiler.loadPrecompiledBundle` while fixing the first issue. Namespace preservation and deterministic execution order are boot invariants that should not live in two parallel implementations.

mparrett:
Both issues confirmed; fixed in c774c44.

**P1 — `*ns*` restore + deterministic order.** You had the mechanism right: `runChunk` frames carry no `ExecContext`, so each chunk's `(ns …)` form falls through to `CurrentNS.SetRoot` and mutates the global root, and the hybrid loop iterated the `NSChunks` map, so `*ns*` was left at a process-random namespace that the returned context would deref. The fix saves the pre-boot root and restores it on exit, and iterates `unit.NSOrder` instead of the map so eager execution order is deterministic. Added `TestBootCoreRestoresCurrentNS` (asserts `*ns*` is back at its pre-boot root after boot).

**P2 — `BootCore` under `bootstrap`.** Dropped the `//go:build !bootstrap` constraint. The function compiles in that config — `CoreCompiledLGB` is empty there — so the `len == 0` guard now becomes the live path and returns the documented error instead of an undefined-symbol failure. `go build -tags bootstrap ./...` passes.

On unifying the decode/chunk-execution half with `loadPrecompiledBundle`: agreed, and the PR body already lists it as a follow-up. One thing the unify has to account for: `loadPrecompiledBundle` carries the same latent `*ns*` issue, it also iterates the chunk map without restoring the root. It doesn't surface today because `api.NewContext` re-establishes `*ns*` before any user code runs, whereas BootCore returns straight to the caller. So the shared path should fold in the save/restore rather than adopt the current behavior. I'd rather do that as its own change than widen this PR.

Gate is green on the new commit: gofmt/vet, full `go test ./...`, `-race` on the rt tests, and linux/wasm/plan9 cross-builds.

mparrett:

nnunley:
Coordination note with #438: BootCore's eager hybrid-chunk execution bypasses the on-demand loader in `LookupOrRegisterNS`, which is where #438 re-applies the generated native-primitive adapters after a namespace's bootstrap defs run (e.g. `clojure.string/upper-case`). If a chunk BootCore runs eagerly re-Defs one of those vars, a BootCore'd binary would keep the bootstrap closure and permanently trip the root-deviation counter — native direct-call fast paths would stay on the trampoline (correct behavior, but the perf win is lost). Whichever lands second should call `reapplyGeneratedPrimitives` (or its then-current equivalent) after the eager chunk loop — it's one call, and the "unify the two behind this function" refactor you note in the comments would make it the single place to do it.

mparrett:
Follow-up (the `loadPrecompiledBundle` ↔ `BootCore` unify, and folding in the `*ns*` save/restore rather than adopting current behavior) is now tracked in #506, under Epic #503.

nnunley (approved):

nnunley:
Non-blocking hygiene note from a multi-agent pass (which otherwise came back clean — the `!preexisting` replay condition verified correct, hybrids are by definition preexisting so their chunks do run): `BootCore` relies on the documented "call once at startup" contract with no `sync.Once`. That's consistent with the existing boot-path discipline, so fine as-is — just worth a thought when #506 unifies it with `loadPrecompiledBundle`, since that's the moment a second caller tends to appear.

mparrett:
Superseded by #520, which rebases this onto current `main` and folds it into the #506 unify. Two things pushed it from "merge as-is" to "rebase + unify":

- This branch was 17 commits behind `main` and predates #424 (`rt.LoadCore` + a `runChunk` in `pkg/rt/run.go`). The merge is textually clean, but wouldn't compile — `BootCore`'s own `runChunk` collides with that one.
- Since #438 merged first, the eager-hybrid path needs the `ReapplyGeneratedPrimitives` call from your review. #520 puts it in one shared spine (`rt.LoadCoreBundle`) alongside the `*ns*` restore, so all three core-boot paths share both invariants instead of drifting.

The three commits here are preserved in #520 (rebased) with the unify on top. Closing in favor of #520.

PR #501, opened 2026-07-14
feat(bytecode): opt-in DEFLATE compression for .lgb bundles

mparrett:
The embedded-core follow-up remains stacked as #502.

## What it does

Adds opt-in DEFLATE compression for compiled `.lgb` bodies:

- `lg -c ... -z` and `lg -b ... -z` emit compressed bundles.
- The magic, format version, flags, and capability payload stay plaintext so compatibility and opcode-set checks happen before decompression.
- Compressed bundles use format version 3 and include a declared uncompressed-body size plus a codec byte.
- Uncompressed bundles remain version 2 and retain the pre-change byte encoding.
- Both streaming and byte-backed decoders inflate transparently; the byte-backed path preserves deferred source-map slicing over the inflated buffer.

## Review hardening

The review feedback is folded into the current commit:

- Version 2 explicitly rejects the compression flag at the header, while old decoders reject compressed version-3 bundles as an unsupported version.
- Declared inflated size is capped at 256 MiB; decode rejects oversized declarations, output beyond the declaration, and exact-size mismatches.
- The compressed encoder always closes/finalizes the DEFLATE writer, including after body-write failure, and preserves both write and close errors.
- `docs/guide/usage.md` documents `-z`.

## Validation

- Compression round trips through streaming and byte-backed decode paths.
- Tests cover header-level compatibility rejection, inflate bounds, exact-size validation, and compressor cleanup on failure.
- Current GitHub checks pass, including lint, race, generated artifacts, TinyGo/WASI builds, and CodeQL.

## Policy

Compression remains opt-in. Whether the embedded core should default to compressed is deliberately handled by #502 / #503, where the binary-size and startup-time tradeoff is measured.


mparrett:
Stacked follow-up compressing the embedded core (with the boot-cost measurement): #502.

mparrett:

nnunley:
Ran a multi-agent review over this — the format contract held up everywhere we poked it (plaintext magic/version/flags, old decoders fail loudly on a compressed bundle, all `.lgb` consumers go through the flag-aware decode paths). Two non-blocking hygiene notes for whenever the format discussion settles:

1. The inflate path is `io.ReadAll(flate.NewReader(...))` with no size bound. Not flagging it as a vuln — running an untrusted `.lgb` already means running untrusted bytecode, which can allocate whatever it wants — but a declared-uncompressed-size field + `io.LimitReader` is cheap, gives a better failure mode than OOM, and is easier to add now than after the format ships.
2. In the compressed encode path, an error from the body write/flush skips the flate writer's `Close()`, leaving a truncated stream in whatever the destination writer is. The decoder catches it loudly (missing trailer), so it's cosmetic today — write-then-rename at the callers would make it moot.

nooga (changes requested):
The compression logic itself works — round-trip tests pass, and old decoders reject compressed bundles cleanly with no panic (verified independently). Requesting changes on two things before this lands:

1. **Needs a rebase.** This branches off pre-#424 and still carries #424's original unsquashed commits; `main` has moved ~5000 lines since. Once you strip the #424 carry-over, the actual net-new content is small: `.gitignore`, the `-z` flag in `lg.go`, and `pkg/bytecode/{tags,encoder,decoder,compression_test}.go`. Worth rebasing onto current `main` so the diff reads as just the compression layer, which is what the PR description already claims it is.

2. **The old-decoder-rejects-cleanly behavior is incidental, not contractual.** `readHeader()` never validates the `flags` field against a known-flags mask, and `FormatVersion` stays at 2. An old decoder only fails because the DEFLATE bytes happen not to parse as valid tags/varints deep in string-table decoding — the resulting error ("reading code[9]: EOF") gives zero indication the real cause is a compression flag it predates. Once compressed bundles are out in the wild this matters a lot more than it does today. Please either bump `FormatVersion` for this flag, or add an explicit "reject unrecognized flag bits with a clear message" check in `readHeader`.

Two more from the existing thread that are still open, worth folding into this pass since you're touching this code anyway:
- Unbounded inflate in the byte-backed decode path (`decoder.go:217`, `io.ReadAll(flate.NewReader(...))` with no size bound) — decompression-bomb risk on a corrupted/adversarial bundle. A declared uncompressed-size field + `io.LimitReader` would close it cheaply.
- `Encode()`'s compressed path returns before `fw.Close()` on a `writeBody` error, leaving a truncated DEFLATE stream flushed to `w`. Same class of issue nnunley flagged on #502's `lgbgen` write path.

Also please update `docs/guide/usage.md` — it documents `-c`/`-b` but not the new `-z` flag.

mparrett:
Addressed the review pass in `82d25a3a` and refreshed the PR description to match the rebased implementation.

- Rebased onto current `main`; the old #424 carry-over is gone.
- Made compatibility contractual: compressed bundles are format v3, while plain bundles stay byte-identical v2. A v2 bundle with the compression flag is rejected at the header, and an older decoder rejects v3 as an unsupported version before parsing the body (`TestV2RejectsCompressionFlagAtHeader`).
- Added declared uncompressed-body size framing with a 256 MiB ceiling, a `declared+1` inflate bound, and exact-size validation. Tests cover oversized declarations and output beyond the declaration.
- The compressed encoder now always closes/finalizes the DEFLATE writer after a body-write error and joins write/close errors; this has a regression test.
- Documented `-z` in `docs/guide/usage.md`.

`go test ./...` passes locally, and the current GitHub checks are green (including lint, race, generated artifacts, TinyGo/WASI, and CodeQL).

@nooga, this should cover each requested change; please take another look when convenient. Thanks to @nnunley as well for the two hygiene catches.

nnunley:
Went through nooga's five items against the current head; all five check out. Single commit,
zero behind main, no #424 carry-over. `FormatVersion = 3` with `uncompressedFormatVersion = 2`
**and** the unknown-bit rejection in `readHeader` — he asked for either, this does both.
Inflate is bounded on both paths, the encoder always closes and joins the errors, and `-z` is
documented.

Three nits, none blocking:

1. `lg -version` now reports `lgb: format 3` while the default artifact is still version 2.
   The line's meaning changed from "what we write" to "the newest we read", which is
   defensible, but anything keying on it should be checked.
2. `-z` is accepted and silently ignored outside `-c`/`-b`. #624 validates its analogous
   `-strip` and exits 2; worth matching.
3. Re-encoding a decoded v3 bundle with `FlagCompressed` cleared emits a v3 *plaintext*
   bundle that older decoders reject for no reason, since `encodeFormatVersion` never
   normalizes downward. Sharp edge rather than a bug — the format tolerates it.

**The flag bit collides with #624**, which also takes `1 << 3`, for `FlagDebugSplit`. #624 is
the later PR so it is the one that moves, but I would rather propose the structural fix here,
because this PR already owns the machinery it depends on — `FormatVersion`, and `readHeader`
rejecting unknown bits per version.

The reason "just use `1 << 4`" is not enough: git does not catch the next one. Two branches each
adding a flag, one inside the const block and one as a standalone const further down — the shape
you and #624 actually have — merge **clean**, leaving two names on the same bit with no conflict
and no warning. I ran it to be sure.

Not writing the bit down removes the failure mode:

```go
const (
	FlagConstsBase   uint16 = 1 << iota // 1<<0
	FlagCapabilities                    // 1<<1
	FlagLocalVars                       // 1<<2
	FlagCompressed                      // 1<<3  (this PR)

	flagsEnd // first unused bit; keep last
)
const knownFlags = flagsEnd - 1
```

Values come out byte-identical to the current explicit shifts — I checked all five, including
#624's. A flag is then added by appending a name, and picking a duplicate becomes
unrepresentable rather than discouraged.

The second half is the part your `readHeader` mask is about. A derived `knownFlags` must not
become the *admitted* set, or appending a flag silently widens what v2 accepts:

```go
const (
	v1Flags uint16 = FlagConstsBase | FlagCapabilities
	v2Flags uint16 = v1Flags | FlagLocalVars
	v3Flags uint16 = v2Flags | FlagCompressed
)
```

which answers the nit I would otherwise have raised on its own: as written, every future flag
needs a coordinated edit to the supported set, and that edit is easy to forget. This makes
appending force the decision instead.

**The guard, and where it is enforced.** A pattern nobody checks decays, so this wants a test
rather than a comment. `pkg/bytecode/flags_layout_test.go`, stdlib `go/ast`, parsing its own
package:

| check | what it prevents |
|---|---|
| `flagsEnd` is the last entry in its block | an appended flag landing after the sentinel, so the derived mask and per-version sets stop covering it |
| no `Flag*` constant declared outside that block | exactly the #501/#624 shape — a flag added elsewhere, silently taking a bit already in use |
| every entry after the first carries no explicit value | someone reintroducing `= 1 << 3` and reopening the whole failure mode |

It needs **no workflow entry**: `go test ./...` already runs in the CI `test` job and in the
`go-test` pre-push hook, so adding the file is the whole wiring. That matters more than it
sounds — a `make` target no job invokes is documentation, not a gate.

I wrote it and checked it fails for the right reasons rather than trusting that it would.
Against a correct block, three tests pass; then, one violation at a time:

```
CAUGHT   flag declared outside the block (the #501/#624 shape)
         flag FlagDebugSplit is declared outside the flagsEnd block; a flag declared
         elsewhere can silently take a bit that is already in use
CAUGHT   explicit shift reintroduced inside the block
         flag FlagCompressed carries an explicit value; bits are positional and must be
         left to iota
CAUGHT   sentinel no longer last
         flagsEnd must be the last entry in its const block so appended flags take the
         next bit; found FlagDebugSplit last
```

Happy to hand over the file.

Happy to send this as its own small PR for you and #624 to rebase onto instead — we do not
depend on it, so it is yours to place.

Measured rather than guessed, since I raised it: linking `compress/flate` costs **+34,416 bytes
on `lg` (+0.17%)** and **+17,296 on `lg-runtime` (+0.10%)**, same toolchain, `main@8a8222d0` vs
this head. Not a concern. I did not measure the TinyGo/WASI targets, which link a different
runtime.

mparrett:
@nnunley Thanks for the follow-up. Your non-blocking suggestions and proposal are good. I'll include it here shortly.

mparrett:
Took all three nits plus the structural flag layout in `38888365`.

- `-version` now reports `lgb: format 2 (default write), 3 (max)`.
- Bare `-z` exits 2 with `error: -z requires -c or -b`, matching #624's `-strip` check.
- `encodeFormatVersion` picks the minimum version that admits the module's flags, so clearing `FlagCompressed` on a decoded v3 module re-emits version 2 plaintext.
- Flag bits are `1 << iota` with a `flagsEnd` sentinel and per-version admitted sets (`v1Flags` / `v2Flags` / `v3Flags`). `pkg/bytecode/flags_layout_test.go` gates the three failure modes you listed, plus `knownFlags == v3Flags` so appending a flag forces a version-set edit.

#624 can append `FlagDebugSplit` before `flagsEnd` and add it to the version set that admits it — no bit number to choose. I'll leave the rebase order to whoever lands next.

nooga (approved):
Approving. Everything from my 08-04 review is fixed: rebased, flags version-gated via the `flagsEnd`/iota scheme, inflate bounded at 256 MiB with exact-size checks, `Encode()` always closes the compressor (`errors.Join`), docs updated. Build/vet/`pkg/bytecode`/`test/e2e`/check-generated all pass locally; GitHub's CONFLICTING is the manifest only.

Please rebase onto current main, `make generate`, drop the "Draft." from the body, and I'll merge. Then #745 → #624 → #502 in that order.

PR #510, opened 2026-07-14
rt: generic JVM collection-interface interop (.valAt/.iterator/.assoc/.cons/.nth/.count) on let-go collections

### LINKED ISSUE #509: JVM collection-interface methods (.valAt/.iterator/.assoc/.cons/.nth/.count) don't dispatch on native collections
## Summary

JVM collection-interface methods called via the `.`-form on native let-go
collections don't dispatch — they raise `method-invoke expected Receiver`. This
blocks any Clojure library that reaches for `.valAt`, `.iterator`, `.assoc`,
`.cons`, `.nth`, `.count` (etc.) on its `:clj` reader-conditional branch.

## Repro

```clojure
(.valAt {:a 1} :a)          ; error: method-invoke expected Receiver
(.nth [10 20] 1)            ; same
(let [it (.iterator [1 2])] ; same (no Iterable/Iterator)
  [(.hasNext it) (.next it)])
```

## Why it's generally useful

These are the interfaces every let-go collection already implements internally
(`Lookup`, `Sequable`, `Associative`, `Collection`, `Indexed`, `Counted`), so a
generic handler in `invokeMethodFallback` — dispatched on the value's let-go
interface — makes the JVM method names work on **any** collection, not just the
ones a particular library touches. It's the same shape as the existing generic
`.reduce` fallback. Note `.cons` must map to append (Conj), **not** core `cons`
(prepend), so it needs to sit ahead of the name→fn fallback.

`.iterator` returns a small `java.util.Iterator`-like receiver exposing
`.hasNext`/`.next`.

## Motivation

metosin/malli uses `.valAt` in its map validator, `.iterator` pervasively in
`malli.impl.util/-vmap`, and `.hashCode`/`.longValue` to key its regex cache.

abogoyavlensky:
Resolves: #509

### DIFF
--- a/pkg/rt/host_interop.go
+++ b/pkg/rt/host_interop.go
@@ -105,6 +105,14 @@ func hostCollectionMethod(rec vm.Value, name vm.Symbol, args []vm.Value) (vm.Val
 	if rec == nil {
 		return vm.NIL, false, nil
 	}
+	// Scope this shim to real collection types. vm.String satisfies Counted/
+	// Indexed/Collection but with byte semantics (String.Count is len in bytes),
+	// while core count/nth are rune-based — and java.lang.String isn't
+	// clojure.lang.Counted anyway. Letting strings fall through keeps
+	// (.count "café") = 4 (core), not 5 (bytes). Same carve-out indexed? uses.
+	if _, ok := rec.(vm.String); ok {
+		return vm.NIL, false, nil
+	}
 	switch string(name) {
 	case "valAt": // clojure.lang.ILookup
 		if l, ok := rec.(vm.Lookup); ok {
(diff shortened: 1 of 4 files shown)


nnunley:
One severe catch: **`hostCollectionMethod`'s `count` case silently changes `(.count "str")` from rune count to byte count.** `vm.String` implements `vm.Counted`, and `String.Count()` returns `Int(len(l))` — bytes (`pkg/vm/string.go:46`). Before this PR, `(.count "s")` fell through to the name→fn fallback and hit core `count`, which special-cases strings as `len([]rune(...))` — runes. So `(.count "café")` goes from 4 to 5 with no error. It's also broader than the JVM contract being emulated: `java.lang.String` doesn't implement `clojure.lang.Counted`, so real Clojure throws there rather than answering. Excluding `vm.String` from the interception (the same carve-out `indexed?` uses in #508) — and probably from `nth`/`cons` for the same reason — keeps the shim scoped to actual collection types.

abogoyavlensky:
@nnunley Thanks for the comment, this has been fixed.

nnunley (approved):

PR #512, opened 2026-07-14
rt: add mutable java.util.HashMap and ArrayDeque host-compat types

### LINKED ISSUE #511: No mutable java.util.HashMap / java.util.ArrayDeque
## Summary

let-go has no mutable `java.util.HashMap` or `java.util.ArrayDeque`. Libraries
that use a JVM mutable collection internally on their `:clj` branch can't even
compile the constructor form.

## Repro

```clojure
(java.util.HashMap. 16 0.75)  ; CompileError: Can't resolve ->HashMap in this context
(java.util.ArrayDeque.)       ; Can't resolve ->ArrayDeque
```

## Why it's generally useful

Real mutable JVM collections are a common building block. Two small,
self-contained value types cover the vast majority of uses:

- `java.util.HashMap` — `.putAll` (accumulating/merging) + `.get`.
- `java.util.ArrayDeque` — `.push` / `.pop` / `.peek` / `.isEmpty` (stack-style).

Each is a dedicated `ValueType` + `Receiver`, following the existing
`clojure.lang.PersistentQueue` precedent, so `(type x)` and error messages stay
honest. (They're mutable JVM shims, kept in the `pkg/rt` compat layer rather than
`pkg/vm` alongside the persistent collections.)

## Motivation

metosin/malli's registry uses `(doto (HashMap. …) (.putAll m))` for its
fast-registry (on the critical validation path), and its regex engine uses an
`ArrayDeque` as the CPS backtracking stack.

abogoyavlensky:
Resolves: #511


nnunley:
One severe catch: **`putAll` silently succeeds without copying anything when the source isn't `vm.Sequable`** — and `*hostHashMap` itself has no `Seq()`, so `(.putAll m1 m2)` where `m2` is another `java.util.HashMap` is a silent no-op. Java's `putAll(Map)` copies or throws; here the caller gets success and an unchanged map, which is undetectable data loss. Returning an error for non-Sequable sources (or giving `hostHashMap` a `Seq()`) closes it. Minor adjunct while you're in the file: `h.m` is an interface field written by `putAll` and read by `get` with no synchronization, and let-go futures do run on goroutines — malli's build-once-then-read pattern avoids it today, but an `atomic.Value` would make it safe rather than accidentally safe.

abogoyavlensky:
@nnunley Thanks for the comment, this has been fixed.

nnunley (approved):

PR #514, opened 2026-07-14
rt: JVM static methods & host-class markers

### LINKED ISSUE #513: Common JVM static methods & instance? host-classes unresolved (createOwning, createWithCheck, System/arraycopy, Long/parseLong, …)
## Summary

A batch of common JVM static methods and `(instance? Class x)` host-classes are
unresolved, so libraries that reference them on their `:clj` branch fail to
compile. All can be backed by real let-go primitives.

## Repro

```clojure
(LazilyPersistentVector/createOwning (object-array 0))          ; Can't resolve
(clojure.lang.PersistentArrayMap/createWithCheck (object-array 4)) ; Can't resolve
(System/arraycopy src 0 dst 0 3)                                ; Can't resolve System/arraycopy
(Array/newInstance java.lang.Object 3)                          ; Can't resolve
(Util/hashCombine 1 2)                                          ; Can't resolve
(Long/parseLong "42")                                           ; Can't resolve
(UUID/fromString "…")                                           ; bare UUID unresolved
(instance? java.util.Map {})                                    ; Can't resolve java.util.Map
(instance? Pattern #"x")                                        ; bare Pattern unresolved
```

## Why it's generally useful

These are real, resolvable JVM statics/classes, each mapped to a let-go
primitive:

- `LazilyPersistentVector/createOwning` → `vec`
- `PersistentArrayMap/createWithCheck` → build a map from a flat array, throwing
  on a duplicate/odd key
- `System/arraycopy` → overlap-safe, range-checked element copy
- `Array/newInstance` → `object-array`
- `Util/hash` + `Murmur3/hashLong` → let-go `hash`; `Util/hashCombine` → the
  standard Clojure combine
- `Long/parseLong` / `Integer/parseInt` / `Float/parseFloat` / `Double/parseDouble`
  → `strconv`; bare `UUID/fromString` → reuse `vm.ParseUUID`
- `(instance? …)` markers: `java.util.Map`, `CharSequence`, bare `Pattern`

Static-holder namespaces are created **without** an auto-refer to `clojure.core`
(a small `defStaticNS` helper), so defining `Util/hash` does not print a shadow
`WARNING` at every `lg` startup.

## Motivation

metosin/malli reaches all of these across its registry, entry parser, regex
cache, and string coercion.

abogoyavlensky:
Resolves: #513

### DIFF
--- a/pkg/rt/host_jvm_statics.go
+++ b/pkg/rt/host_jvm_statics.go
@@ -120,11 +120,13 @@ func installJVMStatics(ns *vm.Namespace) {
 		if !ok {
 			return vm.NIL, fmt.Errorf("parseLong expects a string")
 		}
-		n, err := strconv.ParseInt(string(s), 10, 64)
+		// strconv.Atoi (returns int) + MakeInt matches let-go's own parse-long
+		// and avoids an int64->int conversion (CodeQL "incorrect conversion").
+		n, err := strconv.Atoi(string(s))
 		if err != nil {
 			return vm.NIL, err
 		}
-		return vm.Int(n), nil
+		return vm.MakeInt(n), nil
 	})
 	parseFloat := mustWrap(func(vs []vm.Value) (vm.Value, error) {
 		s, ok := vs[0].(vm.String)
--- a/pkg/rt/host_jvm_statics.go
+++ b/pkg/rt/host_jvm_statics.go
@@ -1,0 +1,158 @@
+/*
+ * Copyright (c) 2021-2026 Marcin Gasperowicz <xnooga@gmail.com>
+ * SPDX-License-Identifier: MIT
+ */
+
+package rt
+
+import (
+	"fmt"
+	"strconv"
+
+	"github.com/nooga/let-go/pkg/vm"
+)
+
+// defStaticNS returns (creating if needed) a bare namespace that holds host
+// static members (e.g. `Util/hash`, `Long/parseLong`). Unlike DefNSBare it does
+// NOT auto-refer clojure.core, so defining a member whose name overlaps a core
+// var (`Util/hash` vs `clojure.core/hash`) does not print a shadow WARNING at
+// every `lg` startup. These namespaces only carry static members; they never
+// resolve core names, so the missing refer is harmless.
+func defStaticNS(name string) *vm.Namespace {
+	name = resolveNSAlias(name)
+	nsMu.RLock()
+	if e := nsRegistry[name]; e != nil {
+		nsMu.RUnlock()
+		return e
+	}
+	nsMu.RUnlock()
+
+	ns := vm.NewNamespace(name)
+	nsMu.Lock()
+	nsRegistry[name] = ns
+	nsMu.Unlock()
+	return ns
+}
+
+// installJVMStatics registers JVM static methods and (instance? Class x) markers
+// that Clojure libraries reach on their :clj branches. Motivated by metosin/malli
+// (registry, regex cache, entry parser, string coercion) but each is a general,
+// real host-interop resolution backed by a let-go primitive.
+func installJVMStatics(ns *vm.Namespace) {
+	// --- (instance? Class x) host-class markers ---
+	// Map literals evaluate to vm.MapType (let-go.lang.Map), so match that.
+	RegisterHostClass("java.util.Map", vm.MapType)
+	RegisterHostClass("CharSequence", vm.StringType)
+	RegisterHostClass("Pattern", vm.RegexType) // bare java.util.regex.Pattern
+	RegisterHostClass("java.util.AbstractList", vm.Symbol("java.util.AbstractList"))
+	RegisterHostClass("java.util.Vector", vm.Symbol("java.util.Vector"))
+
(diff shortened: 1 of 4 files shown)


mparrett:
Thanks for the PR! Your approach is clean and has nice test coverage. The Go-level TestJVMStatics hits every shim.

There are two CI failures to address:

### 1. lint: staticcheck ST1005 (host_jvm_statics.go:76)
```
    return vm.NIL, fmt.Errorf("Duplicate key: %s", k)
```
Capitalized error string. This casing mirrors Clojure's "Duplicate key: …" message. If that parity is intentional, use `//nolint:staticcheck // matches JVM message`; otherwise lowercase it.

### 2. CodeQL high-severity: int64 to int narrowing (host_jvm_statics.go:127, parseLong result)
```
    n, err := strconv.ParseInt(string(s), 10, 64)
    ...
    return vm.Int(n), nil   // vm.Int is `type Int int`: platform width
```
I don't believe this is a false positive. On a 32-bit lane (e.g. TinyGo wasm target) `vm.Int(int64)` truncates. The codebase already has the right helper: `longCompatValue(v int64) vm.Value` (lang.go:8614) boxes as Int on 64-bit, BigInt on 32-bit, and is what Long/MAX_VALUE already uses. parseLong should return `longCompatValue(n)` which would fix CodeQL and make the shim correct on 32-bit.

## Worth a check - same class, CodeQL didn't catch

Two more sites have the identical int64 to int narrowing, just not on a taint path CodeQL tracks:
- :102 — `vm.Int(int64(vm.HashValue(...))): HashValue` is `uint32`, values > 2^31 overflow negative on 32-bit.
- :107 — `hashCombine` does int64 arithmetic then narrows.

On stock 64-bit Go all three are lossless, so not a bug on the primary target. But for consistency with parseLong's fix and 32-bit targets, they'd ideally box the same way. Lower priority.

**Reviewed shims:**
- `createWithCheck`: dup detection via before/after RawCount() is correct (a replaced key leaves count unchanged); odd-length + non-array guards present; test covers dup + odd.
- `System/arraycopy`: bounds-checked, reads whole source into a buffer before writing so overlapping same-array copies are correct (matches java.lang.System.arraycopy).
- `Array/newInstance`: ignores the class arg (fine, let-go arrays are untyped), negative-size guarded.
- `defStaticNS`: deliberately skips clojure.core auto-refer to avoid a startup shadow WARNING on Util/hash. Reasonable; the RUnlock/Lock TOCTOU is harmless (runs once at init).

**Minor non-blocking notes:**
- `Integer/parseInt` is aliased to parseLong (bitSize 64), so it accepts values beyond int32 that the JVM would reject. Probably ok for a compat shim.
- `hashCombine` isn't bit-identical to the JVM (Java wraps at 32 bits; this is int64, and >> 2 on a negative differs). Only matters if something depends on exact JVM hash values. malli uses it for internal caching, so within-let-go determinism is all that's needed.
- `ns.Lookup("vec").(*vm.Var).Deref()` will hard-panic if `installClojureCompatAliases` ever runs before core vec is defined. Ordering is currently fine (tests pass), but latent fragility.

pkg/rt/host_jvm_statics.go
```diff
@@ -0,0 +1,158 @@
...
+				return vm.NewObjectArray(int(n)), nil
+			}
+		}
+		return vm.NIL, fmt.Errorf("Array/newInstance expects (class, int)")
+	}))
+
+	// --- clojure.lang.Util/hash + Murmur3/hashLong -> let-go hash; Util/hashCombine
+	// -> the standard Clojure combine.
+	hashOf := mustWrap(func(vs []vm.Value) (vm.Value, error) {
+		return vm.Int(int64(vm.HashValue(vs[0]))), nil
```
mparrett, line 102: Note: `vm.Int(int64(vm.HashValue(...))): HashValue` is `uint32`, values > 2^31 overflow negative on 32-bit.

pkg/rt/host_jvm_statics.go
```diff
@@ -0,0 +1,158 @@
...
+	})
+	hashCombine := mustWrap(func(vs []vm.Value) (vm.Value, error) {
+		a, _ := vs[0].(vm.Int)
+		b, _ := vs[1].(vm.Int)
+		return vm.Int(int64(a) ^ (int64(b) + 0x9e3779b9 + (int64(a) << 6) + (int64(a) >> 2))), nil
```
mparrett, line 107: Note: `hashCombine` does int64 arithmetic then narrows.

abogoyavlensky:
@mparrett Thanks, the linting and CodeQL issues have been fixed.

mparrett (approved):
Both fixes look good and CI is green.

- `duplicate key` lowercase clears ST1005.
- `strconv.Atoi` + `MakeInt` sidesteps the int64 to int conversion entirely, so CodeQL's happy. Noting it trades behavior on a 32-bit lane: instead of silently truncating, a `>2^31` value now returns a parse error rather than round-tripping as a 64-bit long. This is fine for the 64-bit target; flagging since it's a behavior change vs the `longCompatValue`/BigInt route.

LGTM.

nnunley (approved):

PR #516, opened 2026-07-14
Host compat/compile stubs

### LINKED ISSUE #515: Compile-only stubs needed for JVM-only surfaces (java.time, java.util.concurrent, clojure.lang.IDeref/IFn)
## Summary

Some JVM-only surfaces can't run meaningfully under let-go, but a library still
references them on its `:clj` branch — and because the reference must *compile*,
an unresolved symbol blocks the whole namespace from loading. These need
**compile-only stubs**: forms that resolve so the namespace compiles, but fail
loudly / degrade if actually executed.

## Repro

```clojure
;; java.time formatter built at LOAD time via a method chain:
(-> (DateTimeFormatterBuilder.) (.appendPattern "…") (.toFormatter)) ; Can't resolve
(FutureTask. task)  ; Can't resolve ->FutureTask
TimeUnit/MILLISECONDS ; Can't resolve
;; a deftype implementing these interfaces won't compile:
clojure.lang.IDeref  ; Can't resolve
clojure.lang.IFn     ; Can't resolve
(BigDecimal. "1.0")  ; Can't resolve
(URI. "…")           ; Can't resolve
```

## Why it's the right shape

Real support would mean pulling in java.time / java.util.concurrent — out of
scope. Instead:

- **java.time**: a chainable stub whose builder methods (`appendPattern`,
  `toFormatter`, `withZone`, …) return self so the load-time construction
  threads; `.parse`/`.format` throw. `ChronoField/*` are markers.
- **Timeout / threads**: `FutureTask.`, `Thread.`, `TimeUnit/MILLISECONDS` as
  load-only loud stubs.
- **`clojure.lang.IDeref` / `IFn`**: registered as `Protocol`s so a `deftype`
  implementing them compiles (the generated methods are never dispatched — let-go
  uses `-deref`/`-invoke` — so such a value fails loudly if used).
- **`BigDecimal.` / `URI.`**: loud stubs (a caller's try/catch usually degrades
  these to a pass-through).

Each throws with a clear "not supported under let-go" message rather than
returning a plausible-but-wrong value.

## Motivation

metosin/malli's `malli.transform` builds `DateTimeFormatter`s at load time and
has date/decimal/URI coercion paths; `malli.impl.util/-run` has a bounded-timeout
helper; and `borkdude/dynaload`'s `LazyVar` `deftype` implements
`clojure.lang.IDeref`/`IFn`.

abogoyavlensky:
Resolves: #515


mparrett:
Nice. The compile-only-stub approach fits the problem: a form that resolves so the namespace compiles but fails loudly if run, rather than a plausible-but-wrong value. Feedback and review findings:

Two of the three CI failures are inherited and clear when #514's fixes are shipped, since `host_jvm_statics.go` is identical to #514:

- **ST1005**  capitalized `"Duplicate key"` string
- **CodeQL**  the `parseLong` int64 to int narrowing

**Specific to this PR:**

- **gofmt** `host_jvm_stubs.go:30`: the aligned method block (lines 30–33) pads with several spaces before `{`; gofmt doesn't align consecutive func declarations, so `gofmt -w` settles it.

The proposed mechanism passed validation. I traced the stub dispatch against the runtime:

- `chainStub` satisfies `Receiver` (`Value` + `InvokeMethod`), and `.method` dispatch in `lang.go` routes through `rec.InvokeMethod`, the same path `String`'s methods use, so the builder-methods-return-self threading works
  as intended.
- `NewProtocol(name, methods)` matches the `IDeref`/`IFn` registrations, and `theHostDateStubType` satisfies `ValueType`.

I didn't find any correctness issue in the stubs and the loud errors carry clear messages.

**Two test suggestions (non-blocking)**

- The `IDeref`/`IFn` test asserts the symbols resolve, but #515's scenario is a `deftype` implementing them compiling (dynaload's `LazyVar`). A one-line `deftype` compile-check would cover the case that was broken.
- Every stub is defined under both `Foo.` and `->Foo`, but the tests only hit the `Foo.` form. The `->` form is what `deftype`/`defrecord` construction emits, so assert the two resolve to the same value.

Minor: `theHostDateStubType.String()` returns `"java.time.stub"` while `chainStub.String()` returns `"#<java.time.stub>"`. Harmless, surfaces only in debug printing.

abogoyavlensky:
@mparrett Thanks, the linting and CodeQL issues have been fixed.

mparrett (approved):
Looks good. CI's green now.

- gofmt alignment fixed.
- Statics fix matches #514 (lowercased `duplicate key`, `Atoi`/`MakeInt` for the narrowing), so the same 32-bit-parseLong tradeoff carries over; fine for the 64-bit target.

The two test suggestions (a `deftype` implementing `IDeref`/`IFn` compiling, and the `->Foo` alias forms) are non-blocking, so no objection on moving this forward. LGTM.

nnunley (approved):

Issue #445, opened 2026-07-12
perf-pr A/B is ungateable as a single shot; median-of-N makes it a real gate


mparrett:
## Summary

`perf-pr.yml` runs a same-runner base-vs-head A/B and reports deltas but never
fails — the workflow comment notes a budget gate "could come later, once the
observed noise floor is known." I measured that floor by running the A/B on a
no-op PR (a comment-only change, so base and head are perf-identical and every
reported delta is pure noise), repeated on real `ubuntu-latest` runners.

Two findings:

1. **A single-shot A/B can't be gated.** Roughly 1 run in 4 blows a cluster of
   memory-bound micro-benchmarks to 25–40% while the anchor stays flat. To never
   false-fire on a genuine no-op you'd need a budget above 40%, which can't catch
   a real 30% regression.

2. **The anchor is register-only, so it's blind to the noise that matters.**
   `BenchmarkRatchetAnchor` is a pure-register loop with no memory traffic. It
   normalizes CPU-tier differences but not memory-bandwidth contention, and the
   benchmarks that blow out are the memory-bound ones.

The fix is variance reduction, not a bigger budget. **Median-of-N interleaved
A/B** turns the ungateable 40% floor into a gate at ~8%. N=5 is not enough (it
false-fired on 1 of 4 real runs); **N=7 at an 8% budget holds** with a
worst-family false-positive rate ≤1.5% and no per-benchmark carve-outs. A
memory-aware second anchor would remove even that residual and let the whole
fleet gate tight.

## Method

- A no-op PR: one comment added to `pkg/vm/bench_test.go`. The merge-base and the
  head are byte-identical in every benchmark, so a correct gate must report zero
  regressions. Any nonzero delta is measurement noise.
- Add the `perf` label, read the full 53-row `pr-fast` table (all four
  `pkg/vm -tags gogen_ir` families the profile covers), repeat to build a
  distribution. Deltas are `ratio_to_anchor` based, as the existing check is.
- All runs landed on quiet runners: the anchor moved ≤0.3% between base and head
  every time. So the noise below is *not* a case of the anchor catching a slow
  runner — it slips past the anchor entirely.

## Finding 1 — a single-shot A/B is heavy-tailed

Four independent same-code A/B runs:

| Run | runner | anchor Δ | max \|Δ\| | p95 \|Δ\| | families >5% |
|----:|--------|---------:|----------:|----------:|-------------:|
[...]

PR #517, opened 2026-07-15
perf(ci): interleaved repeat A/B as an informational shadow check

mparrett:
Implements step 1 from #445, in the shape the review there settled on: land the median-of-N as a shadow check first, not a gate. It also closes out the memory-aware second anchor (step 2) as a dead end.

## Why

The single-shot `perf-pr.yml` A/B is too heavy-tailed to gate on. On a no-op PR (base and head byte-identical in every benchmark, so every delta is pure noise) roughly 1 run in 4 blows a cluster of memory-bound micros to 25-40% while the register-only anchor stays flat. To never false-fire on a genuine no-op you'd need a budget above 40%, which can't catch a real 30% regression. A bigger budget can't fix a variance problem; more samples can.

The per-family median of N interleaved base/head snapshots absorbs up to ⌊(N-1)/2⌋ contaminated cycles. That turns the ungateable 40% floor into a would-gate in single digits.

## What it adds

- `scripts/ab_repeat.py` runs N interleaved base/head snapshots across two pre-built worktrees (each via `bench-ratchet -profile pr-fast ... snapshot`), then reports the per-family median delta and whether any family would gate at each candidate budget.
- `.github/workflows/perf-pr-repeat.yml` is opt-in via a `perf-repeat` label, N=7 default. It runs alongside the existing single-shot `perf-pr.yml` (untouched) so the two can be compared during the shadow phase.

## Informational, not a gate

The issue's step 1 said "turn on the gate." The review that followed, after a cross-check against the sibling Paserati workflow, narrowed that: accept the diagnosis, proceed with median-of-N, but shadow it before enforcing it. The specificity is studied; the power isn't, and the workflow-level false-positive rate hasn't been validated held-out at N=7. So this check never fails a PR on a perf verdict. Its exit code reflects only harness integrity (below). Enabling a required check is a follow-up, once the shadow data meets an explicit target.

## Non-obvious choices

- ABBA counterbalancing (odd cycles base-first, even head-first). A fixed base-then-head order lets first-vs-second-position drift (warmup, cache, thermal) masquerade as a consistent head regression, a systematic bias the median can't remove. Alternating cancels it. This came out of porting the same idea to Paserati, where a fixed order produced a 1-in-N false positive on a no-op PR.
- Exit code reflects integrity, never a perf verdict. Families are categorized rather than intersected: `comparable` (exactly N base and N head observations), `MISSING`/`NEW` (present on every cycle of one side only, i.e. a renamed or dropped benchmark, reported instead of silently vanishing), and `FLAKY` (present on some cycles but not all N). Flaky presence means the two halves aren't comparable, so the deltas can't be trusted, and that fails the job. A perf delta never does.
- The decision unit is the whole workflow. The check reports "would any family gate" at 6/8/10%, one Bernoulli sample toward `P(any-gate | no-op)`, rather than a per-family rate. A per-family rate undercounts the failure a maintainer sees, since any one of ~53 families crossing budget fails the check.
- Full evidence trail. Every base/head snapshot (with raw samples), cycle order, per-cycle anchor, CPU model, Go version, and SHAs are kept in the aggregate and uploaded, so the shadow-phase false-positive and power analysis can run offline.

## The memory-aware second anchor (issue step 2) is dropped

I prototyped it and it doesn't hold up, for a structural reason. `go test -bench` runs benchmarks sequentially, so the anchor and each family are measured seconds apart. Noisy-neighbor contention is non-stationary at that timescale, so dividing by the anchor injects its volatility rather than cancelling the family's. On one contended run, normalization made the worst family's |median| worse (4.09% to 13.23%). A working version would need concurrent measurement of anchor and family, a much larger harness change. Median-of-N alone is the fix.

## No-op validation

Labeling this PR runs a same-code A/B (it touches only `scripts/` and the workflow, no `pkg/**` or `cmd/**`), so every would-gate is a false positive. One no-op run on a fork (N=7, AMD EPYC 9V74, go1.26.4, ~49 min):

| budget | would-gate |
|---|---|
| 6% | 1 (`VectorConj/PersistentVector/10`, median +6.67%) |
| 8% | 0 |
| 10% | 0 |

Integrity: 53 comparable families, 0 missing, 0 new, 0 flaky. Anchor drift ≤0.33% per cycle (a quiet draw). This matches #445: N=7 holds at 8%, and the lone 6% straggler is `VectorConj`, one of the memory-bound families the issue flagged for exclusion or a wider budget.

The median doing its job on this run: `VectorConj/ArrayVector/100` had a worst cycle of +58.53%, a single contaminated snapshot that would hard-gate a single-shot A/B, which the median of 7 absorbed to +5.33%, under the 8% budget. One run is one sample, not the held-out validation a gate needs, but the mechanism behaves as designed on real CI.

## Next (not in this PR)

Per the review's path: collect gate-level false-positive and power data in shadow mode (P(any-gate | no-op) across runner draws, plus detection rate for injected 10/15/30% regressions), then enable a required check only after those numbers meet a target. Right-size N and the timeout from observed runtime.

### DIFF
--- a/.github/workflows/perf-pr-repeat.yml
+++ b/.github/workflows/perf-pr-repeat.yml
@@ -1,0 +1,140 @@
+name: Perf PR (repeat A/B)
+
+# INFORMATIONAL shadow check — a variance-reduced same-runner A/B. A single-shot
+# base-vs-head A/B on shared CI runners is too heavy-tailed to gate on (~1 in 4
+# runs blows a memory-bound cluster to 25-40% while the register-only anchor
+# stays flat; nooga/let-go#445). This runs N INTERLEAVED base/head snapshots
+# across two pre-built worktrees and reports the per-family median-of-N delta and
+# whether any family WOULD gate at each candidate budget. Interleaving (ABBA)
+# also cancels slow time-separation drift.
+#
+# It never fails the PR on a perf verdict — the job's exit code reflects only
+# harness integrity (a flaky / non-comparable measurement). The point of this
+# phase is to collect the false-positive and provenance data needed to calibrate
+# a real gate before enforcing one, per the review on #445. Runs alongside the
+# single-shot perf-pr.yml (separate `perf-repeat` label) so the two can be
+# compared during the shadow phase.
+
+# Opt-in via the `perf-repeat` label (same-repo PRs run the workflow from the PR
+# branch, so this need not live on the default branch). workflow_dispatch is kept
+# for the eventual upstream form, where it will sit on main.
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, labeled, unlabeled]
+    paths:
+      - 'pkg/**'
+      - 'cmd/**'
+      - 'scripts/ab_repeat.py'
+      - '.github/workflows/perf-pr-repeat.yml'
+  workflow_dispatch:
+    inputs:
+      pr:
+        description: 'PR number to benchmark'
+        required: true
+      n:
+        description: 'repeat count (interleaved cycles)'
+        default: '7'
+      budgets:
+        description: 'candidate budgets % to report would-gate for (comma-separated)'
+        default: '6,8,10'
+
+permissions:
+  contents: read
+
+concurrency:
+  group: perf-pr-repeat-${{ github.event.pull_request.number || github.event.inputs.pr }}
+  cancel-in-progress: true
+
+jobs:
+  bench:
+    if: github.event_name == 'workflow_dispatch' || contains(github.event.pull_request.labels.*.name, 'perf-repeat')
+    runs-on: ubuntu-latest
+    # N=7 × pr-fast (count 3 × 500ms) is 14 snapshots plus two `make lowered`
+    # builds. Measured ~49 min on a quiet EPYC 9V74 draw; the cap leaves headroom
+    # for a noisier runner, where the memory-bound families (and so the whole
+    # suite) run slower. Right-size from observed runtime during the shadow phase.
+    timeout-minutes: 90
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+          fetch-depth: 0
+          persist-credentials: false
+
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
+
+      - name: Resolve base and head
+        id: refs
+        env:
+          EVENT: ${{ github.event_name }}
+          PR_FROM_EVENT: ${{ github.event.pull_request.number }}
+          PR_FROM_INPUT: ${{ github.event.inputs.pr }}
+          BASE_FROM_EVENT: ${{ github.event.pull_request.base.ref }}
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set -euo pipefail
+          if [ "$EVENT" = "workflow_dispatch" ]; then
+            PR="$PR_FROM_INPUT"
+            [[ "$PR" =~ ^[0-9]+$ ]] || { echo "pr must be numeric" >&2; exit 1; }
+            base_ref="$(gh api "repos/${GITHUB_REPOSITORY}/pulls/${PR}" --jq .base.ref)"
+          else
+            PR="$PR_FROM_EVENT"
+            base_ref="$BASE_FROM_EVENT"
+          fi
+          git fetch --no-tags origin "$base_ref" "refs/pull/${PR}/head:pr-head"
+          head_sha="$(git rev-parse pr-head)"
+          base_sha="$(git merge-base "origin/${base_ref}" pr-head)"
+          echo "base=${base_sha}" >> "$GITHUB_OUTPUT"
+          echo "head=${head_sha}" >> "$GITHUB_OUTPUT"
+          echo "Base (merge-base): ${base_sha}"
+          echo "Head:              ${head_sha}"
+
+      - name: Build base and head worktrees
+        env:
+          BASE: ${{ steps.refs.outputs.base }}
+          HEAD: ${{ steps.refs.outputs.head }}
+        run: |
+          set -euo pipefail
+          git worktree add --force ../wt-base "$BASE"
+          git worktree add --force ../wt-head "$HEAD"
+          # lowered (gogen_ir) tree must exist in each worktree before bench.
+          ( cd ../wt-base && git submodule update --init --recursive && make lowered )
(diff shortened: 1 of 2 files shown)


mparrett:
Sample run:

```
=== interleaved repeat A/B — N=7, profile=pr-fast (INFORMATIONAL, never gates) ===
runner: AMD EPYC 9V74 80-Core Processor  4 CPU  go1.26.4
base 1a098b3e9007  head 1b0c86787352  order BH HB BH HB BH HB BH
anchor stability per cycle (base/head ns/op, should stay flat — it measures compute, not memory):
  cycle 1: base 1.407  head 1.406  Δ-0.0%
  cycle 2: base 1.407  head 1.407  Δ-0.0%
  cycle 3: base 1.407  head 1.412  Δ+0.4%
  cycle 4: base 1.412  head 1.407  Δ-0.4%
  cycle 5: base 1.408  head 1.412  Δ+0.3%
  cycle 6: base 1.408  head 1.408  Δ+0.0%
  cycle 7: base 1.407  head 1.411  Δ+0.3%

family                                    median%  worst% >prim
----------------------------------------------------------------
VariadicInvoke/5Args                        +2.31   -4.72     0
MapAssoc/GoMap-Lookup/100                   +1.00   +7.64     0
SeqIteration/ArrayVector/01000              +0.88  +14.16     1
SeqIteration/ArrayVector/10                 +0.85   +4.08     0
SeqIteration/PersistentVector/100           +0.53   +9.54     0
FuncInvoke/Native                           +0.34   +2.41     0
VectorCreation/ArrayVector/1000             +0.29  -19.28     0
MapAssoc/HAMT-Dissoc/100                    +0.27  +23.23     1
VectorConj/PersistentVector/100             +0.23  +11.00     1
FuncInvoke/Direct                           +0.17   +7.10     0
VariadicInvoke/1Arg                         +0.17   +4.23     0
VectorCreation/ArrayVector/10000            +0.15  +12.84     1
... 41 more (full table in aggregate.json)

  budget    6%:  clean  
  budget    8%:  clean  
  budget   10%:  clean  
confirm K=2/7 @ 10% (secondary): clean

(no-op PR: every would-gate above is a false positive; each is one sample toward P(any-gate).)
```

https://github.com/nooga/let-go/actions/runs/29386707875/job/87261410616?pr=517

nnunley:
Does the CI set GOMEMLIMIT? Have you checked to see how much alloc churn is happening in the build containers?

nnunley (approved):

nnunley:
Non-blocking hygiene note from a multi-agent pass (design otherwise verified: read-only `pull_request` perms, isolated worktrees, exit code tied to harness integrity not perf verdicts): a failed `go` subprocess or a bench-ratchet JSON shape change dies in `ab_repeat.py` with a raw traceback. The workflow's error capture contains it and the job is informational, so purely cosmetic — a try/except with a one-line "cycle N, side X failed" would just make red runs cheaper to triage.

mparrett:
> Does the CI set GOMEMLIMIT? Have you checked to see how much alloc churn is happening in the build containers?

Great questions. Right now the workflow sets neither `GOMEMLIMIT` nor `GOGC` so we're on defaults in the runner. I haven't yet profiled alloc churn in the build containers specifically.

The design leans on normalization rather than a pinned GC to keep the noise down. Each side is measured against its own per-run anchor (`ratio_to_anchor`), cycles are counterbalanced ABBA, and the reported number is the median over N=7. GC pauses that hit base and head roughly equally within a cycle divide out of the ratio, and the counterbalancing cancels order effects that a fixed base-then-head sequence would otherwise smear into a phantom head regression.

Pinning `GOMEMLIMIT`/`GOGC` would still shrink the residual cycle-to-cycle variance. I think this is a good lever... I'm thinking fold it into the held-out false-positive and power study on #445 (which needs to happen before this becomes a required gate) rather than bolt it onto the informational check here, since the ratio normalization already carries most of the weight. If you feel strongly about it, I'll consider adding it here. In either case, I'll make sure it's noted on #445.

mparrett:
> Non-blocking hygiene note from a multi-agent pass (design otherwise verified: read-only `pull_request` perms, isolated worktrees, exit code tied to harness integrity not perf verdicts): a failed `go` subprocess or a bench-ratchet JSON shape change dies in `ab_repeat.py` with a raw traceback. The workflow's error capture contains it and the job is informational, so purely cosmetic — a try/except with a one-line "cycle N, side X failed" would just make red runs cheaper to triage.

Done, thanks. `813613a` wraps the bench-ratchet subprocess and the JSON parse in `snapshot()`, so a failed `go run` or an unexpected JSON shape now prints one `::error:: <side> cycle N/M failed: ...` line instead of a raw traceback. It still exits nonzero afterward, since a missing sample is a harness-integrity failure, not a perf verdict. The run goes red, but regibly.

nnunley:
I thought we were using edn for results from bench-ratchet. Are we now mixed mode? Once again, non-blocking.

mparrett:
> I thought we were using edn for results from bench-ratchet. Are we now mixed mode? Once again, non-blocking.

Oof, this might have slipped- good catch. I'll address this in a follow-up. Taking a look now, actually.

mparrett:
> > I thought we were using edn for results from bench-ratchet. Are we now mixed mode? Once again, non-blocking.
> 
> Oof, this might have slipped- good catch. I'll address this in a follow-up. Taking a look now, actually.

Thanks again for the good question. Acknowledged non-blocking. I took a deeper look, and bench-ratchet has been JSON since it landed (#119): `baseline.json`, the `.jsonl` captures, the immutable snapshot `.json`. It has `-format json` report mode. This PR doesn't touch any of that; it only adds `scripts/ab_repeat.py` (+ the workflow), and that script reads the JSON `snapshot` already writes.

The EDN lives in the lg-side lane: `fanout-baseline.edn` / `ir-stress-baseline.edn`, produced by `scripts/fanout-ratchet.lg`. So the repo _is_ mixed-mode, but along a tool-language line that predates this PR: a Go tool serializes JSON, an `.lg` harness serializes EDN. #517 sits entirely on the bench-ratchet/Go side, so JSON matched the tool it wraps.

Your question does make me wonder, should we make the whole perf surface speak EDN? I'm open to it, but that'll be a larger effort we should track separately.

mparrett:
@nnunley Merged this to take an easy one out of the growing PR queue, but still interested in your thoughts on "whole perf surface speaks EDN" — happy to chat about that anytime.

PR #518, opened 2026-07-15
chore: gitignore .gomodcache and drop the committed toolchain lock

nnunley:
A GOMODCACHE pointed into the repo once leaked a Go toolchain download lock file (`.gomodcache/cache/download/golang.org/toolchain/@v/v0.0.1-go1.26.0.darwin-arm64.lock`) into version control, so the directory re-materializes in every fresh checkout/workspace. Ignores `.gomodcache/` and removes the stray file. No build impact — nothing reads it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -229,3 +229,6 @@ openspec/changes/*
 lg451
 lg458
 lgmain
+
+# Local Go module cache (GOMODCACHE pointed into the repo — never commit)
+.gomodcache/


mparrett (approved):
LGTM

- Confirmed deleted file is stray.
- Ignoring .gomodcache/ doesn't collide with the real cache (`GOMODCACHE=` elsewhere)
- Trailing slash pattern is correct, ignores the directory wherever it appears. (Note: if you wanted it anchored to repo root you'd want `/.gomodcache/` but I think your initial intent here is better)

Issue #504, opened 2026-07-14
TinyGo as the size-optimized runtime lane — findings, blocker, and the hard boundary


mparrett:
Part of the footprint & constrained-targets epic (#503). #466 adds the standard-Go WASI target and notes in `docs/guide/tinygo.md` that TinyGo isn't supported, blocked at init on TinyGo 0.41's incomplete `reflect`. This issue is the fuller picture: a prototype gets much further than the doc-note suggests, and the remaining boundary is specific.

## Why TinyGo

It's the size-and-memory lane that standard Go can't reach. On the same program (legmacs), a prototype build measured:

| target | artifact | resident memory |
|---|---|---|
| stock Go, native | 18.9 MB | — |
| stock Go, wasm (`GOOS=js`) | 7.3 MB | — |
| **TinyGo, wasm** | **1.5 MB** | — |
| stock Go, native `linux/arm64` | 17 MB | 11.5 MB |
| **TinyGo, native `linux/arm64`** | **9.9 MB** | **3.5 MB** |

~5x smaller wasm and ~3.3x less resident memory native, at a ~1.3x compute penalty. That's the case for embedding let-go on edge/constrained devices and in size-sensitive wasm, where stock Go's runtime is the ceiling.

## The init-time blocker, and a workaround

TinyGo 0.41's `reflect` is incomplete: the method-reflection used by `NativeFn.Box` (variadic detection and call assembly) traps at init, because native-fn boxing runs `MustBox` at package init and every bundle hits it on boot. Build-tag-splitting that init path — stubbing the reflect-based box so it errors only if invoked, rather than at load — clears the trap. With that plus a couple of goroutine-stack and output-routing fixes, a prototype legmacs **builds, boots, and takes input** under TinyGo on current `main`, in both lanes below.

TinyGo's `-target=wasm` is `GOOS=js`, so the existing browser host seam (the Go `HostWriter` output path and the key-source input ring) works unchanged — no new host shim.

## Hard boundary: 32-bit `int` on the wasm lane

[...]

Issue #503, opened 2026-07-14
Epic: Footprint & constrained targets


mparrett:
Umbrella for shrinking what let-go ships and widening where it runs: binary size, the embedding surface, and alternate runtimes (WASI, TinyGo, small native). Distinct from lowering/codegen (#258), which makes code native and fast, and from runtime performance (#464), which keeps the running VM's hot paths fast. Here the metric is bytes on disk, resident memory, and whether let-go links and boots on a given target at all. The one real intersection is boot cost: compressing the embedded core (#502) trades startup time for size, and that tradeoff is called out where it lands.

The goal is a single roll-up so footprint and target-reach work is visible and prioritizable, rather than living inside whichever PR happened to need it.

## Compiler-free / embedding surface

Running precompiled bytecode without linking the reader, compiler, and resolver: a smaller binary and a smaller, more stable embedding surface.

- **#424** runtime-only build: a `lg-runtime` that executes precompiled `.lgb` with no compiler in the binary.
- **#499** `rt.BootCore()`: compiler-free `clojure.core` load, so an AOT-lowered or embedded binary can reach the `.lg`-defined core without the compiler's eval path.

## Bundle size

The `.lgb` bundle format and the embedded core are stored uncompressed; compressing them trims the shipped artifact and every binary.

- **#500** tracking issue: `.lgb` bundles carry no compression (3–4.4x headroom).
- **#501** opt-in DEFLATE compression for program bundles (`lg -c -z` / `-b -z`).
- **#502** compress the embedded `core_compiled.lgb` (the universal per-binary win), with the boot-cost measurement.
- **#631** strip debug sections (source maps + local-var tables) from the embedded core and `-w` program bundles — the complementary lever to #502 (removes the data rather than compressing it; the two compose).

## Alternate runtimes & platforms

Targets beyond native and browser-`js`, where footprint, the host capability model, and cross-target build discipline matter most.

- **#466** build as a WASI (`GOOS=wasip1`) module: standalone wasm, host gated capabilities, standard-Go 64-bit `int`.
- **TinyGo**: the size-optimized runtime lane (much smaller wasm, lower native RSS), and the boundary it hits. Tracking issue: #504.
[...]

Issue #464, opened 2026-07-13
Epic: Runtime performance & concurrency


nnunley:
Umbrella for keeping the running VM fast and thread-safe — the hot paths of the
bytecode interpreter and its shared runtime data structures, as distinct from
lowering/codegen (#258) or Clojure-semantics parity (#257). Behavior is
unchanged under these; only performance and concurrency correctness move.

The goal is a single roll-up so runtime regressions and concurrency-hygiene
fixes are visible and prioritizable, rather than living only inside the PR that
happened to fix them.

## Known runtime regressions

- **`Var.Deref` ~8× regression on the perf timeline.** The #203/#210
  structured-concurrency rework routed every `^:dynamic` var deref through a
  walk of the single shared `globalBindingStack`, regressing serial deref and
  contending the shared head under parallelism. No issue was filed at the time;
  this epic is its home.
  - **#459** — Phase 1: per-var root-binding gate; unbound/non-dynamic vars skip
    the shared stack (lock-free single atomic load).
  - **#462** — Phase 2: per-var binding chains + membership registry → **O(1)**
    bound-var deref regardless of how many other vars are bound; kills the
    shared-head contention.

## Concurrency-correctness fixes

- **#463** — `ArrayVectorSeq.ChunkedFirst` data race: it wrote a scratch field
  on a goroutine-shared persistent seq node. Introduced by #397's chunked
  optimization; `-race` isn't run on `pkg/vm` in CI, so it slipped in.

## Workflow

File runtime performance and concurrency issues here as sub-issues. Each child
should stand on its own (repro or benchmark + path to fix); this epic tracks the
set and its progress. Perf regressions surfaced by the
`nooga.github.io/let-go/perf` timeline belong here when the cause is in the
running VM rather than the lowering pipeline.

PR #522, opened 2026-07-16
feat: TinyGo support: small-footprint WASM and native builds

mparrett:
# TinyGo support: small-footprint WASM and native builds

Part of #503 (footprint & constrained targets). Implements the TinyGo lane whose scope, measurements, and 64-bit-`int` boundary are laid out in #504 - this is the concrete PR that issue's A sk invited.
Draft: it's the proposal for the scope decision, not a merge request yet.

## Why

Compiled with TinyGo, let-go fits a niche the stock Go build doesn't: much smaller binaries and a fraction of the memory, for two audiences —

- **the browser** — WASM bundles small enough to ship inline, and
- **constrained native targets** — edge, embedded, Raspberry Pi-class devices where RSS is the binding constraint.

Measured on this branch (linux/arm64):

| | Stock Go | TinyGo |
|---|---|---|
| `lg` binary (stripped) | 13 MB | **9.9 MB** (3.0 MB gzipped) |
| RSS, trivial eval | 11.5 MB | **3.5 MB** |
| Cold start | ~20 ms | ~10 ms |

WASM bundles shrink further: a self-contained editor app goes from 7.3 MB (stock) to **1.5 MB** (TinyGo).

## What's in it

The changes are build-tag-gated (`//go:build tinygo`), so the stock build is untouched.

- **Build gates + minimal stubs** for the subsystems TinyGo's stdlib can't provide: an `os` namespace subset, a `net/http` stub (TinyGo's `net/http` doesn't link), syscall and reflect shims. Existing portable `_other.go` fallbacks are reused where possible, so the gates cover both the WASM (`GOOS=js`) and native (`GOOS=linux`) TinyGo paths.
- **Reflect-limitation shims.** TinyGo leaves several `reflect` operations unimplemented (they panic): `Type.IsVariadic`/`NumIn`/`In` and `Value.Call`. let-go boxes typed Go functions and dispatches Go methods through those, which trapped at boot. This splits the affected paths so TinyGo installs non-reflecting equivalents:
  - `NativeFn.Box` installs a stub that returns a clear error only if a reflect-boxed function is invoked, instead of trapping every program at startup.
  - boxed-method dispatch hand-registers the small set of methods programs reach for via the `.` operator (currently `time.Time.Sub`). An unshimmed method errors at the call site naming TinyGo's limitation, rather than reporting the type has no methods.
- **A pure-Go hash** backing the `xxh3` namespace under TinyGo. The `xxh3` dependency is assembly-only on arm64/amd64 with no pure-Go build path, so TinyGo can't link it; this substitutes an FNV-1a + splitmix64 hash. It is **not** bit-compatible with xxh3 — see limitations.
- **Bundler support** in the `-w` (WASM) path: a `LETGO_USE_TINYGO=1` switch that invokes TinyGo with a configurable stack size, and `LETGO_TINYGO_PANIC` to surface boot panics during debugging. (A stack-size override is required: the interpreter's recursion overflows TinyGo's default goroutine stack.)

## Validation

- The `lg` CLI builds and runs under native TinyGo (linux/arm64) — evaluates expressions, the `xxh3` namespace works, benchmarks pass.
- Two real apps build and run to WASM under TinyGo: a terminal editor (boots, renders, takes input) and a terminal roguelike (title, seeded world-gen, dungeon render, movement, combat).
- Benchmark suite, native, best-of-3 vs the stock build:

  | Workload | Stock | TinyGo | Ratio |
  |---|---|---|---|
  | fib (tree-recursive) | 1.995 s | 3.712 s | 1.86× |
  | tak (recursive) | 1.994 s | 3.979 s | 2.00× |
  | loop-recur (1M) | 0.080 s | 0.085 s | 1.06× |
  | map-filter pipeline | 0.026 s | 0.013 s | 0.50× |
  | reduce (1M) | 0.079 s | 0.088 s | 1.11× |
  | persistent-map (10k) | 0.038 s | 0.030 s | 0.79× |
  | transducers (100k) | 0.059 s | 0.083 s | 1.41× |

  Sustained recursive compute is ~2× slower (TinyGo's call overhead and simpler optimizer); short-lived and data-pipeline work is competitive or faster, because the leaner cold start dominates short runs. The profile fits startup-bound, memory-light workloads.

## Known limitations

- **WASM `int` is 32-bit.** TinyGo's wasm target uses a 32-bit `int` (stock Go's wasm uses 64-bit). let-go's `Int` follows the platform, so 64-bit integer work overflows or truncates on WASM — checked arithmetic errors, wrapping arithmetic wraps at 2³². Native TinyGo (arm64/amd64) is unaffected (64-bit). Programs doing 64-bit hashing/bignum work need care on the WASM target.
- **`xxh3` is not bit-compatible under TinyGo.** The substitute hash is deterministic and well-distributed but produces different digests, so a TinyGo build is its own determinism domain (a seed won't reproduce a stock build's output). Fine for standalone deployments; not for cross-build reproducibility.
- **No Unix domain sockets, limited `os.ProcessState`.** TinyGo's stdlib lacks both, so the nREPL server (Unix-socket transport) and some process introspection are unavailable natively.
- **Native macOS doesn't link** (upstream TinyGo, tinygo-org/tinygo#4794 — Darwin routes syscalls through libSystem assembly TinyGo can't link). Linux is the native target.

## CI

A `tinygo-wasi-build` job builds the runtime-only binary and boots it under wasmtime. The reflect trap is a runtime failure, so unlike the existing build-only `wasip1-build` guard this job runs the module (`-v` exercises the full core boot) to catch a regression.

## Concurrency boundary

This lane is single-threaded — wasm has no threads, `runtime.NumCPU()`/`GOMAXPROCS` are 1 — so goroutines interleave but don't run in parallel (the stock-Go wasm build included). The cooperative channel surface holds up on native Go channels (`go*`, `>!`/`<!`, `chan`, `timeout`, `pipe`, `promise-chan`), but `alts!`/`alts!!` hit `reflect.Select`, which TinyGo doesn't implement, so they panic — the same reflect gap as interop. And `pmapv`/`NumCPU` fanout degrade to sequential (correct results, no speedup). Documented in `docs/guide/tinygo.md`; a fuller cross-target treatment is left to the parallel-lowering note (#464).

## Notes for review

The commits are organized by concern and can be read/merged in logical order. The build gates and the stock build are independent — nothing here changes behavior for a normal `go build`. Happy to split this into smaller PRs (WASM enablement, reflect shims, native gates, hash) if that's easier to review.


nnunley (commented):
Ran a multi-agent review over the branch — architecture looks sound (the build-tag split for TinyGo's reflect subset is clean), and nothing critical is introduced by this PR. One theme worth addressing before merge: **the new tinygo-tagged code fails silently in several places**, which will make TinyGo builds painful to debug.

- `pkg/rt/os_tinygo.go` (~26, 38) and `pkg/vm/boxed_reflect_tinygo.go`: `Wrap()` error returns are discarded, so a broken builtin installs silently. Same pattern in `wasm.go`'s `fd_write` regex patch of `wasm_exec.js` — if TinyGo ever reshuffles that file, the patch no-ops and console output just disappears with no diagnostic. A returned error (or at least a loud warning) in these spots would save future-you an afternoon.
- Related nits in the same vein: `getenvStub` unconditionally returns `""` (worth a comment or a warning on first call so users know env vars are dead on TinyGo builds), and `LETGO_TINYGO_STACK/PANIC/GC` are passed through to tinygo unvalidated, so a typo surfaces as an opaque tinygo error rather than a let-go one.

Separate from this PR: the review also confirmed two pre-existing data races in `pkg/vm/boxed.go` (the unsynchronized `BoxedTypes` map and the non-atomic `Hash()` cache). Your diff only relocates code around them, so no action needed here — I'll file it as its own issue.

pkg/rt/os_tinygo.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+// os.go's init() that registers this installer is //go:build !tinygo, so the
+// tinygo build must register its own — without this the os namespace is never
+// installed and os/getenv, os/exit, os/args resolve to nil (e.g. xsofy's
+// seed boot-param crashed "nil is not a function" calling os/getenv).
+func init() { RegisterInstaller(installOsNS) }
+
+func installOsNS() {
+	exitFn, _ := vm.NativeFnType.Wrap(func(vs []vm.Value) (vm.Value, error) {
```
nnunley: `Wrap` error discarded — if wrapping fails, `exit` installs as a nil/broken fn and the failure only surfaces at call time with a confusing error. Since this runs at namespace init, a `panic(err)` (or logging + skipping the Def) would localize it. Same pattern at line 38.

pkg/rt/os_tinygo.go
```diff
@@ -0,0 +1,57 @@
...
+		code, ok := vs[0].(vm.Int)
+		if !ok {
+			return vm.NIL, fmt.Errorf("os/exit expected Int")
+		}
+		os.Exit(int(code))
+		return vm.NIL, nil
+	})
+
+	getenvStub, _ := vm.NativeFnType.Wrap(func(vs []vm.Value) (vm.Value, error) {
+		return vm.String(""), nil
```
nnunley: Unconditionally-empty `getenv` will read as "env var not set" rather than "env vars unsupported on TinyGo" to any .lg code probing configuration. TinyGo's WASI target does support `os.Getenv` — if that works under your build config it'd be strictly better; otherwise a one-time stderr warning (or a doc comment here explaining why not) would prevent a silent-misconfiguration hunt.

pkg/vm/boxed_reflect_tinygo.go
```diff
@@ -0,0 +1,58 @@
...
+	"reflect"
+	"time"
+)
+
+var timeType = reflect.TypeOf(time.Time{})
+
+func reflectMethods(t reflect.Type) map[Symbol]*NativeFn {
+	if t == timeType {
+		// (.Sub a b) — method value semantics: receiver is the first arg.
+		sub, _ := NativeFnType.Wrap(func(vs []Value) (Value, error) {
```
nnunley: Same discarded `Wrap` error as os_tinygo.go:26 — broken method installs silently under the tinygo tag.

wasm.go
```diff
@@ -132,14 +133,47 @@ func buildWasm(ctx *compiler.Context, nsRes *resolver.NSResolver, src string, ou
...
+	var build *exec.Cmd
+	if useTinyGo {
+		fmt.Println("building wasm with tinygo...")
+		// -stack-size: TinyGo's wasm target auto-sizes goroutine stacks and falls
+		// back to a 64KB default it can't bound against let-go's unbounded
+		// interpreter recursion (Frame.Run→Func.Invoke→Frame.Run). Deep recursion
+		// overflows the stack into adjacent static memory (the reader's macros map),
+		// which manifested as the "3-minute hang". Override via LETGO_TINYGO_STACK
+		// for sweeping. See docs-xsofy/tiny-let-go-the-stack-overflow-hunt.md.
+		stack := os.Getenv("LETGO_TINYGO_STACK")
```
nnunley, line 152: `LETGO_TINYGO_STACK` / `LETGO_TINYGO_PANIC` / `LETGO_TINYGO_GC` pass straight through to the tinygo invocation, so a typo (`64kb`, `pring`) surfaces as an opaque tinygo flag error. Minor, but a quick validate-or-warn here would attribute the mistake to the env var by name.

nnunley (commented):
Following up on my earlier summary with the same items anchored in context (plus one correction: I claimed the `fd_write` patch fails silently — it doesn't, `patchTinyGoStdout` already warns on stderr when the regex misses. Withdrawn.)

mparrett:
@nnunley Thanks for the thorough pass; glad the build-tag split held up.

Addressed all four items in `91952e7` (on the rebased branch):

- Discarded `Wrap` errors (`os_tinygo.go`, `boxed_reflect_tinygo.go`): both now fail loud at namespace init. `os/exit` and `os/getenv` go through the existing `mustWrap`, and `.Sub` panics on a wrap error instead of installing a nil method. Now, a broken builtin surfaces at init with the cause rather than at call time.
- `os/getenv`: you were right this was more than a nit. It ignored the key and returned `""` for everything, so env-based config (xsofy's seed boot-param included) was silently dead on TinyGo. It's now wired to real `os.Getenv`, verified working on the WASI/native targets; the wasm target has no environ, so it still returns `""` with no regression.
- `LETGO_TINYGO_STACK/PANIC/GC`: now validated with a warn-and-pass-through, so a typo is attributed to the env var by name instead of surfacing as an opaque tinygo flag error. I kept it a warning rather than a hard fail so a valid value a newer tinygo adds still works.

**Other notes:**
- Acknowledge the withdrawal of the `fd_write` / `patchTinyGoStdout` point. No change there.
- The two `boxed.go` races were addressed in #544
- Rebased the branch onto main (it had gone conflicting). The only conflict was the regenerated `interop_xxh3.go` header. Resolved to keep #522's hand-added `!tinygo` build tag on top of main's `-opaque-structs` regeneration.

wasm.go
mparrett: Thanks. Addressed in 91952e7

pkg/vm/boxed_reflect_tinygo.go
mparrett: Addressed in 91952e7

pkg/rt/os_tinygo.go
mparrett: Addressed in 91952e7

pkg/rt/os_tinygo.go
mparrett: Addressed in 91952e7

pkg/rt/interop_xxh3_tinygo.go
```diff
@@ -0,0 +1,116 @@
...
+ * xxh3/HashSeed); it is not for security or cross-implementation reproducibility.
+ * The namespace keeps the `xxh3` name so callers (xsofy) build unchanged.
+ */
+
+package rt
+
+import "github.com/nooga/let-go/pkg/vm"
+
+// fallbackHash64 mixes bytes into a seed. FNV-1a absorbs the input; a
+// splitmix64/Murmur3 finalizer then diffuses it so small input changes avalanche
```
nnunley: you probably should publish this as murmur3, rather than xxh3.  xxh3 is a specific hashing algorithm.

pkg/rt/interop_xxh3.go
```diff
@@ -1,4 +1,10 @@
+//go:build !tinygo
```
nnunley, line 1: Seems like we should add issues to:
1. add support for tinygo based lginterop
2. add support for go annotations from the lginterop command and/or the defs.edn file. Since this will get nuked the next time someone regenerates the bindings.

pkg/vm/boxed_reflect_tinygo.go
```diff
@@ -0,0 +1,63 @@
+//go:build tinygo
+
+/*
+ * TinyGo does not implement reflect.Type.Method / reflect.Value.Call, so
```
nnunley, line 4: So... Because lginterop has the full signatures of the go-module, you should be able to generate stubs for tinygo to be filled in later.

The -native route should actually able to import pure go packages, and generate the delegating calls for you.

nnunley (changes requested):
So... While I think this is good, I have concerns with editing generated files, given that it's going to be nuked the next time someone regenerates bindings.

mparrett:
@nnunley Both points taken; addressed in f76767d, plus the two issues.

**murmur3, not "xxh3":** agreed — the fallback shouldn't claim an algorithm it isn't, and on reflection the `xxh3` name wasn't worth keeping under TinyGo at all. The fallback is now a real MurmurHash3 (x64-128, low word) published as a `murmur3` namespace, and TinyGo builds have no `xxh3` namespace — callers that want the portable hash call `murmur3/*` (we'll switch xsofy's TinyGo lane accordingly). `murmur3` is registered on all builds, not just TinyGo: that lets the stock suite exercise it (verified against reference vectors from python `mmh3` and `twmb/murmur3`), and gives programs one hash with identical values on stock Go, TinyGo native, and TinyGo wasm.

**Editing generated files:** filed the two issues you asked for — #576 (lginterop generates TinyGo-compatible bindings from the signatures it already has, folding in your delegating-calls suggestion) and #577 (build-constraint pass-through from the lginterop command / a defs entry, so the `!tinygo` tag survives regeneration). Until #577 lands, this branch carries a guard: `TestInteropXxh3KeepsTinygoBuildTag` fails in the stock suite if a regeneration drops the hand-added tag. The tinygo-wasi CI job would catch that too, but as a runtime trap at namespace install with nothing pointing at the build tag; the guard names the cause at ordinary test time.

A bonus fix from the cross-width validation: seed args above 2^31-1 read as BigInt on TinyGo wasm (32-bit int) but Int on stock builds, so the same seed literal worked on one and type-errored on the other. `hashSeedArg` now accepts both; `(murmur3/HashStringSeed "xsofy" 3735928559)` returns the same value on stock native and TinyGo wasi.

pkg/rt/interop_xxh3_tinygo.go
mparrett: Done in f76767d — real MurmurHash3 (x64-128, low word) now, published as `murmur3`; the `xxh3` namespace is gone under TinyGo.

pkg/rt/interop_xxh3.go
mparrett: Filed: #576 (TinyGo-compatible lginterop bindings) and #577 (build-constraint pass-through). Until #577 lands, f76767d adds a stock-suite guard test that fails if a regeneration drops the tag.

pkg/vm/boxed_reflect_tinygo.go
mparrett: Folded into #576.

mparrett:
Two follow-up commits from running a full game (xsofy) in the browser on this branch, end to end:

- `5731ba8` — a registry for hand-shimmed boxed methods (`vm.RegisterBoxedMethods`), plus `rt` registering `LGWriter` `WriteString`/`String`/`Flush`. Embedded core's `print-method` calls `(.WriteString w …)` on every collection print, so `pr-str`/`pprint` of any map or vector hit the fail-loud method error at call time. The registry exists because `pkg/vm` can't special-case `rt` types without an import cycle; new shims are now one map entry from any package.
- `fe22045` — `LETGO_TINYGO_OPT` to override the hardcoded `-opt=z`, completing the `LETGO_TINYGO_*` family. For the record, `z` is the right default: in an A/B on the game, `-opt=2` cost ~600 KB of bundle and changed nothing measurable at runtime.

One note for rebase time rather than another commit now: current main's `pkg/bundle` (the #234 extraction) calls `os.SameFile`, which TinyGo's `os` doesn't implement, so this branch stops compiling under TinyGo once rebased. I have the fix staged (a build-tagged `sameFile` helper in the branch's existing pattern) and will fold it into the rebase.

CI-relevant validation on the branch as pushed: both test modes green, the wasi runtime boots under wasmtime, and the full TinyGo build round-trips `pr-str` of nested collections.

nnunley (approved):
Reviewed via multi-agent lens pass across all 29 files, focusing on build-tag counterpart coverage. Verified: no missing symbol definitions across tag combos, term_pty tag logic correct, reflect/os shims panic at init rather than silently no-op'ing at call time, and CI exercises the TinyGo bootstrap under wasmtime. murmur3-vs-xxh3 hash non-determinism across builds is documented, not silent. Suggestions only (env-var docs, shimmed-type comments) — non-blocking. LGTM.

Issue #523, opened 2026-07-16
Discussion: raw TCP/UDP sockets — a `net` namespace


mparrett:
## Motivation

let-go already speaks the network at the HTTP layer: `http/serve` wraps `net/http`, and the client side (`http/get`, `http/post`, `http/request`) wraps `net/http`'s client. What's missing is the layer below — raw TCP and UDP.

Raw sockets unlock a class of programs that HTTP can't express: custom line/binary protocols, TCP proxies, small daemons, a Redis-style server, a game or chat server, anything speaking a wire format that isn't HTTP. Go's `net` package makes the underlying support cheap to expose, so the open question is the API shape and the constraints, not the plumbing.

Filing this as a discussion because the design has a few forks worth settling before anyone writes code.

## Sketch

A `net` namespace, native-only, roughly:

- `(net/listen "tcp" ":7070")` → a listener object
- `(net/accept listener)` → a connection object (blocks)
- `(net/dial "tcp" "host:port")` → a connection object
- connections are readable/writable and closeable, with deadlines
- UDP as datagram send/recv on a bound socket

Connections would be boxed values wrapping Go's `net.Conn` / `net.Listener`. let-go already has boxed readers (`newLGReader`, used for the HTTP client's `:as :stream`), so the read side has a precedent to build on.

### 1. Connection object, not fire-and-forget

`http/serve` is `http.ListenAndServe` under the hood: it blocks forever, returns no handle, and hides the socket entirely behind the handler. That's the right call for HTTP. It's the wrong call for raw sockets, where the connection *is* the thing the program holds — you need explicit read/write/close and deadlines, not a callback that owns the loop.

So the proposal is a first-class connection/listener object with read/write/close/set-deadline, rather than another blocking `serve`. This is the main design decision and the reason to talk before building.

### 2. WASM is a hard constraint

Go's `net` doesn't exist under `GOOS=js` or `wasip1`. let-go also runs in the browser via WASM, so a `net` namespace has to be build-tagged native-only: absent from the WASM build, and any core `.lg` that `require`s it would break the WASM bundle. That gating needs to be designed in from the first commit, not bolted on. Worth deciding up front how a native-only namespace should present itself when it's simply not there (missing var vs. a stub that errors on use).

### 3. Blocking and concurrency

`accept` and `read` block. How that composes with let-go's concurrency story (goroutines? a thread per connection? something else) decides whether the API feels natural or fights the runtime. This is the part I have the least conviction on and most want the team's read on.

[...]

PR #526, opened 2026-07-16
rt: add net and bencode namespaces (TCP client + bencode framing)

### LINKED ISSUE #525: Add net and bencode namespaces (TCP client + bencode framing over a conn)
Add two small, generic namespaces so pure `.lg` can speak **bencode-over-TCP** (what nREPL uses). Today lg's lg-facing stdlib has AF_UNIX sockets (`unix/connect`), HTTP, and pods over stdio — but no TCP client, so an nREPL client can't be written in pure `.lg`.

**`net`** (portable, no build tag; patterned on `pkg/rt/unix_linux.go`):
- `(net/dial host port)` → conn, 3s connect timeout
- `(net/write! conn str-or-bytes)` → nil (raw bytes)
- `(net/read! conn max-bytes)` → byte-array, nil on clean EOF
- `(net/close! conn)` → nil

**`bencode`** (over a `net` conn; persistent per-conn `*bencode.Decoder`):
- `(bencode/write! conn value)` → nil
- `(bencode/read! conn)` → decoded value, nil on clean EOF
- `(bencode/read! conn timeout-ms)` → same, with a "timeout" error via SetReadDeadline

Conversion (both directions): String ↔ byte string, keyword → name, Int ↔ int, vector/list ↔ list, map (String/keyword keys) ↔ dict with String keys; nil/bool/float error on write.

Reuses the vendored `github.com/zeebo/bencode`. Go tests cover dial/read/write, bencode round-trips over a real TCP pair, conversion edge cases, and timeout recovery (`go test ./pkg/rt -race` clean). A branch is ready (`pkg/rt/net.go`, `pkg/rt/bencode.go`, `pkg/rt/net_test.go`).

abogoyavlensky:
Resolves: #525

### DIFF
--- a/pkg/rt/bencode.go
+++ b/pkg/rt/bencode.go
@@ -1,0 +1,221 @@
+/*
+ * Copyright (c) 2026 let-go contributors
+ * SPDX-License-Identifier: MIT
+ */
+
+/*
+ * bencode namespace — message framing over a net conn.
+ *
+ *   (bencode/write! conn value)       → nil. Encodes one bencode value.
+ *   (bencode/read! conn)              → decoded value, nil on clean EOF. Blocks.
+ *   (bencode/read! conn timeout-ms)   → same, but errors with a message
+ *                                       containing "timeout" if nothing arrives
+ *                                       in time (SetReadDeadline, cleared after).
+ *
+ * Value conversion (pinned both directions):
+ *
+ *   lg → bencode                         bencode → lg
+ *   String        → byte string          byte string → String
+ *   keyword       → its name             int         → Int
+ *   Int           → int                  list        → vector
+ *   vector/list   → list                 dict        → map with String keys
+ *   map           → dict (String/keyword keys)
+ *   nil/bool/float → error (no bencode representation)
+ */
+
+package rt
+
+import (
+	"errors"
+	"fmt"
+	"io"
+	"net"
+	"time"
+
+	"github.com/nooga/let-go/pkg/vm"
+	"github.com/zeebo/bencode"
+)
+
+func init() { RegisterInstaller(installBencodeNS) }
+
+// bencodeFromLG converts an lg value into a Go value the bencode encoder
+// accepts. nil/bool/float are rejected: bencode has no representation for them
+// and silently coercing would corrupt nREPL messages.
+func bencodeFromLG(v vm.Value) (interface{}, error) {
+	switch v.Type() {
+	case vm.StringType:
+		return string(v.(vm.String)), nil
+	case vm.KeywordType:
+		return string(v.(vm.Keyword)), nil // :op → "op"
+	case vm.IntType:
+		return int64(v.(vm.Int)), nil
+	case vm.MapType, vm.PersistentMapType:
+		return bencodeMapToDict(v)
+	case vm.ArrayVectorType, vm.PersistentVectorType, vm.ListType:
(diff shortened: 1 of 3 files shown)


nooga (approved):
Approving as a stop-gap to unblock nREPL client work in pure .lg — the client-side TCP/bencode primitives here are solid (real TCP-pair tests, -race clean, sound conn/mutex design).

One thing to iterate on next, not blocking this merge: this has no WASM build-tag gating (net.go says "no build tag"), which breaks with the established `unix_linux.go/unix_other.go` and pods.go/pods_wasm.go pattern of a native impl + a stub under the js/wasm tag returning a clear "not supported in the browser" error. Under `GOOS=js`, `net.Dial` is backed by Go's admitted fake in-process network, so a browser build's net/dial won't fail cleanly today. Let's follow up with a native/stub split, and close the loop with @mparrett  on how this lands alongside the broader #523 design (server/UDP, API shape) so we don't end up with two generations of the same namespace.

mparrett:
Thanks @nooga — agreed on the gating, and happy to own that follow-up. I'll do the native/stub split on `net.go` matching the `unix_linux.go`/`unix_other.go` + `pods.go`/`pods_wasm.go` shape: real impl under `!js && !wasip1`, a stub under the wasm tag returning a clear "net not supported in the browser" error so a `GOOS=js` build fails loudly instead of leaning on Go's fake in-process network. Small PR, on top of this one once it lands.

On not ending up with two generations: I read #526 as the client slice of the same `net` namespace, not a parallel one — `dial`/`read!`/`write!`/`close!` are the foundation, and the #523 items (server `listen`/`accept`, UDP) extend the same namespace additively rather than replacing it. So no second generation as long as we hold the connection-object shape #526 already has, which matches what #523 argued for.

The one thing worth settling before more surface accretes is verb naming: `net` uses `dial`/`write!`/`read!`/`close!` while `unix` uses `connect`/`send`/`recv`/`close`. Two socket namespaces reading differently. No strong preference on which set wins — I'd rather converge now than after server/UDP double it. Happy to fold that into the gating PR if you want a single cleanup pass.

nooga:
In Clojure `!` is used to denote a side-effect. It's not 100% - `println` doesn't have a bang, but many functions that mutate or side-effect do. I'd be leaning towards `write!`/`read!`/`close!` naming convention across both NSes.

mparrett:
Works for me — bangs it is, `net` keeps `write!`/`read!`/`close!`.

One thing to decide on the "across both NSes" part: `unix` already shipped with `send`/`recv`/`close` (no bangs), so aligning it is a breaking rename of a released namespace, not a fresh choice. Options as I see them:

- **Hard rename** `send`→`write!`, `recv`→`read!`, `close`→`close!`. Cleanest end state; fine if `unix` is new enough that nothing depends on it yet — your call, since you know the users.
- **Alias with deprecation**: add the `!` names, keep the old ones working with a warn-on-use for a release or two, drop them later. Safest, a few extra lines.

I'd default to the alias path unless you're comfortable breaking `unix` outright. Either way I'll roll it into the gating PR so the convergence lands in one pass.

(Leaving the lifecycle verbs alone — `unix/connect` vs `net/dial`, `listen`/`accept` — unless you want those aligned too; the bang convention only speaks to the read/write/close ops.)

nooga:
Fine with aliasing.

Issue #357, opened 2026-06-29
AOT lowering: float param constrained only by float arithmetic is typed `int` → non-compiling Go (float64 + int)


mparrett:
## Summary

When a function parameter is only ever used in float arithmetic, `lg-compile` infers the **return** type as `float64` but leaves the **parameter** typed `int`, so the lowered Go does `float64 + int` and doesn't compile. The float constraint reaches the result but isn't back-propagated to the parameter.

Found on `v1.11.0` (`f9048d899a740364b2aeb9a3ee046f8ea9956c5b`) while AOT-compiling a Mandelbrot escape-time kernel.

## Minimal repro

`repro.lg`:
```clojure
(ns repro)

(defn add-to-float [a]
  (loop [z 1.5]
    (if (> z 100.0)
      z
      (recur (+ z a)))))
```

```
./lg scripts/lg-compile out github.com/nooga/let-go/out repro.lg
go build ./out/...
```

## Generated Go

```go
func add_to_float(ec *vm.ExecContext, arg0 int) float64 {  // <- arg0 typed int
	var z float64
	var v7 bool
	var v12 float64
	z = 1.5
	for {
		v7 = z > 100
		if v7 { break } else {
			z = z + arg0   // <- float64 + int
			continue
		}
	}
	v12 = z
	return v12
}
```

## Error

```
invalid operation: z + arg0 (mismatched types float64 and int)
```

## Expected

`arg0` inferred as `float64` (it flows only into float ops and into the `float64` result), or an `int → float64` coercion emitted at the use site. The return type is already correctly `float64`, so the float-ness is known at the result — it just isn't propagated back to the parameter.

## Why it matters

This is the blocker for AOT-lowering idiomatic numeric kernels. The natural escape-time function

```clojure
(defn escape [cx cy mi]
  (loop [zx 0.0 zy 0.0 i 0]
    (let [zx2 (* zx zx) zy2 (* zy zy)]
      (if (or (>= i mi) (> (+ zx2 zy2) 4.0)) i
        (recur (+ (- zx2 zy2) cx) (+ (* 2.0 (* zx zy)) cy) (inc i))))))
```

[...]

PR #530, opened 2026-07-16
feat(ir): accept numeric param hints (^double/^long) in AOT lowering (#357)

mparrett:
Re: #357. This takes the hint-based path; automatic inference of a primitive param type isn't sound (see Why), so it doesn't fully close the issue — the inference question stays open.

## Why

A float parameter constrained only by float arithmetic was typed `int`, so the lowered Go did `float64 + int` — which doesn't compile, and if it did would truncate the argument at the parameter boundary. That second part makes it a correctness bug, not just a build failure: `escape(-0.75, 0.10)` lowered with an `int` param computes against `(0, 0)` and returns the wrong iteration count.

Pure inference can't reach `float64` soundly here. `(add-to-float 3)` is legal, so the param is genuinely int-or-float, and the sound join of that is boxed — the fixpoint experiment on #357 lands on `vm.Value`, not `float64`. This is the same wall Clojure hit, and its answer is the hint, not inference: `^double`/`^long` declares the primitive, and callers coerce at the boundary.

## What

Adopt that model.

- `build-fn` accepts a `^double`/`^long` param — reified as `(with-meta a {:tag …})` — binding the symbol and routing the tag into the fn's `:arg-types` (double/float → `:float`, long/int → `:int`). Previously the arg-pattern matcher threw `unsupported arg pattern` on any `:tag`-annotated param.
- `infer-arg-types` leaves an explicitly-hinted param alone, so the hint is authoritative instead of being overridden by the op-based `:int` backward pass.
- Call-site coercion for mixed int args already works via `coerce-arg-to-type`, so nothing new is needed there.

`(defn escape [^double cx ^double cy mi] …)` now lowers to `Escape(ec, arg0 float64, arg1 float64, arg2 int) int`, compiles, and runs with correct (untruncated) values. Multi-arity comes along for free: the AOT `:go` path rebuilds each arity as a single-arity `defn` through `build-fn`, so `(defn scale ([^double a] …) ([^double a ^double b] …))` lowers to native `Scale_1`/`Scale_2`.

AOT-path only — the `ir.*` pipeline is excluded from the runtime bundle (#356), so no runtime-bundle footprint.

## Scope

Two things this deliberately leaves alone:

- The first-class multi-arity **closure** (`:multi-fn-template`) stays boxed. It lowers through `MakeNativeMultiArity`/`BoxNativeFn`, a runtime arity-dispatch across a `vm.Value` boundary, so a typed inner param only relocates the box into the reflect adapter rather than removing it — the callee is chosen at runtime, the same reason `invokePrim` doesn't engage through a boxed `IFn`.
- `^long` used in float arithmetic still needs use-site widening (an `l2d` at the op), which is `lower-go` territory.

## Tests

Two regression tests in `pkg/ir/lisp_lower_go_test.go`: single-arity via `build-fn`, and multi-arity via the `lower-ns-to-go` seam (the path that decomposes arities, distinct from `build-fn`'s own closure representation). Full `go test ./...`, the linux/wasm/plan9 cross-builds, `-race` on `pkg/rt`, and the generated-manifest check are green.

### DIFF
--- a/pkg/rt/core/ir/build.lg
+++ b/pkg/rt/core/ir/build.lg
@@ -1273,9 +1273,34 @@
           (nil? n)    (recur (drop 2 b) out)
           :else       (recur (drop 2 b) (conj out n v)))))))
 
+;; A type-hinted param — `^double a` — reaches here reified as the form
+;; (with-meta a {:tag (quote double)}). Bind the symbol and carry the tag out to
+;; :arg-types so the param lowers to a native Go type (#357). Clojure's model:
+;; the hint is a declaration, callers coerce at the boundary (coerce-arg-to-type).
+(defn- hinted-arg? [x]
+  (and (seq? x) (= 'with-meta (first x)) (symbol? (second x))))
+
+(defn- hinted-sym [x] (second x))
+
+(defn- hinted-tag [x]
+  (let [m (nth x 2 nil)
+        m (if (and (seq? m) (= 'quote (first m))) (second m) m)
+        t (when (map? m) (get m :tag))]
+    (if (and (seq? t) (= 'quote (first t))) (second t) t)))
+
+;; Numeric hint -> lattice type. double/float -> :float, long/int -> :int.
+;; Class hints (e.g. ^Foo) fall through to nil (no native numeric type).
+(defn- tag->arg-type [t]
+  (let [n (when t (name t))]
+    (cond
+      (contains? #{"double" "float"} n) :float
+      (contains? #{"long" "int"} n)     :int
+      :else                              nil)))
+
 (defn- expand-fn-args [args-vec body-forms]
   "Detect & rest and destructuring patterns in args-vec.
-   Returns nil if nothing to expand, or {:flat-args [...] :body (...) :variadic? bool}."
+   Returns nil if nothing to expand, or
+   {:flat-args [...] :body (...) :variadic? bool :arg-tags [tag-or-nil ...]}."
   (let [amp-pos   (loop [i 0 remaining args-vec]
                     (if (empty? remaining)
                       -1
@@ -1287,12 +1312,21 @@
         rest-sym   (when variadic? (nth args-vec (inc amp-pos) nil))
         has-destructure? (some (fn [x] (not (symbol? x))) fixed-args)]
     (when (or variadic? has-destructure?)
-      (let [result (loop [remaining fixed-args flat-args [] let-binds []]
+      (let [result (loop [remaining fixed-args flat-args [] let-binds [] arg-tags []]
                      (if (empty? remaining)
-                       {:flat-args flat-args :let-binds let-binds}
+                       {:flat-args flat-args :let-binds let-binds :arg-tags arg-tags}
                        (let [x (first remaining)]
-                         (if (symbol? x)
-                           (recur (rest remaining) (conj flat-args x) let-binds)
+                         (cond
+                           (symbol? x)
+                           (recur (rest remaining) (conj flat-args x) let-binds (conj arg-tags nil))
+
+                           (hinted-arg? x)
+                           (recur (rest remaining)
+                                  (conj flat-args (hinted-sym x))
+                                  let-binds
+                                  (conj arg-tags (hinted-tag x)))
+
+                           :else
                            (let [gs (gensym "p__")
                                  binds (cond
                                          (vector? x) (expand-vector-pattern gs x)
@@ -1300,16 +1334,21 @@
                                          :else (throw (str "unsupported arg pattern: " (pr-str x))))]
                              (recur (rest remaining)
                                     (conj flat-args gs)
-                                    (into let-binds binds)))))))
+                                    (into let-binds binds)
+                                    (conj arg-tags nil)))))))
             final-flat-args (if variadic?
                               (conj (:flat-args result) rest-sym)
                               (:flat-args result))
+            final-arg-tags  (if variadic?
+                              (conj (:arg-tags result) nil)
+                              (:arg-tags result))
             body (if (seq (:let-binds result))
                    (list (apply list 'let* (vec (:let-binds result)) body-forms))
                    body-forms)]
         {:flat-args final-flat-args
          :body body
-         :variadic? variadic?}))))
+         :variadic? variadic?
+         :arg-tags final-arg-tags}))))
 
 (defn- return-hinted-arity-vector? [form]
   (and (seq? form)
@@ -1469,6 +1508,14 @@
             (ir/add-terminator! f final-blk :return
                                 [last-val]
                                 nil)))
+        ;; Route numeric param hints (^double/^long) into :arg-types so the param
+        ;; lowers to a native Go type instead of vm.Value (#357). infer-arg-types
+        ;; leaves an explicitly-typed param alone, so the hint is authoritative.
+        (when-let [tags (and expanded (:arg-tags expanded))]
+          (when (some some? tags)
+            (swap! f assoc :arg-types
+                   (mapv (fn [t] (or (tag->arg-type t) :unknown))
+                         (take arity tags)))))
         f)
       ;; Multi-arity path: preserve each arity as a nested template
       ;; with a shared capture layout. The outer function (arity 0)
--- a/pkg/rt/core/ir/build.lg
+++ b/pkg/rt/core/ir/build.lg
@@ -1289,9 +1289,14 @@
     (if (and (seq? t) (= 'quote (first t))) (second t) t)))
 
 ;; Numeric hint -> lattice type. double/float -> :float, long/int -> :int.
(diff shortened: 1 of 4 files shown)


pkg/rt/core/ir/build.lg
```diff
@@ -1469,6 +1508,14 @@
             (ir/add-terminator! f final-blk :return
                                 [last-val]
                                 nil)))
+        ;; Route numeric param hints (^double/^long) into :arg-types so the param
+        ;; lowers to a native Go type instead of vm.Value (#357). infer-arg-types
+        ;; leaves an explicitly-typed param alone, so the hint is authoritative.
+        (when-let [tags (and expanded (:arg-tags expanded))]
```
nnunley, line 1524: It strikes me that there should be some sort of hard error at the point of this type coercion that should error out if a non-numeric type (other than the any/vm.Value) has been found through inferencing.

pkg/rt/core/ir/build.lg
```diff
@@ -1273,9 +1273,34 @@
...
+
+(defn- hinted-tag [x]
+  (let [m (nth x 2 nil)
+        m (if (and (seq? m) (= 'quote (first m))) (second m) m)
+        t (when (map? m) (get m :tag))]
+    (if (and (seq? t) (= 'quote (first t))) (second t) t)))
+
+;; Numeric hint -> lattice type. double/float -> :float, long/int -> :int.
+;; Class hints (e.g. ^Foo) fall through to nil (no native numeric type).
+(defn- tag->arg-type [t]
```
nnunley, line 1295: You probably should test what happens when the detected type is non-numeric -- some will need numeric conversion, and some should flat out throw exceptions at lowering. Alternatively, if it's incompatible, the type should be ignored but there should be a warning.

mparrett: Good call.

mparrett:
Pushed `2131f93`/`b0fa820` — took the ignore-plus-warn path for both threads. A hint with no native param type now warns (`WARNING: <fn>: ignoring unsupported param type hint …`) and stays `vm.Value` instead of silently dropping to `:unknown`. `tag->arg-type` is guarded so a non-Named shape like `^{:tag [long]}` warns-and-boxes rather than throwing on `(name t)`, and there's a test asserting the warning actually fires. Left a hard throw for genuinely-incompatible types as a follow-up, so a `^String` param still works boxed.

mparrett:
Validated end-to-end on a mandelbrot. The `^double`-hinted escape kernel lowers to `Escape(ec, arg0 float64, arg1 float64, arg2 int)` and, run as a native binary, matches the bytecode VM's escape counts **cell-for-cell across all 2337 cells** on a bit-identical grid (step `0.0625`, so coordinates parse to the same float64 bits on both sides).

It also surfaced a sibling gap, now filed as #534: the grid helpers — `(* 0.0625 col)` with an unhinted `int` param — don't lower (`0.0625 (untyped float constant) truncated to int`). That's the arithmetic-emission half of #357, distinct from the param-hint work here. `typeinfer` types the result `:float` correctly, so it's a use-site `float64()` coercion missing in `lower_go`.

<details><summary>native-lowered render</summary>

```
................:::::::::---------------:::::::::::::::::
...............:::::::---------------------::::::::::::::
..............::::::-------------===+*+===---::::::::::::
.............:::::-------------====+*@#*+===---::::::::::
.............::::-------------====++*%-  ====---:::::::::
............:::-------------======++#%- #+====----:::::::
............::-------------======++*-. *@*++===----::::::
...........::-------------=====++**%#    #+++==-----:::::
...........::------------=====+***#%:   =%#*++==----:::::
..........::-----------=====++:@ %  =   - .#*#@==----::::
..........:-----------====+++*@: :          # .+=-----:::
..........:----------==+++++*#@:           @  -*+=----:::
..........---------==++++++**@:%              @*+=-----::
.........:-------==*#**+***#%*               =@#+=-----::
.........-----===++*=%#%%##%%.                #%+==-----:
.........---====++*#.@:. +@@ =                 .*==-----:
.........-=====+++*#@=   +=.-                  -+==-----:
.........=====+++*#%-*      #                  %+==-----:
.........====+**#%..@       *                 #*+==-----:
.........++*#**##@# =       .                 %*+==-----:
........                                     .#*+==------
.........++*#**##@# =       .                 %*+==-----:
.........====+**#%..@       *                 #*+==-----:
.........=====+++*#%-*      #                  %+==-----:
.........-=====+++*#@=   +=.-                  -+==-----:
.........---====++*#.@:. +@@ =                 .*==-----:
.........-----===++*=%#%%##%%.                #%+==-----:
.........:-------==*#**+***#%*               =@#+=-----::
..........---------==++++++**@:%              @*+=-----::
..........:----------==+++++*#@:           @  -*+=----:::
..........:-----------====+++*@: :          # .+=-----:::
..........::-----------=====++:@ %  =   - .#*#@==----::::
...........::------------=====+***#%:   =%#*++==----:::::
...........::-------------=====++**%#    #+++==-----:::::
............::-------------======++*-. *@*++===----::::::
............:::-------------======++#%- #+====----:::::::
.............::::-------------====++*%-  ====---:::::::::
.............:::::-------------====+*@#*+===---::::::::::
..............::::::-------------===+*+===---::::::::::::
...............:::::::---------------------::::::::::::::
................:::::::::---------------:::::::::::::::::
```
</details>

pkg/rt/core/ir/build.lg
nnunley: The comment isn't entirely true. There's also the [:dtype T] semantics to support go structs. I guess we shouldn't worry about that now. I've got a deeper set of fixes in the wings.

nnunley (approved):

pkg/rt/core/ir/build.lg
mparrett: Added the guard — `assert-hint-agrees-with-inference!` in `infer_arg_types.lg`. At the point where the hint would override op-based inference, if the backward pass proved the param a concrete non-numeric type it now throws with the fn, arg index, hint, and inferred type, instead of letting the numeric hint stand.

It can't fire today by construction: the backward pass yields only `:int` or `:unknown` (`int-constraint-ops`), so a concrete non-numeric param type never arises. So it's a tripwire for when arg inference grows concrete param types — the type-inference widening you mentioned on #536 — where a hint that contradicts the body would surface here rather than miscompile downstream. `TestLowerGoNumericHintGuardInertWhenInferenceAgrees` pins the inert case: a `^double` param used in arithmetic infers `:int`, which is numeric, so the guard leaves the hint alone and the param still lowers to `float64`.

pkg/rt/core/ir/build.lg
mparrett: Took the ignore-plus-warn path. A recognized non-numeric hint (`^String`) and a non-Named tag shape (`^{:tag [long]}`) both warn — `WARNING: <fn>: ignoring unsupported param type hint ^…` — and leave the param `vm.Value` rather than mis-routing to a native numeric param. Covered by `TestLowerGoNonNumericHintStaysBoxed` (asserts the warning fires and the param stays boxed) and `TestLowerGoNonNamedHintWarnsAndBoxes`.

mparrett:
Process note, not a code change. I rebased this onto current `main` after #536 landed, and it was bumpier than a normal replay. A concurrent session of ours had added the cross-type `=` fix to the #536 branch, and #530 and #536 both touch `generated.sums` and the same test file (`lisp_lower_go_test.go`), so the rebase needed a hand-merged union of the two PRs' added tests rather than a clean apply.

I've checked the result closely: the test-file merge is addition-only (none of #536's merged tests were dropped), `build.lg` and `infer_arg_types.lg` are byte-identical to the pre-rebase versions, and `go test ./...`, the linux / js-wasm / plan9 cross-builds, `-race` on `pkg/ir` + `pkg/rt`, and the generated-manifest check are all green.

Given the concurrency and that hand-merge, I'd appreciate a second set of eyes before we merge — here, and retroactively on the now-merged #536, since the two evolved together. If anything in the merged test file or the `=` / widening interaction looks off, I'm happy to adjust.

nooga (approved):
Re-reviewed after the follow-up commit addressing nnunley's and mparrett's feedback. Verified both fixes directly: `tag->arg-type` no longer crashes on a non-Named tag shape (`^{:tag [long]}`), confirmed via `TestLowerGoNonNamedHintWarnsAndBoxes`; and the new `assert-hint-agrees-with-inference!` guard is a real (if currently inert) tripwire — traced that the backward-inference pass can only yield `:int`/`:unknown` today so it can't false-fire, and the negative-case test (`TestLowerGoNumericHintGuardInertWhenInferenceAgrees`) confirms it doesn't over-fire on the motivating escape-kernel case. Non-numeric hints now warn instead of silently dropping, verified via the `*out*`-capturing test. Full test suite green on the updated branch.

PR #532, opened 2026-07-16
docs(guide): wrapping Go packages via lginterop — database/sql worked example

nnunley:
Adds `docs/guide/go-interop.md` (indexed in `docs/README.md`): how to wrap a Go package for let-go, prompted by a community question about feeding HoneySQL's `[sql-string & params]` output into `database/sql` Exec/Query.

Covers:
- The existing two-layer pattern, with the in-tree exemplar named: generated raw namespace (`cmd/lginterop` → `pkg/rt/interop_xxh3.go`) + `.lg` veneer (`pkg/rt/core/hash.lg`).
- What boxing/reflection already handles with no wrapper: boxed Go pointers, reflective method dispatch including variadics, struct write-back, per-element slice conversion — so `Exec`/`Query`/`Rows` cursor methods map directly, and HoneySQL output feeds them with a single `apply`.
- The one genuine seam: `Scan(dest ...any)` writes through caller-allocated pointers, which immutable let-go values can't provide. Documented as a ~20-line `ScanRow` shim that the lg-based gogen emitter can generate (out-param template), keeping the wrapper lg-first rather than hand-maintained Go.
- Drivers as a host/build decision (blank import): embedding host imports it, or a build tag like `pkg/glplat`'s.
- A complete `db/query` + `db/exec!` veneer sketch, including the realize-before-close cursor subtlety.

`docs-frontmatter` check passes locally.

### DIFF
--- a/docs/guide/go-interop.md
+++ b/docs/guide/go-interop.md
@@ -72,9 +72,9 @@ needs `(require 'hash)` first:
 
 Mechanically, `cmd/lginterop` is a two-stage pipeline: the Go binary scans the
 target package and extracts its exports, then drives `scripts/lginterop.lg`
-(via the `lg` binary) to render the Go source. You don't need to know that to
-use it, but it explains why the tool wants to run from the repo root with a
-built `lg`.
+(via the `lg` binary, which the tool builds itself) to render the Go source.
+You don't need to know that to use it, but it explains why the tool wants to
+run from the repo root.
 
 Related docs: [Embedding let-go in Go](embedding-in-go.md) covers the host
 side (running let-go inside your own Go program), which is the other way to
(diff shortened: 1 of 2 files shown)


nooga (approved):

docs/guide/go-interop.md
```diff
@@ -0,0 +1,215 @@
...
+| `db` veneer: `query`, `exec!`, `with-db` | `pkg/rt/core/db.lg` (or a userland `.lg` library) | ~1 screen |
+
+## Is anyone doing DB work with let-go yet?
+
+Not that is visible in-tree or in the compat corpora: xxh3 is the only
+lginterop-wrapped package on `main`, and the HoneySQL compat work (#494)
+stops at `format` — generating SQL strings, not executing them. A
+`database/sql` wrapper along the lines above would be the first, and the
+`[sql-string & params]` shape means HoneySQL output feeds it with a single
+`apply`.
```
nnunley: Remove this editorial - it's not durable information

nnunley: Removed in dcd18338 — the section is gone. In its place the doc gained a durable `cmd/lginterop` reference (both modes, all flags) and a link to the Usage guide for building/running `lg`, so everything the interop doc leans on is now documented in-tree.

nooga (approved):

nooga:
Reads better now, thanks!

nnunley:
Heads-up on merge order: the latest commit documents external-package mode as working (`-opaque-structs`, header round-trip, e2e golden guard), which is true once #537 lands — merge #537 first, then this.

PR #533, opened 2026-07-16
feat(rt): gate net/bencode off js/wasm; add unix bang-name aliases

mparrett:
## What

Follow-up to #526, doing the two things the review there asked for.

**WASM gating.** `net.go` and `bencode.go` shipped with no build tag, so they compiled into the browser build too. Go's `GOOS=js` net stack is an in-process fake — `net/dial` there resolves against nothing and "succeeds", so a browser build never fails cleanly. This build-tags both files native (`!js`) and adds browser stubs (`net_wasm.go`, `bencode_wasm.go`) that register the same namespace surface with functions returning a clear "not supported in the browser" error. Matches the existing `pods.go`/`pods_wasm.go` and `unix_linux.go`/`unix_other.go` split. Requiring `net`/`bencode` still works in the browser; the error only fires if a program reaches for the network.

**Naming convergence (#526).** `net` uses the Clojure side-effect bang convention — `write!`/`read!`/`close!`. This adds the same names to `unix` as the canonical verbs and keeps `send`/`recv`/`close` as deprecated aliases: they print a one-time notice (`unix/send is deprecated; use unix/write!`) and delegate. `unix_other.go` mirrors the new names so the namespace surface is identical across platforms.

## Notes

- The deprecation notice is warn-once per name via `sync.Once`, so it's quiet after the first call and race-free under the `-race` CI added in #524.
- Aliases point at the existing `send`/`recv`/`close` implementations unchanged — no behavior change to the ops themselves, just the names and the notice.
- Lifecycle verbs (`unix/connect` vs `net/dial`, `listen`, `accept`) are left alone; the bang convention only speaks to the read/write/close ops.

## Test

`TestUnixBangAliases` (linux): checks the canonical and deprecated names all resolve, round-trips a message through `unix/write!` → `unix/read!` over a `socketpair`, and exercises a deprecated alias to confirm it still delegates. Ran the full `pkg/rt` suite on Linux with `-race` (net, bencode, unix all green, deprecation notice observed); builds clean on `js/wasm`, `wasip1/wasm`, linux, darwin, and plan9.

### DIFF
--- a/pkg/rt/bencode.go
+++ b/pkg/rt/bencode.go
@@ -1,10 +1,14 @@
+//go:build !js
+
 /*
  * Copyright (c) 2026 let-go contributors
  * SPDX-License-Identifier: MIT
  */
 
 /*
- * bencode namespace — message framing over a net conn.
+ * bencode namespace — message framing over a net conn. Native only; js/wasm
+ * gets a stub (bencode_wasm.go) since it frames over a net conn, which the
+ * browser doesn't have.
  *
  *   (bencode/write! conn value)       → nil. Encodes one bencode value.
  *   (bencode/read! conn)              → decoded value, nil on clean EOF. Blocks.
--- a/pkg/rt/bencode_wasm.go
+++ b/pkg/rt/bencode_wasm.go
@@ -1,0 +1,37 @@
+//go:build js && wasm
+
+/*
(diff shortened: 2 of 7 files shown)


mparrett:
cc: @abogoyavlensky

nooga (approved):
No notes

PR #536, opened 2026-07-16
fix(ir): widen int operand to float64 in mixed int/float native lowering

### LINKED ISSUE #534: AOT lowering drops the int→float coercion in mixed int/float arithmetic
Native `:go` lowering of a mixed int/float arithmetic op emits the operation without coercing the int operand to `float64`, producing non-compiling Go. `typeinfer` already types the *result* `:float`; the gap is in the emission.

## Repro

```clojure
(defn scale-col [col] (* 0.0625 col))
```

`lg-compile` lowers this to:

```go
func ScaleCol(ec *vm.ExecContext, arg0 int) float64 {
	var v2 float64
	v2 = arg0 * 0.0625   // int * untyped-float-const
	return v2
}
```

`go build`: `0.0625 (untyped float constant) truncated to int`.

## Diagnosis

`typeinfer` gets it right — the result var is `float64` and the fn returns `float64` (the concrete-float-operand contagion rule fires). The gap is purely in `lower_go`'s arithmetic emission: when the op's result type is `:float` but an operand is `:int`, the int operand needs a `float64(...)` wrap. Expected: `v2 = float64(arg0) * 0.0625`. It's the same `l2d` widening Clojure emits at a mixed-primitive arithmetic site — the compiler coerces the narrower operand up.

## Scope

Emission-only in `lower_go`; `typeinfer` needs no change. The existing mixed-operand handling for `/` (native `/` vs the runtime helper) is a model for where the operand-type check goes.

## Relation

- Sibling to #357: that's the param-*typing* half (a float-only param typed int); this is the arithmetic-*emission* half (an int operand meeting a float). Even with #357's param hints (#530), a `^long` param used in float arithmetic still hits this.
- Not covered by #438 (native-primitive generation) — verified it still repros on current `main`.

## Impact

Any unhinted int used in float arithmetic — coordinate scaling (`x * step`), ratios, an `int` counter times a float — fails to lower. It surfaced while validating #530 on a mandelbrot: the escape kernel lowers and runs correctly (2337-cell match against the bytecode VM), but the grid helpers `(* 0.0625 col)` hit this.

Happy to take it — the fix looks localized to the arithmetic emission.

mparrett:
Closes #534. This is the use-site-widening half flagged in #530's scope: the `l2d`-at-the-op case that param hints don't reach.

## Why

`typeinfer` already promotes a mixed int/float arithmetic op to `:float` by numeric contagion, so the result local is declared `float64`. But the native binary path emitted the operands bare, so an int operand against a float lowered to `int <op> float64`, which Go rejects. Against an untyped float *constant* it's a hard compile error rather than a type mismatch:

```clojure
(defn cx-of [col] (+ -2.5 (* 0.0625 col)))
```

lowered to `arg0 * 0.0625`, and `go build` fails with `0.0625 (untyped float constant) truncated to int`. The same defect hit mixed *comparisons* — `(< col 0.5)` with an int param emits `col < 0.5`, equally invalid.

It surfaced validating #530 on a mandelbrot: the `^double`-hinted escape kernel lowers and runs correctly (2337-cell match against the bytecode VM), but the grid helpers `(* 0.0625 col)` (an unhinted int column times a float step) wouldn't build.

## What

Widen the int side with `float64(...)` in the native binary path, matching the VM's int→float promotion. `cx-of` now lowers to `float64(arg0) * 0.0625`.

The check keys on the operand specs, not the result type (one operand `int`, the other `float64`), so it covers the comparison case too (`float64(col) < 0.5`), where the result is `bool` rather than `float64`.

The widen is unconditional at this site because only int/float64 can reach it mixed: `quot` lowers native for int/int only, `div` for float/float only, and bitwise/shift ops for int/int only — every other operand mix already routes through a `vm.*Value` helper before this branch. So there's no op here where promoting the int to float would be wrong.

Emission-only; `typeinfer` is untouched. `generated.sums` is regenerated (it digests the touched `.lg` source); `core_compiled.lgb` is unaffected, since `ir.lower-go` is the on-demand AOT tool, not an embedded namespace.

## Tests

Two regression tests in `pkg/ir/lisp_lower_go_test.go`: mixed arithmetic and mixed comparison, each asserting the int operand is widened and the op stays native (no boxed `*Value` helper). Full `go test ./...`, the linux and js/wasm and plan9/amd64 cross-builds, `-race` on `pkg/ir` and `pkg/rt`, and the generated-manifest check are green.

### DIFF
--- a/pkg/ir/lisp_lower_go_test.go
+++ b/pkg/ir/lisp_lower_go_test.go
@@ -211,6 +211,35 @@ func TestLowerGoStrictMixedIntFloatComparisonWidensIntOperand(t *testing.T) {
 	}
 }
 
+// Regression for #534: the int→float widening is sound for arithmetic and
+// ordering, but NOT for `=`. Clojure `=` is category-sensitive across the
+// numeric tower — (= 1 1.0) is false — whereas widening the int to float64 and
+// comparing with native Go == makes `(= col 1.0)` return true at col=1,
+// diverging from the bytecode VM. A mixed concrete int/float `=` must route to
+// rt.EqValue (which delegates to vm.ValueEquals), not native ==.
+func TestLowerGoStrictMixedIntFloatEqUsesEqValueNotNativeEq(t *testing.T) {
+	ensureLoader()
+
+	fn := buildLispIR(t, `(defn eqf [col] (= col 1.0))`)
+	seedArgTypes(t, fn, "[:int]")
+	optimizeLispIR(t, fn)
+	result := lowerGo(t, fn, ":strict")
+
+	if got := result.ValueAt(vm.Keyword("status")); got != vm.Keyword("lowered") {
+		t.Fatalf("expected :lowered status for mixed int/float =, got %v (reason=%v)",
+			got, result.ValueAt(vm.Keyword("reason")))
+	}
+	rendered := bindAndRenderGoDecl(t, result)
+	if !strings.Contains(rendered, "EqValue") {
+		t.Fatalf("expected a mixed int/float = to route to rt.EqValue (Clojure value-equality), got:\n%s", rendered)
+	}
+	// A native `==` here would be the divergence: float64(col) == 1.0 is true at
+	// col=1, but Clojure (= 1 1.0) is false.
+	if strings.Contains(rendered, "==") {
+		t.Fatalf("expected NO native == for a cross-type = (it would violate Clojure's category-sensitive =), got:\n%s", rendered)
+	}
+}
+
 // Regression for PR #235 review (robustness): a genuinely ambiguous :number
 // result — e.g. (+ x x) where x is only known to be {int,float} — has no native
 // Go type, so go-type-spec must give it a lowering target (vm.Value) instead of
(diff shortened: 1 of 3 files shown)


nooga (approved):
Verified beyond the diff: built the branch and reproduced the exact motivating case — `(defn cx-of [col] (+ -2.5 (* 0.0625 col)))` with col seeded `:int` lowers to `float64(arg0) * 0.0625 + -2.5`. Took it further than the PR's own string-contains assertions: extracted the rendered Go into a standalone program and ran it (`CxOf(10) = -1.875`, correct). Also confirmed directly against go-type-spec that int/float64 really are the only pair that can reach this branch mixed (no `:float32` case exists anywhere in that function), so the unconditional widen is sound, not just asserted. Full test suite, -race on pkg/ir + pkg/rt, and the js/wasm + plan9/amd64 cross-builds are all clean on my end too.

nnunley:
I think this is fine. I have some ideas for how to improve the type-inferencing logic to widen types down the road, but this should be a good intermediate change.

nnunley (approved):

PR #537, opened 2026-07-16
fix(lginterop): rebuild the smart-wrapper emitter on gogen AST constructors

### LINKED ISSUE #535: lginterop external-package mode is broken: scripts/lginterop.lg calls removed gogen/parse-stmts
## Symptom

The documented external-package invocation fails on current main:

```
$ go run ./cmd/lginterop -packages github.com/zeebo/xxh3 -out pkg/rt
lginterop: github.com/zeebo/xxh3: lg failed: exit status 1
error: compiling def value  (lginterop-*.lg:129 build-wrapper-body)
caused by: Can't resolve gogen/parse-stmts in this context
lginterop: generated 0/1 package(s) in pkg/rt
```

`pkg/rt/interop_xxh3.go` therefore cannot be regenerated as documented — it's frozen at whatever the emitter last produced before the drift.

## Root cause

`scripts/lginterop.lg` predates gogen's move to a form-based API: it builds wrapper bodies as Go source *strings* and fed them through `gogen/parse-stmts`, which no longer exists (current `pkg/rt/gogen/gogen.lg` exposes `walk-stmt-form`/`walk-expr-form`/`->expr` over s-expression forms, no string parsing). The resolution failure happens while compiling `build-wrapper-body` itself, so **both** smart and non-smart paths die even though only smart mode calls it. This is what #467's "stale gogen/parse-stmts" exclusion was pointing at — the corpus exclusion hid the bitrot rather than tracking it.

Primitives mode (`-primitives`, pure-Go emitter) is unaffected — verified: regenerates `pkg/rt/zz_primitives_generated.go` byte-identically.

## Fix direction

Rewrite `build-wrapper-body` (and any other string-building emission) to construct gogen forms directly, matching the current API. The emitter stays in lg — this is repair, not a rewrite into Go (see #531 S1, which also relocates it into `cmd/lginterop` via embed + in-process eval; fixing the drift is a prerequisite for S1's byte-identical golden test).

Verification bar: `go run ./cmd/lginterop -packages github.com/zeebo/xxh3 -out pkg/rt` succeeds and leaves `interop_xxh3.go` byte-identical.

nnunley:
Fixes #535. The external-package mode died with the form-based gogen refactor: `build-wrapper-body` assembled Go statements as strings and fed them through `gogen/parse-stmts`, which no longer exists — the whole script failed to compile, so `lginterop -packages` generated nothing and `interop_xxh3.go` was frozen.

**The emitter stays in lg** — wrapper bodies, arity checks, and unbox/box conversions are now constructed directly as go/ast nodes via the gogen constructors. lginterop is a gogen-based lowering to Go, not a string templater.

Also aligns tool output with the committed conventions, which turned out to be hand-edits the tool never produced:
- `func init() { RegisterInstaller(installXNS) }` installer pattern (matching async.go/bencode.go/…)
- generated-by header that records the exact invocation **flags**, so regenerating from the header's own command round-trips
- output filename `interop_<alias>.go` (was `github_com_zeebo_xxh3.go`)
- new `-opaque-structs` flag: skips `vm.RegisterStruct` so struct types stay `vm.Boxed` and dispatch methods reflectively — xxh3's `Hasher` relies on this (registering would flatten it to a field-only Record and break the `hash` veneer's `.WriteString`/`.Sum64` dispatch)

With those, `pkg/rt/interop_xxh3.go` is adopted as **pure tool output**: regeneration is byte-identical (cmp-verified round trip), guarded forever by a new e2e golden test so gogen API drift can never silently break the emitter again.

The rewrite also lowers clean, so `scripts/lginterop.lg` re-enters the ir-stress corpus (29/29 fixtures; full corpus 2409/2420 — the 11 failures are pre-existing on main, verified by A/B running the corpus at base).

Verification:
- `go test ./pkg/... ./test/` green; `make check-generated` ok
- hash veneer semantics unchanged (opaque `Hasher` method dispatch verified by running the doc examples)
- `-smart` output compiles (typed wrappers with arity checks + unbox/box)
- `make ir-stress`: corpus grows by 29 fixtures, no new failures


mparrett (approved):
Approving — I verified the load-bearing claims on the PR head (a single commit atop current main, d1494ac):

- `TestLginteropXxh3RoundTrip` passes; regeneration is byte-identical.
- The `-smart` path works: I generated `-smart -opaque-structs` output, parse-checked it with gofmt, and type-checked it by swapping it into `pkg/rt` and building. The AST-built wrapper bodies are a faithful port of the old string semantics — I compared them branch-by-branch against `unbox-snippet`/`box-expr` (arity checks, `vm.String` type-assert with error, `ToInt`+cast, `vm.Int(int64(...))` boxing all match).
- Every gogen constructor call matches the arity of the natives registered in `pkg/rt/gogen.go`.
- This also resolves everything raised on #535: the hand edits are now tool output (`RegisterInstaller` init, `-opaque-structs`), the byte-identical bar is real via the golden test, the corpus exclusion is reverted, and repair-vs-delete is answered by porting `-smart`.

Scope note: I didn't independently run `make ir-stress` or the full suite; my verification was the golden test, the smart-path probe, and the diff analysis above.

Three suggestions, none blocking:

1. **The `-smart` path has no CI fence.** The golden test covers only the non-smart xxh3 output, and the ir-stress corpus lowers the script without executing `build-wrapper-body` — so the new wrapper-body construction is verified only manually. The check is cheap: generate `-smart` to a temp dir, parse, type-check (the three steps I ran). A second leg on the e2e test would give it the same guard the non-smart path has now.
2. **Alias-based filenames can collide.** `interop_<alias>.go` derived from the package's last segment means `github.com/a/util` and `github.com/b/util` in one invocation silently overwrite each other; the old full-path name couldn't. Distinct deps.edn aliases work around it, but a collision check would fail loudly instead.
3. **The Hasher rationale left the repo.** The old in-file comment explained why the structs stay opaque (registering flattens them to field-only Records and breaks `.WriteString`/`.Sum64` dispatch in the hash veneer). That rationale now lives only in this PR's body; the flag help documents the mechanism but not the xxh3 case. One sentence in the e2e test comment would keep it discoverable.

nooga (approved):
Verified beyond reading the diff: built the branch and ran the new TestLginteropXxh3RoundTrip myself (genuinely regenerates and byte-diffs clean against the committed file), generated -smart output standalone and confirmed it actually compiles, re-ran the ir-stress lower-go check on scripts/lginterop.lg directly (29/29, matching the PR body), and make check-generated + the full test suite (go test ./... -count=1) are clean.

One minor, non-blocking gap: the generated-by header records -packages/-smart/-opaque-structs but not a custom alias override (possible via deps.edn's :gointerop config, though unused today by xxh3), so "regenerating from the header's own command round-trips" is true for the default-alias case but not guaranteed if a future package uses an alias override. Worth a follow-up, not blocking this PR.

nnunley:
Ran a multi-agent review on this branch (5 specialists + adversarial verifier). The emitter rewrite itself came back clean — the Logic pass verified every unbox/box/arity/result branch faithful to the old string spec, and the Contract pass verified the installer pattern, filename, 6-arg call sync, `hash.lg` coverage, and that the e2e test's flags match the committed header exactly.

Three verified findings, all **pre-existing** defects in the touched `cmd/lginterop/main.go` (confirmed present in the parent commit), fixed in 617ef7e per the touched-code policy:

1. **Temp-script leak** — `writeGenScript`'s `lginterop-*.lg` was never removed (the cleanup `defer` was commented out for debugging); one orphan per run. Now removed on exit, with `LGINTEROP_KEEP_SCRIPT=1` preserving the debug affordance.
2. **`ensureLgBinary` race + silent staleness** — stat-then-build into the shared `./lg` raced concurrent invocations (the new e2e test adds a second builder under parallel `go test`), and an existing-but-stale binary was reused silently — the exact drift class the round-trip golden test exists to catch. Now always builds (incremental, so cheap) into a unique temp name and renames atomically; verified with a concurrent double-invocation smoke test, no temp leftovers.
3. **Alias-collision silent overwrite** — two entries sharing an alias wrote the same `interop_<alias>.go`, second clobbering the first. Now refused up front: `alias "xxh3" used by both github.com/zeebo/xxh3 and example.com/fork/xxh3 — outputs would collide`, exit 1.

Rejected by verification (with proof): header newline injection (every path is gated by `go/types` `Import` before header construction) and the `fmt`-import gating (exact — every smart wrapper emits `fmt.Errorf` in its arity check, so the import is present iff used).

Round-trip remains byte-identical after the fixes; e2e + rt + language suites green, `make check-generated` ok.

mparrett:
Re-verified the review-pass commit (97d7ec9) — approval stands. What I ran on the new head:

- Collision refusal: `-packages` with a duplicate alias exits 1 with the promised message; the check shares `defaultAlias` with `generatePackage`, so it can't drift from the real filename logic.
- Temp-script cleanup: a generation run leaves `$TMPDIR` unchanged — and the three orphans already sitting in mine from running the old code earlier today are the leak's own evidence.
- Concurrent double-invocation: both exit 0, byte-identical outputs, zero `lg-build-*` leftovers in the repo root.
- Golden test passes; manual regeneration is still byte-identical.
- The `generated.sums` bump looked odd for a commit that touches no `.lg` files, but it's required: the genmanifest spec hashes `cmd/lginterop/*.go` too.

One residual edge, take it or leave it: the collision check keys on the raw alias, but filenames pass through `goPackageToFileName`, which collapses `/`, `.`, and `-` to `_` — so aliases like `go-json` vs `go.json` still silently collide. Keying the seen-map on `goPackageToFileName(alias)` makes it exact in one line.

That closes suggestion 2 from my review; 1 (a CI leg for `-smart`) stays open, and 3 (the Hasher rationale in-repo) is covered by #532's guide once it merges.

mparrett (approved):
Re-approving on 97d7ec9 so the formal review covers the latest head — verification details in my comment above (all three review-pass fixes checked live; merge with current main tested clean, golden and genmanifest tests green).

PR #541, opened 2026-07-17
fix(lgbgen): stage --target=go output, install atomically, add completeness sentinel

nnunley:
## Problem

`lgbgen --target=go` deleted the whole `pkg/rt/core_go_lowered/` tree and rewrote it in place. Lowering takes minutes (typeinfer), so a killed or crashed run left a torn, **gitignored** tree — invisible to `git`/`jj status` — that broke every subsequent `-tags gogen_ir` build with a wall of `no required module provides package .../core_go_lowered/...` errors. The tree only self-repaired on the next full regeneration, which made the failures look flaky (this bit the #429 pre-push gate repeatedly before being root-caused).

## Fix

- **Staged install**: generation writes to `core_go_lowered.stage/`, and only on success installs into place via per-file renames. The exposed window shrinks from the whole lowering to milliseconds, and a failed/killed run leaves the previous consistent tree fully intact (previously it left rubble).
- **Completeness sentinel**: a `.lgbgen-tree.sum` file (sha256 per generated file) is written as the last install step and removed as the first step of a reinstall, so its validity positively asserts a complete tree: `genmanifest.WriteTreeManifest` / `CheckTreeManifest`.
- **Actionable failure**: `TestGogenAOTDiff` validates the sentinel before building with `-tags gogen_ir`. A torn tree now fails in 0.01s with `run make generate (or make lowered) and retry` instead of the module-resolution wall.

## Verification

- Full `./test/e2e` (incl. `TestLoweringDeterminism`, which regenerates via the new staged path), `./pkg/genmanifest`, `./cmd/lgbgen` — green.
- Torn-tree simulation (deleted `edn/edn.go` from a valid tree): `TestGogenAOTDiff` fails immediately with `missing file edn/edn.go ... (torn tree)`.
- `generated.sums` refreshed (lgbgen sources are generator inputs); `make check-generated` green.

### DIFF
--- a/cmd/lgbgen/main.go
+++ b/cmd/lgbgen/main.go
@@ -15,6 +15,7 @@ package main
 import (
 	"flag"
 	"fmt"
+	"io/fs"
 	"os"
 	"os/signal"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


nooga (approved):
Verified beyond the diff: regenerated the tree via the new staged path, then reproduced the exact torn-tree repro from the PR body (deleted core_go_lowered/edn/edn.go) — TestGogenAOTDiff failed in 0.01s with the actionable "torn tree" message instead of the old module-resolution wall. Regenerated again, confirmed clean pass and no leftover .stage dir. check-generated + full test suite green.

One follow-up worth doing (not blocking): pkg/genmanifest/tree.go's new WriteTreeManifest/CheckTreeManifest has no direct unit tests — only exercised via TestGogenAOTDiff's happy path in CI and your manual torn-tree verification. Given the sibling staleness_test.go already establishes the pattern for this package, a small tree_test.go covering each CheckTreeManifest failure branch (missing sentinel, missing file, stray file, checksum mismatch) would make this regression-proof rather than relying on manual verification.

PR #543, opened 2026-07-17
test(lginterop): e2e leg for -smart output; defeat the test-cache blind spot

mparrett:
Follow-up to #537 (the suggestion left open in review): the committed xxh3 file is non-smart, so `TestLginteropXxh3RoundTrip` never executes `build-wrapper-body`'s typed unbox/box/arity emission — the `-smart` emitter had no CI coverage. The new leg generates the smart variant and type-checks it against the real `pkg/rt` by swapping it over the committed file with `go build -overlay`: no tree mutation, no new dependency, and the wrappers must compile, not just parse.

Falsifying the new test surfaced a second gap, fixed here too. Both tests reach the emitter through subprocesses (`go run ./cmd/lginterop` → lg → the script), so the Go test cache never sees `scripts/lginterop.lg` or `cmd/lginterop/*.go` as inputs — an emitter edit returned a stale cached pass locally, the drift class these tests exist to catch (CI is cold-cache and unaffected). `trackInputs` reads those files in-test so the cache keys on their content.

Verification:

- Both tests green fresh and cached; full `./test/e2e/` green.
- Falsification: sabotaging an exercised emitter branch (`vm.ToInt` → `vm.ToIntTypo`) fails the new leg with `undefined: vm.ToIntTypo` — without `-count=1`, proving the cache now invalidates on script edits. Reverting goes green again.
- First sabotage attempt hit the `bool` boxing branch and passed — xxh3 has no bool in any signature, so that branch is never emitted. Coverage is bounded by xxh3's API surface: the leg exercises the type branches it reaches (string, uint64, error-returning results); the rest stay uncovered until a wrapped package uses them.

### DIFF
--- a/test/e2e/lginterop_regen_test.go
+++ b/test/e2e/lginterop_regen_test.go
@@ -17,7 +17,11 @@ import (
 // emitter through subprocesses (`go run ./cmd/lginterop` → lg → the .lg
 // script), and the Go test cache only tracks files the test binary itself
 // opens — so an edit to scripts/lginterop.lg returns a stale cached pass.
-// Reading the inputs here makes the cache key on their content.
+// Reading the inputs here makes the cache key on their content. This covers
+// the high-churn emitter surface (the script + the driver's Go sources),
+// not the full subprocess input closure: lg's own VM/compiler/gogen sources
+// and the embedded core bundle are not keyed, so a codegen-affecting change
+// there can still hit a stale cached pass locally (CI runs cold-cache).
 func trackInputs(t *testing.T, patterns ...string) {
 	t.Helper()
 	for _, pat := range patterns {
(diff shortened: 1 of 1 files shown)


nooga (approved):
Verified beyond the diff: ran both e2e legs fresh (pass), then reproduced the falsification claim myself — sabotaged the emitter (ToInt -> ToIntTypo in scripts/lginterop.lg) and re-ran WITHOUT -count=1; the smart-output test genuinely failed with undefined: vm.ToIntTypo, confirming trackInputs really does defeat the test-cache blind spot rather than silently returning a stale pass. Reverted and confirmed green again. Full test suite green. The go build -overlay technique for type-checking generated output without mutating the tree is a nice, low-risk pattern.

Issue #528, opened 2026-07-16
Epic: Maintenance debt — retire superseded mechanisms, reduce process weight


mparrett:
Umbrella for behavior-preserving cleanup: mechanisms that have been superseded but never retired, logic that exists in several copies, and build/CI weight that outlived its reason. Distinct from runtime performance and concurrency (#464), lowering/codegen (#258), and Clojure-semantics parity (#257): nothing here changes what programs do — only how much code and process we maintain to do it.

The recurring pattern, across otherwise-unrelated corners: a migration lands and the predecessor stays. Each leftover is small; together they're where a lot of the "why are there two of these?" questions come from. A roll-up makes them visible and cheap to veto before anyone spends PR effort.

The lists below are the instances I've verified so far — exemplars of the pattern plus known finds, not an exhaustive inventory. New sightings accumulate here as sub-issues.

**Gate for everything under this epic: behavior unchanged.** Anything that would redraw a behavior line (however small) gets its own issue and discussion first.

## Superseded but still alive

- Per-file `//go:embed` source vars (`AsyncSrc`, `IRBuildSrc`, `IROpsSrc`, the 19 IR pass vars, and friends) superseded by the `all:core` FS embed in `core_embed_fs.go`. Roughly 33 vars with zero references (including the whole files `irpasses.go`, `irbuild.go`, `irops.go`, `asyncembed.go`), and each embeds a second copy of its core source into every non-bootstrap binary. Only `CoreSrc` is live.
- ~~`scripts/lginterop.lg`, superseded by `cmd/lginterop`~~ **Corrected** (see comments): it's the live codegen back end — `cmd/lginterop` reads it from the repo root on every invocation (`writeGenScript`, main.go:419). The item is now code homing: `//go:embed` the macro library (or move it under `cmd/lginterop/`) so the binary is self-contained and the pairing is visible from both sides.
- The `wasm/` playground: pre-LetGoHost design (xterm 5.3, bare `window.Eval`, no COI/worker path) maintained in parallel with the `pkg/rt/wasm` assets that `lg -w` ships. `pages.yml` deploys the old one as the homepage. Proposal: regenerate the playground from the current assets and delete `wasm/main.go` + `wasm/index.html`. (This one touches the public playground, so it's a "needs a call" item: the page would gain the COI/worker boot path.)
[...]

PR #544, opened 2026-07-17
fix(vm): synchronize BoxedTypes cache and Boxed.Hash() lazy cache

### LINKED ISSUE #540: pkg/vm/boxed.go: unsynchronized BoxedTypes map + Boxed.Hash() lazy cache (latent data races)
Surfaced during the multi-agent review of #522 (which only relocates code around these — the races predate it on main).

Two unsynchronized lazy-init sites in `pkg/vm/boxed.go` will trip `-race` (and can misbehave on weak memory) if boxed values are ever touched from multiple goroutines:

1. **`BoxedTypes` map** (`valueType`, ~line 100/126): a plain global `map[string]*aBoxedType` read-then-written with no lock. Two goroutines boxing the same (or different) types concurrently is a concurrent map write — a runtime panic, not just a detector warning. `sync.Map` or a `sync.RWMutex` around the lookup/insert fixes it.

2. **`Boxed.Hash()` cache** (~lines 56–59): `if !n.hashed { n.hash = ...; n.hashed = true }` — a data race when two goroutines hash the same boxed value. Benign-ish today (both compute the same value), but `hashed=true` can be observed before `hash` is written under reordering, and it fails `-race`. `sync.Once`-per-value is heavy; a `uint64` with atomic load/store packing `hashed<<32|hash` (0 = uncomputed sentinel handling) or just an atomic.Uint64 of the hash+flag would do.

Today the VM is effectively single-goroutine in most paths, so this is latent rather than urgent — but boxed values embedded in IR vectors (per the `Hash()` comment) get hashed during lowering, and any future parallel lowering/eval would hit both immediately.

mparrett:
Fixes #540.

Two lazy-init sites in `pkg/vm/boxed.go` race under concurrent access. Both are latent today: the VM is effectively single-goroutine on most paths. But they trip `-race`, and become real the moment boxed values are touched from more than one goroutine, which is what the parallel-lowering direction in #464 would introduce (boxed values embedded in IR vectors get hashed during lowering, per the existing `Hash()` comment).

## What was racing

**`BoxedTypes` global map.** `valueType()` did a lock-free read-then-write on a plain `map[string]*aBoxedType`. Two goroutines boxing types concurrently is a concurrent map write — a runtime panic, not just a detector warning.

**`Boxed.Hash()` cache.** The `if !hashed { hash = …; hashed = true }` pattern is a data race on two fields. It's benign in value terms (both goroutines compute the same hash), but `hashed = true` can be observed before `hash` is written under reordering, so a reader can return a zero hash.

## The fixes

**Map:** a `sync.RWMutex` around lookup/insert. The method-table build (the reflection loop, the expensive part) stays outside the write lock; the insert rechecks the key under the lock so a concurrent caller that already inserted wins and every caller shares one `*aBoxedType` per type. Read hits — the common case after warm-up — take only the read lock.

**Hash:** a single `atomic.Uint64` replacing the `uint32` + `bool` pair. The computed flag is packed into bit 32, so `0` means uncomputed and a legitimate hash of `0` still reads as cached (the sentinel subtlety called out in the issue). Concurrent first-computes are harmless — same input, same value, idempotent store.

This also retires a workaround: `pkg/rt/net.go` warmed the `*netConn` type into the cache during single-threaded install because the map was unsafe, with a comment noting the real fix belonged in `pkg/vm`. That's this change, so the warm-up is gone.

## Test

`boxed_race_test.go` adds two `-race` regression tests that fail before this change and pass after: N goroutines box the same freshly-seen type at once (the map insert race), and N race to first-compute `Hash()` on one shared `Boxed` (the cache race). Both are `-short`-compatible and fast, so they run under the race job added in #524 (`go test -race -short ./pkg/vm/...`).

Verified: race tests fail on `origin/main`, pass here; full `-race -short` suite over `pkg/vm` + `pkg/rt` green; `gofmt` and `go vet` clean; builds cross `GOOS=linux`, `GOOS=js/wasm`, `GOOS=plan9`.

Refs #464, #528.

### DIFF
--- a/pkg/rt/net.go
+++ b/pkg/rt/net.go
@@ -141,13 +141,6 @@ func installNetNS() {
 		return vm.NIL, nil
 	})
 
-	// Warm vm.NewBoxed's type cache for *netConn now, while installers run
-	// single-threaded. The cache (vm.BoxedTypes) is an unsynchronized map, so
-	// the first boxing of a not-yet-seen type must not race; after this, every
-	// net/dial only reads the cache. (The map itself is shared by all Boxed
-	// types across the VM — a broader fix belongs in pkg/vm, not here.)
-	_ = vm.NewBoxed(&netConn{})
-
 	ns := vm.NewNamespace("net")
 	ns.Def("dial", dialFn)
 	ns.Def("write!", writeFn)
--- a/pkg/vm/boxed.go
+++ b/pkg/vm/boxed.go
@@ -8,6 +8,8 @@ package vm
 import (
 	"fmt"
 	"reflect"
+	"sync"
+	"sync/atomic"
 )
 
 type aBoxedType struct {
@@ -28,10 +30,13 @@ func (t *aBoxedType) Box(value any) (Value, error) {
 }
 
 type Boxed struct {
-	value  any
-	typ    *aBoxedType
-	hash   uint32
-	hashed bool
+	value any
+	typ   *aBoxedType
+	// hash caches the String-based hash lazily. It packs the "computed" flag
+	// into bit 32 so a legitimate hash of 0 still reads as cached: 0 means
+	// uncomputed, otherwise the low 32 bits hold the hash. Atomic so
+	// concurrent Hash() callers don't race (both compute the same value).
+	hash atomic.Uint64
 }
 
 // Type implements Value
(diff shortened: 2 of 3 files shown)


nooga (approved):
Best-verified of this batch: copied the new boxed_race_test.go onto unfixed origin/main and reproduced a genuine data-race failure under -race on the Hash() cache (write/write + read/write), confirming this is a real regression test, not vacuous. Ran both new race tests 20x under -race on this branch with no flakes. Checked the bit-packing by hand — 1<<32 | hash is always non-zero even when hash==0, so the uncomputed sentinel can never collide with a legitimate zero hash. gofmt/go vet clean, -race -short over pkg/vm + pkg/rt green, full suite green, js/wasm + plan9/amd64 cross-builds clean. The net.go workaround retirement is exactly the right cleanup now that the root cause is fixed.

PR #548, opened 2026-07-17
fix(compiler): honor require aliases before the core/* resolution shortcut

### LINKED ISSUE #529: Compile-time resolution fails for a require alias named `core` (shadowed by clojure.core)
Aliasing a required namespace as `core` makes compile-time symbol resolution through that alias fail: the alias appears to be shadowed by the auto-referred `clojure.core`. Any other alias name works.

## Repro

Two namespaces on one source path.

`repro/leaf.lg`:

```clojure
(ns repro.leaf)
(defn hi [] 42)
```

`repro/usec.lg`:

```clojure
(ns repro.usec (:require [repro.leaf :as core]))
(defn f [] (core/hi))
(println (f))
```

Run:

```
lg -source-paths . repro/usec.lg
```

Result:

```
error: Can't resolve core/hi in this context
```

Change the alias from `core` to anything else (e.g. `:as leaf`) and it compiles and prints `42`. Same failure via `-e`:

```
lg -source-paths . -e "(ns u (:require [repro.leaf :as core])) (core/hi)"
```

## Notes

- Reproduces on released **lg 1.11.1** and on latest build from main branch.
- The var exists at runtime — `(resolve 'repro.leaf/hi)` returns it; only compile-time resolution through the `core` alias fails.
- Likely `clojure.core` (auto-referred) shadows the `core` alias when the compiler resolves `core/…`.

Found in practice when a namespace aliased a project's `<x>.core` namespace as `core` and called one of its fns in a compiled body (a `deftest`). Renaming the alias, or resolving the var at runtime, works around it.

nnunley:
Fixes #529.

## Root cause

In the compiler's qualified-symbol path, any symbol whose namespace part was literally `core` (`rt.NameCoreNS`) short-circuited to the global core namespace before the current namespace's alias table was consulted. So `(:require [repro.leaf :as core])` registered the alias fine — runtime `(resolve 'repro.leaf/hi)` worked — but compile-time resolution of `core/hi` only ever looked at clojure.core and failed with "Can't resolve core/hi in this context". Any other alias name worked.

## Fix

The qualified-symbol case now calls `CurrentNS().Lookup(symVal)` first, which resolves aliases before anything else (matching Clojure's alias-wins semantics), and falls back to the global-core-ns lookup only when that misses and the namespace part is `core`. The fallback is kept because it still serves `(ns ...)` expansion before refers exist, and macroexpansions that reference private core vars with a qualified name.

## Verification

- New regression test `test/ns/alias_named_core_test.lg`: fails before the fix, passes after. It checks both that `core/join` resolves through the alias at all (`join` doesn't exist in clojure.core), and that `core/reverse` — a name present in **both** namespaces — resolves to the aliased lib, proving the alias shadows core rather than merely falling back.
- The issue's exact repro (`lg -source-paths . repro/usec.lg` and the `-e` variant) now prints `42`.
- Plain `core/inc`, `clojure.core/inc`, and unqualified resolution are unchanged.
- Full local suite green (`go test ./...` + the vendored Clojure compat suite at its recorded pin).

### DIFF
--- a/pkg/compiler/compiler.go
+++ b/pkg/compiler/compiler.go
@@ -421,20 +421,15 @@ func (c *Context) compileForm(o vm.Value) error {
 		symVal := o.(vm.Symbol)
 		// If qualified like ns/sym
 		if sns, inner := symVal.Namespaced(); sns != vm.NIL {
-			// Resolve core/* via global core ns so (ns ...) expansion works before refers
-			if string(sns.(vm.Symbol)) == rt.NameCoreNS {
-				target := rt.NS(rt.NameCoreNS)
-				v := target.Lookup(inner.(vm.Symbol))
-				if v == vm.NIL {
-					return c.compileError(fmt.Sprintf("Can't resolve %s in this context", symVal))
-				}
-				varn := c.constant(v)
-				c.emitWithArg(vm.OP_LOAD_VAR, varn)
-				c.incSP(1)
-				return nil
-			}
-			// Non-core qualified: honor aliases and refers in current ns
+			// Honor aliases and refers in current ns first — an alias literally
+			// named `core` must shadow the auto-referred core ns (issue #529).
 			v := c.CurrentNS().Lookup(symVal)
+			if v == vm.NIL && string(sns.(vm.Symbol)) == rt.NameCoreNS {
+				// Fallback: resolve core/* via the global core ns so (ns ...)
+				// expansion works before refers, and so macroexpansions may
+				// reference private core vars qualified.
+				v = rt.NS(rt.NameCoreNS).Lookup(inner.(vm.Symbol))
+			}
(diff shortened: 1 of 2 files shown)


nooga (approved):
Verified beyond the diff: reproduced the exact bug on pre-fix main (copied the new test over — fails with the precise "Can't resolve core/join in this context" from the issue), then confirmed it passes clean on this branch, including the harder half (core/reverse, present in both the aliased lib and real clojure.core, correctly resolves to the aliased lib — proving the alias genuinely shadows core, not just fills a gap). Traced that the fix reuses the same CurrentNS().Lookup(symVal) path already used for every other alias, so no new resolution logic is introduced. Full pkg/compiler suite and go test ./... -count=1 green.

PR #549, opened 2026-07-17
fix(lgbgen): stage --target=go output and install atomically with a completeness sentinel

nnunley:
## Problem

`lgbgen --target=go` cleaned the real `pkg/rt/core_go_lowered` tree first and then regenerated into it in place. Lowering takes minutes (typeinfer-bound), so a killed or crashed run left the real tree half-deleted for that whole window — and because the tree is gitignored, the damage was invisible to `jj`/`git status` while breaking every subsequent `go build -tags gogen_ir` with a wall of "no required module provides package .../core_go_lowered/..." errors until the next full regeneration.

## Change

- **Staged install**: generation now writes into a `core_go_lowered.stage` sibling and swaps into place only on success. The swap (`installGeneratedTree`) removes the old sentinel first, cleans previously-generated (bannered) files — same orphan-cleanup contract as before — renames staged files in, and installs a fresh sentinel last. Renames are same-filesystem, so the inconsistency window is the file-move loop, and it is positively detectable.
- **Completeness sentinel** (`pkg/genmanifest/tree.go`): `.lgbgen-tree.sum` lists a sha256 per generated file and is written as the last step of a successful install. `CheckTreeManifest` verifies every listed file is present with matching contents, so consumers can distinguish "complete tree" from "torn or partially written tree".
- The manifest is computed from the **stage** dir, not the real dir: it lists exactly what this generation wrote. The real tree legitimately hosts co-tenant packages installed by other tools (`scripts/gogen-trampoline.lg` lowers `test/gogen` fixtures into it, e.g. `examples/deftype_hyphen_field`, so their module imports resolve); the checker deliberately tolerates unlisted files, otherwise every parity run would invalidate the sentinel.
- `TestGogenAOTDiff` validates the sentinel before its `-tags gogen_ir` build, so a torn or missing tree fails with an actionable "run `make generate`" message instead of cryptic module-resolution errors.

## Verification

- 5 new unit tests in `pkg/genmanifest/tree_test.go`: round-trip, co-tenant tolerance, torn-tree detection, modified-file detection, missing-sentinel detection.
- Full regeneration through the staged path, then `make check-generated` green; sentinel verified to stay valid with a simulated co-tenant package present.
- Full local suite green (`go test ./...` incl. the e2e harnesses + the vendored Clojure compat suite at its recorded pin).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -191,6 +191,10 @@ pkg/rt/.core_compiled.lgb.committed
 # the untagged build and shipped bytecode binary never need them.
 pkg/rt/generated.provenance
 pkg/rt/core_go_lowered/
+# Staging sibling that --target=go generates into before the atomic swap into
+# core_go_lowered/. A successful install renames it away; ignore any residue
+# from an interrupted run so it never gets committed.
+pkg/rt/core_go_lowered.stage/
 /lg_gogen_ir.go
 /cmd/lgbgen/main_gogen_ir.go
 /pkg/ir/zz_gogen_ir_wire_test.go
--- a/cmd/lgbgen/main.go
+++ b/cmd/lgbgen/main.go
@@ -976,13 +976,23 @@ func runGoTarget(outDir, codeDir string) {
 // installGeneratedTree moves the staged generation into the real output dir:
 // the completeness sentinel is invalidated first, previously-generated
 // (bannered) files in realDir are removed — same orphan-cleanup contract as
-// before staging existed — then every staged file is renamed into place and a
-// fresh sentinel is written last. Rename is cheap and same-filesystem
+// before staging existed — then every staged file is renamed into place and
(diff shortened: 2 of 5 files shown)


nooga (approved):
This closes the unit-test follow-up I asked for on #541, and catches something real in the process: I reproduced a genuine bug in the already-merged #541 code — its CheckTreeManifest flagged any co-tenant file (e.g. what scripts/gogen-trampoline.lg legitimately installs alongside the generated tree) as "stray", causing a false-positive failure. Confirmed that repro fails on main today, and confirmed this branch fixes it (co-tenant file present, TestGogenAOTDiff passes) while still correctly catching a genuine torn tree (re-ran the edn.go-deletion repro — still fails with the right "torn tree" message). All 5 new tree_test.go tests pass, check-generated and the full suite are green. Good catch.

PR #550, opened 2026-07-17
docs(guide): document lg-runtime, net/bencode; clarify numeric vs dispatch hints

nnunley:
Documents three v1.12-cycle features that shipped without user-facing guide coverage, so the release notes can point at real docs.

## Changes

- **`guide/usage.md`** — new "Compiler-free deployment: lg-runtime" section covering `go build ./cmd/lg-runtime`, running `.lgb` directly, and `lg -b -bundle-base lg-runtime` for a standalone-and-compiler-free bundle (#424). Invocation verified against `lg.go` (`-bundle-base` flag) and `cmd/lg-runtime/main.go`.
- **`guide/net.md`** (new) — `net/dial`/`write!`/`read!`/`close!` and `bencode/write!`/`read!` (incl. the timeout arity), plus the js/wasm gating. API checked against `pkg/rt/net.go` and `pkg/rt/bencode.go` (#526, #533).
- **`guide/clojure-compatibility.md`** — the "type hints are NOT yet consulted" note now distinguishes *dispatch* hints (`^String`, still not consulted) from *numeric* hints (`^double`/`^long`, honored by the AOT native-Go lowering pipeline since #530). Previously a reader would conclude numeric hints do nothing.
- **`docs/README.md`** — index row for the new net page.

Docs-only; no code changes.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -61,6 +61,7 @@ A subdir is earned when a cluster of related docs justifies one; one-off cross-c
 | Running as a WASI (`GOOS=wasip1`) module | `guide/wasi.md` |
 | let-go under TinyGo (status: not currently supported) | `guide/tinygo.md` |
 | nREPL server + editor setup | `guide/nrepl.md` |
+| TCP client + bencode framing (`net`, `bencode`) | `guide/net.md` |
 | Clojure compatibility: namespace table + differences | `guide/clojure-compatibility.md` |
 | Running, compiling, WASM, project mgmt (lgx) | `guide/usage.md` |
 | IR fixup / link pass | `design/els2023-ir-fixup-audit.md` |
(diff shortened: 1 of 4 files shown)


nooga (approved):

mparrett (approved):
The net/bencode signatures check out against `pkg/rt/net.go` and `pkg/rt/bencode.go` on main: dial's 3s timeout, write! taking String or byte-array, read! returning nil on clean EOF, and the 2-arg read! timeout are all faithful. Two small notes on `net.md`:

**Gating is browser-wasm only.** The page says net and bencode are "gated off js/wasm builds, where sockets don't exist." The build tags are `//go:build !js` (real impl) and `//go:build js && wasm` (stub), so the gate is `GOOS=js` specifically. Under `GOOS=wasip1`, net.go compiles and TCP works, since wasip1 has real sockets. The guide already has a wasi page, so "js/wasm" here could read as if wasi is stubbed too. Suggest "gated off browser (`GOOS=js`) builds." The wording is inherited from net.go's own header comment, so worth fixing in both if you touch it.

**write! also encodes a bare keyword.** The description lists keywords only as map keys, but `bencodeFromLG` encodes a standalone keyword value to its name (`:foo` to `"foo"`). Minor, only flag if you want the value list complete.

Neither blocks. LGTM once the js/wasm line is tightened.

PR #552, opened 2026-07-17
bench: VM-vs-AOT comparison — lower fixtures to native Go, drop joker/go-joker/gloat/fennel

nooga:
## What

Reworks the external benchmark suite around the question that actually matters now: **what does the IR-lowering pipeline buy over the bytecode VM on identical code?** The matrix is now:

| leg | what runs |
|---|---|
| **let-go** | plain build, bytecode dispatch |
| **let-go AOT** | `lg_bench`-tagged build with the same fixtures IR-lowered to native Go, dispatched via registered overrides |
| **babashka** | plain `.clj` scripts |
| **clojure JVM** | plain `.clj` scripts via `clj -M` |

joker, go-joker, gloat, and fennel comparisons are removed (support dirs deleted), and `benchmark/README.md` is rewritten for the new suite.

## How to run / repeat

Prereqs: Go, `hyperfine`, python3; optionally `bb` and `clj` (auto-detected, skipped if absent).

```bash
# full run — builds both binaries, runs everything sequentially,
# regenerates benchmark/results.md (takes several minutes; keep the machine idle)
bash benchmark/run.sh

# iterate on one benchmark (more runs, tighter σ, no results.md rewrite)
bash benchmark/run.sh fib
```

`run.sh` builds the plain binary, then `benchmark/aot/build.lg` lowers every fixture in `benchmark/aot/src/bench/*.lg` via `ir.passes.pipeline/lower-ns-to-go`, wires the lowered packages into one `lg_bench`-tagged binary through a generated `zz_lg_bench_gen.go`, verifies **every** fixture dispatches native (a silent bytecode fallback fails the build), and cleans up everything it generated. Full instructions, layout, and the add-a-benchmark recipe are in `benchmark/README.md`.

## How the AOT leg works

- `benchmark/aot/src/bench/*.lg` — the 7 workloads as fixtures (`(ns bench.<name>)` + 0-arg `run`), same code as the `.clj` scripts.
- VM and AOT legs run the *identical* driver + fixture — only dispatch differs. Wall-clock includes startup + namespace load for both.
- Mechanism mirrors `scripts/gogen-trampoline.lg`.

## Results (Apple M1 Pro, hyperfine, full table in benchmark/results.md)

| Benchmark | let-go | let-go AOT | babashka | clojure JVM |
|---|---|---|---|---|
| fib | 2.424s | **0.113s** | 1.935s | 0.603s |
| tak | 2.432s | **0.094s** | 1.939s | 0.649s |
| loop-recur | 70.9ms | **10.6ms** | 69.0ms | 501ms |
| map-filter | **10.7ms** | 12.4ms | 22.7ms | 398ms |
| persistent-map | **23.0ms** | 24.0ms | 29.3ms | 534ms |
| reduce | 41.0ms | 68.2ms | **37.6ms** | 376ms |
| transducers | 60.9ms | 49.4ms | **30.8ms** | 431ms |

Headline: lowered fib/tak run **~21–26x faster than the VM** and ~5–7x faster than Clojure JVM wall-clock — the typed-native emission turns them into raw Go int recursion. Kept honest: seq-heavy benches are at parity (lazy-seq machinery runs through the runtime either way) and **reduce is currently slower lowered (41→68ms)** — worth its own investigation.

## Notes / follow-ups

- Emitter gap found while wiring this up: a fn that lowers to a *fully typed* signature (e.g. `func(ec *vm.ExecContext) int`) gets no boxed wrapper and no `init()`/`RegisterGoOverrides` entry, so it can't be drained as an override. `bench.loop-recur` works around it by splitting the typed loop into its own defn behind a boxed `run` wrapper (documented in the fixture and README). Probably deserves an issue.
- `reduce` lowering regression (41→68ms) deserves an issue too.


mparrett:
Reran the suite independently on an M2 (idle box) and read the generated Go for each fixture. The numbers reproduce closely, and both follow-up notes you flagged check out. One reframing is worth adding to the table.

## Reproduction

Ratios match yours closely despite different silicon and a slower host, with tight σ once the machine was idle:

| bench | your M1 (VM→AOT) | your ratio | my M2 (VM→AOT) | my ratio |
|---|---|---|---|---|
| fib | 2.424→0.113s | 21.4× | 3.00→0.139s | 21.5× (±0.6) |
| tak | 2.432→0.094s | 25.9× | 3.02→0.117s | 25.8× (±1.6) |
| loop-recur | 70.9→10.6ms | 6.7× | 73.5→6.9ms | 10.6× (±1.2) |
| persistent-map | 23.0→24.0ms | ~1.0× | 22.8→21.8ms | 1.05× |
| transducers | 60.9→49.4ms | AOT 1.2× | 58.0→51.2ms | AOT 1.13× |
| map-filter | 10.7→12.4ms | VM 1.16× | 7.9→11.3ms | VM 1.43× |
| reduce | 41.0→68.2ms | VM 1.66× | 46.9→74.8ms | VM 1.59× (±0.04) |

## Two mechanisms under one column

Reading the emitted Go, the fixtures split cleanly:

**Lowered to native code (fib, tak, loop-recur).** The hot code becomes native Go over unboxed `int`: `arg0 <= 1`, `arg0 - 1`, a direct `Fib(ec, arg__3)` recursion, `for { acc = acc + i; i = i + 1 }`. Only the leaves box. That is where the 20–26× comes from.

**Trampolined, not lowered (reduce, map-filter, persistent-map, transducers).** `Run` becomes a native fn, but every hot op is still a runtime invoke with a boxed var lookup. `reduce.go` in full:

```go
arg__4, _ = ec.Invoke(range, {vm.Int(1000000)})   // boxed seq, same as VM
if rt.NativePrimsIntact() {
    v6, _ = rt.Reduce3(ec, ec.Deref(LookupVar("clojure.core","+")), vm.Int(0), arg__4)
} else {
    v6, _ = ec.Invoke(reduce, ...)
}
```

The reduction never lowers — it calls the identical runtime `rt.Reduce3` with a boxed `+`. Even the mapped `#(* % %)` lowers to `rt.MulValue(arg0, arg0)` on a boxed `vm.Value`. So these rows do the same work as the VM, wrapped in a native-fn dispatch plus a `NativePrimsIntact()` branch, which can only tie or lose.

That answers your reduce note: the emitted code shows the reduction was never lowered, so the 1.6× is the wrapper's structural cost. No separate investigation needed. The parity rows are the same story — the bodies run identical runtime code either way. So the sharper headline is narrower: native lowering is a large, reproducible win for scalar-recursive kernels; seq/reduce/transduce bodies trampoline through the runtime and don't benefit. Splitting the table's framing along that line would make the reduce row read as expected, not anomalous.

## The typed-signature gap

Confirmed in `loop_recur.go`: the emitter does produce a typed `func SumLoop(ec) int`, but `RegisterGoOverrides` only registers `"run"` (the boxed wrapper) — there's no `"sum-loop"` entry, so the typed leaf can't be drained on its own. That's the gap you described, an emitter limitation, not a fixture quirk. Worth its own issue, and I'd rank it above the reduce note: a hot typed leaf fn is the case you'd most want to lower cleanly, and today it needs the hand-split.

## Smaller notes

- The `transducers` VM cell in the committed `results.md` is 60.9 ± 21.3ms (35% σ); my idle VM run was 58.0 ± 0.9ms. Looks like contention on that capture — the ratio holds, but worth re-running that row before the table gets cited.
- Each workload lives in two hand-maintained copies (`benchmark/<name>.clj` for bb/clj, `aot/src/bench/<name>.lg` for both let-go legs) with nothing checking they match. All seven are byte-identical today, so this is prospective, but a normalize + diff in `build.lg`, or a line in the add-a-benchmark recipe, would keep the cross-runtime comparison accurate.
- Minor: no `trap` to clean the wireup + lowered tree on a mid-build Ctrl-C. Both are gitignored now so it's cosmetic.

The native-dispatch guard in `build.lg` (fail the build if any fixture silently falls back to bytecode) is the right call — it's what keeps an AOT benchmark accurate, and it's solidly built-in.

mparrett (approved):
Left a more detailed response as a comment. A few minor things for consideration, but nothing blocking on my side. LGTM.

mparrett:
>  Splitting the table's framing along that line would make the reduce row read as expected, not anomalous.

Might be worth discussion?

On why the seq rows are unfused (correcting an earlier version of this comment — I had the fusion pass wrong): `fusion.lg` carries the full rewrite, `(reduce g init (map f coll))` → `(transduce xform (completing g) init coll)`, and `*enable-fusion*` defaults to true, so it's on — thanks @nnunley for the flag pointer. The four seq rows are unfused because they don't match the pass's shape (consumer `reduce`/`into` directly over producer `map`/`filter`/`remove`/`keep`): `reduce` and `persistent-map` reduce over `range`, `map-filter` has a `take` between the reduce and the chain, `transducers` is already a `transduce`. So the flag is a no-op for this suite — building the AOT binary both ways, the `reduce` row is identical (57.0 vs 57.5ms).

On a fixture that does match, `(reduce + 0 (map inc (range 1M)))`, the lowered transduce form ran slower than unfused (124→174ms; VM 119ms). It looks like it loses the chunked-reduce fast path for a cheap body, so the allocation saving doesn't convert to wall-clock. Small sample and a deliberately cheap body, not a general claim on the pass — but it means fusion isn't the lever that moves these lowered seq rows.

Separately, filed the typed-signature override gap from your PR notes as #554.

PR #555, opened 2026-07-18
docs(design): reference for IR pipeline dynamic vars

nnunley:
Adds `docs/design/ir-dynamic-vars.md` — a single index for the IR pipeline's `^:dynamic` configuration surface, which was undocumented and scattered across `core.lg`, `passes/pipeline.lg`, `passes/inline.lg`, `passes/fusion.lg`, `passes/typeinfer.lg`, and `lower_go.lg`.

## Why

The IR compile/lowering pipeline is configured almost entirely through dynamic vars rather than flags. Only `*strict-structured?*` has an env seed (`LG_STRICT_STRUCTURED`); everything else is `binding` / `set!`. There was no index, so the knobs were effectively discoverable only by grepping.

## Contents

The doc separates the two kinds of var:

- **Knobs you set** — grouped into compilation-mode control (`*ir-compile*`, `*ir-compile-verbose*`, `*ir-compile-fallback-log*`, `*target*`), pass toggles/tuning (`*enable-fusion*`, `*enable-inline*`, `*max-unroll*`, `*typeinfer-max-drains*`, `*strict-structured?*`, `*direct-calls-disabled?*`, `*pass-trace*`), and cross-package/exported-wrapper control for `--target=go` (`*emit-exported-wrappers*`, `*cross-pkg-registry*`, `*wrapper-target-names*`, `*export-name-overrides*`, `*deftype-ctor-types*`). Each with default + semantics + `file:line`.
- **Per-compile state** — the ~13 `nil`/empty-initialized vars the pipeline rebinds as it runs (traversal cursor, registries, emission bookkeeping, type/dispatch context), listed so they're discoverable but clearly marked as not-a-settings-surface.

Semantics are distilled from each var's in-source docstring. Added a row to the `docs/README.md` topical map. Frontmatter follows the design-doc convention (`human-verified` left blank per the docs-status contract).

Docs only.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -63,6 +63,7 @@ A subdir is earned when a cluster of related docs justifies one; one-off cross-c
 | nREPL server + editor setup | `guide/nrepl.md` |
 | Clojure compatibility: namespace table + differences | `guide/clojure-compatibility.md` |
 | Running, compiling, WASM, project mgmt (lgx) | `guide/usage.md` |
+| IR pipeline dynamic vars (knobs + per-compile state) | `design/ir-dynamic-vars.md` |
 | IR fixup / link pass | `design/els2023-ir-fixup-audit.md` |
 | Parallel IR lowering + determinism | `design/parallel-lowering-and-type-cache.md` |
 | Runtime I/O, host decoupling | `design/runtime-io-host-decoupling.md` |
--- a/docs/design/ir-dynamic-vars.md
+++ b/docs/design/ir-dynamic-vars.md
@@ -1,6 +1,6 @@
 ---
 status: active
-last-verified: 2026-07-17
+last-verified: 2026-07-19
 authoritative-for:
   - ir-dynamic-vars
 human-verified:
@@ -26,6 +26,11 @@ Two kinds of var live here:
 For build tags, environment variables, and `lgbgen` CLI flags, see the
 companion sections in the guide; this page covers the dynamic vars only.
 
+Every knob here changes what the pipeline converts to IR (or how it lowers
+it), so a change to one is only as good as its coverage evidence. Measure it
+with the ir-stress harness — see [Verifying a var change](#verifying-a-var-change)
+at the end of this page and `scripts/ir-stress.md` for the harness itself.
+
 ## Compilation-mode control
 
 The entry knobs that decide whether IR compilation runs at all and which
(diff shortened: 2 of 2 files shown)


mparrett (commented):
This is a helpful index! I went looking for exactly this reviewing #552 and ended up grepping all six files. Two notes.

The `*enable-fusion*` entry is right (deforestation, on by default). Worth flagging that `fusion.lg`'s own header comment still reads "v1 recognizes the shape only … No rewrite" — stale, since `fuse!` / `rewrite-stages!` do the rewrite. The doc gets it right; the source header is what misled me on #552. Might be worth fixing the header too, here or separately.

The gap: `*ir-compile*` is the knob callers actually set, and the entry gives its semantics but not its cost, which is the one var with a real amortization story. Pipeline load + per-fn compile is a one-time cost, so in a quick check it pays off only amortized, and only for allocation-bound work: ~5–12× fewer allocations on alloc-heavy workloads (persistent-map, transducers, reduce), converting to wall-clock (≈−13%/run, break-even ~64 runs on persistent-map) when heap churn dominates — but a small net loss on compute-bound code (fib, loop-recur) that never breaks even. Note `reduce` over `range` drops ~78% allocations with no fusion applicable, so most of it is the pipeline cutting per-element boxing, not `*enable-fusion*`. A one-line "cost: pays off amortized, for alloc-bound work" under that entry would save the next reader the surprise I had.

Caveat: gctrace GC-cycle proxy on one machine, so `BenchmarkIRCompile` is the real instrument — can share the runs if useful.

nnunley:
Updated in `daaca44`, covering both your notes plus the coverage-verification gap.

**New "Verifying a var change" section.** Every knob on this page decides what converts to IR or how it lowers, so a change to one is a coverage change until measured — and the page previously said nothing about how to measure. It now documents ir-stress as the instrument: `make ir-stress` (census over `scripts/ir-stress-corpus.edn`), `make ir-stress-gate` (ratcheted against `docs/perf/ir-stress-baseline.edn`), `make jank-stress` (Clojure surface the internal corpus misses), `make parity-full` (both modes × both engines, bucket-by-symbol). Plus which mode answers which claim — `lower-go` for AOT conversion, `ir-compile` for eval-mode, `trace` when a tuning knob produces `:stress/timeout` — and a before/after bucket-diff recipe over the `LG_STRESS_LOG` TSV, with `make ir-stress-rebaseline` for intentional coverage moves. Cross-links `scripts/ir-stress.md`. Per-knob pointers added to `*enable-fusion*`, `*enable-inline*`, and `*max-unroll*`.

**Your `*ir-compile*` cost gap** — agreed, that was the missing half of the entry, and it's now a callout under the compilation-mode table: one-time pipeline+compile cost, pays back only amortized and only on alloc-bound work (~5–12× fewer allocs, ≈−13%/run, break-even ~64 runs on persistent-map), a small net loss on compute-bound code that never breaks even, and the note that `reduce` over `range` drops ~78% with no fusion applicable — so most of it is per-element boxing, not `*enable-fusion*`. Attributed to you with the single-machine gctrace-proxy caveat stated and `BenchmarkIRCompile` named as the real instrument. I'd take those runs if you're willing to share them — happy to fold firmer numbers in.

**The stale `fusion.lg` header** — you're right that it's misleading ("v1 recognizes the shape only … No rewrite" while `fuse!`/`rewrite-stages!` clearly rewrite), and it's worth fixing. Doing it separately rather than here: `pkg/rt/generated.sums` is a single aggregate digest over all `.lg` sources, so even a comment-only edit trips the staleness gate and drags a `make generate` run into what is otherwise a docs-only PR. I'll send it as its own change.

Verified locally: `python3 scripts/docs_frontmatter_hook.py --check` passes, every `make` target named exists, the TSV field order in the recipe matches the harness (`file⇥defn⇥bucket⇥ms`), and the section anchors resolve.

mparrett (approved):

mparrett (approved):
Read through — matches the pipeline behavior on main. Merging.

PR #556, opened 2026-07-18
fix(test): self-heal a drifted clojure-test-suite submodule before the compat suite

nnunley:
## Problem

jj does not update git submodules on checkout. So a jj working copy can sit at an older `test/clojure-test-suite` revision than the parent repo's gitlink pins. The local pre-push gate then runs `TestClojureTestSuite` against a **stale** suite and fails on assertions that pass on CI (which runs `git submodule update --init`).

This is exactly what happened after #429 bumped the pin `41290a6` → `bd610c81` (which already carries upstream's `:lg` order-tolerant branches): a jj checkout left the submodule at `41290a6`, so `cons`/`cycle`/`mapcat` failed locally while CI was green. (See the closed #553, which misdiagnosed this as a dropped patch.)

## Change

Add `syncJankSubmodule()`, called from `TestMain` right before `applyJankPatches()`:

- Reads `git submodule status test/clojure-test-suite` (leading char: `' '` in sync, `'+'` drifted, `'-'` uninitialized).
- In sync → no-op fast path. Uninitialized → left to the existing skip. Drifted → warn to stderr and run `git submodule update --init` to **self-heal** to the pinned SHA.
- **Best-effort**: never fails the run; a still-drifted suite fails loudly on its own assertions, which beats a silent stale pass.
- Leaves the submodule at the pin afterward (unlike `revertJankPatches`, which undoes test-time overlays — the pin is the authoritative state).
- jj ignores the submodule, so checking out a different SHA there creates no jj snapshot churn.

## Verification

Forced the submodule to the old `41290a6`, then ran `TestClojureTestSuite/{cons,cycle,mapcat}`: the guard warned, ran `git submodule update`, self-healed to `bd610c81`, and the tests passed — with the submodule left at the pin. `go vet ./test/` clean.

Test-harness only; no runtime change.

### DIFF
--- a/test/fanout_ratchet_test.go
+++ b/test/fanout_ratchet_test.go
@@ -7,6 +7,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"sync"
 	"testing"
 )
 
@@ -28,12 +29,10 @@ func TestMain(m *testing.M) {
 		panic("build lg: " + err.Error() + "\n" + string(out))
 	}
 	repoRoot, _ = filepath.Abs("..")
-	syncJankSubmodule()
-	appliedPatches := applyJankPatches()
 	code := m.Run()
 	// Leave the submodule worktree as we found it — these are test-time
 	// overlays, not checkout changes the run should persist.
-	revertJankPatches(appliedPatches)
+	revertJankPatches(jankPatches)
 	os.RemoveAll(tmp)
 	os.Exit(code)
 }
@@ -81,6 +80,7 @@ func TestPatchOverlayApplyIdempotentAndRevert(t *testing.T) {
 }
 
 func TestJankSuiteCoversLetGoUnicodeScalar(t *testing.T) {
+	ensureJankSuite()
 	fixture := filepath.Join("clojure-test-suite", "test", "clojure", "core_test", "char.cljc")
 	got, err := os.ReadFile(fixture)
 	if os.IsNotExist(err) {
@@ -92,6 +92,26 @@ func TestJankSuiteCoversLetGoUnicodeScalar(t *testing.T) {
 	if !bytes.Contains(got, []byte(`:lg      (= (first "𐅧") (char 65895))`)) {
 		t.Fatal("patched fixture lacks the let-go Unicode scalar expectation")
 	}
+}
+
+// jankSuiteSetup gates suite preparation so it happens once, and only when a
+// test or benchmark that actually reads the suite runs. TestMain must not do
+// this eagerly: it executes before the -run filter applies, so an eager sync
+// would move the submodule checkout for every `go test ./test` invocation,
+// including runs that never touch the suite.
+var (
+	jankSuiteSetup sync.Once
+	jankPatches    []string
+)
+
+// ensureJankSuite syncs the submodule to its pinned SHA and overlays the :lg
+// fixture patches. Call it at the top of anything that reads clojure-test-suite.
+// Patches are reverted by TestMain after the run.
+func ensureJankSuite() {
+	jankSuiteSetup.Do(func() {
+		syncJankSubmodule()
+		jankPatches = applyJankPatches()
+	})
 }
 
 // syncJankSubmodule brings the clojure-test-suite working tree to the commit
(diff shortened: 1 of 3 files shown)


mparrett:
I found two issues while reviewing this change at `b5693c5`:

1. **[P2] Avoid changing the submodule for unrelated tests**

   `syncJankSubmodule()` is called from `TestMain`, which runs before Go applies the `-run` test filter. As a result, every `go test ./test` invocation can persistently move the submodule checkout—even when `TestClojureTestSuite` is not selected. Also, `+` from `git submodule status` covers any checkout that differs from the gitlink, including an intentional developer checkout.

   I reproduced this by checking out `41290a6` and running only `TestFanoutShow`; the command moved the submodule to `bd610c81` and left it there. This should be scoped to the compatibility suite, or the original checkout should be restored when the compatibility suite is not explicitly being run.

2. **[P3] The diagnostic reports the checkout SHA as the pin**

   For a `+` status, `git submodule status` prints the submodule's currently checked-out SHA. Consequently, `pinned` contains the drifted revision. In the reproduced case, the message was:

   ```
   note: test/clojure-test-suite drifted from pin 41290a61131a
   ```

   but the gitlink actually pins `bd610c81`. The code should either read the index gitlink SHA separately or phrase this as “checkout X differs from pin.”

Validation otherwise passed: `go vet ./test/` and focused `cons`, `cycle`, and `mapcat` compatibility tests.

nnunley:
Both addressed in 8c0521c:

- **P2**: suite setup (sync + patch overlay) moved out of `TestMain` into a `sync.Once` (`ensureJankSuite()`) invoked from the three suite consumers (`TestClojureTestSuite`, `TestJankSuiteCoversLetGoUnicodeScalar`, the two suite benchmarks). Your repro — `-run TestFanoutShow` — no longer touches the submodule at all. Patch revert stays in `TestMain` and is a no-op when setup never ran. An intentional `+` checkout still gets moved *when the compat suite itself runs*, which I think is right: the suite's assertions are only meaningful against the pin.
- **P3**: the note now reads the pin from the index gitlink (`git ls-files -s`) and prints `checkout X differs from pin Y` — the SHA `submodule status` prints for `+` was indeed the drifted checkout, as you said.

Verified: `go vet ./test` clean; `-run TestFanoutShow` leaves the submodule untouched; suite entry skips cleanly where the submodule is uninitialized.

mparrett:
The code changes themselves look correct. I independently verified:

- Unrelated tests leave a drifted submodule untouched.
- Suite access self-heals to the pinned revision.
- The diagnostic reports checkout and pin SHAs correctly.
- Patch cleanup leaves the submodule clean.
- `go vet ./test` passes.

**Remaining before merge:** Rebases/conflict resolution + CI pass.

mparrett (approved):
Changes look good. Just needs generated.sums replacement + CI green.

nnunley:
Force-pushed a cleanup that rewrites this PR down to its intended scope. Heads up: the prior approval was against the old tree, so this needs a fresh look.

**What was wrong.** The base commit had been captured by a whole-working-copy snapshot and carried far more than the submodule self-heal:
- a committed **16,529-line build artifact**, `pkg/rt/core_go_lowered.stage/core/core.go` (the lgbgen staging tree — already `.gitignore`d on main at line 197, soon an OS tempdir via #614);
- two **unrelated, unmerged IR/compiler features** — a `maybeIRCompileDefFnArg` / `compile-def-fn-value` def+name\* IR-compile seam, and a `lower.lg` block-junk *agreement* fix — neither of which touches the submodule.

**What this push does.** Reduced to the self-heal and only that:
- `test/fanout_ratchet_test.go` — `syncJankSubmodule()` brings the clojure-test-suite worktree to the pinned gitlink before the compat suite runs (jj doesn't update submodules on checkout);
- `zz_compat_test.go` / `zz_bench_test.go` — route their suites through it via `ensureJankSuite()`.

Full diff is now **+65/−2 across 3 test files**, zero non-test changes. Rebased on current main, builds clean, `go test ./test/` green.

The two IR/compiler features were **preserved, not dropped** — extracted to a separate branch (they're the only copy; not in main) for their own PR against the grammar/yamlstar effort, where the block-junk fix can be reviewed against #579's overlapping `lower.lg` rework.

mparrett (approved):
LGTM.

PR #557, opened 2026-07-18
feat(gogen): self-contain gogen as embedded core source (#425 Item 2)

mparrett:
Closes the #425 Item 2 gap: shipped `lg` binaries now embed the `gogen` macro layer, so the AOT lowering path resolves it without any `LG_SOURCE_PATHS` / cwd plumbing.

## Why

`gogen` is the `.lg` macro layer the whole go-lowering path compiles through: `scripts/lg-compile`, the IR-gen specs (`ir_ops` / `ir_bridge` / `ir_data`), and `ir.lower-go`'s emitter all `(require 'gogen)`. It shipped as an external classpath dir (`deps.edn {:paths ["pkg/rt/gogen"]}`), so any `lg-compile` invoked from the wrong cwd (or with a binary predating the gogen native ns) died with `Can't resolve gogen/ns->go-pkg`. That is the turnkey-AOT blocker Item 2 tracks: it gates every kernel-AOT build.

## How

Embed `gogen.lg` into shipped binaries and serve it to the resolver as an auxiliary embedded source (`rt.EmbeddedSource`), so `(require 'gogen)` resolves with no external source path. Two constraints shaped the mechanism — the first is the non-obvious one:

1. **gogen stays out of `pkg/rt/core/`.** The embedded-core tree (`//go:embed all:core`, enumerated by `EmbeddedNSNames`) is the universe the self-hosting bootstrap (`cmd/lgbgen`) walks. gogen is the AOT Go-emitter, not core content, and enrolling it there is enough to break the interpreted-vs-native lowering fixpoint: `TestLoweringDeterminism` starts seeing `clojure.core/apply` (and `str`) lowered as a direct native call in one pass and a trampoline in the other. So gogen is embedded from `pkg/rt/gogen/` via a dedicated `//go:embed` (`pkg/rt/gogen_src.go`) and registered in an `auxEmbeddedSources` map that `EmbeddedSource` checks first — resolvable like embedded core, invisible to the lowering universe.

2. **The embed is gated `//go:build !bootstrap`.** During the self-hosting bootstrap (lgbgen, `-tags bootstrap`) gogen must resolve exactly as it does today — from the classpath (`deps.edn`), which lgbgen always has since it runs from the repo root. Making gogen embedded-resolvable in the bootstrap binary is itself what perturbs the fixpoint; excluding the embed there keeps cold and hot lowering byte-identical. Shipped binaries (no bootstrap tag) get the embed and the self-containment.

`deps.edn` keeps its classpath entry (the bootstrap relies on it). The change is bootstrap-neutral: `core_compiled.lgb`, `core_go_lowered/`, and `generated.sums` are unchanged.

It also drops source-path plumbing that was dead for the shipped path: the Makefile / `generate.lg` `--source-paths pkg/rt/gogen` (`generate.lg` loads the specs in-process on `./lg`, which now embeds gogen) and the IR-spec run-hints. (`test/language_test.go` keeps its classpath entry — it runs under `-tags bootstrap` too, where the embed is absent.)

## Verification

- A shipped `lg` runs a full `lg-compile` of `examples/aot/cross-package` and the mandelbrot AOT kernel from an arbitrary cwd with no `LG_SOURCE_PATHS`, byte-identical Go (escape → native `Escape(ec, float64, float64, int)`).
- A new `!bootstrap` `pkg/resolver` regression test resolves gogen under empty search paths — the pre-Item-2 failure mode.
- `TestLoweringDeterminism` cold == hot; `go test ./...` green under both `-tags bootstrap` and the default build; linux / js-wasm / plan9 / wasip1 cross-builds pass; gofmt clean.

## Scope

This is #425 Item 2 only (self-contain gogen). Item 1 (generate the native-entry frame) and Item 3 (signal the silent perf-cliff) are separate.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -85,11 +85,12 @@ pkg/rt/core_go_lowered/ir/lower_go/lower_go.go: $(CORE-LG-FILES) $(LGBGEN-SOURCE
 # scripts/generate.lg: the three Go-gen files (op_generated.go,
 # ir_bridge_generated.go, ir/data/generated.lg), the core_compiled.lgb bundle,
 # and the lowered-Go tree. Requires ./lg, so it builds first. The orchestrator
-# uses os/sh + os/exec* to run ./lg on the pkg/rt/gogen sources and to run
+# uses os/sh + os/exec* to run ./lg on the IR-gen specs (which require the
+# embedded gogen macros, #425) and to run
 # `go run -tags bootstrap ./cmd/lgbgen [--target=go]` for the bundle/lowered
 # tree. (Replaces the former generate-ir-{ops,bridge,data}.sh shell scripts.)
 generate: build
-	./lg scripts/generate.lg --go "$$(command -v go)" --lg ./lg --source-paths pkg/rt/gogen
+	./lg scripts/generate.lg --go "$$(command -v go)" --lg ./lg
 
 # Short commit for `-X main.commit` so SHA-pin require-letgo checks can fire on
 # `make` builds. Release builds get this from goreleaser; a bare `make` build
--- a/pkg/ir/ir_bridge.lg
+++ b/pkg/ir/ir_bridge.lg
@@ -1,7 +1,7 @@
 ;; Generate pkg/rt/ir_bridge_generated.go — Lisp/Go bridge primitives
 ;; for IR types. See docs/superpowers/specs/2026-05-22-defgostruct.md.
 ;;
-;; Run with: lg -source-paths pkg/rt/gogen pkg/ir/ir_bridge.lg
+;; Run with: lg pkg/ir/ir_bridge.lg
 ;;
 ;; Each struct is declared via defgostruct. Two emission paths:
 ;;
--- a/pkg/ir/ir_data.lg
+++ b/pkg/ir/ir_data.lg
@@ -5,7 +5,7 @@
 ;; types, defirdata declares Lisp-native data types (atom-backed maps)
 ;; and emits the boilerplate accessors / mutators / constructors.
 ;;
-;; Run with: lg -source-paths pkg/rt/gogen pkg/ir/ir_data.lg
+;; Run with: lg pkg/ir/ir_data.lg
 ;;
 ;; Spec shape per type:
 ;;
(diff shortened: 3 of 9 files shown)


nnunley:
Multi-agent review sweep result: **no blockers from me** — this is in good shape and, importantly, roadmap-compatible.

What was verified:
- Committed artifacts (`core_compiled.lgb`, `core_go_lowered/`, `generated.sums`) are byte-identical to main — the self-containment is genuinely inert to the runtime artifacts.
- The `//go:build !bootstrap` gating is correct and the bootstrap path has no circular dependency on the artifact it produces.
- Cross-platform builds (linux, js/wasm, wasip1) pass.
- Alignment with where the IR work is headed (#528, #574, #575): keeping gogen *outside* `core/` for determinism is the right call, and nothing here fossilizes op tables in a place the defop catalog can't later replace. 👍

Two non-blocking notes:
1. `auxEmbeddedSources` (core_embed_fs.go:30) is a bare map that's safe **only because registration happens exclusively from `init()`** (gogen_src.go:28) — init happens-before everything, so no race as written. Worth a comment on the map stating that invariant, because the first person to add a runtime `RegisterEmbeddedSource` call would introduce a data race with resolver reads.
2. An empty/corrupted embedded source currently fails quietly through the resolver fallback chain; a loud error naming the ns would make future embed bugs much cheaper to find.

mparrett:
Both addressed in ffc28a5.

1. Documented the invariant on `auxEmbeddedSources`: bare lockless map, safe only because registration is init-only (happens-before every resolver read), with an explicit warning against adding a runtime `RegisterEmbeddedSource`.
2. Registration now goes through `registerEmbeddedSource`, which panics naming the namespace when the source is empty — so a `//go:embed` that never populates fails loudly at startup instead of surfacing later as a stray `Can't resolve <ns>/...`.

I scoped the empty-source guard to the aux-embed path (what this PR adds); an empty file on the `core/` path is a separate pre-existing case I left alone. Thanks for the roadmap read too — good to know keeping gogen out of `core/` lines up with the defop-catalog direction.

nnunley (approved):
Reviewed via multi-agent lens pass with per-finding verification. Embed path/dependency-closure correct, bootstrap isolation preserved (//go:build !bootstrap), auxEmbeddedSources written only in init() before resolver reads, empty-embed panics loudly instead of silently falling back, and gogen stays invisible to the lowering fixpoint. Regression test exercises resolution with empty search paths. LGTM.

PR #559, opened 2026-07-18
perf(vm): use the transient growth reserve in assocTransient inserts

mparrett:
`ensureEditable` deliberately over-allocates ("avoid frequent realloc during transient batch ops"), but the new-entry branch of `assocTransient` never used the spare capacity: every insert into an owned node allocated a fresh exact-size array and copied all slots. This is the path every map literal (`NewArrayMap`) and `into` take.

The change: when the owned array has room, insert in place with an overlap-safe `copy` shift. When the reserve is exhausted, reallocate with headroom capped at the node's structural max (32 entries / 64 slots).

To be precise about what improves: this removes the per-insert allocation and full-array copy. The in-place shift still moves the tail on every insert (about half the node on average, bounded by the 32-entry fan-out), so slot copying is reduced, not eliminated — allocations are what amortize across a batch.

Benchmarks (interleaved A/B against main, n=8, Apple M2):

| bench | time | B/op |
|---|---|---|
| TransientMapAssocInsertBatch1K | −24% | −52% |
| DepsTransientMapTransientSet64x16 | −27% | −52% |
| DepsTransientMapVector64x16 | −8% | −16% |
| overwrite benches | unchanged | unchanged |

The overwrite benches staying flat is expected: they never hit the new-entry branch.

**Tradeoff:** the headroom is bounded but persists. A transient-built map keeps up to cap−len unused slots per multi-insert node in its frozen form. Measured on a 512-entry `NewArrayMap` build: ~27% capacity slack, vs ~0 before this change (about 5.7 KB extra on that map). Single-insert nodes waste nothing. So this trades some steady-state footprint on batch-built maps for the construction-time win above. If that balance doesn't fit let-go's typical workloads, an alternative is trimming the array on `persistent!`, which keeps the build win and drops the slack at the cost of one final copy per node.

Tests: the mutation boundary is pinned directly — deterministic begin/middle/end in-place inserts (driving `assocTransient` with crafted hashes for position control), the capacity-exhaustion realloc boundary including a descending-position insert across it, and a persisted-map-with-spare-capacity mutated through a second transient without the frozen map observing anything.

Safety notes: in-place growth only ever mutates a privately-owned backing array (any node reachable from a persisted map or another transient fails the edit-token check in `ensureEditable` and gets cloned first). Nothing reads node arrays by capacity, and every slot in the extended region is written or preserved before the new length is exposed, so stale values past the old length are never observable. Collision buckets are a separate node type and don't reach this code.

### DIFF
--- a/pkg/vm/persistent_map.go
+++ b/pkg/vm/persistent_map.go
@@ -407,13 +407,33 @@ func (n *hmapBitmapNode) assocTransient(edit *atomic.Bool, shift uint, hash uint
 	*addedLeaf = true
 	nEntries := bits.OnesCount32(n.bitmap)
 	editable := n.ensureEditable(edit)
-	// Need to grow the array
-	newArray := make([]any, 2*(nEntries+1))
-	copy(newArray, editable.array[:2*idx])
-	newArray[2*idx] = key
-	newArray[2*idx+1] = val
-	copy(newArray[2*(idx+1):], editable.array[2*idx:2*nEntries])
-	editable.array = newArray
+	need := 2 * (nEntries + 1)
+	if cap(editable.array) >= need {
+		// Use the growth reserve (ensureEditable over-allocates): shift the
+		// tail right in place and insert. copy is overlap-safe, and every
+		// slot in [0, need) is written or preserved, so stale values beyond
+		// the old length are never observed.
+		arr := editable.array[:need]
+		copy(arr[2*(idx+1):], arr[2*idx:2*nEntries])
+		arr[2*idx] = key
+		arr[2*idx+1] = val
+		editable.array = arr
+	} else {
+		// Reserve exhausted — reallocate with headroom so allocations (and
+		// their full-array copies) amortize across a batch of inserts.
+		// Note the in-place branch above still shifts the tail on every
+		// insert, so slot copying stays O(tail) per insert (bounded by the
+		// 32-entry node fan-out); what this removes is the per-insert
+		// allocation and full-array copy. A bitmap node holds at most 32
+		// entries (64 slots), so cap the reserve there: no unbounded spare
+		// capacity rides along into the persisted map.
+		newArray := make([]any, need, min(2*need, 64))
+		copy(newArray, editable.array[:2*idx])
+		newArray[2*idx] = key
+		newArray[2*idx+1] = val
+		copy(newArray[2*(idx+1):], editable.array[2*idx:2*nEntries])
+		editable.array = newArray
+	}
 	editable.bitmap = n.bitmap | bit
 	return editable
 }
--- a/pkg/vm/persistent_map_transient_boundary_test.go
+++ b/pkg/vm/persistent_map_transient_boundary_test.go
@@ -1,0 +1,163 @@
+/*
+ * Copyright (c) 2026 let-go contributors
+ * SPDX-License-Identifier: MIT
+ */
+
+package vm
+
+import (
+	"fmt"
+	"sync/atomic"
+	"testing"
+)
+
+// The in-place transient insert reuses an owned node's spare capacity, so
+// its correctness hangs on two boundaries: the shift must be right for any
+// insert position (begin/middle/end), and spare capacity on a persisted
+// node must never be reused by a later transient (ensureEditable must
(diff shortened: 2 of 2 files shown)


mparrett:
A note on the repeat A/B step summary before it reads as a regression: it flags `MapAssoc/HAMT-Lookup` (+11–14% median) and `SeqIteration/List` (+8–19%) as would-gate families. Three checks say that's runner/layout sensitivity rather than an effect of this diff:

- No mechanism: the `HAMT-Lookup` fixture is built with persistent `Assoc` (bench_test.go), which never goes through `assocTransient` and carries none of the capacity slack described in the PR text, and the measured path (`ValueAt`) is untouched by this change. `SeqIteration/List` doesn't involve maps at all.
- Cross-PR control: #560's run — a disjoint diff touching only `lazy_seq.go`, with dead-flat aggregate cycles — elevates the same `SeqIteration/List` families (+7–8%), and its `HAMT-Lookup` sits at +2–4%.
- Doesn't reproduce: a local interleaved A/B on exactly these families (n=8, darwin/arm64, pinned base/head binaries) is flat — HAMT-Lookup p=0.28–0.63, List p=0.37–0.82.

My read: any real edit to `pkg/vm` shifts code layout for the whole package, and the pr-fast micro-benches are sensitive to that on the CI runner. I've noted a related calibration observation on #445.

nooga (approved):
Reviewed adversarially with a focus on the spare-capacity hazard. Approving.

What I verified:

- **Invariant preserved**: `len(array) == 2*popcount(bitmap)` holds after both new branches — the in-place path slices to exactly `need` (pkg/vm/persistent_map.go:416), the realloc path makes `len==need` with slack only in `cap` (persistent_map.go:430). I swept every consumer of `hmapBitmapNode.array`: `nodeSeq`/`each` iterate `len(n.array)/2`, `cloneAndSet`/`cloneAndSet2`/`removePair`/persistent `assoc` all size and copy by `len`, and downstream consumers (map `Seq`/`entries`, `Equals` via `each`, `Hash` via `Seq`, `String`, the bytecode encoder's `writeMapConsts` at pkg/bytecode/encoder.go:665, transit) all go through those. Nothing reads `cap` or reslices beyond `len`, so slack slots are never observed.

- **Overlap-safe shift**: `copy(arr[2*(idx+1):], arr[2*idx:2*nEntries])` (persistent_map.go:417) shifts right with overlapping dst>src; Go's `copy` has memmove semantics per the spec, so this is correct, and dst/src lengths match exactly. Every slot in `[0, need)` is written or preserved including the two slots newly exposed past the old `len`, so stale spare-region values can't leak even after a collision push-down. The new `TestAssocTransientInsertPositions` pins begin/middle/end insert positions.

- **Ownership**: in-place mutation only happens on `ensureEditable`'s result, which requires pointer equality with this transient's `&t.edit` (persistent_map.go:354, transient.go:22-29). Each transient embeds its own `atomic.Bool`, and nodes holding `edit` pointers keep them GC-live, so token addresses can't be recycled while frozen structure still references them. A later transient over a persisted map with slack always clones — pinned by `TestTransientPersistedNodeNotReused`.

- **Bounds**: a bitmap node maxes at 32 entries, and a new-entry insert into a full node is unreachable (the bit would be set), so `need <= 64` and `min(2*need, 64) >= need` always holds.

- Ran `go build ./...`, `go test ./pkg/vm/ -count=1 -race`, `go test ./test/ -count=1`, and the clojure-test-suite run — all green.

Non-blocking observations:

- The acknowledged footprint trade-off is real: `Persistent()` (transient.go:130) freezes nodes with up to `cap-len` slack per multi-insert node. If it ever matters, a trim walk at persistent! (reallocating exact-size arrays for owned nodes) would recover it at the cost of one extra copy per owned node; note that `cloneAndSet` already sheds slack on the first persistent-mode write to a node, so slack decays under later persistent updates anyway.
- `ensureEditable` only grants +2 headroom on first clone (persistent_map.go:360), so the second insert into a freshly-cloned node still takes the realloc branch before the doubling kicks in. Starting the clone at `min(2*(len+2), 64)` could save one realloc per node per transient, but that changes memory behavior for single-insert nodes, so fine as is.

PR #561, opened 2026-07-18
perf(bench-ratchet): robust sampling — warmup-discarding median + anchor-drift warning

nnunley:
## Problem

Single cold samples made the ratchet lie. The committed baseline was captured with `-count=1`, no warmup, and mean aggregation. Its anchor (1.847 ns/op) landed in the slow mode of a **bimodal distribution** the historical `docs/perf/.runs/` captures show clearly on an M3:

```
  1.0 ns: ########### (11)          <- P-core (true CPU speed)
  1.1 ns: ###...##### (42)
  1.8 ns: ################### (19)  <- E-core / throttled captures
  1.9 ns: ############### (15)
```

Since every benchmark is reported as a multiple of the anchor, a subsequent clean run reads as +38%/+31%/+17% phantom regressions while raw wall times actually improved across the board (2026-07-18; suite compile ~2.83s→2.25s, IRCompile ~21.4ms→14.4ms in wall time while the report screamed REGRESSION).

## Changes (hyperfine-style)

- **`reduceSamples`**: same-named `-count` reps reduce by discarding the FIRST rep as warmup (the suite bench visibly climbs 264→319 ms across in-process reps) and taking the **median** of the rest instead of the mean — one GC pause or contention spike no longer skews the stored baseline. Single-rep captures (legacy `.jsonl`, explicit `-count 1`) pass through unchanged.
- **`defaultCount` 1 → 4**, and the suite jobs' `count: 1` pins → 4: warmup + 3 kept samples, the minimum for a meaningful median (a median of 2 doesn't exist). The old pin rationale ("variance is negligible") is what this incident falsified; a suite pass is ~2.4 s today, so the default gate grows by roughly a minute.
- **Anchor-drift warning** in check mode when the anchor moves >15% on the same machine class — the anchor divides every ratio, so a drifted anchor scales every verdict by the same bogus factor. This would have flagged the incident on first run.

Raw per-rep samples are still retained in the baseline for forensics; per-sample ratios compute against the reduced anchor.

## Not in this PR

`docs/perf/baseline.json` still carries the polluted anchor and needs a recapture (`bench-ratchet update -force`) from a clean checkout on a quiet machine — left out so the recapture can be reviewed on its own.

## Verification

- `go test ./cmd/bench-ratchet` green (updated `TestAggregateFromFileRetainsSamples`, new `TestReduceSamplesWarmupAndMedian`, `TestSuiteJobsPinCountToOne` → `...ToFour` with rewritten rationale); `go vet` clean.
- Full `make bench-ratchet` run with the new sampling: anchor medians stably at ~1.01 ns across reps (vs single-sample scatter before).
- Note: the local pre-push gate's `go test -timeout 60s ./...` cannot fit the e2e suite (~175 s) on the capture machine — e2e verified passing directly with no timeout; gate bypassed for the push.

### DIFF
--- a/cmd/bench-ratchet/main.go
+++ b/cmd/bench-ratchet/main.go
@@ -291,6 +291,16 @@ func main() {
 		}
 	}
 
+	// The reduction needs at least 3 reps for a true median (count >= 4 gets
+	// warmup-discard + median, count 3 a plain median-of-3). Below that it
+	// degenerates to a single sample; runs that accept the tradeoff for
+	// wall-clock reasons still work, but the degradation must never be silent.
+	if effCount < 3 && mode != "show" {
+		fmt.Fprintf(os.Stderr, "bench-ratchet: WARNING: -count %d cannot form a median; "+
+			"use -count >= 3 (>= 4 adds a discarded warmup rep). Results are more outlier-sensitive.\n",
+			effCount)
+	}
+
 	if *outPath == "" {
 		*outPath = defaultRunPath()
 	}
@@ -940,12 +950,21 @@ func captureOnePackage(pkg string, count int, benchtime, timeout, tags string, f
 // is taken. Median, not mean: a single contention spike or GC pause
 // skews an average straight into the stored baseline, which is how a
 // concurrently-running pass once manufactured six phantom regressions.
-// With one rep (legacy captures, -count 1) the value passes through.
+//
+// Exactly three reps take the plain median of all three: the median
+// already rejects the systematically-slow cold rep, whereas discarding
+// it first would leave two samples and force a mean — the estimator
+// this function exists to avoid. This keeps -count 3 callers (the
+// perf-timeline ubuntu leg, whose measured budget can't fit count 4,
+// see #573) on a true median. With one rep (legacy captures, -count 1)
+// the value passes through.
 func reduceSamples(vals []float64) float64 {
 	if len(vals) == 0 {
 		return 0
 	}
-	if len(vals) > 1 {
+	// n==2 still discards the cold rep (a single warm sample beats a mean
+	// that averages the cold rep in).
+	if len(vals) > 3 || len(vals) == 2 {
 		vals = vals[1:]
 	}
 	s := append([]float64(nil), vals...)
@@ -959,8 +978,8 @@ func reduceSamples(vals []float64) float64 {
 
 // aggregateFromFile reads a .jsonl of StreamRecord lines and returns
 // a Baseline computed from them. Same-named records (multiple -count
-// repetitions, in capture order) reduce via reduceSamples: first rep
-// discarded as warmup, median of the remainder.
+// repetitions, in capture order) reduce via reduceSamples: median, with
+// the first rep discarded as warmup when enough reps allow it.
 func aggregateFromFile(path string) (MachineBaseline, error) {
 	f, err := os.Open(path)
 	if err != nil {
@@ -1705,6 +1724,17 @@ func compareAndReportMarkdown(baseline, current MachineBaseline, budget float64,
 		anchorDrift := (current.Anchor.NSPerOp - baseline.Anchor.NSPerOp) / baseline.Anchor.NSPerOp
 		fmt.Printf("- anchor: baseline `%.3f ns/op`, current `%.3f ns/op` (`%+.1f%%`)\n\n",
 			baseline.Anchor.NSPerOp, current.Anchor.NSPerOp, anchorDrift*100)
+		// The markdown reports (perf-pr.yml / perf-wasm.yml) are exactly the
+		// surface the phantom-regression incident is meant to expose, so the
+		// same >15% anchor-drift warning the text path emits must render here —
+		// otherwise a polluted capture publishes a trustworthy-looking verdict.
+		if anchorDrift < -0.15 || anchorDrift > 0.15 {
+			fmt.Printf("> **⚠️ anchor moved %+.1f%% on the same machine class — normalized.**\n", anchorDrift*100)
+			fmt.Printf("> The ratios below are scaled by this factor and are **not trustworthy** —\n")
+			fmt.Printf("> likely a polluted capture (CPU scheduling / thermal / concurrent load) on\n")
+			fmt.Printf("> one side. Re-run on a quiet machine; if the drift persists, recapture the\n")
+			fmt.Printf("> baseline (`bench-ratchet update -force`) from a clean checkout.\n\n")
+		}
 	} else {
 		fmt.Println()
 	}
--- a/cmd/bench-ratchet/main_test.go
+++ b/cmd/bench-ratchet/main_test.go
@@ -76,6 +76,7 @@ func TestReduceSamplesWarmupAndMedian(t *testing.T) {
 		{"empty", nil, 0},
 		{"single passes through", []float64{7}, 7},
 		{"two drops warmup", []float64{100, 40}, 40},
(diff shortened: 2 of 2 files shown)


cmd/bench-ratchet/main.go
```diff
@@ -1401,6 +1427,20 @@ func compareAndReport(baseline, current MachineBaseline, budget float64, format
...
 		fmt.Printf("anchor: baseline %.3f ns/op, current %.3f ns/op (%+.1f%%)\n",
 			baseline.Anchor.NSPerOp, current.Anchor.NSPerOp, anchorDrift*100)
+		// The anchor is the divisor for EVERY normalized ratio: on identical
+		// hardware+toolchain it should be stable to a few percent. A large
+		// drift means the anchor ran in a different CPU regime on one side
+		// (P-core vs E-core scheduling, thermal throttle, a concurrent
+		// load) — and every REGRESSION/ok verdict below is scaled by the
+		// same bogus factor. Observed in practice as a bimodal ~1.06 vs
+		// ~1.85 ns anchor on an M3 manufacturing phantom +38% regressions.
+		if anchorDrift < -0.15 || anchorDrift > 0.15 {
```
mparrett, line 1496: [P1] Emit the anchor-drift warning in Markdown reports

`compareAndReport` returns immediately for `format == "markdown"`, so this new block never executes for either `.github/workflows/perf-pr.yml` or `.github/workflows/perf-wasm.yml`; both invoke the tool with `-format markdown`. Those are the reports meant to expose the phantom-regression incident, but they will still publish apparently trustworthy verdicts with no warning when the anchor moves more than 15%. Please move this check into shared logic or add the same warning to `compareAndReportMarkdown`.

mparrett:
Before this and #565 both land, there's a composition worth a look: `.github/workflows/perf-timeline.yml` calls bench-ratchet with an explicit `-count 3` (plus `-benchtime 2s -timeout 30m`, under the job's `timeout-minutes: 90`). Neither PR touches the other's file, so both merge green; the interaction only shows up combined.

Two pieces:

1. The `-count 3` in the workflow. With the new `reduceSamples` that means discard the warmup rep, then average the remaining 2 — the even-length branch this PR's own comment calls out ("a median of 2 doesn't exist") — so timeline snapshots would quietly use a weaker estimator than the count-4 median-of-3 the baseline uses, and the anchor-drift warning would be comparing across two estimators. Fixed on #565 now (cefa144): the workflow passes `-count 4`, which is the new default's median-of-3 once this lands and one extra plain-mean sample until then, so it's safe in either merge order. If you also want bench-ratchet to clamp or reject counts below 4 so no caller can reach that corner, no objection here — your call.

2. The suite pins going 1 → 4, which live in bench-ratchet (and its test), so the workflow can't adjust them. That adds 18 suite executions per `-full` leg (6 variants × 3 extra reps). The last completed ubuntu leg took ~65 min against the 90-min job timeout, and each variant runs under the single `go test -timeout 30m`. I couldn't get a per-rep suite cost from CI (every timeline run in the last 40 is cancelled or queued — the jam #565 addresses), so I can't say whether the new counts fit; the 38-min M3 full capture in #564 is the only timing for the new profile, and CI runners are slower. A leg that exceeds the go-test timeout fails its bench invocation; one that exceeds `timeout-minutes` is killed. Either way the timeline stops recording snapshots — the state #565 exists to get us out of.

To get a real number I've started a timing run on a fork branch cut from this PR's head: one ubuntu `-full` leg at the new counts, with publish disabled and both timeouts raised so it measures the true duration instead of getting killed (https://github.com/mparrett/let-go/actions/runs/29653784883). I'll post the duration here when it finishes. If it fits under the stock timeouts with headroom, nothing else needed. If it's tight, my one position is that the fix belongs in bench-ratchet (smaller suite reps for the timeline profile, a flag — whatever shape you prefer) rather than in raising the workflow timeouts, since longer legs spend more of the macOS runner pool whose starvation caused the jam in the first place.

mparrett:
Probe result, and it corrects my earlier comment in two ways.

One ubuntu `-full` leg at the new counts: **103.9 min** end to end, 101.5 min in the capture step — past the stock `timeout-minutes: 90`. Run (publish disabled, timeouts raised so it could finish instead of getting killed): https://github.com/mparrett/let-go/actions/runs/29653784883

The corrections:

1. The 1 → 4 suite pins I centered are fine. All six suite variants at count 4 finish in under a minute each, ~4 min combined. I withdraw that concern.

2. The blowout is `-count` scaling on the main `pkg/vm` bench invocation: ~97 min at count 4, versus a ~65-min whole leg at count 3. Which means the workflow-side "fix" I pushed (`-count 4`, cefa144) is what breaks the budget — and it rode #565's merge a few minutes before the probe finished, so current main is in the dies-at-90-min state. #573 reverts the workflow to `-count 3`, the measured-fit envelope.

That leaves the estimator gap this PR set out to close: at count 3, `reduceSamples` lands in the discard-warmup-then-average-of-2 branch. One small option on this side would square it: for n=3, take the plain median of all 3 (discard-then-median only from n≥4). The median already rejects the cold first rep, the timeline keeps its ~65-min legs, and no caller lands on an average of 2. If you'd rather standardize on count 4, the ubuntu leg needs a bigger `timeout-minutes` (~120), and I'd want a macos-14 leg measured before trusting the scarce pool with it — your M3 took 38 min, and the runner M1s are slower.

nnunley:
Went with your median-of-3 option — it's the right shape: `406ab13` makes `reduceSamples` take the plain median of all three at exactly n=3 (the median already rejects the cold rep; discard-then-reduce would leave the mean-of-2 this PR exists to kill). n≥4 keeps warmup-discard + median-of-rest, n=2 still discards the cold rep. So the timeline's `-count 3` legs stay inside the measured 90-min envelope *and* get a true median — no timeout changes, no macos-pool exposure.

The count guard adjusted to match: stderr warning only below `-count 3` (where no median is formable), so no caller can land on a degenerate estimator silently, per your "clamp or reject" suggestion — warning rather than reject, since count-1 legacy captures are still legitimate.

Thanks for burning a fork runner on the probe — the count-scaling-on-pkg/vm attribution (vs the suite pins) is exactly what we'd have misdiagnosed without it. One estimator caveat now recorded here for posterity: count-3 timeline snapshots (median-of-3-with-cold-rep) and count-4 baselines (median-of-3-post-warmup) are close but not identical estimators; the anchor-drift warning compares across them. If that ever matters in practice, stamping the count into snapshot metadata is the durable fix.

mparrett:
Quick check on this P1: at `406ab13`, the Markdown path still returns before the anchor-drift warning runs, and both perf workflows use `-format markdown`. So CI still won’t display the warning this PR is intended to add. 

Could we add the same warning to `compareAndReportMarkdown`—ideally through shared logic with a small regression test—before merging? The new median-of-three changes otherwise look good.

mparrett (changes requested):

cmd/bench-ratchet/main.go
nnunley: Fixed — added the same >15% anchor-drift warning to `compareAndReportMarkdown` (rendered as a GitHub blockquote right under the anchor line), so the markdown reports the perf workflows publish now carry the warning. Rebased onto current `main` too (conflict-free).

mparrett (approved):
Verified latest head `147c033`:

- P1 fixed in Markdown output and preserved by both workflows’ `sed` extraction.
- All GitHub checks green; PR is mergeable.
- Targeted tests, `go vet`, and diff checks pass.
- No remaining high-priority findings.

LGTM!

PR #563, opened 2026-07-18
perf(vm): gate lookup-stats noters on an atomic before the mutex

mparrett:
`noteLookup` and `noteNamespaced` take the global `lookupStatsMu` on every `Namespace.Lookup` and `Symbol.Namespaced` call just to discover that stats are disabled — so the production configuration pays an always-taken global mutex, and a cross-goroutine serialization point, for a feature that is off.

The change: make the enabled flag an `atomic.Bool` checked before locking. The noters return on a single atomic load when stats are off; when they're on, the flag is re-checked under the lock, so the enabled path records exactly as before (a concurrent disable between the fast check and `Lock` still drops the sample, same as today). The mutex still guards the stats maps.

One behavioral note: `SetLookupStatsEnabled(false)` is no longer a synchronous barrier — a noter already past its fast-path check can record one final sample after the call returns. The previous always-locked version quiesced synchronously; neither current caller (boot-time enable, exit-time snapshot) depends on that.

Measured (interleaved A/B against main, n=8, Apple M2, new benchmarks included in the PR):

- `Namespace.Lookup`, stats disabled: 24.4 ns → 16.4 ns (−33%, p=0.000)
- same, 8-way parallel: 132 ns → 92 ns per op (−30%, p=0.001) — the serialization on the stats mutex is gone; the remainder is the namespace's own read locks
- `Namespace.Lookup`, stats enabled: unchanged (p=0.88)

### DIFF
--- a/pkg/vm/lookup_stats.go
+++ b/pkg/vm/lookup_stats.go
@@ -5,6 +5,7 @@ import (
 	"slices"
 	"strings"
 	"sync"
+	"sync/atomic"
 )
 
 type LookupStats struct {
@@ -50,8 +51,13 @@ func writeLookupCounts(b *strings.Builder, prefix string, m map[string]uint64) {
 }
 
 var (
-	lookupStatsMu      sync.Mutex
-	lookupStatsEnabled bool
+	lookupStatsMu sync.Mutex
+	// lookupStatsEnabled is atomic so the noters can bail without taking
+	// lookupStatsMu — they sit on Namespace.Lookup and Symbol.Namespaced,
+	// where an always-taken global mutex is a cross-goroutine serialization
+	// point paid even with stats off (the production case). The mutex still
+	// guards the stats maps themselves.
+	lookupStatsEnabled atomic.Bool
 	lookupStatsGlobal  = LookupStats{
 		NamespacedBySym: map[string]uint64{},
 		LookupByNS:      map[string]uint64{},
@@ -60,9 +66,7 @@ var (
 )
 
 func SetLookupStatsEnabled(enabled bool) {
-	lookupStatsMu.Lock()
-	lookupStatsEnabled = enabled
-	lookupStatsMu.Unlock()
+	lookupStatsEnabled.Store(enabled)
 }
 
 func ResetLookupStats() {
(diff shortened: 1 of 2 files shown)


nooga (approved):
Reviewed for correctness, focusing on the lock protocol and the acknowledged quiescence change. Approving.

**Lock protocol** (pkg/vm/lookup_stats.go): the invariant that matters — the stats maps are only ever touched while holding lookupStatsMu — still holds everywhere. noteNamespaced (line 104) and noteLookup (line 122) both re-check lookupStatsEnabled.Load() after acquiring the mutex, so a noter that passed the fast-path gate before a disable cannot mutate the maps unless it also observes enabled under the lock. SnapshotLookupStats (line 82) and ResetLookupStats (line 72) take the same mutex, so a straggler cannot tear or corrupt a snapshot: a snapshot issued after SetLookupStatsEnabled(false) blocks until any in-flight noter releases the lock, and at worst includes that one final sample. That is exactly the behavior change the PR declares, and it is bounded to one sample per in-flight noter.

**Caller audit**: SetLookupStatsEnabled is called from exactly one production site — pkg/compiler/eval.go:98-100, enable-at-boot under LG_LOOKUP_STATS — and from the new bench file. No production code ever calls SetLookupStatsEnabled(false), so nothing depends on the old synchronous-quiesce semantics. The only snapshot consumer is emitRuntimeStats at lg.go:381-385, which reads under the mutex at exit; a late sample there is at most a count off-by-one in a debug histogram, not a correctness issue.

**Verification**: go build ./... clean; go test ./pkg/vm/ -count=1 -race passes (37s). Since the PR adds only benchmarks, I also wrote a throwaway stress test — 8 goroutines hammering noteLookup/noteNamespaced against a loop of SetLookupStatsEnabled toggles, SnapshotLookupStats, and ResetLookupStats — and it is race-detector clean. I did not run the benchmarks per repo policy on parallel runs, so I am taking the -33%/-30% numbers on faith; the mechanism (skipping a global mutex on the Namespace.Lookup hot path) makes them plausible.

**Non-blocking observations**: (1) In BenchmarkNamespaceLookupStatsEnabled the two defers run in LIFO order, so ResetLookupStats executes while stats are still enabled — harmless single-goroutine, just slightly odd ordering. (2) The comment at lookup_stats.go:109-114 is accurate but worth keeping in sync if a disable-then-assert caller ever appears; a doc comment on SetLookupStatsEnabled itself stating the non-barrier semantics would put the contract where future callers will read it.

PR #567, opened 2026-07-18
perf(vm): typed Keyword/Int fast paths at the top of valueEquiv

mparrett:
Every map-key comparison through `valueEquiv` paid a double ~20-case `isComparable` type-switch before reaching a verdict, and an unequal pair of keywords additionally murmur-hashed both strings just to produce the negative. Front-load the two dominant key types with exact reductions of the general path: a keyword only ever equals a keyword with the same string, and `NumEq` on two Ints is `==`. A non-Int second operand still falls through, so `Int`×`Float`/`BigInt` equivalence stays with the numeric path. This is the same pattern the keyword-keyed map lookup already proves in `findKeyword`.

Measured (interleaved A/B against main, n=8, Apple M2, benchmarks included in the PR):

- `valueEquiv` Keyword-equal −50%, Keyword-unequal −91% (30 ns → 2.6 ns — the negative previously hashed both keywords)
- Int-equal −64%, Int-unequal −75%
- map-level `MapAssoc/HAMT-Lookup` (Int keys): −14–17% across 10/100/1000 entries (p ≤ 0.01)

One measured cost: the mixed `Int`×`Float` pair now pays the two failed asserts before reaching `NumEq` — +1.2 ns (+13.7%) on that path. The bench suite pins that pair's semantics (Ints and Floats are never map-key equivalent, matching Clojure `=`) so a future fast path can't silently short-circuit it.

The same tradeoff applies more mildly to other key types: a comparison whose first operand is neither a keyword nor an Int now pays two failed type assertions before the previous logic — a small constant on the collection-key path in exchange for the keyword/int win.

### DIFF
--- a/pkg/vm/hash.go
+++ b/pkg/vm/hash.go
@@ -141,6 +141,23 @@ func mixFinishLen(h uint32, length uint32) uint32 {
 // valueEquiv tests if two Values are equivalent for map key purposes.
 // Uses hash as a fast negative check, then structural comparison.
 func valueEquiv(a, b Value) bool {
+	// Typed fast paths for the two dominant map-key types, ahead of the
+	// double isComparable type-switch below. Both are exact reductions of
+	// the general path: a keyword only ever equals a keyword with the same
+	// string (the == identity check plus the hash-negative and Equals
+	// fallbacks all collapse to that), and NumEq on two Ints is ==. A
+	// non-Int b deliberately falls through — Int×Float/BigInt equivalence
+	// belongs to the NumEq path. Neither type can be NIL or EmptyList, so
+	// hoisting past that check is safe.
+	if ak, ok := a.(Keyword); ok {
+		bk, ok := b.(Keyword)
+		return ok && ak == bk
+	}
+	if ai, ok := a.(Int); ok {
+		if bi, ok := b.(Int); ok {
+			return ai == bi
+		}
+	}
 	if (a == NIL && b == EmptyList) || (a == EmptyList && b == NIL) {
 		return true
 	}
--- a/pkg/vm/value_equiv_bench_test.go
+++ b/pkg/vm/value_equiv_bench_test.go
@@ -1,0 +1,58 @@
+/*
+ * Copyright (c) 2026 let-go contributors
+ * SPDX-License-Identifier: MIT
+ */
+
+package vm
+
+import "testing"
+
+// Direct benchmarks for valueEquiv's dominant map-key type pairs. The
+// map-level benchmarks (MapAssoc/*) exercise these through the trie; these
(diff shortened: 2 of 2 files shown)


nooga (approved):
Reviewed adversarially for map-key equality semantics, since valueEquiv defines them. Approving — both fast paths are exact reductions of the old path.

**Keyword path** (pkg/vm/hash.go:152-155): returning `ok && ak == bk` immediately (false for any non-Keyword b) matches the old behavior exactly. Keyword has no `Equals` method, is not Sequable/Seq, and `IsNumber` is false, so the old path for a Keyword lhs could only succeed via the interface identity check — which requires identical dynamic type. Checked the specific mixed pairs: Keyword×Symbol (same name) was already false via the hash-negative check (Keyword.Hash adds 0x9e3779b9 to the string hash, keyword.go:38 vs symbol.go:39; and even on collision there is no Equals to fall to), Keyword×String and Keyword×NIL likewise ended at `return false`.

**Int path** (pkg/vm/hash.go:156-160): Int×Int → `==` matches NumEq's Int/Int arm (numbers.go:1038-1040) and the old identity fast path. Critically, a non-Int rhs falls through with no early return, so Int×BigInt stays true via NumEq (numbers.go:1041-1042) and Int×Float stays false (NumEq's Int arm has no Float case) — Clojure's `(= 1 1N)` / `(not (= 1 1.0))` semantics preserved. The latter is pinned at the language level by test/numeric_tower_test.lg:56.

**Hoisting past the NIL/EmptyList check** is safe as the comment claims: Keyword (string kind) and Int (int kind) can never compare equal to the `*Nil`/EmptyList pointers.

**Callers**: audited all valueEquiv call sites — persistent_map.go (find/assoc/dissoc/collision nodes), persistent_queue.go:169, record.go:172, constpool.go:264. All are pure equality checks; none depend on the old evaluation order or the hash-negative side path. The `=` builtin uses rt's separate `equals`, unaffected.

**Verified**: `go build ./...`, `go test ./pkg/vm/ -count=1`, `go test ./test/ -count=1`, and `go test ./test/ -run TestClojureTestSuite` all pass on the branch. Also executed the new benchmark assertions once (-benchtime=1x): all five sub-benches pass, including the Int-vs-Float fall-through guard.

Non-blocking observation: the new assertions in pkg/vm/value_equiv_bench_test.go live inside `Benchmark*` functions, so a plain `go test ./pkg/vm/` never executes them — the Int×Float and keyword-mismatch pins only run under -bench (the .lg suite does cover Int×Float). A tiny `TestValueEquivMixedTypes` unit test would pin these in the default test run.

nnunley (approved):

PR #568, opened 2026-07-18
perf(vm): box-free keyword lookup in protocol method dispatch

mparrett:
`Protocol.Lookup` fetched the method fn via the generic `ValueAt`, which boxes the method keyword into a `Value` (a string-kind interface conversion — one heap allocation) on every protocol-method call, twice when the `AnyType` default was consulted, and compares through `valueEquiv`. Route the lookup through the existing `ValueAtKeywordOr` instead: bare `Keyword` in, `findKeyword`'s typed compare, no allocation.

Dispatch semantics are unchanged: impl-map keys are keywords by construction (`Extend`'s contract), and a non-keyword key never matched under the boxed path either. The nil-target branch now shares `lookupIn` instead of duplicating its logic inline.

Measured (interleaved A/B against main, n=8, Apple M2, benchmarks included in the PR):

- full protocol-method call on an extended type: 65 ns → 38 ns (−41%), and 16 B/op down to zero
- lookup through the `AnyType` fallback (two maps consulted): 76 ns → 35 ns (−54%), and 32 B/op down to zero
- the dispatch path no longer allocates in either shape

Remaining per-call cost is the method-name hash (`Keyword.Hash` recomputes the murmur mix per lookup); caching or interning that is a representation question this change leaves alone.

### DIFF
--- a/pkg/vm/protocol.go
+++ b/pkg/vm/protocol.go
@@ -80,15 +80,7 @@ func (p *Protocol) Lookup(methodName Symbol, target Value) (Fn, bool) {
 	key := Keyword(methodName)
 
 	if target == NIL {
-		if p.nilImpl != nil {
-			v := p.nilImpl.ValueAt(key)
-			if v != NIL {
-				if fn, ok := v.(Fn); ok {
-					return fn, true
-				}
-			}
-		}
-		return nil, false
+		return p.lookupIn(p.nilImpl, key)
 	}
 
 	vt := target.Type()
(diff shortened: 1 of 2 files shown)


nooga (approved):
Reviewed for semantic equivalence between the old boxed lookup and the new typed path; approving.

What I verified:

- **Match semantics are identical.** Old path: `ValueAt` → `hashValue(key)` → `Keyword.Hash()` (Keyword implements Hashable, pkg/vm/hash.go:23) → trie `find` with `valueEquiv`. New path: `ValueAtKeywordOr` (pkg/vm/persistent_map.go:808) → `k.Hash()` directly → `findKeyword` with typed `==`. Same hash, so the same trie traversal; and for a Keyword probe key, `valueEquiv` can only ever match an identical stored Keyword (the `a == b` comparable fast path; the numeric, `Equals`, and Seq-fallback branches don't apply to Keywords — pkg/vm/hash.go:143-190), which is exactly what the typed compare accepts. This holds even in the collision-node scan (persistent_map.go:483) and even if a non-keyword key were somehow present — it didn't match before and doesn't now.
- **Keys are keywords by construction.** Every impl-map producer emits Keyword keys: the `extend-type`/`extend-protocol`/`reify` macros all build pairs via `(keyword mname)` (pkg/rt/core/core.lg:2365, 2408, 2454; deftype/defrecord expand to `extend-type`), and the Go-side `protoImplMap` takes a `vm.Keyword` (pkg/rt/ions.go:322). `extend-type*` accepts an arbitrary map, but per the previous point that's not a behavior change.
- **Missing-key and default ordering unchanged.** Both old and new return NIL on miss, which `lookupIn` treats as not-found; the exact-type → AnyType fallback ordering in `Lookup` (pkg/vm/protocol.go:86-98) is untouched.
- **nil-target refactor is behavior-preserving.** The old inline branch (nil map → false, NIL value → false, non-Fn value → false, Fn → true) maps one-to-one onto `lookupIn` (protocol.go:108-118), and the no-impl error message in `invokeIn` (protocol.go:186-192) still fires identically since `Lookup`'s (nil, false) contract is unchanged.
- **Tests.** `go build ./...`, `go test ./pkg/vm/ -count=1`, and `go test ./test/ -count=1` all pass. The pre-existing `keyword_lookup_test.go` parity tests (ValueAtKeywordOr vs ValueAtOr, Keyword.Hash vs hashValue) directly pin the equivalence this change relies on.

Non-blocking nit: the comment on `BenchmarkProtocolLookupFallback` says the fallback path 'consulted two impl maps per call', but in that benchmark the exact-type map is nil so `lookupIn` early-returns before any map fetch — under the old code the second boxing still happened at the `lookupIn` call site, so the alloc claim stands, but 'two impl maps' slightly oversells what the bench exercises. Not worth blocking on.

PR #569, opened 2026-07-18
perf(vm): strconv.Itoa and plain concat for Int/Keyword String()

mparrett:
`Int.String` went through `fmt.Sprintf("%d")` and `Keyword.String` through `fmt.Sprintf(":%s")` — reflection-based formatting plus an interface-box of the operand on every call, on methods that sit under `str`/`pr-str` and any render path that stringifies values. `strconv.Itoa` and a two-string concat produce identical output.

Measured (interleaved A/B against main, n=8, Apple M2, benchmarks included in the PR): `Int.String` 42.9 ns → 16.4 ns (−62%), `Keyword.String` 35.3 ns → 11.5 ns (−67%). The benchmark also shows the keyword path dropping to zero allocations, though part of that is the non-escaping benchmark shape — in escaping positions the concat still allocates its result, just without fmt's extra operand box.

### DIFF
--- a/pkg/vm/int.go
+++ b/pkg/vm/int.go
@@ -6,8 +6,8 @@
 package vm
 
 import (
-	"fmt"
 	"reflect"
+	"strconv"
 )
 
 type theIntType struct {
@@ -64,5 +64,5 @@ func (l Int) Unbox() any {
 }
 
 func (l Int) String() string {
-	return fmt.Sprintf("%d", int(l))
+	return strconv.Itoa(int(l))
 }
--- a/pkg/vm/keyword.go
+++ b/pkg/vm/keyword.go
@@ -46,7 +46,7 @@ func (l Keyword) Unbox() any {
 }
 
 func (l Keyword) String() string {
-	return fmt.Sprintf(":%s", string(l))
+	return ":" + string(l)
 }
 
 func (l Keyword) Arity() int {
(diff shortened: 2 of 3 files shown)


nooga (approved):
Verified output identity and ran the suites — this is a sound, mechanical swap.

**Output identity.** `Int` is `type Int int` and `Keyword` is `type Keyword string`, so both replacements are exact: `strconv.Itoa(int(l))` matches `fmt.Sprintf("%d", int(l))` and `":" + string(l)` matches `fmt.Sprintf(":%s", string(l))` byte-for-byte. I property-tested this against 1M random ints plus edge cases (0, ±1, MaxInt, MinInt) and keyword edge cases (empty string, `ns/name`, unicode, embedded quotes/NUL) — all identical. Namespaced keywords need no special handling since the namespace is just part of the underlying string (pkg/vm/keyword.go:48).

**No collateral.** keyword.go correctly retains the `fmt` import (still used by `Box` and `Invoke`); int.go correctly drops it. `go build ./...`, `go test ./pkg/vm/ -count=1`, and the full `.lg` language suite (`go test ./test/ -count=1`) all pass, so every consumer of these `String()` outputs (str, pr-str, printing, error messages) is unaffected. The new bench file (pkg/vm/scalar_string_bench_test.go) compiles and its `len(...) == 0` guard keeps the result live.

**Non-blocking.** I didn't re-run the benchmarks, but the claimed −62%/−67% is plausible for dropping fmt's reflection path, and the description is appropriately honest that the keyword zero-alloc figure is partly the non-escaping bench shape. The remaining `fmt.Sprintf` `String()` methods in pkg/vm (Protocol, Scope, TransientMap) are cold debug-print paths and fine to leave as-is.

PR #570, opened 2026-07-18
fix(bench-ratchet): report out-of-scope baseline entries as one line, not MISSING rows (stacks on #561)

nnunley:
Extracted from the `catalog-dispatch-facets` stack — it's a standalone reporting fix that was riding along with unrelated IR work, and it's newly relevant: every ratchet report during the 2026-07-18 baseline investigation was flooded with **133 MISSING rows** (baseline entries outside the current run's scope), burying the handful of real verdict lines. This collapses them to a single summary line.

**Stacked on #561** (rebased onto it cleanly, same two files, disjoint regions — aggregation vs reporting; `go test ./cmd/bench-ratchet` green on the combined tree). This branch contains #561's commit; only the top commit is under review. Land #561 first and this becomes a one-commit diff.

Together with #561 (sampling) and #564 (baseline recapture), this completes the ratchet-trustworthiness set: clean samples, clean priors, readable reports.

### DIFF
--- a/cmd/bench-ratchet/main.go
+++ b/cmd/bench-ratchet/main.go
@@ -79,8 +79,10 @@ const (
 	// gate. These metrics are deterministic (no CPU-dependent noise), so the
 	// budget is tight — it only absorbs rare boundary effects (e.g. a map
 	// resize landing one element differently).
-	allocBudget      = 0.02
-	defaultCount     = 1
+	allocBudget = 0.02
+	// 4 reps minimum: the first is discarded as warmup by reduceSamples,
+	// leaving >=3 for a meaningful median (a median of 2 doesn't exist).
+	defaultCount     = 4
 	defaultBenchtime = "1s"
 	defaultTimeout   = "10m"
 	defaultTags      = "gogen_ir"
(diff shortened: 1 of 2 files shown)


cmd/bench-ratchet/main.go
```diff
@@ -1383,9 +1412,46 @@ func printBaseline(b MachineBaseline) {
...
+			if !ok {
+				continue
+			}
+			rest = r2
+		} else if strings.Contains(rest, " [") {
+			continue
+		}
+		// go test -bench matches per slash-segment and the job filters are
+		// family-anchored, so match the family segment only.
+		family := rest
```
mparrett, line 1441: [P2] Preserve slash-specific `-filter` semantics

This truncates the benchmark name to its family before applying the original regexp. For a supported manual filter such as `-filter "^BenchmarkMapAssoc/HAMT-Assoc/100$"`, `go test` does run that exact sub-benchmark (it applies slash-separated regexp segments), but `jobsSelect` tests the whole regexp only against `BenchmarkMapAssoc` and returns false. If that targeted sub-benchmark disappears, the report will classify its baseline row as out of scope and suppress the `MISSING` signal this change is meant to preserve. Suggest reproducing Go test’s slash-segment matching here, or derive scope from an equivalent matcher.

mparrett (approved):
One comment to follow up on. Let me know if it makes sense. Thanks!

PR #571, opened 2026-07-18
perf(vm): divide-free fast path for checkedMulInt overflow guard

mparrett:
Every non-zero `Int`×`Int` multiply paid one or two signed divisions as its overflow check — roughly a 20–40-cycle guard on a 3-cycle multiply, sitting on the `OP_MUL` dispatch fast path. The change adds a pre-check: when both operands fit in half the platform int width, the product cannot overflow, so the divide-based checks run only for wide operands. The bound derives from the int width (the int32 range on 64-bit builds, int16 on 32-bit) rather than being hard-coded, so 32-bit-int targets stay correct.

Measured (interleaved A/B against main, n=8, Apple M2, benchmarks in the PR; ambient load widened the intervals, paired deltas significant): common in-range operands 8.3 ns → 3.6 ns (−57%, p=0.000); wide operands 7.2 ns → 10.3 ns (+44%, p=0.021) from the four failed guard compares. Wide checked multiplies are the rare shape: products between 2^31 and 2^63 that don't overflow, and overflow-prone hash-mixing code uses `unchecked-multiply`, which never enters this function.

Correctness is pinned by an exhaustive boundary test against a `math/big` oracle: the guard edges and neighbors, the int extremes, and the historical `minInt`×−1 case, all operand pairs cross-checked for both the overflow verdict and the product.

### DIFF
--- a/pkg/vm/checked_mul_test.go
+++ b/pkg/vm/checked_mul_test.go
@@ -1,0 +1,62 @@
+/*
+ * Copyright (c) 2026 let-go contributors
+ * SPDX-License-Identifier: MIT
+ */
+
+package vm
+
+import (
+	"math/big"
+	"testing"
+)
+
+// TestCheckedMulInt checks checkedMulInt against a math/big oracle over a
+// grid of adversarial operands: the fast-path guard boundary (mulGuardMin/
+// Max and neighbors), the int extremes, and the historical minInt×-1 edge.
+func TestCheckedMulInt(t *testing.T) {
+	vals := []Int{
+		0, 1, -1, 2, -2, 3, 255, 256, -255, -256,
+		mulGuardMax, mulGuardMax - 1, mulGuardMax + 1,
+		mulGuardMin, mulGuardMin + 1, mulGuardMin - 1,
+		maxIntValue, maxIntValue - 1, minIntValue, minIntValue + 1,
+		maxIntValue / 2, minIntValue / 2, mulGuardMax * 2, mulGuardMin * 2,
+	}
+	for _, a := range vals {
+		for _, b := range vals {
+			got, ok := checkedMulInt(a, b)
+			ref := new(big.Int).Mul(big.NewInt(int64(a)), big.NewInt(int64(b)))
+			fits := ref.IsInt64() && Int(ref.Int64()) >= minIntValue && Int(ref.Int64()) <= maxIntValue
+			if fits != ok {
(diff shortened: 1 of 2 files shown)


nooga (approved):
Reviewed the guard math, edge cases, and tests. Approving.

**Bound derivation** (pkg/vm/numbers.go:16-18): `intBits = 32 << (^uint(0) >> 63)` matches the width of `Int` (`type Int int`, pkg/vm/int.go:53), so the guard is derived from the same width as the value it protects. On 64-bit the guard is [−2^31, 2^31−1]; worst cases are (−2^31)·(−2^31) = 2^62 < 2^63−1 and (−2^31)·(2^31−1) = −(2^62−2^31) > −2^63, so no product of two in-guard operands can overflow, including the asymmetric-min corner. On 32-bit int the guard becomes [−2^15, 2^15−1] with worst case 2^30 < 2^31−1. I cross-compiled and vetted `pkg/vm` with `CGO_ENABLED=0 GOARCH=386` to confirm the constant expressions are valid at that width.

**minInt edge cases**: `minIntValue` is far outside the guard and the fast path requires *both* operands inside it, so minInt×−1 (and −1×minInt) still reach the explicit check at numbers.go:45 unchanged. The slow path below the new guard is byte-for-byte the previous code, and its truncated-division inequalities remain correct for all four sign quadrants (Go's truncation toward zero acts as ceil for the negative quotients, which is the safe direction in each branch).

**Test** (pkg/vm/checked_mul_test.go): the oracle grid includes mulGuardMax/Min and both neighbors on each side, int extremes and their neighbors, zero, ±1, and wide values (mulGuardMax·2), giving 576 pairs that cross the guard boundary in both operands and both signs. It asserts both the overflow verdict against math/big and the product value when ok, and the `IsInt64() && ...` short-circuit plus the min/maxIntValue range check keep the oracle correct on 32-bit builds too. The minInt×−1 pair is covered via −1 × minIntValue.

**Call sites**: only NumMul (numbers.go:181) and OP_MUL (vm.go:1182) call checkedMulInt; both consume just (product, ok), and the fast path returns the identical result for every input it captures (including 0×anything-in-guard), so no observable semantic change.

**Ran**: `go build ./...` and `go test ./pkg/vm/ -count=1` — pass. I did not re-run benchmarks; the claimed numbers are plausible given the divide-vs-compare tradeoff and the fast path is the common case.

Non-blocking nit: the two range checks per operand could be folded into one unsigned comparison each (`uint(a)+uint(-mulGuardMin) <= uint(mulGuardMax-mulGuardMin)`), but the compiler may already do this and it costs readability.

PR #572, opened 2026-07-18
feat(rt): surface host capability — RGBA frame sink (graphics, #255)

mparrett:
The graphics half of #255 (Model B): a thin `surface` host capability that hands the host a raw RGBA frame, the binary dual of the `*emit*` seam. Graphics needs a buffer carrier — an RGBA frame is too big to escape-encode through the text (`*out*`) or JSON (`*emit*`) seams.

## The capability

```clojure
(surface/present rgba width height)  ; blit an RGBA byte-array (w*h*4, top-left origin)
(surface/available?)                 ; is a real sink wired? -> pick canvas vs another path
```

Fire-and-forget: a frame with no sink attached is dropped. Present-on-call, no retained scene — the rich layer (dirty rects, double-buffering) stays above the seam in guest code, per #255 §5.6.

## Binding discipline

Follows the interface-at-a-dynamic-var pattern the merged I/O seams (`*out*`/`*emit*`/`*keys*`/`*storage*`) use — the three bindings #255 names:

- **native / embedder** — `api.WithSurface(rt.Surface)` binds a Go sink (frame capture in tests, or a native terminal-image / fb0 encoder). Defaults to `nopSurface` (present dropped, `available?` false).
- **WASM** — `HostSurface` forwards to the JS host via `_lgSurface(uint8array, w, h)`: one `CopyBytesToJS`, then a transferable `postMessage` (worker→main, zero-copy). Bound at the `*surface*` root in `rendermain.go`, like `HostEmitter` / `HostKeySource`.

`present` / `available?` are `CtxNativeFn`s dispatching through `*surface*`, so `(binding [*surface* …])` and `api.WithSurface` both work, and the zero value is a silent no-op on every host.

`available?` reflects real sink attachment: `HostSurface.Available()` reads `_lgSurfaceReady`, which the bundle sets when the shell wires `onSurface` (worker mode relays it), not merely "running in a browser bundle."

## Tested

`pkg/api/surface_test.go` covers `WithSurface` frame capture, honest `available?`, and the no-sink no-op path (present validates args, then drops). Native + wasm builds clean.

## Notes

The wire format is RGBA8 / top-left — identical to `ImageData` and to PNG, so a native terminal-image path (PNG → iTerm2 IIP, kitty, sixel) consumes the same bytes untransformed. Two commits: the initial capability, then a reshape onto the var pattern above; the design rationale is in the #255 comment thread, where the audio seam sketch converges on the same interface shape.

Refs #255.


pkg/rt/surface.go
```diff
@@ -0,0 +1,125 @@
...
+		if !ok {
+			return vm.NIL, fmt.Errorf("surface/present expects a byte-array for rgba")
+		}
+		w, ok1 := vs[1].(vm.Int)
+		h, ok2 := vs[2].(vm.Int)
+		if !ok1 || !ok2 {
+			return vm.NIL, fmt.Errorf("surface/present width and height must be Int")
+		}
+		if len(data) < int(w)*int(h)*4 {
+			return vm.NIL, fmt.Errorf("surface/present rgba too small: have %d, need %d", len(data), int(w)*int(h)*4)
```
nnunley: `int(w)*int(h)*4` can overflow (e.g. w=h=2^31 → product wraps ≤ 0), and negative w/h also pass since their product is positive — either way the size check is bypassed and bogus dimensions reach the sink. Suggest validating `0 < w, h <= maxDim` (say 16384) before the multiply. This compounds with the wasm-side comment below: the JS exception a bogus ImageData raises becomes a Go panic.

pkg/rt/surface_js_wasm.go
```diff
@@ -0,0 +1,52 @@
...
+	if fn.IsUndefined() {
+		return
+	}
+	// Copy the Go RGBA bytes into a JS Uint8Array. js.CopyBytesToJS is the one
+	// copy on this path; _lgSurface then transfers the underlying ArrayBuffer to
+	// the main thread without a second copy. Still vastly cheaper than encoding a
+	// PNG: this is the throughput reason graphics needs a buffer seam (#255).
+	arr := js.Global().Get("Uint8Array").New(len(rgba))
+	js.CopyBytesToJS(arr, rgba)
+	fn.Invoke(arr, w, h)
```
nnunley, line 43: `js.CopyBytesToJS` and `fn.Invoke` propagate JS exceptions as Go panics — a throwing `_lgSurface` callback (or the bad-dimension case above) takes down the whole guest runtime. Worth a `defer recover` in `Present` that drops the frame instead of crashing; graphics frames are the one place where silent drop is the correct failure mode.

nnunley (commented):
Ran a multi-agent review sweep over the open PRs tonight (verified findings only). Design-wise this lands exactly on the dynamic-var seam pattern I'd hoped for — `PresentVia`/`binding` composition and the "sink attached, not just browser" fix for `available?` are both right. Two verified robustness issues inline; the rest (try/catch around the main-thread relay in lg-host-core.js, postMessage failure handling in worker mode) are in the same vein: a misbehaving guest or detached canvas should drop frames, never kill the runtime.

nnunley:
CI triage: both failing jobs (build, race) are the same single failure — `TestAssembleHTMLGolden` in `pkg/rt/wasm` (golden 25170 bytes vs 27414 got), i.e. the shell's new surface wiring changed `AssembleHTML` output without refreshing the golden. `go test ./pkg/rt/wasm -update` and commit the refreshed golden alongside the review items above; everything else in the run passes.

pkg/rt/surface.go
mparrett: Fixed in 124b3ab. Dimensions are bounded to 1..16384 before the multiply (16384²·4 = 2³⁰, so `w*h*4` can't overflow) and zero/negative reject; the length check is now exact `w*h*4` per ImageData semantics. The rgba arg also requires a real byte-array — a String used to coerce through `asBytes` against the documented contract. A table test covers string/int rgba, zero/negative/capped/overflow-scale dims, and under/oversized buffers, asserting none of them reach a live sink.

pkg/rt/surface_js_wasm.go
mparrett: Fixed in 124b3ab. The recover landed one level up, in `PresentVia`, so a faulty native embedder `Surface` gets the same drop-the-frame guarantee as a throwing `_lgSurface` callback (the wasm path routes through it). Also guarded the three JS seams (worker sender, worker→main relay, main-thread sender) with try/catch matching the emit relay — the relay runs on a main-thread event that Go recovery in the worker can't reach. A panicking-sink test pins the contract: the frame drops with no error, and later presents and ordinary evaluation continue.

mparrett:
Both review threads are addressed in 124b3ab (threads resolved, review re-requested). Validation: `surface/present` requires a real byte-array (String no longer coerces), bounds dimensions to 1..16384 before the multiply so `w*h*4` can't overflow, and demands exactly `w*h*4` bytes. Failure containment: `PresentVia` recovers around the sink call, and the three JS seams (worker sender, worker→main relay, main-thread sender) get try/catch matching the emit relay. New tests: a 9-case invalid-frame table asserting nothing reaches a live sink, and a panicking-sink test proving frames drop and execution continues. gofmt/vet, the `pkg/{rt,api,rt/wasm}` suites, and linux/wasm/plan9 cross-builds are green locally.

nnunley (approved):
Reviewed via multi-agent lens pass with focus on the host-capability security boundary. Verified: dimension bounds cap at 16384 (prevents w×h×4 int32 overflow), exact buffer-size check, capability gated behind *surface* with per-Run push/pop, and panic isolation on faulty sinks (fire-and-forget). No sandbox-escape or OOB-write path found; defaults to nopSurface. LGTM.

PR #573, opened 2026-07-18
ci(perf-timeline): revert -count to 3 — a count-4 ubuntu leg measures 104 min, past the 90-min timeout

mparrett:
The count-4 bump (cefa144) rode #565's merge a few minutes before the timing probe it was waiting on finished. The probe answers it: one ubuntu `-full` leg at count 4 runs **103.9 min** end to end (101.5 min in the capture step), past `timeout-minutes: 90` — on current main every timeline leg dies at the kill, which would re-stall the timeline #565 just unjammed. Probe run (fork, publish disabled, timeouts raised so it could finish): https://github.com/mparrett/let-go/actions/runs/29653784883

The breakdown matters for #561: the six suite variants at count 4 cost ~4 min combined (each finishes in under a minute) — the 1 → 4 suite pins are not the problem. The cost is the main `pkg/vm` bench invocation at `-count 4`: ~97 min, and it scales with count. The last count-3 leg fit in ~65 min.

This restores `-count 3`, the measured-fit envelope, and leaves the estimator question on #561 (count 3 lands in the new `reduceSamples` average-of-2 branch once that merges — see the discussion there).

### DIFF
--- a/.github/workflows/perf-timeline.yml
+++ b/.github/workflows/perf-timeline.yml
@@ -126,13 +126,15 @@ jobs:
           snapshot_tmp="${RUNNER_TEMP}/${snapshot_name}"
           raw_tmp="${RUNNER_TEMP}/${snapshot_name%.json}.jsonl"
 
-          # -count 4 matches bench-ratchet's warmup-discarding median sampling
-          # (#561): rep 1 is discarded, leaving a median of 3. A count of 3
-          # would reduce to an average of 2 there; under pre-#561 mean
-          # aggregation it's simply one more sample. Safe in either merge order.
+          # -count 3: a count-4 ubuntu leg measured 103.9 min end to end
+          # (pkg/vm bench alone ~97 min), past timeout-minutes: 90 — every
+          # timeline leg would die at the kill. Count 3 keeps the leg in the
+          # measured ~65-min envelope. Note for the #561 sampling rework:
+          # count 3 lands in reduceSamples' discard-warmup-then-average-of-2
+          # branch there; the estimator question lives on that PR.
           go run ./cmd/bench-ratchet \
             -full \
-            -count 4 \
+            -count 3 \
             -benchtime 2s \
             -timeout 30m \
             -out "${raw_tmp}" \


mparrett:
@nnunley — 5-line revert, checks green; each timeline leg on main currently burns ~90 runner-minutes before dying at the timeout. Stamp when you get a sec?

nnunley (approved):

nnunley:
Ran a multi-agent review sweep tonight; this one intersects the sampling work, so laying out where I'd like it to land.

The 104-min data point is real and the timeout has to be respected — no argument there. The problem is what count-3 does to the estimator: with the warmup-discarding reduction in #561, count-3 leaves **2 post-warmup samples, so the "median" degenerates to a mean-of-2** — which forfeits exactly the outlier robustness the sampling change exists for, and CI runners are the *noisiest* environment we measure in, where sample count matters most.

Alternatives I'd prefer over a blanket revert, in order:

1. **Raise `timeout-minutes` to 120 for the ubuntu leg.** The measured 104 min fits with headroom, and GitHub allows up to 360. This keeps count-4 everywhere and is a one-line change.
2. **Per-OS counts**: keep count-4 on macos-14 (which fits) and drop only ubuntu to 3, recorded as a known caveat.
3. Shard the ubuntu leg by package if 120 min ever stops being enough.

Independent of which we pick: **snapshots should record the sampling method/count in their metadata.** Once count-3 and count-4 snapshots coexist in the timeline, future analysis can't distinguish a real perf shift from an estimator change. I'm also planning to add a min-count warning/enforcement to bench-ratchet itself on the #561 branch so this degradation can never happen silently again.

If option 1 works for you, happy to see this PR repurposed into the timeout bump + metadata field.

mparrett:
@nnunley — follow-up data that changes option 2's premise: **the macos-14 leg has never fit the 90-min timeout, at any count.**

The arm64 leg was added in #363 (2026-07-06). Since then, all but one of its ~40 runs have run the full 90 minutes and died at the `timeout-minutes: 90` kill ("The job has exceeded the maximum execution time of 1h30m0s" in the job annotations), including the runs after this PR merged at count 3. The one exception was cancelled by hand at 64 min, so it produced no snapshot either. The current main tip's run shows the shape — ubuntu leg green in 82 min, macos leg cancelled at exactly 90: https://github.com/nooga/let-go/actions/runs/29668554565 (the seven runs before it repeat it).

Two consequences:

- `perf-data` has **zero arm64 snapshots**; every machine-keyed file on the branch is amd64. The Apple-Silicon line on the perf page has been empty since the leg landed, and each push to main spends 90 macos runner-minutes producing nothing; that burned pool is also what fed the >24h queue starvation #565 had to work around.
- "Keep count-4 on macos-14 (which fits)" doesn't hold — count 3 doesn't fit either. We don't know the macos envelope, because no leg has ever finished; all we know is >90 min at count 3.

This also sharpens #581 item 6, which reads the red runs as the runner-pool queue cancel: the queue starvation is real (legs sit queued for hours first), but the terminal cause is the timeout — the job-level timestamps show each macos leg *running* for exactly 90 minutes before the kill. So moving the leg to a nightly schedule alone wouldn't fix it; a scheduled leg still dies at 90 min unless the timeout moves too.

Given the estimator constraint from #561 (count 3 is already the floor before the median degenerates to a mean-of-2), cutting the macos count is the wrong lever. Proposal:

1. **Raise `timeout-minutes` on the macos-14 leg now** (say 150; GitHub allows 360). One-line unblock, keeps count 3, and the first completed leg tells us the real envelope.
2. **Probe the true macos runtime on a fork**, the same way the 104-min ubuntu number was measured, and right-size the timeout from data.
3. If the measured envelope turns out ugly, or macos runner supply makes ~2h legs flaky, fall back to a scheduled cadence for the arm64 leg — daily on the tip plus backfill — instead of per-push (the #581 item 6 shape, with the raised timeout riding along).

One more #561 implication: until a macos leg completes, estimator and metadata policy only affects amd64 data in practice. So if snapshots grow a sampling-method field (agreed, they should), the arm64 lane can adopt it from its first real snapshot; there's no mixed history to worry about on that side.

I'll open the timeout one-liner and kick off the fork probe unless you'd rather fold this into the #561 branch.

Issue #575, opened 2026-07-18
Epic: block interface as shared liveness metadata + block-arg erasure (EPIC-018)


nnunley:
Companion of #574 (EPIC-017) — the cross-block/locals half of the IR representation work, made visible upstream. Dates refer to the internal iteration log.

## Summary

Block-args are a low-level accommodation in the shared IR: the stack VM lacks native locals, so every cross-block value must be threaded through branch-target args, and `lower.lg` **rejects** any direct cross-block ref (`check-cross-block!`). That rejection is why the AOT inline splice is unusable on the bytecode path (fold-unroll ships an outline-to-sibling workaround instead), and it forces the DUP-heavy branch-arg materialization that dominates lowered-code size (see #574's 2× finding).

The epic builds shared liveness metadata on the block interface, then gives the stack VM real locals, then erases block-args where a single-source value makes them redundant.

## Story ledger

**Done (foundation, backend-independent):**
- STORY-0068 — `ir.passes.liveness`: shared per-block liveness (in/out/kill) as the substrate for everything below. *(done, ITER-0066)*
- STORY-0069 — relocate/deduplicate the existing ad-hoc reachability helpers onto it (`lower_go`'s `live-nids` was the one genuine duplicate). *(done, ITER-0067)*
- STORY-0070 — single-**value** resolver: trace a block-arg to its unique source value when all in-edges agree (scope refined from single-source to single-value). Retiring `check-cross-block!` rides on this. *(done, ITER-0068)*
- STORY-0071 — block-arg classification + census over the corpus: how many args are erasable, what the win is. *(done, ITER-0069)*

**Pending (backend-coupled; sequencing guard: hold until STORY-0060 / EPIC-016 re-check verdict — if that flips to GO for an index-RPN mainline VM, the stack-VM half here is superseded):**
- STORY-0072 — `OP_LOAD_LOCAL`/`OP_STORE_LOCAL` in the stack VM: frame locals so cross-block values stop living on the operand stack.
- STORY-0073 — resolver **lowering**: `lower.lg` consumes cross-block refs via locals; retires `check-cross-block!` and the junk-below accounting class with it.
- STORY-0074 — retire the fold-OUTLINE workaround: the general inliner (not just fold-unroll) unblocks on the bytecode path.

## Why this matters now

[...]

Issue #574, opened 2026-07-18
Epic: IR normalization + execution-form cleanup (EPIC-017)


nnunley:
Making an internal roadmap epic visible upstream (referenced from #528 discussion; companion of the EPIC-018 issue). Dates refer to the internal iteration log.

## Summary

The index-RPN execution spike (ITER-0057) showed that the IR's **representation mechanics — not the execution model — dominate interpreted performance**: removing string dispatch, per-inst guards, and dead-param edge writes moved the numeric slice from 0.88× to 1.03–1.04× geomean (1.23–1.28× at scale). This epic holds the representation follow-on work: normalize/canonicalize the op catalog, strip stack-VM accommodations before execution, and rework frame allocation.

## Root cause (architecture)

The design already has the right levels: `ir.structurize` captures a target-independent control tree (`:if`/`:loop`/`:seq`/`:break`/`:continue`/`:return`/`:tail`/`:goto`), and `lower_go` consumes it. Two holes break the layering:
- **`structurize` is incomplete** — `:try` is not a structural node, so try/catch leaks to the op level and `lower_go` special-cases it outside the tree walk.
- **The bytecode backend never adopted the structural level** — `lower.lg` goes straight from the low CFG to stack bytecode, bypassing `structurize`. This is the single root cause behind the cross-block splice rejection and `:try` being unlowerable on the bytecode path.

**Fresh supporting evidence (2026-07-18):** the stack lowering's junk-below accounting proved fragile in practice (per-pred disagreement corrupts OP_RECUR drop counts → slice underflows), and IR-lowered bytecode measures ~2× the instruction count of the direct compiler on branch-heavy code, ~67% of the excess being branch-arg DUP_NTH shuffle. Both are consequences of the missing structural level.

## Physical-layout findings (audit vs Carbon SemIR / indexed-RPN), ranked

1. **Packed inst records** — insts are boxed persistent vectors with nested refs vectors vs ~16-byte tagged unions with inline operand words + side-table spill (top measured cost). #558's type side table is the interim step on this arc.
2. Canonicalization/interning (STORY-0058, reframed: `:gt`→`:lt` normalization cheap now; `:inc`/`:dec` respecialization gated on the structural level).
3. Mutation-by-tombstone → freeze-after-construction (STORY-0059, **done**: pre-execution cleanup pass).
4. Blocks as (start,len) spans via post-optimize dense renumbering.
[...]

Issue #256, opened 2026-06-18
set! mutates var root instead of throwing when no thread-local binding (Clojure deviation)


nnunley:
## Summary

`(set! x v)` on a var with no active thread-local binding silently mutates the var's **root** instead of throwing, deviating from Clojure — which throws `Can't change/establish root binding of: x with set`.

Source: the `OP_SET_VAR` handler in `pkg/vm/vm.go` — the no-binding case falls through to `varrd.SetRoot(val)` (the code self-labels this a "PERMISSIVE DEVIATION").

## Why it isn't a one-line guard

let-go uses var-root `set!` as a **general mutation primitive**, which Clojure never does. Flipping the fallthrough to throw panics core load on the first `defmethod`:

- `defmethod` (`pkg/rt/core/core.lg:2601`) expands to `(set! name (defmethod* name ...))` — a root-level `set!` with no binding (e.g. `print-method`, core.lg:2608).
- All 23 dynamic-var `set!` sites live in `pkg/rt/core/test.lg` (`register-test`, `register-fixtures`, the top of `run-tests`) and also `set!` at root.

Clojure equivalents never touch the var: `defmethod` mutates the `MultiFn` via `addMethod`; `clojure.test` binds `*report-counters*` via `binding` + a ref and mutates the ref.

## Path to fix

1. Rework `defmethod` to mutate the multimethod object (`addMethod`-style) rather than `set!`-ing the var root.
2. Move the `test.lg` harness onto `binding`/refs instead of root `set!`.
3. Then make `OP_SET_VAR` throw when `setBinding` returns false (no active thread-local binding).

## Context

Surfaced while addressing the Codex review on #251 (def/declare promise semantics). The `DefStub`/`IsBound` LGB-parity fix landed in that PR; this is the remaining, larger deviation it exposed.

mparrett:
@nnunley Thanks for opening this.

Do you have experience with sub-issues in github? I'm not sure but I'm wondering if it would be helpful to create a parent issue for ancillary review findings like this one, with the individual findings as sub-issues. Or maybe it just complicates things. Let me know what you think!

nnunley:
It looks like we can add new issues to an Epic, as wel las move existing issues under them.

PR #579, opened 2026-07-19
perf(ir): re-emit ordinary var loads instead of copying them; add lowering-shape ratchet

nnunley:
## Why

The IR pipeline is an *optimizing* pipeline, so "it compiles correctly" isn't the bar — it has to pay for itself. It doesn't, on compute-bound code. Measured on `main`, identical source through both paths:

| workload | runtime (IR vs plain) | instructions |
|---|---|---|
| `fib 21` | **1.135× slower** | 18 → 24 (+33%) |
| `loop`/`recur` sum | 0.99× | 16 → 18 |

Six of `fib`'s 24 instructions were `DUP_NTH` — pure stack shuffle. `materialize!` copies each operand to the top of the stack and strands the original below as junk; the trailing `RECUR` in a function containing no loop is that junk's cleanup.

## What this fixes

One cause was addressable without restructuring. `lower.lg` excluded **every** `:load-var` from cheap-load re-emission, so a var used twice was materialized once and copied with `DUP_NTH` per extra use. `lower_go.lg` already solved this more precisely — restrict materialize-once to **set!-TARGET** vars, whose two loads could otherwise straddle the `set!` and read different roots. Ordinary var loads can simply be re-emitted per use, which is exactly what the plain compiler does and what the VM's load-time snapshot semantics expect.

This PR mirrors that rule into the bytecode backend.

## Effect

- `fib` 24 → 23 instructions, 6 → 4 `DUP_NTH`; `LOAD_VAR` count now matches the plain compiler exactly.
- **ir-stress native-lowering failures 20 → 11.** The entire `:validate/cross-block-ref` bucket (8) retires — a re-emitted load no longer crosses blocks. Baseline rebaselined per the ratchet's own rule.
- `make parity-quick`: **PARITY** (6311/6311 assertions, both engines). Full `./test` green, `pkg/{ir,vm,compiler}` green, `make check-generated` clean.
- **Runtime unchanged** (fib 1.14× vs 1.135×, within noise). Re-emission trades a stack copy for a var deref, so this is a coverage and semantics win, not yet a speed one. Stating that plainly rather than claiming a perf win the numbers don't support.

## What it doesn't fix

The residual shuffle is emission **order**: the call argument is computed before the callee is loaded, so it has to be copied above it, stranding the original. That isn't a peephole — it traces to the bytecode backend never having received **structural lowering**. `structurize` is required by `lower_go.lg`, `dominance.lg`, and `passes/pipeline.lg`, but *not* by `lower.lg`, which still walks raw blocks by stack position. Without the control tree it can't know consumer order or that a value is dead after a use in a sibling arm, so it copies defensively.

That's EPIC-017 (#574) and EPIC-018 (#575, STORY-0072/0073: per-edge stack maps, block-arg erasure).

## The ratchet

`test/ir_lower_stack_discipline_test.lg` pins the emitted **shape**, not just results:

- `fib` ≤ 23 instructions, ≤ 4 `DUP_NTH` (target: 18 / 0 — the plain compiler)
- `sumloop` ≤ 18 instructions, ≤ 8 `DUP_NTH` (target: 16 / 0)
- non-`set!` var loads re-emit per use, matching the plain compiler

Upper bounds, not equalities, following the same discipline as `ir-stress-gate` and `bench-ratchet`: an improvement passes with room to spare, then the numbers tighten in the same commit. `sumloop` is the sharper case — 8 of 18 instructions are shuffle, i.e. the loop body spends nearly half its budget rearranging the stack rather than computing.

### DIFF
--- a/docs/perf/ir-stress-baseline.edn
+++ b/docs/perf/ir-stress-baseline.edn
@@ -1,9 +1,9 @@
-{:failed 20
- :total 2393
- :passed 2373
- :captured "2026-07-11"
+{:failed 11
+ :total 2431
+ :passed 2420
+ :captured "2026-07-19"
  :how "LG_STRESS_PASSES=1 make ir-stress (corpus = every shipped/test/example .lg)"
  :note
  "ITER-0021 lowering-coverage ratchet. `make ir-stress-gate` fails when the\n  corpus failure count exceeds :failed — i.e. a form newly fails to lower to\n  native Go and falls back to bytecode / vm dispatch (a trampoline, not a\n  direct/native call). The ratchet only tightens: after a genuine coverage\n  gain (fewer failures), the baseline re-tightens and locks the improvement in.\n  NEVER loosens the gate after a coverage loss without human investigation —\n  it's always a real regression, and the gate's role is to prevent PR merges\n  from masking a regression.\n\n  Corpus is discovered from every shipped/test/example .lg, so upstream growth\n  moves :total. :failed stays 20 in already-known bucket classes; no new mode."
  :buckets
- {":unresolved/a" 1, ":validate/cross-block-ref" 8, ":unresolved/_" 1, ":build/unrecognized-form" 1, ":unresolved/n" 7, ":other | unsupported function body shape" 1, ":unresolved/read-json" 1}}
+ {":unresolved/a" 1, ":build/unrecognized-form" 1, ":unresolved/n" 7, ":other | unsupported function body shape" 1, ":unresolved/read-json" 1}}
--- a/pkg/rt/core/ir/lower.lg
+++ b/pkg/rt/core/ir/lower.lg
@@ -252,6 +252,146 @@
         (contains-after? refs i (nth refs i)) false
         :else (recur (inc i))))))
 
+;; --- set!-target vars ------------------------------------------------
+;;
+;; A :load-var may be re-emitted at each use (one LOAD_VAR, no junk) exactly
+;; as the plain bytecode compiler does — it emits a fresh LOAD_VAR per use and
+;; the VM snapshots the root when the load executes. That is unsafe only for a
+;; var this same function set!s: two loads straddling the set! would read
+;; different roots, where a single materialized load holds the pre-set! value.
+;; So materialize-once is restricted to set!-TARGET vars, mirroring
+;; lower_go.lg's load-var-snapshot? rule. A blanket exclusion of every
+;; :load-var (the pre-2026-07-19 behavior) forced a DUP_NTH copy per extra use
+;; of ordinary vars, which is how self-recursive fns like fib lowered to 33%
+;; more instructions than the compiler this pipeline is meant to beat.
+
+(defn- var-key-of-symbol [aux]
+  (let [sym (symbol aux)]
+    (str (or (namespace sym) "core") "/" (name sym))))
+
+(defn- var-key-of-var-value [v]
+  ;; A *vm.Var value stringifies as "#'ns/name"; strip the #' and re-key.
+  (let [sym (symbol (subs (str v) 2))]
+    (str (or (namespace sym) "core") "/" (name sym))))
+
+(defn- var-key-of-ref [f ref]
+  ;; The var a set!'s TARGET operand denotes. build-set! emits that target as a
+  ;; Var-valued :const (aux = the *vm.Var), NOT a :load-var — so matching only
+  ;; :load-var here (the pre-fix behavior) never recognized a real set!, and the
+  ;; snapshot protection below silently did nothing. Mirror lower_go.lg's
+  ;; var-key-of-ref: accept both a :load-var (aux symbol) and a Var-valued
+  ;; :const, so the bytecode + Go lowerers agree on which loads are set! targets.
+  (let [op (ir/op ref f)]
+    (cond
+      (= :load-var op) (var-key-of-symbol (ir/aux ref f))
+      (= :const op)    (let [v (ir/aux ref f)] (when (var? v) (var-key-of-var-value v)))
+      :else nil)))
+
+(defn- set-var-target-keys [l]
+  "Var keys this fn set!s, computed once and cached on the lowering state."
+  (let [cached (:set-var-keys @l)]
+    (if (some? cached)
+      cached
+      (let [f    (f-of l)
+            keys (reduce
+                   (fn [acc bid]
+                     (reduce (fn [a nid]
+                               (if (= :set-var (ir/op nid f))
+                                 (let [refs (ir/refs nid f)
+                                       k    (when (pos? (count refs))
+                                              (var-key-of-ref f (nth refs 0)))]
+                                   (if k (conj a k) a))
+                                 a))
+                             acc (ir/block-insts bid f)))
+                   #{} (ir/blocks f))]
+        (swap! l assoc :set-var-keys keys)
+        keys))))
+
+(defn- load-var-snapshot? [l nid]
+  "True for a :load-var whose var this fn set!s — the one case that must be
+   materialized once rather than re-emitted per use."
+  (let [f (f-of l)]
+    (and (= :load-var (ir/op nid f))
+         (contains? (set-var-target-keys l) (var-key-of-symbol (ir/aux nid f))))))
+
+;; --- vars live across an effect --------------------------------------
+;;
+;; Re-emitting a :load-var at each use matches the plain compiler ONLY while
+;; the var's root can't change between the load and that use. A same-fn set!
+;; is caught above; a CALL is the other mutator — the callee may set! the var
+;; (or def/redefine it). The plain compiler snapshots the root at the def site
+;; (it stores let-bound values into locals BEFORE any following call), so a use
+;; after a mutating call must read the pre-call root. Re-emitting reads the
+;; POST-call root instead — `(let [x mv] (mutate!) x)` returned the mutated
+;; value. So a load live across an effectful op is body-emitted at its def and
+;; DUP_NTH-copied per use, exactly as the pre-2026-07-19 blanket rule did, but
+;; now only for the loads that actually straddle an effect — fib/sumloop callee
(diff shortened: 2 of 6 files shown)


mparrett:
The set!-target carve-out doesn't fire: `build-set!` resolves the target at build time and emits it as a `:const` whose aux is the `*vm.Var` — never a `:load-var`. So `set-var-target-keys` matches nothing, the snapshot set is always empty, and every var load re-emits per use, including across a `set!`.

Repro on this branch (e475b0f):

```clojure
(def pv 1)
(defn probe [] (let [a pv] (set! pv 99) (if (> a 0) (+ a pv) (- a pv))))
```

`probe` returns 100 via the plain compiler and via main's IR path, 198 on this branch — the deferred cheap-load re-emits `a`'s load at its use sites, which sit after the `SET_VAR`:

```
[:LOAD_CONST 0 user/pv]
[:LOAD_CONST 1 99]
[:SET_VAR]
[:LOAD_VAR 0 user/pv]   ; "a", re-read post-set!
...
```

The PR's fixtures never trip it because fib/sumloop don't `set!` anything, so the ratchet numbers stand.

`lower_go.lg`'s version of the rule already handles this shape: `var-key-of-ref` accepts both a `:load-var` target and the Var-valued `:const`, keying the latter via `var-key-of-var-value`. Mirroring those two helpers plus the `:const` arm restores the carve-out. The straddle shape above is probably worth pinning in `ir_lower_stack_discipline_test.lg` next to the re-emission test, since it's the P1 semantics the new code comment describes.

mparrett (changes requested):
Assuming the auto review is accurate, this needs an update before merging. Let me know if it's off or not.

mparrett (changes requested):
The optimization does what it says on its own fixtures — the stack-discipline ratchet and the `fib` `LOAD_VAR` parity reproduce at the stacked head, and the test suite is green. But the relaxed guard is unsound for a `set!` performed by a callee, and the divergence is observable today. Minimal repro, both compilers in one file:

```clojure
(def mv 1)
(defn mutate! [] (set! mv 2) 100)

(defn probe-bc [] (let [v mv] (+ (mutate!) v)))

(set! *ir-compile* true)
(defn probe-ir [] (let [v mv] (+ (mutate!) v)))
(set! *ir-compile* false)

(set! mv 1)
(probe-bc)  ;; => 101  (v fixed at bind time)
(set! mv 1)
(probe-ir)  ;; => 102  (v read after the callee's set!)
```

On main both return 101. At this PR's head the IR path returns 102: `should-body-emit-cheap?` defers the single-use load (it isn't its consumer's first ref), so the re-emitted `LOAD_VAR` executes after `mutate!` — the let-init moved across the call. The multi-use flavor diverges the same way: `(let [v mv] (+ (mutate! v) (idf v)))` returns 2 on main and 3 at the PR head, because the second use's re-emitted load runs after the first argument's call. The case the guard does cover — a `set!` in the same fn between bind and use — still snapshots correctly; I probed that too.

Why the rationale doesn't hold: the plain compiler re-emits one `LOAD_VAR` per *source deref site*, and a let binding is one deref whose read the local pins at bind time. The IR path re-emits per *use of one inst*, and a let-bound deref is exactly one inst with uses elsewhere, so re-emission moves the read to wherever the use materializes. Those are the same thing only when the source had a deref at each use site. `set-var-target-keys` can't see any of this: `set!` mutates a var root from any fn (#256), so a same-function scan can't establish that a root is stable across a call.

`lower_go.lg` has the same rule with the same same-fn guard, so the Go backend presumably shares the cross-fn hole (I didn't build an AOT repro). Its header comment names the bytecode VM's snapshot-at-load behavior as the reference it diverges from — after this PR that reference behavior is gone on the bytecode side too. If the rule is kept anywhere, I'd decide that at a shared level rather than per backend.

Two smaller things from tracing this:

- `lower.lg`'s scan only recognizes a set! target that is a `:load-var` ref; `lower_go`'s `var-key-of-ref` also handles the Var-valued `:const` operand ("mirrors set-var's value-expr operand handling"). If that operand shape can reach `lower.lg`, the guard misses the set! and the same-fn case breaks too; if it can't, a comment saying why would keep the two scans mirrored.

- `ir.passes.mutability` has the same evidence base (this fn's `:set-var` insts plus a short list of var-mutating builtins), so CSE's `stable-load-var?` merges derefs that a callee's `set!` can distinguish. Pre-existing and out of this PR's scope, but the same assumption now lives in three places.

A direction that keeps the win and the semantics, offered as a sketch: treat re-emission as sound only when the load doesn't cross an effect — defer or re-emit only when no `:call` or `:set-var` sits between the load's def position and the materialization point, and stop CSE-merging `:load-var` so each source deref keeps its own inst. `fib` keeps its 2 `LOAD_VAR`s and 0 `DUP_NTH` (each load sits next to its consumer with only arithmetic between); the let shapes pin at bind position; a multi-use value crossing a call pays a `DUP_NTH`, the correct price for a snapshot. The deeper fix is the frame-locals story in #575 — a store at the def site gives bind-time semantics and removes the shuffle in one move — but that's gated, and this PR shouldn't ship unsound in the meantime.

Happy to turn the repro into a regression test on the PR branch if useful.

mparrett:
Correction to my review above: the sentence claiming the same-fn case "still snapshots correctly" is wrong, and my earlier comment on this PR already showed why — `build-set!` emits the set! target as a Var-valued `:const`, so `set-var-target-keys` is always empty and the carve-out never fires. That comment's repro diverges at e475b0f as described (198 vs 100; re-verified). The same-fn shape I probed returned the right value because its layout happens not to defer the load, not because the guard fired.

The two findings compose rather than compete: mirroring `var-key-of-ref`'s `:const` arm restores the same-fn carve-out, but the cross-fn repro in my review still diverges after that fix — that fn contains no `set!` at all, so a same-function scan is empty no matter how the target is keyed. The effect-barrier (or frame-locals) direction is still needed for the general case.

nnunley:
Implemented the direction you sketched — re-emission is gated on not crossing an effect, and CSE no longer merges `:load-var` across one — on **both** backends, not per-backend.

**`lower.lg` (bytecode)**
- `load-var-crosses-effect?`: a `:load-var` live across an effectful op (`:call`/`:tail-call`/`:set-var`/`:def`, scanned *strictly between* the def and each use so the consuming call itself doesn't count) is body-emitted at its def and `DUP_NTH`-copied per use. Cross-block uses are treated conservatively.
- `cse.lg`: a var-root barrier (a non-pure `:call`, `:set-var`, or `:def`) drops earlier `:load-var` congruences from the per-block table, so two derefs straddling an effect stay distinct insts rather than merging into one value that materialize-once would then collapse.

**`lower_go.lg` (AOT)**
- The same `load-var-crosses-effect?`, folded into a `load-var-materialize?` predicate that drives all three snapshot-local lifecycle sites (declare / assign / reference) so they stay in lockstep. You were right that the Go backend shared the hole.

Your two repros now match the plain compiler on both paths:
- `(let [v mv] (+ (mutate!) v))` → `101 / 101`
- `(let [v mv] (+ (mutate1! v) (idf v)))` → `2 / 2`

`fib` is unchanged — 2 `LOAD_VAR`, 0 `DUP_NTH` (callee loads sit adjacent to their invoke with only arithmetic between); the shuffle ratchet and `LOAD_VAR` parity still hold. New parity tests in `ir_lower_stack_discipline_test.lg` cover the same-fn `set!` and the cross-call cases against the bytecode baseline; `TestGogenAOTDiff` + `TestLoweringDeterminism` are green, so the AOT tree matches bytecode.

The two smaller items:
- `lower.lg`'s `var-key-of-ref` now also handles the Var-valued `:const` set-target operand (it previously matched only a `:load-var` ref, so the same-fn snapshot was silently inert — that's now genuinely exercised). `lower_go.lg` already had this shape.
- The CSE `stable-load-var?` merge you flagged is the barrier change above. I scoped it to CSE rather than touching the shared `ir.passes.mutability` evidence base, since `rebinds-var-roots?` consumes the same analysis for the baked-direct-call fallback.

One deliberately-narrow edge, called out for transparency: a load used **only as a call's callee** (`(f (g x))` where `g` reassigns `f`) is excluded from materialization — it re-resolves at call time (re-emit / `CachedVarFn`) on both backends, which is pre-existing behavior and outside the value-load re-emission this PR changed; materializing it just produced a dead Go local. Happy to close that too if you'd rather, but it isn't the regression this PR introduced.

On the shared-level point: the rule is duplicated per backend for now to close the divergence; a shared home (alongside the #575 frame-locals story) is a sound follow-up rather than something to hold this on.

mparrett (approved):
Verified this independently in a fresh worktree off f6ff14d24, not just from the PR description.

Reran all five of my straddle probes from the 07-20/21 review (`(let [v mv] (+ (mutate!) v))` and the multi-use variant). Both now return 101/101 and 2/2, matching the plain compiler — the case that previously diverged to 102/3 is fixed. Disassembly confirms the mechanism you described is what fires: `LOAD_VAR mv` → `INVOKE mutate!` → `DUP_NTH` × 2, i.e. the load is materialized once at the def and copied per use instead of re-emitted per use. `lower_go.lg` carries the same `load-var-materialize?` predicate, not just `lower.lg`, so the Go backend gets the same fix rather than inheriting the old hole.

I also tried to break the excluded edge you called out (`(f (g x))` where `g` reassigns the var bound to `f`) on both main and this branch. Both resolve `f`'s var before evaluating `g`'s call on both backends, so there's no divergence to find. Agreed it's out of scope for this PR.

New `ir_lower_stack_discipline_test.lg` pins the exact scenario from my review. Full suite: 901 pass / 0 fail, plus `TestGogenAOTDiff` and `TestLoweringDeterminism` green.

Blocking merge, unrelated to correctness: `gh pr view` currently shows `mergeable: CONFLICTING`. Merging `origin/main` locally, it's a single conflict in `pkg/rt/generated.sums` — the auto-generated manifest hash, from main moving since this branch forked. Rebase + `make generate` should clear it.

Approving — the miscompile is fixed and the fix generalizes across both backends. Rebase and this is good to merge from my side.

PR #580, opened 2026-07-19
feat(ir): *ir-compile-strict* + a bytecode-path coverage census (70.7% vs the 99.5% we were measuring)

nnunley:
Stacks on #579.

## Why

Two blind spots hid the same thing: how much does *not* lower on the path `*ir-compile*` actually drives at runtime.

**1. The `defn` fallback was silent.** The hybrid fallback itself is right — without it a single unlowerable defn aborts a whole corpus. But "falls back" and "falls back *silently*" are separable, and only the first is a feature. `*ir-compile-verbose*` reports without failing, so nothing could gate on it: you pay full IR compile cost, run bytecode anyway, and nothing says so.

**2. The coverage ratchet only ever measured `lower-go`.** `corpus` mode hardcoded `:lower-go`, so a form that stopped lowering to *bytecode* moved nothing. `:validate/cross-block-ref` and friends are Go-path buckets; the bytecode path had no census and no gate.

## What's here

- **`*ir-compile-strict*`** (default `false`) — rethrows instead of downgrading, naming the defn. It catches what `*strict-structured?*` cannot: that var is scoped to structured-control-flow drift in the **Go** backend and never observes a bytecode-path failure.
- **A `:lower` mode** — `try-lower-bytecode`, the exact analog of `try-lower-go`: build IR + lower via `ir.lower`, no eval, no interning. Selectable with `LG_STRESS_BACKEND`, plus `make ir-stress-bytecode{,-gate,-rebaseline}` against its own baseline file.
- The `ir-compile` census mode now binds `*ir-compile-strict*` — **without it an eval that silently downgraded returned `:ok` and the census scored a coverage failure as a pass.**
- `classify-error` folds lowering failures (which carry an instruction index) into `:lower/not-on-stack/<op>` and `:lower/unsupported-op/<op>`. Without folding the census emits 130+ single-count rows instead of a diagnosis.

## First capture of the unmeasured path

| backend | coverage | failures |
|---|---|---|
| `lower-go` | 99.5% | 11 / 2432 |
| `lower` (what `*ir-compile*` runs) | **70.7%** | **713 / 2432** |

```
:lower/not-on-stack/{call,eq,gte,block-arg,...}   535
:lower/unsupported-op/:try                        144
:lower/unsupported-op/:div                         21
```

## Why the 535 exist

They're self-inflicted. The IR is **indexed RPN**: `:insts` is a vector indexed by `InstId`, and every operand in `:refs` is already an `InstId`. The index is in hand at the moment of lowering. `lower.lg` discards it and reconstructs a **stack depth** instead (`value-stack-pos`, `DUP_NTH n`). Depth is fragile where the index is stable — it shifts as the stack moves, and a value defined in another block, or displaced when an earlier operand was materialized, has no computable depth at all. The information was never missing; lowering dropped it and then couldn't recompute a substitute.

The VM offers no alternative today. Its entire addressing vocabulary is `LOAD_ARG` / `LOAD_CLOSEDOVER` (read-only, indexed) and `DUP_NTH` (depth-addressed), and `Frame` carries `stack / args / closedOvers` with **no writable slots** — so lowering has nowhere to spill what it can't keep in position. That's also why `:try` has no emission case: try/catch needs values to survive a handler boundary the stack can't express.

The same root cause produces the surplus `DUP_NTH` shuffle measured in #579 (`fib` 23 instructions vs the plain compiler's 18; a `loop`/`recur` sum spending 8 of 18 instructions on shuffle). The plain compiler never meets the problem — a post-order AST walk maps onto a stack by construction.

## Prior art: fmpl

Compared against `~/development/fmpl`'s IR specs (`specs/indexed-rpn-conversion.md`, `fmpl-core/src/vm_internal/frame.rs`). The instructive part is what fmpl **doesn't** have: there is no bytecode layer. It executes the indexed IR directly, with `Frame { values: Vec<Value> }` where `values[i]` is the result of instruction `i`, and operands carried as indices in the instruction itself (`Add { lhs: InstrIndex, rhs: InstrIndex }`).

So fmpl never performs the index→stack-depth conversion that generates all 535 of our failures. That's a point in favour of EPIC-016 (IR-direct VM) over merely bolting slot opcodes onto the stack VM: the conversion doesn't get *fixed*, it ceases to exist. If we do keep the bytecode layer, the minimum is `OP_STORE_LOCAL`/`OP_LOAD_LOCAL` against a frame slot array, `InstId → slot`, block-args becoming "both predecessors write the same slot", and a peephole that keeps produce-then-immediately-consume on the stack so density doesn't regress.

Traps worth carrying into either design, from fmpl's handling: closures must snapshot captured **values** not slot references; `OP_RECUR` must explicitly write parameter slots before looping rather than assuming stack depth; catch handlers want their own slot scope rather than sharing the try block's.

## Gates

Both ratchets green against their own baselines. `./test`, `pkg/ir`, `pkg/vm` green. `make check-generated` clean. Five tests for the strict knob, including one pinning that the default stays permissive so shipping code keeps the hybrid fallback.


mparrett (commented):
Read and tested this at bb38cf0 (both new test files pass here). The strict rethrow embeds the original error text, so `classify-error` still buckets wrapped failures (verified), and the census binding of `*ir-compile-strict*` is what makes the ir-compile numbers mean anything: without it a silent downgrade scores as a pass, as the body says. The fmpl comparison reads as the strongest input to the EPIC-016 decision so far. Three notes, none blocking:

1. **This inherits the #579 set!-guard miscompile** (details in the comment on #579; the repro fails identically at bb38cf0, and I separately validated that adding the `:const` arm restores parity with the discipline ratchet still green). Consequence for this PR: the 70.7% bytecode census was captured with the inert guard, and `try-lower-bytecode` doesn't eval, so a form that lowers-but-miscompiles scores as covered. Re-capture `ir-stress-bytecode-baseline.edn` once the #579 fix is in; set!-heavy forms may move.
2. `try-lower-bytecode`'s `cond` falls through to `:ok` for any map shape it doesn't recognize, so a future `compile-form` status variant would be silently scored as a pass — the failure mode this PR is built to prevent. Cheap hardening: make the `:else` classify (or throw) instead of passing, so an unknown shape is loud.
3. Doc nit, one line: the `ir-stress.lg` header comment says `LG_STRESS_BACKEND=ir-compile` is the bytecode census, but the Makefile targets and results table use `lower`; the two modes differ (lowering-only vs eval) and the comment blurs which is which.

mparrett:
Correction to my review above: my statement that the `*ir-compile-strict*` binding makes the `ir-compile` census meaningful is wrong for the current harness.

`try-ir-timed` installs the binding inside the thunk executed by `future*`. That binding lives in the future's child execution context, while `eval` compiles and macroexpands the `defn` through the root execution context. The macro therefore sees the default `*ir-compile* = false`, uses the ordinary compiler, and the census records `:ok`.

On `bb38cf0`, the same unlowerable form produces:

```text
foreground strict eval:        :threw
strict binding inside future:  :ok
strict binding outside future: :threw
```

The practical result is visible on `pkg/rt/gogen/gogen.lg`: `lower` reports 11/26 passing with 15 `:lower/not-on-stack/call` failures, while `ir-compile` reports 26/26 passing.

This does not invalidate the direct `lower` census or its 70.7% baseline. It means the other advertised blind spot, the silent fallback in the eval-shaped census, remains open. Please either move the bindings outside the thunk so they surround the call to `try-with-timeout`, or make `eval` context-aware, and add a regression test exercising strict evaluation through `future*`. This is blocking for the strict-census half of the PR.

Separately, the branch still inherits the #579 set!-guard miscompile noted earlier, so its bytecode baseline should be recaptured after that fix.

mparrett (dismissed):
Verified locally at bb38cf03: `make build`, the .lg suite via `go test -run TestRunner ./test/` (both new test files run and pass), and `make ir-stress-bytecode-gate` reproduces the committed baseline exactly: 713/2432, identical bucket counts, gate green. The long note in `ir-stress-bytecode-baseline.edn` is the clearest writeup of the not-on-stack failure class anywhere in the tree, and it lives next to the numbers it explains.

A few findings, none blocking:

1. The `:multi-fn-template` branch in `try-lower-bytecode` looks unreachable. Under `*target* :bytecode`, multi-arity success comes back as a `make-multi-arity` value (which lands in the `:else :ok` arm), and a failed arity throws out of `chunk->fn` before any partial shape can escape; the template shape only arises under the go target (lambda-lift and the per-arity template path both gate on `go-target?`). Harmless as written, but it reads as partial-fallback accounting the bytecode path doesn't have. Either drop it or mark it as try-lower-go symmetry. A corollary I checked while tracing this: because a failing arity throws on this path, the strict knob has no partial-arity blind spot.

2. `:else :ok` scores unknown result shapes as passes. Today that's right: the remaining shapes are a chunk or a MultiArityFn, and the string error channel is handled explicitly. But if `compile-form` grows a new failure-status map, the census silently counts it as coverage and the ratchet locks it in. Inverting the polarity (enumerate the success shapes, classify everything else) would give the census the same fail-loud posture as the strict knob. Low priority.

3. Strict placement checks out: the throw fires at macroexpansion inside the existing catch and escapes the `or`, so the bytecode fallback never runs. And since the wrapper string embeds the original error, `classify-error`'s substring regexes still bucket wrapped failures, which is what keeps the `ir-compile` census buckets comparable with the `lower` backend's.

4. Where is the gate meant to run? Neither `ir-stress-gate` nor the new `ir-stress-bytecode-gate` appears in the workflows; both are Makefile-only. Consistent with the sibling, but the PR's own argument is "nothing in CI can gate on it" — is the intent local pre-PR discipline for now, with CI wiring as a follow-up? A line in the Makefile comment saying where the gate is expected to run would settle it.

5. `LG_STRESS_BACKEND=ir-compile` in corpus mode evals and interns the whole corpus into one runtime, unlike the two no-eval backends, so cross-file interference (redefs, ns state) can skew its buckets. One line in `ir-stress.md` would keep anyone from reading its numbers as directly comparable to `lower`'s.

6. An aside that belongs to #579 underneath: after the rebaseline to 11, `ir-stress-baseline.edn`'s `:note` still says ":failed stays 20 in already-known bucket classes" — the note rides through rebaseline verbatim. The file is tool-maintained, so the fix is in the rebaseline path (or the note should stop naming the number).

The strict-mode census correction — binding `*ir-compile-strict*` in the `ir-compile` backend so silent downgrades stop scoring as passes — is the load-bearing fix here; anyone who trended the old ir-compile numbers should know they were an overcount.

mparrett:
Correction to my review above: its closing claim — that binding `*ir-compile-strict*` in the `ir-compile` census makes silent downgrades stop scoring as passes — is wrong for the harness as shipped, as my earlier comment on this PR already established. Re-verified at bb38cf0: a strict foreground eval of an unlowerable defn throws, and the identical eval through `future*` (the `try-with-timeout` path the census actually runs) returns `:ok`. My point about `classify-error` bucketing wrapped failures holds only once the bindings move outside the thunk or `eval` becomes context-aware.

(Edit: I've since dismissed the approval — this branch also carries #579's confirmed miscompile, so a standing green review was wrong regardless. The `lower` census, its baseline, and the knob itself still hold; re-review comes after the future*-binding fix and the #579 fix underneath.)

mparrett:
@nnunley Gentle nudge, though I'm sure you're already tracking it. This is the current bottleneck for the #613 stack, and the blocking future*-binding scope issue (strict-census half) looks unaddressed since the 07-21 review; the 07-22 push was a rebase. Let me know if I can provide an exact repro/fix shape or a future*-scoped regression test if that would help unblock it!

mparrett:
As offered — here's the repro/fix shape and a future*-scoped test. Adopt, adapt, or toss as you see fit; you're closer to this harness than I am.

**Repro** (matches what you'd expect from your own three-line probe): an unlowerable defn throws in the foreground under strict, but returns :ok through future* when the binding sits inside the thunk — which is exactly how the ir-compile census runs it, so the downgrade scores as a pass.

**Fix shape.** `try-with-timeout` spawns the thunk with `future*`, and future* inherits the *spawning* context's dynamic bindings — so binding around the `try-with-timeout` call (not inside the thunk) is what the child eval sees. Keep the thunk a bare eval and wrap only the ir-compile mode:

```clojure
(let [thunk (cond
              (= measurement-mode :lower-go) (fn [] (try-lower-go form))
              (= measurement-mode :lower)    (fn [] (try-lower-bytecode form))
              :else (fn [] (try (eval form) :ok
                                (catch e (ir.coverage/classify-error e)))))
      t0 (System/nanoTime)
      result (if (= measurement-mode :ir-compile)
               ;; Bind OUTSIDE future*: a binding installed inside the thunk lands
               ;; in the child ctx only AFTER eval has macroexpanded the defn through
               ;; the root ctx at *ir-compile*=false — the silent downgrade the census
               ;; then scores :ok. future* snapshots the spawning ctx's bindings, so
               ;; binding here is what the child eval actually inherits.
               (binding [*ir-compile* true *ir-compile-strict* true]
                 (try-with-timeout thunk per-fixture-timeout-ms))
               (try-with-timeout thunk per-fixture-timeout-ms))
      t1 (System/nanoTime)
      ms (double (/ (- t1 t0) 1000000.0))]
  ...)
```

**Regression test shape** — pins the future* path specifically, so this can't silently regress again:

```clojure
(deftest ir-compile-strict-through-future
  (testing "binding inside the future thunk does NOT take (the bug)"
    (is (= :ok
           (deref (future* (fn [] (try (binding [*ir-compile* true *ir-compile-strict* true]
                                          (eval unlowerable-form))
                                       :ok (catch _ :threw))))))))
  (testing "binding around the future* call IS observed by the child eval (the fix)"
    (is (= :threw
           (binding [*ir-compile* true *ir-compile-strict* true]
             (deref (future* (fn [] (try (eval unlowerable-form) :ok (catch _ :threw))))))))))
```

(`unlowerable-form` = the loop/if-in-recur one you already have in ir_compile_strict_test.lg.) The first case documents the trap; the second is the guarantee. Happy to open a small PR against your branch if that's easier than patching inline.

mparrett (commented):
LGTM, but the minions seem pretty dug in about the lingering review items.

> PR still claims (in its commit message + census docs) to close the silent-fallback blind spot while that half doesn't actually work. As is, it'd do so untested, so it could rot further.

> Would want to see: de-scoped claim (document ir-compile mode as not-strict-through-future*, softened commit message, file a follow-up) OR the future* fix + regression tests.

nnunley:
Addressed the lingering items — the ir-compile-mode blind spot is now fixed and tested, not de-scoped.

**Fail-loud `try-lower-bytecode` (your finding #2, raised both rounds).** The shape classification moved into `ir.coverage/classify-bytecode-result`, which now scores `:ok` ONLY for shapes it can name as a success and classifies everything else, mirroring the `*ir-compile-strict*` posture the PR is built on:
- String → error channel, classify.
- non-map native value (`*CodeChunk` / `*MultiArityFn`) → `:ok`.
- `{:status :lowered}` → `:ok`; `{:status :fallback}` → classify `:reason`.
- **any other map → classified LOUD** (`:census/unexpected-result-shape …`), so a future `compile-form` status variant can't silently score as coverage.

**Unreachable `:multi-fn-template` (your finding #1).** Dropped from the bytecode path — you were right that a multi-arity success returns a `*MultiArityFn` (non-map, now `:ok`) and a failed arity throws out of `chunk->fn`, so the template shape only arises under `:go`. `try-lower-go` keeps its version.

**Tests.** Extracting the classifier into `ir.coverage` (the coverage-classification SSOT) made it unit-testable — `test/ir_compile_strict_test.lg` now has three deftests covering the success shapes, the failure channels, and the load-bearing case: an unknown status map classifies loud, never `:ok`.

**Baseline recaptured post-#579 (your finding #1, other half).** #579 merged, so I rebaselined `ir-stress-bytecode-baseline.edn`: **713 → 719** (`:total` 2432 → 2466 with corpus growth). The +6 is exactly the movement you predicted — all in `:lower/not-on-stack/*` and `:lower/unsupported-op/*`, i.e. `lower.lg`'s stack-position materialize, which #579's effect-crossing materialization (body-emit + DUP_NTH) puts more pressure on. Verified none of the +6 are from this change: zero landed in the new `:census/unexpected-result-shape` bucket; they're all real lowering errors from the `catch` path. Gate is green at 719.

Left `try-lower-go`'s own `:else :ok` as-is for now (you flagged it "today that's right, low priority") — making the established lower-go census fail-loud is a separate change that'd want its own rebaseline. Ran `/simplify` over the diff: clean.

mparrett (approved):
Approving — the load-bearing half (the gated `:lower` bytecode census) is fail-loud and solid now, and that's what matters for the ratchet. The `classify-bytecode-result` extraction with the loud `:else` for unknown status maps is exactly right, and the post-#579 rebaseline (713→719, all real lowering errors) checks out.

The agent pulled the branch and probed the `future*` item empirically, so here's what the agent saw. You can take it if helpful here, a follow up, or leave it.

- The fix (binding outside `future*`) works: the unlowerable form throws through the census path once the binding wraps `try-with-timeout`.
- Applied to the full `ir-compile` census it surfaces +47 previously-hidden downgrades (90→137 fail), all in `:lower/not-on-stack/*` and `:lower/unsupported-op/*`. Before the fix there are zero `:lower/*` buckets — i.e. the strict binding is currently inert in that mode.
- But even fixed, `ir-compile` shows ~47 lowering failures against the `:lower` census's 719 on the same corpus. The two don't reconcile: `*ir-compile*` eval only attempts IR-lowering on a subset, while `try-lower-bytecode` forces it on everything. So `ir-compile` is a fuzzier, ungated user-load signal and `:lower` is the real census.

Given that, I'm fine either way. Apply the 10-liner if you want the strict knob live in `ir-compile` too, or leave `ir-compile` as a non-strict diagnostic and let `:lower` carry the census. Your call; you're closer to it. One soft nudge **if you leave it**: consider trimming the commit-message line that says the `ir-compile` strict binding stops a downgrade from scoring `:ok`, since as measured it currently doesn't — just so the record matches. Not blocking.

The 10-liner, for reference (verified on the branch — flips the census 90 → 137):

```diff
diff --git a/scripts/ir-stress.lg b/scripts/ir-stress.lg
--- a/scripts/ir-stress.lg
+++ b/scripts/ir-stress.lg
@@ -565,19 +565,17 @@
                 (= measurement-mode :lower-go) (fn [] (try-lower-go form))
                 (= measurement-mode :lower)    (fn [] (try-lower-bytecode form))
                 :else
-                ;; *ir-compile-strict* is essential here, not incidental: the
-                ;; defn macro's fallback is silent, so without it an eval that
-                ;; quietly downgraded to the bytecode compiler returns :ok and
-                ;; the census scores a coverage FAILURE as a pass. Strict mode
-                ;; turns that downgrade into the throw classify-error buckets.
-                (fn [] (try
-                         (binding [*ir-compile* true
-                                   *ir-compile-strict* true]
-                           (eval form))
-                         :ok
-                         (catch e (ir.coverage/classify-error e)))))
+                (fn [] (try (eval form) :ok
+                            (catch e (ir.coverage/classify-error e)))))
         t0 (System/nanoTime)
-        result (try-with-timeout thunk per-fixture-timeout-ms)
+        ;; *ir-compile-strict* must be bound OUTSIDE future*: eval macroexpands
+        ;; the defn in the child ctx, and a binding installed inside the thunk
+        ;; lands too late — the downgrade already ran and scores :ok. future*
+        ;; inherits the spawning ctx's bindings, so binding here is what the
+        ;; child eval actually sees.
+        result (if (= measurement-mode :ir-compile)
+                 (binding [*ir-compile* true *ir-compile-strict* true]
+                   (try-with-timeout thunk per-fixture-timeout-ms))
+                 (try-with-timeout thunk per-fixture-timeout-ms))
         t1 (System/nanoTime)
         ms (double (/ (- t1 t0) 1000000.0))]
     (when (and (= result :stress/timeout) autotrace-on?
```

Nice work on the fail-loud classifier and the rebaseline.

Issue #358, opened 2026-06-29
AOT lowering: native homogeneous collections


mparrett:
**TL;DR: the boxed half is the bigger win — ~83× measured.**

## Context

AOT lowering (`lg-compile`, v1.11.0) turns scalar arithmetic into native Go: a Mandelbrot escape kernel lowered to a pure `float64` loop with zero `ec.Invoke`/`rt.*Value`, ~56× faster than interpreted (modulo the param-typing gap in #357). This issue is about the other half of real programs — collection and string transformation — which can't lower at all today. Filing as an investigation to gauge appetite and pick a direction, not to push a finished design.

## Root cause

`lower_go.lg`'s `go-type-spec` maps inferred types to Go types, and the native vocabulary is just:

```
:int -> int   :float -> float64   :bool -> bool   [:dtype T] -> *T
;; everything else:
:vector :map :set :list :cons :lazyseq :seq -> "vm.Value"
:ratio :number :string :unknown :any :nil [:union ...] -> "vm.Value"
```

So any code over vectors, maps, or strings stays `vm.Value`: ops become `rt.AddValue` and friends, calls become `ec.Invoke(rt.CachedVarFn(...))` trampolines, and every step boxes and allocates. One `vm.Value` poisons the chain: `(float i)` returns boxed, cascading the surrounding arithmetic back onto `rt.*Value`.

## Measured prize

The boxed half benefits more than the arithmetic half. A faithful hand-written Go port of the mandelbrot demo's sixel encoder (the part that won't lower), using plain `[]int` and `strings.Builder` (what a collection-lowering would emit), benched against the interpreted demo and byte-equivalent (27794 vs 27695, a float32-boundary diff):

| Phase | Interpreted | Native `[]int` | Speedup | Lowers today? |
|---|---|---|---|---|
| Compute | ~97 ms | 1.23 ms | ~79× | yes (AOT: 56× measured) |
| **Encode** | ~120 ms | **1.45 ms** | **~83×** | **no** |
| Frame | ~217 ms | 2.69 ms | ~81× | partial |

The encoder builds per-color masks (`transient` vec → `assoc!`/`nth` → `persistent!`), then strings them (`str`/`subs`/`conj`). It's the biggest beneficiary, because it's the most boxing-heavy. Even its integer `bit-or` work boxes today: the masks live inside a vector, so `nth` returns `vm.Value` and `assoc!` stores one.

## Two design options

[...]

PR #582, opened 2026-07-19
bench: document the two AOT mechanisms behind the results table

mparrett:
## What

Adds a "Reading the AOT column" section under the Performance table in `benchmark/results.md`, splitting the seven fixtures by what the AOT build does with them: three lower to native Go (fib, tak, loop-recur — the 20-26x wins), four only get a native `run` wrapper around the same boxed runtime calls the VM makes (reduce, map-filter, persistent-map, transducers).

## Why

One undifferentiated AOT column makes the reduce row (1.7x slower under AOT) read as an anomaly. The emitted Go shows it isn't: the reduction never lowers — `reduce.go` calls the identical `rt.Reduce3` with a boxed `+`, plus wrapper dispatch — so parity-or-worse is the expected result for those rows until #270/#358 reach them. Naming the split gives the table a stable interpretation as those issues land, and the map-filter flip in #578 (now a small AOT win) reads the same way: wrapper-cost noise around parity, not lowering.

Continues the read from the #552 review thread. Issue/PR links stay out of the doc itself — README and the benchmark docs carry none, so the section names the boundary plainly and the references live here. Framing only; numbers are unchanged from the #578 recapture.

No urgency on review — docs-only, whenever convenient.

### DIFF
--- a/benchmark/results.md
+++ b/benchmark/results.md
@@ -55,3 +55,16 @@ Clojure JVM times include full JVM startup (~350-500ms) which dominates short be
 | tak | 2.397s ± 0.022s (1.0x) | **93.7ms ± 0.6ms** (0.0x) | 1.908s ± 0.037s (0.8x) | 0.613s ± 0.046s (0.3x) |
 | transducers | 46.5ms ± 0.9ms (1.0x) | 43.0ms ± 1.0ms (0.9x) | **20.3ms ± 0.5ms** (0.4x) | 0.373s ± 0.011s (8.0x) |
 
+### Reading the AOT column
+
+The AOT numbers mix two mechanisms:
+
+- **Lowered to native code** — fib, tak, loop-recur. The hot code compiles to
+  native Go over unboxed ints (direct recursion, plain loops); only the leaves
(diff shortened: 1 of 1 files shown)


nooga:
Silly question, why are those not fully lowered?

nooga (approved):

mparrett:
@nooga nor silly at all!

Two missing pieces: lowering function values (#270) and native collection representations (#358). Until those land, the AOT build and the VM run the same code for those four rows.

fib/tak/loop-recur type all the way down to scalars (int/float locals, arithmetic, direct recursion), so gogen can emit straight Go. `(reduce + …)` instead applies a boxed `+` through `rt.InvokeValue` per element, with no static call target to turn into a direct Go call (#270), over the runtime's persistent/lazy structures, which have no unboxed native shape (#358). With the hot loop spending nearly all its time in runtime calls, the only thing left to emit is the native `run` wrapper around the same `rt.Reduce3` the VM dispatches to — which is why the 1.7x reduce row is wrapper overhead, not a lowering regression.

Issue #581, opened 2026-07-19
CI: coverage gaps, a 3× stdlib compile per PR, and gates documented as CI-run that aren't


mparrett:
An audit pass over CI and the local gate machinery on main @ 9dd835e turned up a cluster of maintenance items. A validation pass against current main @ 2be9033 found the relevant workflow, Makefile, docs, and hook files unchanged. Filing these together because they share one theme: what required CI actually runs has drifted from what the Makefile, docs, and hooks say it runs. Parented under #528.

## 1. Required cross-build gaps (js/wasm, plan9, -race scope)

`go.yml` gates `GOOS=wasip1` only. But:

- `GOOS=js GOARCH=wasm` — the target browser bundles ship — is built by `pages.yml` on push to main and by the opt-in `perf-wasm` PR workflow, but it is not a required PR gate. A normal PR can still break the js/wasm build and surface it post-merge, at deploy.
- plan9 (3 build-tagged files, device-validated support per #461) has no build gate anywhere.
- TinyGo has no lane on main, but main also documents it as unsupported. The proposed TinyGo build-and-boot gate belongs with the support work in #522/#504 rather than being treated as a missing gate for an already-supported target.

A build-only matrix step (`GOOS=js GOARCH=wasm`, `GOOS=plan9 GOARCH=amd64`) on the existing wasip1 job closes the two current supported-target gaps. Both cross-builds pass at 9dd835e; from cold build caches they took about 11s each locally, so the runner cost should remain small.

Also: `-race` covers `./pkg/vm/... ./pkg/rt/...` only. `pkg/nrepl` is a concurrent server surface outside those packages and never sees the race detector. Adding `./pkg/nrepl/...` passes at 9dd835e and took about 2.5s locally once its localhost socket tests were allowed to bind.

## 2. PR CI performs one avoidable extra full lowering pass

Three independent jobs compile the `.lg` stdlib from source:

- `build` — `go generate ./pkg/rt/` (bundle-only)
- `gogen-diff` — `make gogen-diff` via `make lowered` (`--target=go`)
- `generated` — `make check-generated` via `lgbgen --target=both`

The original 50–70 runner-minute estimate was wrong. Measured across four recent `go.yml` runs, total job time was 11.8, 12.6, 10.3, and 11.3 runner-minutes; CodeQL is separate. The jobs run in parallel, so the build job remains the roughly 4–6 minute wall-clock critical path.

[...]

PR #583, opened 2026-07-20
ci(perf-timeline): raise the macos-14 leg timeout to 150 min

mparrett:
The one-liner from the #573 discussion (also #581 item 6).

No macos-14 leg has ever finished under the 90-min timeout: every run since the leg landed (#363, 2026-07-06) ran the full 90 minutes and died at the kill — at count 4 and at count 3 alike ("The job has exceeded the maximum execution time of 1h30m0s" in the job annotations). So `perf-data` has zero arm64 snapshots — the Apple-Silicon timeline has never had a data point — and each push to main spends 90 macos runner-minutes producing nothing.

This raises the ceiling for the macos-14 leg only, via a matrix conditional; the ubuntu leg (61–83 min envelope at count 3) stays at 90. Count stays at 3, per the estimator floor from #561.

A completed fork probe measured the macos-14 leg at ~120.5 minutes end to end (118.6 minutes in the capture step) at count 3. A 150-minute ceiling therefore provides about 25% headroom — proportionally more than ubuntu's 90-minute ceiling over its 83-minute worst case. If macos runner supply still makes ~2h legs flaky, the fallback discussed on #573 is a scheduled cadence for the arm64 leg instead of per-push.

### DIFF
--- a/.github/workflows/perf-timeline.yml
+++ b/.github/workflows/perf-timeline.yml
@@ -43,7 +43,13 @@ jobs:
         # explorer then shows both an amd64 and an Apple-M line per benchmark.
         os: [ubuntu-latest, macos-14]
     runs-on: ${{ matrix.os }}
-    timeout-minutes: 90
+    # The macos-14 leg needs more headroom: at count 3 the ubuntu leg fits
+    # in 61-83 min, but no macos leg has ever finished under 90 — every one
+    # since the leg landed (#363) died at the kill, so perf-data has zero
+    # arm64 snapshots. 150 is a provisional ceiling until the first
+    # completed leg measures the real envelope; right-size from data then.
+    # Evidence + discussion: #573.
+    timeout-minutes: ${{ matrix.os == 'macos-14' && 150 || 90 }}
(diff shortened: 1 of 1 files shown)


mparrett:
The fork probe (`probe/timeline-macos-leg-timing`, Perf Timeline workflow), [completed](https://github.com/mparrett/let-go/actions/runs/29711724655) early this morning — and it validates #583: the macos-14 leg ran **success in ~120.5 minutes** (02:56:58 → 04:57:30). That's comfortably past the old 90-min timeout that was killing it since #363, and inside the new 150-min budget with ~30 min of headroom.

So the caveat is cleared — #583 is green, mergeable, and now probe-validated. It just needs a review/approval to land.

nooga (approved):
Reviewed the diff and surrounding perf-timeline.yml context. The matrix-conditional `timeout-minutes` expression is correct, and the core timing claim is independently confirmed by the fork-probe run in your own comment (~120.5 min, comfortably inside the new 150-min budget). No issues.

PR #588, opened 2026-07-20
ci(release): tolerate unresolvable gogen_ir test imports in go mod tidy

nooga:
## Summary
- The v1.12.0 release job failed: https://github.com/nooga/let-go/actions/runs/29749951271/job/88377503689
- Root cause: `pkg/wasmhost/zz_gogen_ir_wire_test.go` (new since v1.11.1, gated behind `//go:build gogen_ir`) blank-imports `pkg/rt/core_go_lowered/...`, a gitignored tree that's only materialized by `make` targets under `-tags gogen_ir` (see `docs/regenerating-generated-artifacts.md`). It doesn't exist on a fresh checkout.
- `go mod tidy` scans test imports across all build-tag combinations regardless of which tag is active, so it tries to resolve those paths as external modules and fails with `no matching versions for query "latest"` — killing the whole `.goreleaser.yml` `before.hooks` chain.
- Fix: `go mod tidy` → `go mod tidy -e`, which reports but doesn't fail on packages it can't load, while still tidying everything resolvable.

## Test plan
- [x] Reproduced the failure in a fresh clone (`go mod tidy` fails identically outside CI)
- [x] Verified `go mod tidy -e` exits 0 in the same fresh clone without touching go.mod/go.sum
- [x] Verified the rest of the release hook chain still passes afterward: `go generate ./pkg/rt/` and `go test -short -timeout 60s ./...`


.goreleaser.yml
```diff
@@ -4,7 +4,13 @@ project_name: let-go
...
 before:
   hooks:
-    - go mod tidy
+    # -e: pkg/wasmhost's gogen_ir-tagged test file blank-imports
+    # pkg/rt/core_go_lowered/..., a gitignored tree only materialized by
+    # `make` targets under -tags gogen_ir (see docs/regenerating-generated-artifacts.md).
+    # It doesn't exist in a fresh checkout, so tidy's build-tag-agnostic
+    # import scan can't resolve it and errors; -e reports that but still
+    # tidies the real dependencies and exits 0.
+    - go mod tidy -e
```
mparrett: **[P2] Preserve strict module validation**

Changing this hook to `go mod tidy -e` suppresses every package-loading failure, not only the expected generated-tree imports. I verified that an unrelated nonexistent build-tagged import also exits 0, while the normal short release tests do not compile that tag.

The root cause is the accidentally tracked `pkg/wasmhost/zz_gogen_ir_wire_test.go`: it is already gitignored, generated by `lgbgen`, removed by `make clean-lowered`, and documented as an uncommitted artifact. Deleting that generated file from version control makes strict `go mod tidy` pass on a clean checkout without modifying `go.mod` or `go.sum`; `make lowered` recreates it before `gogen_ir` consumers run. Please remove the tracked generated file and retain strict `go mod tidy` instead of weakening this release check globally.

mparrett (changes requested):
The clean-worktree release chain passes with this patch, but this suppresses unrelated package-loading failures too. There is a targeted fix that preserves strict release-time module validation; details inline.

nooga:
Good catch, thanks — you're right, this is a better fix. Pushed:

- Reverted the `.goreleaser.yml` change (back to plain `go mod tidy`).
- Untracked `pkg/wasmhost/zz_gogen_ir_wire_test.go` (`git rm --cached`) — it's generated by `lgbgen`, cleaned by `make clean-lowered`, and already listed in `.gitignore`, but got `git add`-ed back in #379 and has been tracked ever since. That's the actual regression vs v1.11.1.
- Verified in a fresh clone of the branch: plain `go mod tidy` now exits 0 without touching go.mod/go.sum, and `go generate ./pkg/rt/` + `go test -short -timeout 60s ./...` still pass.

`make lowered` will regenerate the file locally for anyone working the `gogen_ir` path.

mparrett (approved):
Changes look good- thanks!

- The requested targeted fix was implemented exactly.
- Strict `go mod tidy` passes without modifying module files.
- The complete release hook passes locally.
- All CI checks are green, including `generated-artifacts`, `gogen-diff`, compatibility suites, and expensive lowering tests.
- No new issues as a result of changes.

PR #595, opened 2026-07-20
feat(wasm): SPSC ring buffer + held-key coalesce for SAB key input

mparrett:
The wasm bundle's key input is a single SAB slot: while the worker is mid-turn with a key pending, `LetGoHost.sendInput` drops every further press (oldest-pending-wins, per the comment in `lg-host-core.js`). For interactive programs that means fast typing and held keys mostly vanish — in a game bundle driving a burst of 10 presses into a busy worker, 0 of 10 were accepted. The slower the program's read loop, the worse it gets.

This replaces the slot with an 8-slot SPSC ring in the same SharedArrayBuffer. The producer writes `slot[writeIdx % 8]`, publishes with `Atomics.store` + `notify`, and never blocks the main thread — it returns false only when the ring is full. The consumer (`HostKeySource.ReadKey`) parks on `Atomics.wait(writeIdx)`, copies the slot before advancing `readIdx` (so a full-ring producer can't overwrite mid-read), then lazily coalesces: adjacent queued slots with identical bytes drain in a single advance, so a held key that outruns the read loop becomes one keypress instead of a backlog of queued turns. Distinct keys are never merged and nothing is reordered.

Same burst against this branch: 7 of 10 accepted (ring capacity while one is in flight), producer wall time for the whole burst 0.07 ms. An earlier measurement of the same design on the pre-#245 host layout also had the consumer side: 12 render ticks for a 10-press held-key burst down to 3.

Layout notes: the ring lives in the existing SAB (256 bytes) — cells 0/1 are read/write indices, slot lengths at byte 64, slot bytes (16 per key) at 72; the protocol comment is duplicated in `keysource_js_wasm.go` and `lg-host-core.js` with a keep-in-sync warning; the two sides ship together in a bundle, so there is no cross-version concern. `key-pending?` keeps its semantics (non-consuming, true iff the ring is non-empty). Goldens regenerated via `go test ./pkg/rt/wasm -update`.

The one behavior change to weigh in review: coalescing means N held-key repeats can execute as one read rather than N. For read-key-driven programs that's the point (input stays live instead of replaying a stale backlog), but if you'd rather ship the ring without coalesce and leave merging to programs, that's a clean 20-line subtraction — the ring stands on its own.

### DIFF
--- a/pkg/rt/keysource_js_wasm.go
+++ b/pkg/rt/keysource_js_wasm.go
@@ -10,8 +10,23 @@
  * — installed at the *keys* root by term_wasm.go's install, resolving the JS
  * globals per call so SAB setup order doesn't matter.
  *
- * SAB layout (Int32Array): [0] key-ready flag, [1] byte count; Uint8Array at
- * offset 8, length 16, holds the key bytes. See pkg/rt/wasm/lg-host.js.
+ * SAB protocol — SPSC ring buffer between the main thread (producer) and
+ * this worker (consumer). Layout MUST stay in sync with the producer in
+ * pkg/rt/wasm/lg-host-core.js; the consts below mirror the JS side. The
+ * producer writes slot [writeIdx % keyCapacity] then advances writeIdx; the
+ * consumer waits on writeIdx, reads slot [readIdx % keyCapacity], then
+ * advances readIdx. The producer never blocks the main thread — when the
+ * ring is full it drops.
+ *
+ * Int32Array view (used cells):
+ *
+ *	[0]  readIdx   — consumer increments after reading a slot
+ *	[1]  writeIdx  — producer increments after writing a slot
+ *
+ * Uint8Array view (full SAB):
+ *
+ *	bytes 64..71   slot lengths (1 byte per slot, 8 slots)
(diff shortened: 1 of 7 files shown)


nooga (approved):
Traced the SPSC ring buffer protocol carefully (ordering, happens-before via Atomics, fullness-check races, coalesce-loop bounds) — no correctness bugs found. Good documentation of a genuinely non-trivial lock-free protocol. Approving; the coalesce-behavior tradeoff you flagged is a product call rather than a defect.

mparrett:
Thanks for the quick review/approval. I've had this knocking around on my local branch for some time. I'll share a playable A/B of xsofy wasm before merging this- it's a difference that must be felt. I also want to make sure this is applicable in the wider scope. :)

nnunley (approved):
Reviewed via multi-agent lens pass with special focus on the SPSC-over-SharedArrayBuffer concurrency. Verified sound: Atomics memory ordering, fullness check (w - r >= CAPACITY), slot-protection invariant, wraparound arithmetic, and the held-key coalesce peek-ahead (writeIdx snapshot refreshed each iteration). Suggestions only (E2E burst test, dropped-key telemetry) — non-blocking. LGTM.

PR #600, opened 2026-07-21
fix(compiler): skip no-value top-level forms (comments compiled to dead bytecode)

mparrett:
While diffing disassembled core bundles between commits, ~650 `LOAD_CONST`/`POP` pairs loading the VOID sentinel turned up — one per top-level comment in the embedded core sources.

The reader returns VOID for comments, `#_` discards, and reader conditionals with no matching branch. Every read loop filters it — collections via `appendNonVoid`, `read-string` via `ReadSkipNoValue`, lgbgen's defn scan — except `CompileMultiple`, which treated it as a real form: pool the constant, emit the pair, and run it through a frame at compile time. Two consequences:

- The shipped core bundle carried 655 dead pairs, replayed on every boot. The regenerated bundle drops from 264,072 to 258,947 bytes (−1.9%).
- A trailing comment clobbered the last value: `(load-string "42 ;; done")` returned VOID rather than 42. Clojure returns 42, so this is a behavior change in the conformance direction, noted in case anything relied on the old result.

One deliberate choice: input holding only no-value forms still evaluates to VOID, not nil, so the REPL keeps echoing nothing for a comment-only line. Tests cover the trailing-comment value, the comment-only sentinel, and emitted-code-size equality with and without interleaved comments.


nnunley (approved):
Reviewed via multi-agent lens pass with per-finding verification. VOID-form skipping is sound: VoidType is a singleton pointer compare, only the reader produces VOID (no false positives), trailing-comment/comment-only/empty-input semantics all verified, and the dead LOAD_CONST(VOID)/POP is correctly eliminated. LGTM.

PR #602, opened 2026-07-21
feat(api): IsIncomplete — sanctioned incomplete-input check for REPL loops

### LINKED ISSUE #587: api: no supported way to detect incomplete input — embedder REPLs string-match error text
## Summary

A REPL loop needs to distinguish "this form isn't finished yet, keep reading lines" from "this input is wrong, report it". let-go has the information: `compiler.IsErrorEOF` reports whether an error's (possibly nested) cause is `io.EOF`. But nothing on the `pkg/api` surface exposes or documents it for this purpose, so embedders building interactive frontends are left inspecting error strings.

## What the errors look like today

Against current `main`:

```go
_, err := lg.Run("(defn foo [x]")
// "Syntax error reading source at (<default>:1:14). ... caused by EOF"
compiler.IsErrorEOF(err)  // true — genuinely incomplete, keep reading

_, err = lg.Run("(]")
// "Syntax error reading source at (<default>:1:3). ... unmatched delimiter ]"
compiler.IsErrorEOF(err)  // false — real error, report now
```

The predicate works and the two cases are distinguishable. But it lives in `pkg/compiler`, is documented as a reader-internals note, and nothing tells an api user it's the sanctioned check.

## What embedders do instead

govertone (https://github.com/jasalt/govertone), which embeds let-go v1.11.1 as the control language of a live-coding music runtime, implements its multi-line REPL prompt like this ([`internal/lisp/runtime.go`](https://github.com/jasalt/govertone/blob/master/internal/lisp/runtime.go)):

```go
func incomplete(err error) bool {
    s := strings.ToLower(err.Error())
    return strings.Contains(s, "eof") || strings.Contains(s, "unmatched")
}
```

This is fragile in both directions: any rewording of reader messages breaks it, and matching "unmatched" misclassifies a hard delimiter error as awaiting-more-input, so `(]` leaves the prompt silently accumulating lines that can never complete. An nREPL server has the same need (clients expect a structured "incomplete" status rather than a parse error), so this isn't specific to one host.

## Proposal

1. Export the predicate on the api surface — `api.IsIncomplete(err) bool`, defined as "the reader consumed all input while inside an open form" — delegating to the existing `compiler.IsErrorEOF`.
2. Document it in the embedding docs next to `Run`, with the REPL loop as the worked example, and note explicitly that unmatched-delimiter errors are *not* incomplete input.

The mechanism already exists and is tested; the change is mostly exporting and documenting it. Can put up the PR directly if that helps.

Candidate for the v1.12.0 requirements list (#545).

mparrett:
Closes #587.

One exported function, `api.IsIncomplete(err) bool`, delegating to `compiler.IsErrorEOF` — which already makes the right distinction: input that ended inside an open form has an `io.EOF` cause, an unmatched delimiter does not. The doc comment carries the worked REPL loop (accumulate lines while incomplete, report otherwise), which is the contract the test pins: open form and open string are incomplete, `(]` is not, and the accumulate-and-retry loop converges.

One behavior surfaced while pinning the contract: comment-only input evaluates cleanly (no error), so only empty/whitespace-only input reads as incomplete alongside open forms. The doc comment states that.

No behavior change anywhere else — this exports and documents an existing predicate at the surface embedders see.

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -237,3 +237,24 @@ func (l *LetGo) Run(expr string) (vm.Value, error) {
 	vm.ReleaseFrame(frame)
 	return result, err
 }
+
+// IsIncomplete reports whether err, as returned by Run, means the input
+// ended in the middle of an open form (e.g. "(defn foo [x]") rather than
+// being wrong. Interactive frontends use it to decide between reading
+// more lines and reporting the error:
+//
+//	v, err := lg.Run(buffer)
+//	if err != nil {
+//		if api.IsIncomplete(err) {
+//			continue // prompt for another line, re-Run the whole buffer
+//		}
+//		report(err)
+//	}
+//
+// Empty and whitespace-only input also read as incomplete, which a REPL
+// loop treats the same way: keep prompting. A hard syntax error such as
+// an unmatched delimiter ("(]") is not incomplete — more input can never
+// complete it — and should be reported immediately.
+func IsIncomplete(err error) bool {
+	return compiler.IsErrorEOF(err)
+}
--- a/pkg/api/incomplete_test.go
+++ b/pkg/api/incomplete_test.go
@@ -1,0 +1,50 @@
+package api_test
+
+import (
+	"testing"
+
+	"github.com/nooga/let-go/pkg/api"
+	"github.com/stretchr/testify/assert"
+)
+
+// TestIsIncomplete pins the REPL-loop contract: input that ends inside an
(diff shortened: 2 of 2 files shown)


nnunley (approved):
Reviewed via multi-agent lens pass with per-finding verification. Incomplete-vs-hard-error classification is correct across the cases that matter (unclosed paren/string, unmatched delimiters, empty/whitespace, comment-only, wrapped errors) — all delegate cleanly to compiler.IsErrorEOF. LGTM.

PR #603, opened 2026-07-21
feat(api): namespace facade — lg.NS with Def/DefShadowing

### LINKED ISSUE #593: api: no namespace management — organizing host functions into namespaces requires `pkg/rt` + `pkg/vm` internals
## Summary

`api.LetGo.Def` defines into the instance's current namespace, and that is the whole namespace surface the api offers. Everything past it — defining into a different namespace, creating one, or shadowing a core-referred name cleanly — means importing `pkg/rt` and `pkg/vm` and calling `rt.NS(...)` plus `vm.Namespace` methods directly. Registering host functions in organized namespaces is the first thing a non-trivial embedder does, so the internals imports show up immediately.

The shadowing case is subtle: `Def` over a name referred in from `clojure.core` works (the local var wins) but emits the Clojure-parity shadow warning; the quiet, sanctioned path is `Exclude`-then-`Def`, matching `:refer-clojure :exclude`. Neither `Exclude` nor a target-namespace `Def` is reachable from `pkg/api`.

## What an embedder's registration code looks like today

govertone (https://github.com/jasalt/govertone), which embeds let-go v1.11.1 as the control language of a live-coding music runtime, installs a synth-patch DSL whose unit names collide with core (`filter` is a Sointu filter unit; it also defines its own `now`). Its registration loop ([`internal/lisp/patch_bindings.go`](https://github.com/jasalt/govertone/blob/master/internal/lisp/patch_bindings.go)) is therefore:

```go
core := rt.NS("music.core")
patchNS := rt.NS("music.patch")
for name, f := range bindings {
    native, err := vm.NativeFnType.Wrap(f)
    ...
    core.Exclude(name)
    patchNS.Exclude(name)
    core.Def(name, native)
    patchNS.Def(name, native)
}
```

Two internal packages plus `api` to register a function table — and the `Wrap`/`BoxValue` boxing that `api.Def` does for the caller has to be redone by hand on this path.

## Proposal

A small namespace facade on the api, wrapping what embedders already reach for:

```go
ns := lg.NS("music.patch")        // resolve or create
ns.Def("filter", goFunc)          // boxes like api.Def does
ns.DefShadowing("now", goFunc)    // Exclude + Def, no warning
```

Three methods cover the observed need; `Refer`/aliasing could follow if asked for. With it, the most common embedding act stops depending on `pkg/rt`/`pkg/vm` internals, freeing those to change shape without breaking embedders. Small enough that I can put up the PR with the discussion.

Related: #585/#586/#587 came out of the same embedding review; this one is the reason an embedder's imports don't stay inside `pkg/api` past the first hundred lines.

Candidate for the v1.12.0 requirements list (#545).

mparrett:
Closes #593.

The three methods from the issue, nothing more: `lg.NS(name)` resolves or creates a namespace (it's `rt.NS` underneath, which is already resolve-or-create) and returns a handle; `Def` boxes with the same `vm.BoxValue` path as `LetGo.Def` and defines into that namespace; `DefShadowing` is the programmatic `(:refer-clojure :exclude [name])` — `Exclude` then `Def`, so an intentional core-name collision is silent and unqualified references in the namespace resolve to the host definition.

The tests pin the three stories: qualified calls into a dedicated host-function namespace, shadowing `filter` (host def wins unqualified, `clojure.core/filter` stays reachable qualified), and resolve-or-create on a never-mentioned name.

With this an embedder's registration code stays inside `pkg/api` — the motivating example in #593 needed `rt`, `vm`, and hand-rolled `NativeFnType.Wrap` boxing for the same job. `Refer`/aliasing are left out per the issue; can follow if there's a use.

### DIFF
--- a/pkg/api/ns.go
+++ b/pkg/api/ns.go
@@ -43,7 +43,17 @@ func (n *NS) Def(name string, value any) error {
 // (:refer-clojure :exclude [name]). The definition is silent and
 // unqualified references in the namespace resolve to it instead of the
 // core var.
+//
+// Boxing happens before any namespace mutation: if value is unboxable the
+// namespace is left untouched, so a failed call is a true no-op rather than
+// leaving name excluded-but-undefined (which would silently strip the
+// unqualified clojure.core/<name> resolution).
 func (n *NS) DefShadowing(name string, value any) error {
+	val, err := vm.BoxValue(reflect.ValueOf(value))
+	if err != nil {
+		return err
+	}
 	n.ns.Exclude(name)
-	return n.Def(name, value)
+	n.ns.Def(name, val)
+	return nil
 }
--- a/pkg/api/ns.go
+++ b/pkg/api/ns.go
@@ -1,0 +1,49 @@
+package api
+
+import (
+	"reflect"
+
+	"github.com/nooga/let-go/pkg/rt"
+	"github.com/nooga/let-go/pkg/vm"
+)
+
+// NS is a handle for registering host values into one namespace. It is
+// the namespace-targeted counterpart of LetGo.Def: same boxing, explicit
+// destination. Obtain one from LetGo.NS.
+type NS struct {
+	ns *vm.Namespace
+}
+
+// NS resolves the named namespace, creating it if it does not exist yet,
+// and returns a registration handle for it. Typical embedder use is
+// grouping host functions into their own namespaces at startup:
+//
+//	patch := lg.NS("music.patch")
+//	patch.Def("oscillator", oscillatorFn)
+//	patch.DefShadowing("filter", filterFn) // collides with clojure.core/filter
+func (l *LetGo) NS(name string) *NS {
+	return &NS{ns: rt.NS(name)}
+}
+
+// Def boxes value exactly as LetGo.Def does and defines it in this
+// namespace. Defining a name that is referred in from clojure.core works
+// (the local var wins) but emits the Clojure-parity shadow warning; use
+// DefShadowing when the collision is intentional.
+func (n *NS) Def(name string, value any) error {
+	val, err := vm.BoxValue(reflect.ValueOf(value))
+	if err != nil {
+		return err
+	}
+	n.ns.Def(name, val)
+	return nil
+}
+
+// DefShadowing defines name in this namespace, first excluding it from
+// the clojure.core auto-refer — the programmatic equivalent of
+// (:refer-clojure :exclude [name]). The definition is silent and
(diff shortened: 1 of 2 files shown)


pkg/api/ns.go
```diff
@@ -0,0 +1,49 @@
...
+	return nil
+}
+
+// DefShadowing defines name in this namespace, first excluding it from
+// the clojure.core auto-refer — the programmatic equivalent of
+// (:refer-clojure :exclude [name]). The definition is silent and
+// unqualified references in the namespace resolve to it instead of the
+// core var.
+func (n *NS) DefShadowing(name string, value any) error {
+	n.ns.Exclude(name)
```
nnunley, line 56: `DefShadowing` isn't atomic: `Exclude(name)` runs before `Def(name, value)`, and `Def` can fail (`vm.BoxValue` returns an error for unboxable values). On that path the name is left **excluded-but-undefined** — the caller gets an error and assumes the whole op was a no-op, but the exclusion has already taken effect, so an unqualified reference that previously resolved to `clojure.core/<name>` now resolves to nothing.

Suggest boxing first and only mutating the namespace once it succeeds, so a failure leaves the namespace untouched:

```go
func (n *NS) DefShadowing(name string, value any) error {
    val, err := vm.BoxValue(reflect.ValueOf(value))
    if err != nil {
        return err
    }
    n.ns.Exclude(name)
    n.ns.Def(name, val)
    return nil
}
```

(Or roll back the `Exclude` if you'd rather keep delegating to `Def`.) Minor, but worth it since the torn state is silent.

nnunley (commented):
One correctness note on `DefShadowing` (inline). Everything else here reads clean — nice, well-documented facade. Happy to approve once the ordering is tightened.

pkg/api/ns.go
mparrett: Fixed in 2a5c754. Boxing now happens before any namespace mutation, so an unboxable value returns the error before `Exclude`/`Def` run — a failed call is a true no-op instead of leaving the name excluded-but-undefined. Went with the inline box-first form rather than rolling back `Exclude`. Added a regression test: an unboxable value (`nil`) makes `DefShadowing` error, and unqualified `filter` still resolves to `clojure.core`.

nnunley (approved):
Reviewed — clean. The `NS` facade reuses `vm.BoxValue`/`ns.Def`/`ns.Exclude`; `DefShadowing` boxes before any namespace mutation, so an unboxable value is a true no-op (no excluded-but-undefined state) — the subtle correctness point, and it's handled. Self-contained, good coverage. LGTM.

Issue #607, opened 2026-07-21
AOT lower-vm: embedded VM-fallback bundle rejected by its own runtime — `unsupported capability mask 0x1 (supported: 0x0)`


nnunley:
## Summary

The AOT **lower-vm** path (AOT lowering + VM fallback for un-lowered code) can
produce a binary whose **embedded VM-fallback bundle fails to decode** against its
own **embedded runtime**:

```
error: unsupported capability mask 0x00000001 (supported: 0x00000000)
```

Reported by Ingy running [`gloat`](https://github.com/) `-E let-go-lower-vm` on
`fib.clj`. That mode **worked before** (14 MB binary, 0.027 s) — this is a
regression. The pure-VM mode (`-E let-go-vm`) still works (~0.43 s, on par with
`bb`), and the fully-pruned native mode (`-Xprune`) works (7.4 MB, 0.015 s). Only
lower-vm is broken.

```
$ gloat -fq -E let-go-lower-vm fib.clj && ls -lh fib && time ./fib
-rwxr-xr-x 1 ingy ingy 13M Jul 21 17:14 fib*
error: unsupported capability mask 0x00000001 (supported: 0x00000000)
```

## Root cause: producer/consumer capability skew

The error is raised in `readCapabilities` at
[`pkg/bytecode/decoder.go:203`](pkg/bytecode/decoder.go#L203) (mechanism added in
#443). Walking the two sides:

- **Producer** — the encoder stamps **every** v2 bundle with
  `Flags: FlagCapabilities, Capabilities: CapOpcodeSet`
  (`pkg/bytecode/encoder.go:284-285`). `CapOpcodeSet = 0x1` (`pkg/bytecode/tags.go:24`).
  So the written cap mask is `0x00000001`.
- **Consumer** — the current decoder declares
  `const supportedCaps = CapOpcodeSet` → `0x1` (`pkg/bytecode/decoder.go:201`),
  and rejects any bits outside that with the message above
  (`caps &^ supportedCaps != 0`).

The reported error says **`supported: 0x00000000`**. A decoder computing
`supportedCaps` from the current tree would report **`0x1`**, not `0x0`. So the
runtime linked into the lower-vm binary is from a build where **`supportedCaps == 0`**
— it predates `CapOpcodeSet` being *accepted* — while the embedded bundle was
**encoded by a newer `lg` that emits `CapOpcodeSet`**. The capability check is
doing its job (catching a skew instead of decoding shifted opcodes into UB); the
[...]

PR #608, opened 2026-07-21
feat(bytecode): actionable hint on capability-mask reject

mparrett:
Addresses the error-ergonomics item (question 3) in #607.

When the decoder rejects a bundle whose capability mask has bits the runtime doesn't support, the error now says what that means and what to do about it:

```
unsupported capability mask 0x00000001 (supported: 0x00000000) — bundle was produced by a newer lg than this runtime; recompile it with a matching lg or run it on a newer runtime
```

Previously it printed only the two masks, which is opaque at the point where users hit it: a pre-#443 runtime (v1.9.0–v1.11.1) handed a v1.12.0+ bundle — the skew behind #607. The opcode-set mismatch path one branch below already carries a "recompile the bundle with a matching lg" hint; this brings the capability reject in line with it.

Scope note: this only improves the message in future runtimes. The shipped v1.9–v1.11 decoders keep printing the terse form, so it changes nothing for the #607 report itself — that one is resolved on the producer side (see the issue).

Extended `TestUnsupportedCapabilityMask` to pin the hint text.

### DIFF
--- a/pkg/bytecode/bytecode_test.go
+++ b/pkg/bytecode/bytecode_test.go
@@ -1146,6 +1146,9 @@ func TestUnsupportedCapabilityMask(t *testing.T) {
 	if !strings.Contains(err.Error(), "unsupported capability mask") {
 		t.Fatalf("expected 'unsupported capability mask' in error, got: %v", err)
 	}
+	if !strings.Contains(err.Error(), "recompile it with a matching lg") {
+		t.Fatalf("expected the recompile hint in error, got: %v", err)
+	}
 }
 
 func TestSupportedCapabilityMask(t *testing.T) {
--- a/pkg/bytecode/decoder.go
+++ b/pkg/bytecode/decoder.go
@@ -200,7 +200,9 @@ func (d *decoder) readCapabilities() error {
 	}
 	const supportedCaps = CapOpcodeSet
 	if caps&^supportedCaps != 0 {
-		return fmt.Errorf("unsupported capability mask 0x%08x (supported: 0x%08x)", caps, supportedCaps)
+		return fmt.Errorf(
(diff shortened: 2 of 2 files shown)


pkg/bytecode/bytecode_test.go
```diff
@@ -1146,6 +1146,9 @@ func TestUnsupportedCapabilityMask(t *testing.T) {
 	if !strings.Contains(err.Error(), "unsupported capability mask") {
 		t.Fatalf("expected 'unsupported capability mask' in error, got: %v", err)
 	}
+	if !strings.Contains(err.Error(), "recompile it with a matching lg") {
```
nnunley, line 1149: Not a great error message yet. Still doesn't tell the user what the minimum version necessary. Or what the missing capabilities are.

nnunley (commented):
This probably need to be reworked a little. This is a user facing error, so they need a bit more guidance than to "try again with a different lg".

From a devx standpoint, we probably need a --version page that includes capabilities on the VM/IR/etc, along with an improved error message.

nnunley (approved):
Reviewed via multi-agent lens pass (logic, error-handling, contract, concurrency, security) with per-finding verification. Clean, focused change; message is accurate, actionable, and consistent with the existing opcode-set error path. LGTM.

nnunley:
So, the agents got ahead of me. Feel free to merge, but a issue should be opened to deal with the devx/end user experience concerns.

mparrett:
> So, the agents got ahead of me. Feel free to merge, but a issue should be opened to deal with the devx/end user experience concerns.

Good points about devx- happy to address in this PR.

mparrett:
@nnunley stepping back from folding the broader DX into this PR — shipping the hint as-is.

Follow-up (named capabilities in the reject + `lg -v` page) is draft #622, stacked here. I'll retarget its base to `main` after this merges.

PR #610, opened 2026-07-22
fix(vm): alias collision guard compares namespace NAMES, not pointers

nnunley:
The require-alias collision guard compared namespace pointers; two loads of the same ns can present distinct pointers, so the guard could miss a real collision (or spuriously fire). Compare namespace names instead.

### DIFF
--- a/pkg/vm/namespace.go
+++ b/pkg/vm/namespace.go
@@ -445,11 +445,17 @@ func (n *Namespace) ReferList(ns *Namespace, symbols []Symbol) {
 // short name for different libs, or (historically) cross-namespace alias-table
 // contamination during dependency loading. Rather than letting the last writer
 // silently win, warn so the conflict is visible. Re-aliasing to the SAME target
-// (idempotent ns reload) and first-time aliasing are silent.
+// (idempotent ns reload) and first-time aliasing are silent. Sameness is by
+// NAME, not pointer: a reload that re-creates the namespace object (e.g. the
+// jank suite benchmark re-running the whole corpus per iteration) is still the
+// idempotent case — pointer comparison made it warn "already points to X,
+// being repointed to X", and since `go test` merges the test binary's stderr
+// into its stdout, that spam landed between a benchmark's name and its
+// metrics line, making benchfmt drop the suite [bytecode] results entirely.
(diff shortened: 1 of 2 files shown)


mparrett:
Nice, clean fix — name equality is the right identity here (a reload re-creating the ns object is the idempotent case pointer comparison wrongly flagged), and it's safe under the lock since Name() is lock-free. One suggestion: a small regression test pinning the guard's behavior — same-name re-alias stays silent, different-name still warns — would lock this in against future churn. (The bundled -short hooks commit should drop out on rebase, so no concern there.)

mparrett:
Reviewed and ran it — this is correct and minimal, no blockers from me.

The load-bearing question is whether `Name()` is a stable, unique identity for a namespace, and it is: `name` is set once from the `NewNamespace` argument and is immutable, every caller passes the fully-qualified name, and the namespace registry keys on it. So two namespaces that share a `Name()` are the same logical namespace — name-equality is exactly the idempotent-reload case, and name-inequality is a real collision. I don't see a false-negative path, and the existing nil guards still cover the `.Name()` calls. The only behavior change is distinct-pointer + same-name going silent; same-pointer and different-name are unchanged, and the unconditional `n.aliases[alias] = target` still repoints to the fresh object on reload, which is what you want.

The tests are the right shape. The different-name positive control (and asserting the warning names both targets) is what keeps the fix from silencing the warning rather than narrowing it. I ran both, plus the full `pkg/vm` suite; all green.

Two optional notes, neither blocking:

1. `captureStderr` writes in `fn()`, then closes the pipe, then `io.Copy`s — with no concurrent reader, output past the pipe buffer (~64KB) would block the writer. Fine for one warning line, but worth a comment that it only holds for small output, or copy in a goroutine started before `fn()` if the helper gets reused.
2. The `os.Stderr` swap is safe because neither test opts into `t.Parallel()`, but that's implicit — a future parallel stderr-writing test in the package would race it.

Good catch on the pointer-vs-logical-identity distinction; the benchfmt-via-merged-stderr note in the comment makes the "why" easy to follow.

mparrett (approved):
Two small non-blocking notes to consider. LGTM.

PR #611, opened 2026-07-22
fix(rt): *out*/*err* handles defer to the CURRENT os.Stdout/os.Stderr per write

nnunley:
The `*out*`/`*err*` IOHandles captured `os.Stdout`/`os.Stderr` at install time. Tests silence output by swapping the package variable (e.g. BenchmarkClojureTestSuite); the captured value made those swaps no-ops, so lg prints leaked between `go test`'s benchmark name and its metrics line and broke benchfmt parsing in cmd/bench-ratchet. Writers now defer to the current package variable per write; Fd()/Sync() still target the original descriptor for term raw mode.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -96,7 +96,7 @@ repos:
       - id: go-test
         name: go test ./... -skip TestClojureTestSuite
         description: Run Go test suite excluding the slow Clojure compat suite (mirrors CI test step)
-        entry: go test -timeout 60s ./... -skip TestClojureTestSuite
+        entry: go test -short -timeout 60s ./... -skip TestClojureTestSuite
         language: system
         types: [go]
         pass_filenames: false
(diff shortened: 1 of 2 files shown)


mparrett:
The change itself looks right: writes re-read os.Stdout/os.Stderr, Fd()/Sync() keep the original file. Matches the benchfmt leak story.

Approving with expected rebase. Hooks commit can/should get dropped.

Extra checks: The change is aimed at the Go-test `os.Stdout = devnull` pattern. Other hosts don’t rely on that; wasm bypasses it entirely.

- **WASM:** Guest `*out*`/`*err*` get `SetRoot`’d to `HostWriter` in `rendermain.go` after install. The new `stdStreamWriter` only lives in the default root that wasm replaces; production output never goes through it.
- **Plan 9 / wasip1:** Same `installIOBuiltins` path as native. Writes follow current `os.Stdout`/`os.Stderr`; `File()` still holds the original descriptor, so term `Fd()`/`Sync()` / plan9’s `h.File() == os.Stdout` check keep the same shape as today.
- **Normal rebinding** (`with-out-str`, nREPL, `(binding [*out* …])`) never touched the package vars — unchanged.

mparrett (approved):
LGTM.

PR #613, opened 2026-07-22
perf(rt): direct-call natives for the hot clojure.core surface

nnunley:
Direct-links lowered Go->Go calls for the hot clojure.core native surface: the var-dispatch boundary stays overridable (with-redefs / ^:dynamic / ^:redef exempt), only the interior is direct-linked to native `pkg.GoFn(args)`.

Benchmarks (Apple M3): gogen `IRCompile` **-3.5% time / -5.6% allocs** (170.1k -> 160.5k, p=0.000). vm `VariadicInvoke` **-22% / -46% time** (an aliasing-correctness fix; +1 alloc).

Stacks on #579 and #580 — this branch includes them, so review after those land.

### DIFF
--- a/pkg/rt/lang.go
+++ b/pkg/rt/lang.go
@@ -409,7 +409,6 @@ func init() {
 	// alphabetically so every file has registered first.
 	installLangNS()
 	installNativeDirectNS()
-	registerBuiltinsModule()
 
 	// Generated primitives intentionally re-Def native versions over the
 	// bootstrap closures — silence the warn-on-core-shadow noise for this
@@ -418,6 +417,11 @@ func init() {
 	vm.SetSuppressShadowWarn(true)
 	RegisterGeneratedPrimitives()
 	vm.SetSuppressShadowWarn(false)
+
+	// Register builtins AFTER primitives so that builtin versions (with better
+	// signatures) override the generated primitive versions. This is the EPIC-012
+	// seam: hot clojure.core functions migrate from primitives to builtins.
+	registerBuiltinsModule()
 	// walk namespace is embedded via coreFS and will be loaded on demand
 }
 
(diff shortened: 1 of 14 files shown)


pkg/rt/lang.go
```diff
@@ -418,7 +417,12 @@ func init() {
 	vm.SetSuppressShadowWarn(true)
 	RegisterGeneratedPrimitives()
 	vm.SetSuppressShadowWarn(false)
-	// walk namespace is embedded via coreFS and will be loaded on demand
+
+	// Register builtins AFTER primitives so that builtin versions (with better
+	// signatures) override the generated primitive versions. This is the EPIC-012
+	// seam: hot clojure.core functions migrate from primitives to builtins.
+	registerBuiltinsModule()
```
mparrett, line 424: Nice!

mparrett:
Overall, this is a really nice PR: the direct-linking approach is carefully scoped, the var-override guard is thoughtful, and the commit messages/tests make the performance intent unusually easy to review. I found two correctness blockers worth fixing before merge:

1. `set-var-target-keys` only recognizes a `:load-var` as the target operand of `:set-var`, but `build-set!` always emits a Var-valued `:const` there. As a result, the new snapshot protection never applies to real `set!` forms. A strict-IR repro returns `[2 2]` while the normal compiler returns `[1 1]`:

```clojure
(def ^:dynamic *probe-v* 1)
(defn snapshot-probe []
  (let [x *probe-v*]
    (set! *probe-v* 2)
    [x x]))
```

`lower_go.lg` already handles both `:load-var` and Var-valued `:const`; the bytecode lowerer needs the same normalization.

2. Registering `builtins.Nth`/`Nth3` as the new AOT targets introduces VM/AOT semantic drift. In particular, `(nth nil 0)` returns nil through the canonical Var path but `builtins.Nth` errors; `(nth [1] :bad :missing)` throws canonically but `builtins.Nth3` returns `:missing`; and an empty lazy seq can return nil instead of throwing/using the supplied default because it is not forced before `First()`. Reusing the canonical implementation, or matching all of its validation and lazy-seq behavior, would keep the optimization transparent.

The existing short suites are green (`go test -short ./pkg/vm ./pkg/rt ./pkg/ir ./test`), so targeted parity tests for these cases would be valuable. Aside from those two issues, the direction and implementation look very good.

mparrett (changes requested):

mparrett:
Curious, what's the variance in the benchmark numbers?

nnunley:
Both blockers are fixed.

**1. `set-var-target-keys` Var-valued `:const` target.** `lower.lg`'s `var-key-of-ref` now normalizes both a `:load-var` ref and the Var-valued `:const` operand `build-set!` actually emits, matching `lower_go.lg`. So the snapshot protection now applies to real `set!` forms. Your repro is in the suite verbatim — `ir_lower_stack_discipline_test.lg` asserts `(let [x *snap-v*] (set! *snap-v* 2) [x x])` returns `[1 1]` on both paths.

While verifying this I found the same-fn guard wasn't the whole story: a **callee** can `set!` the var across a `:call`, which the same-fn scan can't see. That's fixed on the base PR (#579) — `load-var-crosses-effect?` materializes loads live across an effect on both backends, and CSE stops merging `:load-var` across an effect. `(let [v mv] (+ (mutate!) v))` now returns `101` on both paths.

**2. `Nth`/`Nth3` VM/AOT drift.** All three cases now match the canonical Var path:
- `(nth nil 0)` → `nil` (was: error). `EmptyList` still errors, as canonically.
- `(nth [1] :bad)` / `(nth [1] :bad :missing)` → non-integer index errors (was: `Nth3` silently returned the default).
- Empty lazy seq is realized before `First()`, so `(nth empty-lazy 0)` errors and `(nth empty-lazy 0 :missing)` returns `:missing` (was: `nil`).

`pkg/rt/builtins/nth_parity_test.go` pins each case against the canonical `rt.Nth`/`rt.Nth3` behavior. `go test ./pkg/vm ./pkg/rt ./pkg/ir ./test` is green, and `TestGogenAOTDiff` confirms the AOT targets match bytecode.

On the benchmark variance question: I'll re-run the sweep with repeats and post the spread separately rather than inline a single-shot number here.

mparrett:
Re-reviewed at 3ffc72c: both blockers resolved — set-var snapshot (via #579, now on main) and the Nth/Nth3 parity fix, with tests. The new native surface, the variadic aliasing fix, and the guard extension all look good. Code LGTM.

Caveat: this branch still carries #580's census commit, so it can't merge ahead of #580 — whose future*-binding (thunk-scoped strict census) issue is still open. Once that lands and this rebases, I think it's good to go (still curious about the benchmark variance too).

mparrett (approved):
Looks good to merge once 580 lands.

PR #609, opened 2026-07-22
fix(hooks): pass -short to the pre-push go-test gate (match CI)

nnunley:
The pre-push `go-test` hook ran `go test -timeout 60s ./...` **without `-short`**, unlike CI (`.github/workflows/go.yml`). So the expensive lowering harnesses (`TestLoweringDeterminism`, AOT-diff) — which self-skip under `-short` — ran and hit the 60s per-package timeout, blocking every push. Add `-short` to match CI; the slow harnesses stay covered by their dedicated lowering steps.

Verified: `go test -short -timeout 60s ./... -skip TestClojureTestSuite` runs test/e2e in ~20s, whole suite green.

PR #614, opened 2026-07-22
fix(lgbgen): stage the lowered tree in an OS tempdir, copy into place

nnunley:
`lgbgen --target=go` generated into an **in-tree** sibling (`pkg/rt/core_go_lowered.stage`) before the atomic swap. A `SIGKILL`ed run (e.g. a timed-out `make generate`) bypasses the `defer os.RemoveAll` and leaves that dir, which a later `go test/build ./...` then compiles and fails on (`undefined: builtins.Vec`, `rt.NotEq`) because it references natives from a different generation.

Generate into `os.MkdirTemp` (out of the module tree) and **copy** into `core_go_lowered/` — copy, not rename, since the tempdir may be on a different filesystem (EXDEV). The completeness sentinel (#549) is still written last, so a torn copy stays detectable. Nothing inside the module can be left half-written now.

Verified: `make generate` + `make check-generated` green; no in-tree stage dir created.

Stacks on #609 (the pre-push `-short` gate fix) — that `.pre-commit-config.yaml` change in the diff belongs to #609; review after it lands.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -191,9 +191,11 @@ pkg/rt/.core_compiled.lgb.committed
 # the untagged build and shipped bytecode binary never need them.
 pkg/rt/generated.provenance
 pkg/rt/core_go_lowered/
-# Staging sibling that --target=go generates into before the atomic swap into
-# core_go_lowered/. A successful install renames it away; ignore any residue
-# from an interrupted run so it never gets committed.
+# Legacy in-tree staging residue. --target=go now generates into an OS tempdir
+# (out of the module tree) and COPIES into core_go_lowered/, so no in-tree
+# staging dir is created and `go test/build ./...` can never compile a
+# half-written residue from a hard-killed run. This entry only ignores residue
+# from older checkouts / an interrupted pre-fix run so it never gets committed.
 pkg/rt/core_go_lowered.stage/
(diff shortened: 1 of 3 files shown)


mparrett:
**No P0–P2 findings.**

The fix matches the failure mode. Staging in `MkdirTemp` keeps a SIGKILL’d generate out of the module tree; install still drops the sentinel first, cleans, copies, then writes the sentinel last so a torn `core_go_lowered/` stays detectable. Copy (not rename) is the right call for cross-filesystem tempdirs. Mid-install inconsistency in the real tree is the same short window as before. This PR's win is the leftover `.stage/` that `go test ./...` was compiling.

[P3] Refresh install docs that still say “rename” — cmd/lgbgen/main.go (call-site comment ~“fast per-file renames”, and `installGeneratedTree`s godoc: “renamed into place” / “same-filesystem”). The code correctly copyFiles for EXDEV; leaving the old rename story will invite someone to flip it back.

**Test gap:** No automated check that staging is out-of-tree (hard to unit-test SIGKILL). Residual risk is low given the design.

mparrett (approved):
LGTM. Minor/optional: stale rename comments cleanup.

PR #615, opened 2026-07-22
fix(ir/lower-go): compute cross-package :needs-error? under the ns's own deftype/protocol registries

nnunley:
`collect-ns-exports` (the phase-0 cross-package registry builder) lowered each ns WITHOUT binding its `*deftype-ctors*`/`*protocol-methods*`/`*protocol-method-sigs*`, so the cross-pkg `:needs-error?` was a pre-devirtualization snapshot that could disagree with lower-ns-to-go's emitted signature — miscompiling a caller as `v, callErr = OneValueFn(...)` against a 1-value `(vm.Value)` signature (the #281 error-effect-across-SCC gap). Bind the same three registries in collect-ns-exports first, so the cross-package effect matches the emission.

Surfaces when deftype/defprotocol forwarding (next PR) devirtualizes `ops.LowerOp`'s protocol calls to error-free `recv.Method` calls. Verified: gogen_ir build clean, ir-stress +9 (2418/2429), deterministic.

### DIFF
--- a/pkg/rt/core/ir/passes/pipeline.lg
+++ b/pkg/rt/core/ir/passes/pipeline.lg
@@ -1513,31 +1513,51 @@
    are their own callable name — no `LG_` forwarding wrapper), so a cross-package
    call resolves to `pkg.Conj(ec,…)`. :refs are the cross-ns callees it references
    (same-ns calls filtered — those resolve intra-package with no import)."
-  (binding [*ns* (the-ns ns-sym)
-            ir.passes.typeinfer/*typeinfer-max-drains* nil
-            ;; Same pure map bound in lower-ns-to-go so the registry's :go-name and
-            ;; the emitted decl agree on a collision-resolved export name.
-            lower-go/*export-name-overrides* (lower-go/build-export-name-map forms)]
-    (let [cross   (atom {})
-          results (binding [lower-go/*cross-ns-vars-used* cross]
-                    (lower-defn-forms (order-defn-forms forms)))
-          exports (reduce
-                    (fn [m r]
-                      (let [base (when (:decl r)
-                                   (lower-go/registry-entry-from-result r ns-sym pkg-name))]
-                        (if base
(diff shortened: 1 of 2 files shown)


mparrett:
Change looks correct, but needs a rebase. Feel free to re-request when it's ready.

mparrett (approved):
LGTM.

mparrett:
Planning to close this unless you see a problem with the reasoning.

The change is already on main, carried in by #617's squash merge. Merging this PR now would contribute no source changes: the only delta is one line of `pkg/rt/generated.sums`, and that line is a regression, since it sets the digest to this branch's stale value.

It shows as approved and conflicting, which reads as "needs a rebase". That is why it stayed open.

To reproduce:

```
git fetch origin main
git fetch origin pull/615/head:pr615

# the two files this PR changes
git diff --name-only $(git merge-base origin/main pr615)...pr615
#  pkg/rt/core/ir/passes/pipeline.lg
#  pkg/rt/generated.sums

# the source one, compared against main
git diff origin/main pr615 -- pkg/rt/core/ir/passes/pipeline.lg
#  (empty — byte-identical)
```

The second command compares the two trees directly, so an empty result means main's copy already equals this branch's and a merge cannot change it. That leaves `generated.sums` as the whole contribution.

One note in case you go looking for the commit that superseded this: there is none with a matching patch-id. #617 was squash-merged, so the change arrived bundled with others. Git's "patch contents already upstream" verdict during a rebase comes from applying the patch and getting an empty result — content equivalence, not commit identity.

The cross-package `:needs-error?` fix is on main and working. This is bookkeeping on a duplicate PR, not a second look at the change.

nnunley:
Superseded: rebasing this branch onto current `main` produces an empty diff — the cross-package `:needs-error?` fix (deriving the ns's own deftype/protocol registries before lowering) is already present in `main` (pkg/rt/core/ir/passes/pipeline.lg). Closing; nothing left to merge.

PR #617, opened 2026-07-22
fix(lgbgen): forward defprotocol/deftype to lower-ns-to-go so ir.ops devirtualizes

nnunley:
Two correctness fixes in the lgbgen → `lower-ns-to-go` path, plus a small DRY refactor. Not a pure refactor — the branch name predates the fixes.

## 1. fix(lgbgen): forward `defprotocol`/`deftype` forms to `lower-ns-to-go` (deftype-native T9 driver gap)

`lower-ns-to-go` has captured `defprotocol`/`deftype` decl forms since the deftype-native work landed: it emits native interface/struct/ctor/method decls and binds `*deftype-ctors*` / `*protocol-methods*` so the namespace's own defns type `(->T ...)` as `[:dtype T]`, lower ctors to native calls, and devirtualize protocol-method calls on known-concrete receivers to direct `recv.Method(ec, ...)` calls.

But the lgbgen driver's form filter only forwarded `defn`/`defn-`/`defmulti`/`defmethod` — the decl forms never reached the pipeline, the registries bound to `nil`, and every ctor and protocol call silently took the boxed `rt.CachedVarFn` trampoline (ir.ops lowered with no type decls at all). `isTypeDeclForm` now forwards `defprotocol`/`deftype` alongside the others.

Result: the lowered `ir/ops` package gains the `Irop` interface, one struct + `OpInfer`/`OpLowerStmts` methods per op-type, native ctor calls, and direct method dispatch in `infer-op`/`lower-op`. Blast radius beyond `ir.ops`: `core.lg`'s `IFn`/`IDeref` protocols emit two (inert) interface decls; no other core ns has top-level decls.

## 2. fix(ir/lower-go): compute cross-package `:needs-error?` under the ns's own deftype/protocol registries

`collect-ns-exports` (the phase-0 cross-package registry builder) lowered each ns without binding its `*deftype-ctors*` / `*protocol-methods*` / `*protocol-method-sigs*`, so the cross-pkg `:needs-error?` was a pre-devirtualization snapshot that could disagree with the signature `lower-ns-to-go` actually emits — miscompiling a caller as `v, callErr = OneValueFn(...)`. Bind the same three registries here first.

## 3. refactor(lgbgen): extract `formHeadIs` for the form-triage predicates

`isDefnOnly` / `isMultimethodForm` / `isTypeDeclForm` each re-implemented the same "list whose head symbol is in a set" walk — `isTypeDeclForm` was the third copy. Collapse them onto one `formHeadIs(form, names...)` helper. Pure refactor in isolation; lgbgen output byte-identical.

## Verification

- New `TestLoweredOpsDevirtualized` — content-based over the committed artifact; red before fix 1.
- `ir-stress` 2395/2414 `:ok` buckets identical to baseline.
- `go build ./...` + `-tags gogen_ir` clean; `make check-generated` lockstep; `pkg/ir` 362 + `pkg/bytecode` + `test/` green.

### DIFF
--- a/pkg/ir/lowered_ops_devirt_test.go
+++ b/pkg/ir/lowered_ops_devirt_test.go
@@ -1,0 +1,74 @@
+/*
+ * Copyright (c) 2026 Norman Nunley, Jr <nnunley@gmail.com>
+ * Part of the let-go project; see CONTRIBUTORS for full list of authors.
+ * SPDX-License-Identifier: MIT
+ */
+
+package ir
+
+import (
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"testing"
+)
+
+// TestLoweredOpsDevirtualized guards the lgbgen driver half of the deftype
+// native lowering (T9): lower-ns-to-go captures defprotocol/deftype forms,
+// emits native interface/struct/method decls, and binds the registries that
+// devirtualize (->ConstOp) to a native constructor and (op-infer recv ...)
+// to a direct method call — but only if the driver FORWARDS those forms.
+// cmd/lgbgen's form filter used to keep only defn/defmulti forms, silently
+// starving the machinery: ir.ops lowered with every protocol call as a
+// boxed rt.CachedVarFn trampoline and no type decls at all.
+//
+// Content-based over the lowered --target=go artifact, so a filter regression
+// fails here without re-running lgbgen. That tree (pkg/rt/core_go_lowered/) is
+// gitignored and a clean checkout does not contain it — `go generate ./pkg/rt/`
+// only rebuilds core_compiled.lgb, not the --target=go tree — so this SKIPS
+// when the artifact is absent (the default -short CI lane and fresh clones)
+// rather than failing on ENOENT. It runs wherever the tree has been generated
+// (local dev after `make generate`); the same forwarding regression is also
+// caught in CI by the -tags gogen_ir wire build (undefined symbols) and by
+// check-generated.
+func TestLoweredOpsDevirtualized(t *testing.T) {
+	_, thisFile, _, ok := runtime.Caller(0)
+	if !ok {
+		t.Fatal("runtime.Caller failed")
+	}
+	repoRoot := filepath.Dir(filepath.Dir(filepath.Dir(thisFile)))
+	loweredOps := filepath.Join(repoRoot, "pkg", "rt", "core_go_lowered", "ir", "ops", "ops.go")
+	b, err := os.ReadFile(loweredOps)
+	if err != nil {
+		if os.IsNotExist(err) {
+			t.Skip("lowered --target=go tree absent (gitignored; run `make generate`); " +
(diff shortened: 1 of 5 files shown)


pkg/ir/lowered_ops_devirt_test.go
```diff
@@ -0,0 +1,63 @@
...
+// Content-based over the committed artifact (like check-generated), so a
+// filter regression fails here without needing to run lgbgen.
+func TestLoweredOpsDevirtualized(t *testing.T) {
+	_, thisFile, _, ok := runtime.Caller(0)
+	if !ok {
+		t.Fatal("runtime.Caller failed")
+	}
+	repoRoot := filepath.Dir(filepath.Dir(filepath.Dir(thisFile)))
+	loweredOps := filepath.Join(repoRoot, "pkg", "rt", "core_go_lowered", "ir", "ops", "ops.go")
+	b, err := os.ReadFile(loweredOps)
```
mparrett, line 42: [P1] Avoid requiring the gitignored lowered tree in the default test lane

`pkg/rt/core_go_lowered/` is intentionally gitignored and a clean checkout does not contain this file. The build job runs `go generate ./pkg/rt/` before `go test -short ./...`, but that directive regenerates only `core_compiled.lgb`, not the `--target=go` tree, so this read fails with `ENOENT` on every clean checkout (and is the current CI failure). Please move this assertion into a lane that first runs `lgbgen --target=go`, guard it with the appropriate build tag/skip, or otherwise generate the artifact before reading it.

mparrett (changes requested):
The new regression test breaks the normal clean-checkout test lane because it assumes the gitignored lowered tree already exists.

mparrett:
Overall the implementation looks sound. The generated Go tree compiled, the targeted regression test passed after generation, and the generated-artifact gate passed.

pkg/ir/lowered_ops_devirt_test.go
nnunley: The test now skips cleanly when the lowered tree is absent rather than failing: `os.ReadFile` on the artifact, and `t.Skip` when `os.IsNotExist(err)`. So the default clean-checkout lane no longer requires `pkg/rt/core_go_lowered/` — it exercises the assertion only in a lane that has already run `lgbgen --target=go`, and is a no-op otherwise. Corrected the stale "committed artifact" comment too.

mparrett: Skip looks right — resolving.

mparrett:
Title/body still describe a pure `formHeadIs` refactor; the real payload is the deftype/defprotocol forwarding + cross-pkg `:needs-error?` fixes. Needs a rebase onto `main` (regen `generated.sums`, keep main's ir-stress baseline; the hooks `-short` commit is already on main).

mparrett (approved):
Everything addressed. LGTM.

Nit: Title/body still say “pure formHeadIs refactor”

PR #612, opened 2026-07-22
perf(ir): catalog-driven op dispatch — generated OpByKeyword string switch (§1/§2/§2b)

nnunley:
Catalog-driven IR op dispatch: `local-carrying?` as a catalog column (§1), reified infer facets + `:try` lower re-home (§2/§4/§5), and a generated `opByKeywordExact` string switch (§2b).

Benchmark (Apple M3): canonical `OpByKeyword` lookup **−96% time** (57ns → 2.1ns, ~27x) and **alloc-free** (1 → 0), matching TestOpByKeywordCanonicalIsAllocFree; non-canonical falls through to the existing lenient scan.

PR #618, opened 2026-07-22
feat(vm): OP_CALL_SELF for non-tail self-recursive defn calls

mparrett:
## Summary

- Adds `OP_CALL_SELF` for eligible non-tail self-calls in direct, lexical-scope-free top-level `defn` bodies.
- Encodes the defining Var's constant index as a runtime guard. Direct invocation is used only while the Var still selects the current chunk; `def`, `alter-var-root`, `with-redefs`, metadata wrappers, and other overrides fall back to ordinary `ExecContext` invocation.
- Borrows the caller's argument slice for the synchronous child call instead of allocating and copying it. Tail self-calls continue to use `LOAD_VAR` + `OP_TAIL_CALL`.
- Keeps lexically nested/capturing defs, nested lambdas, variadic functions, multi-arity cross-arity calls, and general top-level direct calls out of scope.
- Adds regression coverage for Var redefinition, late-discovered closure captures, and functions nested inside non-function `def` initializers.
- Includes the identity `CapOpcodeSet` migration for the pre-`CALL_SELF` signature and regenerates `core_compiled.lgb`.

Naive Fibonacci uses guarded `CALL_SELF` instructions rather than `LOAD_VAR` + generic `INVOKE` at each recursive arm.

## Rebase onto current `main`

The measurements below were re-run from scratch on `862c7cec`. The previous numbers came from `005a982`, which predates #613's direct-call natives, #612's catalog op dispatch, and #579's var-load change. That is enough movement in the call path that the old figures no longer described this branch.

The rebase turned up one interaction. #579's lowering ratchet asserts the plain compiler and the IR path emit the same number of `LOAD_VAR` ops for `fib`. `OP_CALL_SELF` doesn't load a callee, so `fib-bc` now emits 0 `LOAD_VAR` and 2 `CALL_SELF` against the IR path's 2, and the gate read the saved lookup as an IR regression. `622206e6` counts `CALL_SELF` alongside `LOAD_VAR` on the plain side; the invariant the test pins is unchanged.

The frozen `preCallSelfOpcodeNames` list still matches `main`'s live opcode table exactly (44 names, ending at `FINALLY_END`), so the migration key is still the right one after the rebase.

## A/B on current `main`

Apple M2 (`darwin/arm64`, `GOMAXPROCS=1`), precompiled test binaries, three strict `A → B → B → A` blocks at `-benchtime=1s -count=3`, n=18 per side, analyzed with `benchstat`.

| Workload | Entry mode | vs base |
| --- | --- | ---: |
| non-tail-fib-20 | `new-frame-run` | **9.11% faster** (p=0.000) |
| non-tail-fib-20 | `func-invoke` | **9.92% faster** (p=0.000) |
| direct-call-loop | `new-frame-run` | no established difference (p=0.389) |
| direct-call-loop | `func-invoke` | 1.24% faster (p=0.004) |
| closure-call-loop | `new-frame-run` | no established difference (p=1.000) |
| closure-call-loop | `func-invoke` | no established difference (p=0.481) |

The Fibonacci gain reproduces what the pre-rebase branch measured (9.26% and 9.57%), so it survives the intervening call-path work. Shallow calls move by at most 1.2%, in the faster direction, which fits the eligibility guard: nothing outside a top-level self-recursive `defn` changes shape.

Allocations are identical on both sides, at 280 B/op and 34 allocs/op for `non-tail-fib-20` through both entry modes. The earlier per-recursive-call allocation regression stays gone.

## Bundle compatibility

Checked both directions with raw `lg -c` bundles:

- A `.lgb` compiled by pre-`CALL_SELF` `lg` runs correctly under this
branch's runtime, through the identity migration.
- A `.lgb` compiled by this branch is rejected by a pre-`CALL_SELF` runtime
with `opcode set mismatch: bundle compiled with 45 opcodes … runtime has 44 … recompile the bundle with a matching lg`.

## Relationship to #619, #620, and #642

- #619 (frame reuse per bytecode call tree) is closed rather than merged.
Follow-up measurement put its standalone gain at 2% to 3%, and the combined matrix that used to live in this section described a branch that is no longer going anywhere, so it has been dropped.
- #620 proposes general tail-call elimination. It does not make this PR
redundant: #620 leaves non-tail recursion out of scope, which is what `OP_CALL_SELF` targets. The implementations overlap in VM call dispatch and should share resolver machinery where practical.
- #642 gives function-level `recur` its own op. It touches the adjacent
seam, self-reference emission, but on the tail path, so the two are disjoint today. Whichever lands second rebases on the first.

## Validation

- [x] `go test -short ./...` on the rebased head
- [x] `make check-generated`
- [x] Builds for linux, js/wasm, and plan9/amd64
- [x] Regression tests for redefinition, lexical closure capture, and non-function def initializers
- [x] Bytecode assertions confirm eligible recursive arms emit `CALL_SELF`
- [x] Interleaved A/B on `862c7cec`, n=18 per side, analyzed with `benchstat`
- [x] Allocation parity confirmed with `-benchmem`
- [x] Old-bundle load and new-bundle reject both exercised end to end


mparrett:
Please see the below comment on 620, and the note on 619 for the final determination (cancel 619, keep 618):

- https://github.com/nooga/let-go/issues/620#issuecomment-5051816487

- https://github.com/nooga/let-go/pull/619#:~:text=6.83%25%20slower%20for%20direct%20bytecode%20calls%20and%209.98%25%20slower%20for%20closure%20calls

nnunley (approved):

mparrett:
Rebased onto `main` (`862c7cec`). One commit landed after your approval, so it hasn't been reviewed.

#579's lowering ratchet asserts the plain compiler and the IR path emit the same number of `LOAD_VAR` ops for `fib`. `OP_CALL_SELF` doesn't load a callee, so on this branch `fib-bc` emits 0 `LOAD_VAR` and 2 `CALL_SELF` against the IR path's 2, and the gate reads the saved lookup as an IR regression. `622206e6` counts `CALL_SELF` alongside `LOAD_VAR` on the plain side; the invariant the test pins is unchanged. #579 landed after this branch's last push, so only the rebase could surface it.

The rebase merge driver also left a `core_compiled.lgb` that `check-generated` rejected. Regenerated and amended into the alloc-fix commit.

Green locally: full suite, `make check-generated`, and builds for linux, js/wasm, and plan9/amd64.

Take a look at `622206e6` if you want the ratchet change reviewed; otherwise this is ready to merge.

mparrett:
#645 changes this PR’s posture.

As published, `OP_CALL_SELF`’s direct arm creates a child `Frame` and calls `child.Run()` synchronously. That does not compose with #645: it bypasses the explicit descent path and reintroduces Go-recursive stack growth for exactly the non-tail self-recursive calls #645 makes non-recursive. The branches also conflict in `pkg/vm/vm.go`.

The remaining hypothesis is narrower: whether #618’s static-callee shortcut (skipping Var load, `AsFn`, generic type dispatch, and repeated arity work) has enough incremental value when expressed inside #645’s shared explicit frame-transition machinery, without its own recursive execution path.

I’m measuring that as a three-way comparison:

1. `main`
2. #645
3. #645 plus an adapted static-callee shortcut

Decision gate: if (3) does not materially improve on (2), close #618 and preserve the shortcut only as a possible future micro-optimization. If it does, re-scope it as a small follow-up sharing #645/#620’s transition machinery rather than merging the current opcode implementation unchanged.

No state change here; this PR is already draft.

mparrett:
Closing this PR in favor of #645's explicit-frame architecture. The narrower optimization hypothesis is preserved in #646 with the benchmark data and a post-#620 re-evaluation gate.

The decisive implementation point is that this branch's direct arm calls `child.Run()` synchronously. Merged with #645, that would bypass the shared explicit descent loop and restore Go-recursive stack growth for the non-tail self calls #645 makes stack-safe. A disposable adaptation proved that the shortcut can share #645's transition path and remain stack-safe to 1,000,000 non-tail calls, with allocation parity at 280 B/op and 34 allocs/op.

The incremental performance evidence over #645 was positive but not decision-grade:

- focused 2-second `Frame.Run` ABBA: 4.52% faster median, 8/12 rounds, range -10.51% to +19.83%;
- end-to-end `fib(32)`: 2.41% faster median, 9/15 rounds, range -15.26% to +10.18%;
- non-self control: approximately flat through `Frame.Run`, with a noisy possible 2.22% `Func.Invoke` regression;
- allocations unchanged.

That does not justify carrying a separate opcode, compiler eligibility path, bytecode migration, and overlapping resolver today. #646 keeps the idea available once #645/#620 provide one centralized callee-preparation/frame-transition seam, at which point it can be judged by a fresh three-way benchmark.

The branch is being retained for reference; no code is being deleted.

PR #619, opened 2026-07-22
perf(vm): reuse frames per bytecode call tree

mparrett:
## Summary

- Keeps a bounded nested-frame batch on each pooled top-level root frame.
- Threads that call-tree arena through `Func`, `Closure`, `MultiArityFn`, and nested `OP_INVOKE` execution.
- Reuses nested bytecode frames without taking the global frame-pool mutex; standalone `NewFrame` callers retain their existing bounded pool.
- Preserves steady-state zero-allocation top-level invocation.
- Adds regression coverage for nested arena-threaded invocation, concurrent top-level invocation, zero-allocation top-level reuse, and frame reuse from the arena freelist.

## Relationship to #618 and #620

This PR is independent and based directly on `main`. It improves generic non-tail `OP_INVOKE` execution without introducing or depending on `OP_CALL_SELF`.

- #618 is complementary: its guarded non-tail self-call can obtain child frames from this arena while retaining #618's borrowed argument slice. That temporary composition was tested below.
- #620 explicitly describes #619 as complementary. General tail-call elimination removes frames for tail calls; this arena reduces the cost of non-tail frames that remain. The eventual #620 resolver should be coordinated with `invokeWithArena` so VM callee dispatch is not duplicated across multiple type switches.

This patch removes frame-pool lock traffic from nested bytecode calls but intentionally leaves `Frame.Run` Go-recursive.

## Shared #618/#619 benchmark status

The first harness correction released the benchmark's top-level `NewFrame`, fixing its allocation accounting, but the benchmark still entered through standalone `NewFrame(...).Run()`. That does not exercise the root call-tree arena that #619 acquires and propagates through normal `Func.Invoke`/`Closure.Invoke` entry. It is nevertheless a real secondary product path used by compiler evaluation, the public API, the resolver, runtime execution, and WASM entry points. The earlier Fibonacci table mixed these two entry-path questions and has been removed pending separately labelled reruns.

#621 fixes the historical standalone-frame lifecycle while preserving its benchmark names, and adds separately labelled `new-frame-run` and `func-invoke` entry modes. The same benchmark-only change is applied to `main`, #618, #619, and the temporary combined build. A strict ABBA/`benchstat` matrix through the explicit entry paths is pending a quiet host.

Two shallow-call screens (five ABBA blocks each, n=10 per side) exposed an entry-path distinction. Through standalone `NewFrame.Run`, #619 was 6.83% slower for direct bytecode calls and 9.98% slower for closure calls (`benchstat` p<0.001 for both; all five paired blocks were slower). Through normal `Func.Invoke`, no difference was established for direct calls (p=0.684) or closures (p=0.190). Bytes/op and allocs/op were unchanged. The standalone slowdown is therefore a real workload risk to investigate, while the normal-entry screen found no regression.

## Validation

- [x] `go test -race ./pkg/vm -count=1`
- [x] `go test ./pkg/bytecode ./pkg/compiler ./pkg/vm ./pkg/rt ./pkg/wasmhost -count=1`
- [x] `make check-generated`
- [x] `make test`
- [x] `BenchmarkFuncInvoke/Direct` and `/Closure`: 0 B/op, 0 allocs/op
- [ ] Repeat the strict-ABBA #618/#619 matrix through normal `Func.Invoke` entry
- [ ] Analyze the fully corrected preserved output with `benchstat`
- [x] Shallow direct/closure A/B for standalone and normal invocation entry
- [ ] Investigate or mitigate the standalone `NewFrame.Run` shallow-call regression
- [ ] Add bytecode-to-native/protocol/multifn dispatch probes

### DIFF
--- a/pkg/vm/frame_arena_test.go
+++ b/pkg/vm/frame_arena_test.go
@@ -1,0 +1,127 @@
+package vm
+
+import (
+	"sync"
+	"testing"
+)
+
+func TestFrameArena_NestedSelfInvoke(t *testing.T) {
+	// Countdown via OP_INVOKE self-call exercises the arena-threaded path.
+	consts := NewConsts()
+	body := NewCodeChunk(consts)
+	zero := consts.Intern(Int(0))
+	one := consts.Intern(Int(1))
+
+	fn := MakeFunc(1, false, body)
+	fn.SetName("arena-countdown")
+	fnIdx := consts.Intern(fn)
+
+	body.Append(OP_LOAD_ARG)
+	body.Append32(0)
+	body.Append(OP_LOAD_CONST)
+	body.Append32(zero)
+	body.Append(OP_EQ)
+	br := body.Length()
+	body.Append(OP_BRANCH_TRUE)
+	body.Append32(0)
+	body.Append(OP_LOAD_CONST)
+	body.Append32(fnIdx)
+	body.Append(OP_LOAD_ARG)
+	body.Append32(0)
+	body.Append(OP_LOAD_CONST)
+	body.Append32(one)
+	body.Append(OP_SUB)
+	body.Append(OP_INVOKE)
+	body.Append32(1)
+	body.Append(OP_RETURN)
+	ret0 := body.Length()
+	body.Update32(br+1, int32(ret0-br))
+	body.Append(OP_LOAD_CONST)
+	body.Append32(zero)
+	body.Append(OP_RETURN)
+	body.SetMaxStack(8)
+
+	out, err := fn.Invoke([]Value{Int(5)})
+	if err != nil {
+		t.Fatal(err)
+	}
+	if got := out.Unbox().(int); got != 0 {
+		t.Fatalf("got %v, want 0", got)
+	}
+}
+
+func TestFrameArena_ConcurrentTopLevel(t *testing.T) {
+	consts := NewConsts()
+	chunk := NewCodeChunk(consts)
+	chunk.Append(OP_LOAD_ARG)
+	chunk.Append32(0)
+	chunk.Append(OP_RETURN)
+	chunk.SetMaxStack(4)
+	fn := MakeFunc(1, false, chunk)
+
+	var wg sync.WaitGroup
+	for i := 0; i < 32; i++ {
+		wg.Add(1)
+		go func(n int) {
+			defer wg.Done()
+			for j := 0; j < 100; j++ {
+				out, err := fn.Invoke([]Value{Int(n)})
+				if err != nil {
+					t.Errorf("invoke: %v", err)
+					return
+				}
+				if out.Unbox().(int) != n {
+					t.Errorf("got %v want %d", out, n)
+					return
+				}
+			}
+		}(i)
+	}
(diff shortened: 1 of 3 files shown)


mparrett:
Closing this after follow-up measurement. The branch stays up for reference.

Summary of the follow-up, using fib(32) end-to-end binaries built with `lg -b` (the path embedders ship) plus the #621 in-process entry-mode harness:

- The standalone gain is small: 2-3% on fib through both `lg -b` bundle entry and direct `Func.Invoke` entry, consistent with the earlier normal-entry screen (p=0.063).
- The standalone `NewFrame.Run` shallow-call regression reported in the PR body reproduced and remains unresolved.
- Composed with #618 as published, the result is sub-additive: a plain merge of the two heads measures worse than #618 alone (in-process fib(20): #618 is 13-16% faster than main; the merge gives most of that back). The cause is that `OP_CALL_SELF` takes child frames from the global pool via `NewFrame`, while #618 replaces the recursive-arm `OP_INVOKE`s this arena accelerates — so the combined hot path pays the arena bookkeeping and gets none of the mutex-free reuse. A plain merge does not reproduce the earlier +15% combined figure; that number evidently required CALL_SELF children to come from the arena, and neither head includes that glue. The fix is one line (`child := newFrameIn(f.arena, f.code, rawArgs)`) and restores parity with #618 alone; whether it does better than parity needs a quiet host.
- Entry path does not change any of this: the arena attaches at the first `Func.Invoke` boundary, so bundled binaries and API callers see the same behavior.

Given the small standalone win, the open regression, and that #620's resolver design reworks this same dispatch seam, withdrawing this is cleaner than iterating on it. Two things should carry forward into #620:

1. Mutex-free per-call-tree frame reuse. The pool-mutex cost this PR removed is measurable; the TCE resolver is the right owner of frame acquisition.
2. The #618 interaction as a worked example of the overlapping-dispatch-paths hazard: any specialized call opcode and the resolver must share frame-acquisition machinery, or composition silently regresses.

Numbers and methodology are on #620.

mparrett:
Documentation follow-up after the #644 prototype review:

- The statement that non-recursive `Frame.Run` was tracked by #464 was incorrect. #464 is the runtime performance/concurrency epic and does not specify a trampoline or explicit frame chain; #644 now tracks that work.
- The underlying ownership conclusion still holds, but “the TCE resolver is the right owner of frame acquisition” is too narrow. #620 and #644 both reshape callee preparation and frame transitions. The shared resolver/dispatch seam should own arity selection, captures, frame acquisition, tail-frame reuse, and non-tail child descent so those paths cannot diverge again.
- The 7.9% `fib` result reported from the first #644 explicit-frame prototype is still provisional. Three review findings—multi-arity still taking the Go-recursive path, skipped per-frame trace/profile/allocation lifecycle, and retained error-path frames—have not yet been fixed and re-measured. Until that rerun exists, the number should not be treated as a production throughput claim.

This does not change #619’s disposition. Keeping it closed remains the cleaner choice: its mutex-free arena is useful evidence for allocator policy, but landing it separately would add another owner at the same seam that #620/#644 need to unify.

PR #623, opened 2026-07-23
feat(cmd): bootprobe + lgbstat — core boot and bundle diagnostics

mparrett:
## Summary

Two small diagnostic commands that came out of assessing what lazy core loading or bundle pruning would buy. Posting them because the numbers they produce reframed that question for me, and they seem worth having in-tree for anyone poking at boot or bundle size.

- `cmd/bootprobe` — phase-timed core boot: decode vs replay vs remainder, via `LoadCoreBundle`'s existing `OnPhase` hooks. Boots compiler-free (same surface as `cmd/lg-runtime`) so it isolates the runtime boot path. Optional `-decodes`/`-replays` loops with `-cpuprofile` for profiling either phase.
- `cmd/lgbstat` — byte composition of a `.lgb`: const-type counts, string-table weight, per-source-file chunk attribution (via each chunk's first source-map entry), and debug-section sizes measured by ablation (re-encode with the section stripped, diff against the as-is re-encode). Complements `scripts/lgbdump.lg`, which dumps instructions; this accounts bytes.

## What they show today

Measured on an M2 (darwin/arm64), quiet-host medians.

Boot: decode of the 264KB core bundle is ~95% of `LoadCoreBundle` (~1.4ms); replaying all 802 core defs is ~0.08ms; and the whole thing is small next to process spawn + Go package init (~5ms; `pkg/rt` native registration is the largest init item per `GODEBUG=inittrace`). So: lazy *replay* has nothing to win; decode cost tracks bundle size; the bigger cold-start item is package init, not the bundle.

Bundle: source maps are 17.3% (45.7KB), localvar debug tables 2.8%, and docstrings 2.1% — a release-stripped bundle is ~20% smaller, which I verified end-to-end (210,882 bytes; boots, rt/vm/compiler suites pass). Per-file attribution also shows ~10% of code int32s under `ir.data:lgbgen-bootstrap`, and 508 VOID consts remain in the pool (#600 stripped VOID instructions; these are the pool slots).

## Notes

- Both build on linux, js/wasm, and plan9; gofmt/vet clean.
- No behavior changes — additive commands only.
- Follow-ups these numbers suggest, each separable: a `-strip` option on `lg -c`/`-b` for release bundles, VOID pool-slot compaction, and lazy native registration.

### DIFF
--- a/cmd/bootprobe/main.go
+++ b/cmd/bootprobe/main.go
@@ -1,0 +1,136 @@
+/*
+ * Copyright (c) 2026 Matt Parrett <matt.parrett@gmail.com>
+ * Part of the let-go project; see CONTRIBUTORS for full list of authors.
+ * SPDX-License-Identifier: MIT
+ */
+
+// bootprobe reports where core boot time goes: bundle decode vs definition
+// replay vs everything else, using LoadCoreBundle's OnPhase hooks. It boots
+// compiler-free (same surface as cmd/lg-runtime), so the numbers isolate the
+// runtime boot path.
+//
+// Usage:
+//
+//	bootprobe                              one boot, phase breakdown + counts
+//	bootprobe -decodes 300                 also re-decode the bundle N times
+//	bootprobe -replays 300                 also re-run the core chunk N times
+//	bootprobe -decodes 300 -cpuprofile p   CPU-profile the loop
+//
+// Loops report min/median/max so a quiet-host median is easy to extract;
+// run the binary itself repeatedly to sample process-level variance.
+package main
+
+import (
+	"flag"
+	"fmt"
+	"os"
+	"runtime/pprof"
+	"sort"
+	"time"
+
+	"github.com/nooga/let-go/pkg/rt"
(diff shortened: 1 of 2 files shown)


cmd/bootprobe/main.go
```diff
@@ -0,0 +1,136 @@
...
+	"github.com/nooga/let-go/pkg/rt"
+	"github.com/nooga/let-go/pkg/vm"
+
+	_ "github.com/nooga/let-go/pkg/rt/corefns"
+)
+
+func medianOf(xs []float64) float64 {
+	s := append([]float64(nil), xs...)
+	sort.Float64s(s)
+	return s[len(s)/2]
```
nnunley: Probably should do average of the two medians in the case of xs being an even length.

nnunley (approved):

cmd/bootprobe/main.go
mparrett: Sharp catch. Fixed in https://github.com/nooga/let-go/pull/623/commits/bf19b319311c2dbd84f51b140ba5685da689f596

PR #628, opened 2026-07-24
feat(aot): generate native-entry Go frame from main/-main (#425 Item 1)

### LINKED ISSUE #657: AOT entry frame: recover the lowered signature from gogen as data, not by string-matching rendered Go
`ir.passes.entry-frame/analyze-entry` recovers a lowered function's signature by string-matching the rendered Go source. Two predicates do the work, both in `pkg/rt/core/ir/passes/entry_frame.lg`:

- `go-src-has-func?` (~L97) tests `includes? go-src "func <Name>("` to decide whether the entry lowered at all.
- `go-func-needs-error?` (~L101) locates that same needle, takes the first line from it, finds the first `)`, and looks for `error` in the remainder to decide whether the call site binds two values or one.

Both are correct against the gofmt'd single-line signatures gogen emits today, and both fail conservatively. An unmatched name falls back to the VM entry path, which is slower but correct; an unparseable result list defaults to `needs-error? true`, which surfaces as a Go compile error at the call site rather than as wrong behavior. Raised by @nnunley in review of #628, and not a bug today.

The coupling is the problem. gogen knows each lowered function's name and result list structurally, renders them to text, and entry-frame re-derives them with `includes?` and `index-of`. Nothing links the two ends: a change to gogen's rendering — a multi-line signature for a long parameter list, a named result, a build tag between the doc comment and the `func` — silently degrades entry generation, and the only signal is a slower binary or a compile error in generated code.

Proposal: have the lowering hand the entry's shape back as data alongside the rendered source, at minimum `{:go-name, :lowered?, :results}` or equivalent, and let `analyze-entry` consume that. Keep the string parsing as a fallback for callers holding source and nothing else. That takes both predicates off the critical path and turns a rendering change into a type error instead of a silent downgrade.

No urgency while the output stays gofmt'd — this is a job for whoever is next in gogen's emit path. #628 added a CI lane that compiles and runs a generated frame, so a regression here now fails loudly rather than shipping green.

mparrett:
## Summary
- Closes #657 (structured lowering metadata for entry selection).
- Closes the silent perf-cliff half of #425 Item 1: with `--entry-frame`, `lg-compile` emits a `main.go` native-entry frame when it recognizes `main`/`-main` (including `defn-`), choosing a direct native call when the entry lowered or a namespace-qualified VM `InvokeProgramEntry` fallback when it did not.
- Frame emission is **opt-in** (`--entry-frame`). Without the flag, `lg-compile` writes only lowered packages — historical behavior for orchestrators such as Gloat that own their own executable / shared-library frames.
- Frame sits on `#520` `BootCore` + `rt.LoadProgramNamespaces` (NSOrder load with `ApplyGoOverrides` drain), publishes `*command-line-args*` before namespace replay, and owns `defer rt.ShutdownAllPods()`.
- Entry selection uses structured lowerer metadata (collision-resolved `:go-name`), not rendered-Go text scanning (#657).
- Also prints the Item 3 summary line (`N fns lowered; native entry: Main ✓ / none`) with or without the frame. Fib demo in `examples/aot/native-entry/` reproduces the spike (~16–17× on fib(34)).

## Ownership (Gloat)
- Let-Go owns explicitly requested standalone frames (`--entry-frame`).
- Gloat keeps owning `template/lg-main.go` / shared-lib templates (different MainChunk / override-drain semantics).
- Earlier heads-up: [comment](https://github.com/nooga/let-go/pull/628#issuecomment-5069872390) — corrected now that emission is opt-in.

## Test plan
- [x] `go test ./pkg/rt/ -run 'TestLoadProgramNamespaces|TestRunProgramMainChunk|TestRunExecUnit|TestInvokeProgramEntry'`
- [x] `go test ./test/ -run TestRunner/entry_frame`
- [x] `go test ./test/e2e/ -run TestLgCompileEntryFrameOptIn` (no flag / flag / package parity)
- [x] `./examples/aot/native-entry/build.sh` → `fib.native` prints `55` / `5702887` for fib(10)/fib(34)
- [x] CI green (lint fix pushed in `e21014a7`; re-run pending)
- [ ] Ingy ack on ownership boundary


mparrett:
@ingydotnet A note on gloat's let-go backend, prompted by a recent let-go change:

#628 in let-go adds a runtime-only AOT entry frame: boot core, load the program's namespaces, then enter `-main` natively or fall back to the VM. While checking it wouldn't step on gloat, I saw your two let-go templates `template/lg-main.go` and `template/lg-lib-main.go` both carry the same ~20-line boot loop: 

decode `program.lgb`, replay `NSOrder` skipping `MainChunk`, and drain `rt.ApplyGoOverrides` after each ns. And both boot through `compiler.NewCompiler` + the resolver, which links the whole let-go compiler into the binary.

This PR pulls that loop into three public rt helpers:

- `rt.BootCore()`: the runtime-only boot (from #520), no compiler linked
- `rt.LoadProgramNamespaces(unit)`: the `NSOrder` replay + override drain
- `rt.InvokeProgramEntry(ec, name, argv)`: the VM-fallback entry invoke

Both templates could drop to a couple of lines and shed the compiler/resolver imports, which should shrink those binaries. One gotcha if you adopt it: `LoadProgramNamespaces` deliberately does not skip `MainChunk` the way your loop does (a single-ns program needs those vars installed before the native call), so the fallback goes through `InvokeProgramEntry` rather than a bare `NewFrame(unit.MainChunk)`.

No urgency (the two don't share a code path today and nothing's broken).

1. `lg-compile` now emits its own standalone `main.go` frame. If you ever point it at a gloat build dir, gate one of the two frames so you don't get two `func main()`.
2. I'm could take a stab at the gloat PR moving both let-go templates onto these helpers if you like.

Let us know your thoughts. Thanks!

nnunley (commented):
Reviewed (multi-agent + verification). Implements #425 Item 1; sits at the Go executable shell, no new IR ops, parity unaffected. One item to fix pre-merge:

**Undeduplicated blank imports can emit invalid Go.** In the entry-frame generation the `blanks` vector is built by keeping `:go-pkg` per lowered spec with no `distinct`. If two inputs resolve to the same Go package (multi-file same-namespace, or a duplicate input), the generated import block gets `_ "pkg"` twice → duplicate-import compile failure. Wrapping the collected packages in `distinct` before emitting fixes it.

Robustness nits (safe today — both fall back conservatively): `go-func-needs-error?` (`entry_frame.lg` ~L260) defaults to `true` when a signature spans multiple lines or has non-gofmt spacing (it parses only the first line); `go-src-has-func?` (~L256) uses exact `func <name>(` matching, so non-standard formatting forces a VM fallback instead of the direct native call. Both are fine for gofmt'd output.

mparrett:
Fixed the import bug, rebased, and added the lane that would have caught it. Now at `96b64ef9`.

**Duplicate blank imports** (`c6ddf549`). Confirmed at `scripts/lg-compile:200-204`: `blanks` is a `keep` over every spec collecting `(:go-pkg s)`, and `entry_frame.lg` emitted one `_ "pkg"` line per element. I put the `distinct` at the emission site rather than in `lg-compile` — `:blank-imports` is a documented opt, and a repeated import is wrong whoever passed it, so the invariant belongs with the code writing the import block. `entry_frame_test.lg` covers a plain repeat and a repeated entry-import on the not-lowered path, where it is kept rather than filtered. Both fail on the unfixed tree; I checked by reverting the source and re-running.

**Nothing compiled a generated frame** (`96b64ef9`). This turned out to be the bigger hole. `entry_frame_test.lg` asserts on the emitted *string*, and `examples/aot/native-entry/build.sh` was referenced by no workflow and no make target. A duplicate import is a `go build` failure, so no lane could have caught it, and the same gap covers every other way the frame emits well-formed-looking but invalid Go.

The example now runs in the same job as the expensive lowering e2e: `lg-compile`, embed the `.lgb`, link, then assert both documented outputs (`55` and `5702887`). About a second of wall clock from a clean `out/`, which seemed proportionate next to the determinism harness already in that job. Happy to move it to its own job or a narrower trigger if you'd rather.

**The two parser nits — I left both.** Your read of the failure modes matches mine, and they're the tolerable ones: an unmatched `func <name>(` falls back to the VM entry (slower, correct), and an unparseable result list defaults to `needs-error? true`, which shows up as a Go compile error at the call site rather than wrong runtime behavior. With the new lane, both now fail loudly.

The coupling bothers me more than the parsing. gogen knows the name and result list structurally, renders them to text, and entry-frame re-derives them with `includes?` and `index-of`; tightening the string matching would preserve that. Filed #657 to have the lowering hand the entry's shape back as data, keeping the string path as a fallback, which takes both predicates off the critical path. I'd default to doing that when someone's next in gogen's emit path, but happy to pick it up before this merges if you'd prefer.

.github/workflows/go.yml
```diff
@@ -110,6 +110,22 @@ jobs:
         # full; the fast `make test` loop passes -short to skip them.
         run: go test -v -timeout 600s ./test/e2e/ -run 'TestLoweringDeterminism|TestDeftypeSkeletonNativeLowering|TestLginteropXxh3RoundTrip'
 
+      # The only lane that compiles and RUNS a generated native-entry frame.
+      # test/entry_frame_test.lg asserts on the emitted string, so a frame that
+      # emits well-formed-looking but invalid Go — a duplicate blank import, a
+      # result list that disagrees with the call site — passes every other gate
+      # and ships green. This runs lg-compile, embeds the .lgb, links the
+      # binary, and checks what it prints. Cheap next to the harnesses above:
+      # one lg build plus a two-package go build.
```
nnunley: The breadcrumbs here are are odd.  I guess it's a pattern currently that the LLMs overly document the CI. Perhaps we should have it step back. I thought we had the beginnings of a set of AOT targets. If we have an expectations file for each aot run, along with deadlines, I think this could be made more generic than just running fib.native for testing.

examples/aot/native-entry/build.sh
```diff
@@ -0,0 +1,44 @@
...
+  for c in "$ROOT/lg" "$ROOT/lg-bc"; do [ -x "$c" ] && LG="$c" && break; done
+fi
+[ -n "$LG" ] && [ -x "$LG" ] || { echo "no lg binary — build one or set LG=" >&2; exit 1; }
+
+mkdir -p "$OUT"
+# Fresh-ish out: keep go.sum if present; wipe generated sources.
+rm -f "$OUT"/main.go "$OUT"/program.lgb "$OUT"/fib.native
+rm -rf "$OUT/fib"
+
+echo "==> lg-compile (emits fib.go + main.go native-entry frame)"
```
nnunley: This is specialized for fib.go? Clean up the comments if it's intended to be generic. Or rename if it's only building the fib example.

What would be more amusing, actually, would be to have the harness be written in lg as a way of dogfooding the e2e process.

pkg/rt/core/ir/passes/entry_frame.lg
```diff
@@ -0,0 +1,323 @@
...
+    (or (first (filter (fn [e] (= "-main" (:name e))) found))
+        (first found))))
+
+(defn entry-go-name
+  "Go identifier the lowered entry is emitted under. Public → PascalCase
+   (main/-main both → Main); private → unexported munge (main → main,
+   -main → _main), via lower-go/go-name so the munge (incl. reserved-word
+   handling) stays single-sourced with the lowered emitter."
+  [entry]
+  (if (:private? entry)
```
nnunley, line 151: I think having a private main should probably still be considered a valid main target for a top level form. I'd check the other AOT frameworks to see how they handle it - specifically glojure and jolt.

pkg/rt/core/ir/passes/entry_frame.lg
```diff
@@ -0,0 +1,323 @@
...
+    (lower-go/go-export-name (:name entry))))
+
+(defn- go-src-has-func? [go-src go-name]
+  (and (string? go-src)
+       (clojure.string/includes? go-src (str "func " go-name "("))))
+
+(defn- go-func-needs-error? [go-src go-name]
+  "True when the lowered func's result list includes error. Defaults to true
+   (safer for call sites) when the signature line can't be parsed."
+  (if-not (string? go-src)
```
nnunley, line 172: We probably should be exposing go-parser/go-ast for this analysis, since it can give you stronger contracts for working with this. Inspect the gogen code for examples.

pkg/rt/core/ir/passes/entry_frame.lg
```diff
@@ -0,0 +1,323 @@
...
+
+(defn- go-string-lit [s]
+  (str "\"" (clojure.string/replace
+              (clojure.string/replace s "\\" "\\\\")
+              "\"" "\\\"") "\""))
+
+(defn- emit-argv-setup [arity-shape]
+  (case arity-shape
+    :none
+    "\trt.SetCommandLineArgs(os.Args[1:])\n"
```
nnunley: Use the gogen code for this, instead of emitting strings.

nnunley (dismissed):
So, it looks like this should lean on the code I wrote for gogen - there's a lot of direct string generation that could be handled by it, rather than hand-rolling another copy. The gogen code already has name munging support, too, so that might remove some of the duplication.

mparrett:
Thanks for all the great feedback, I will take a look.

pkg/rt/core/ir/passes/entry_frame.lg
mparrett: Agreed. I think #657 and this suggestion address complementary seams: the lowerer should return the actual emitted function metadata, including its collision-resolved Go name, while entry-frame should use gogen to construct the frame rather than template Go text. Function collision resolution currently lives in `ir.lower-go` rather than gogen, so I’ll preserve that as the source of truth. I’ll also add the small gogen constructs the frame needs. Does that match the scope you intended?

mparrett:
**Phase 0 checkpoint** — rebased onto current `main`.

| | SHA |
|---|---|
| base (`main`) | `1f281df4` |
| head | `1143337a` |

- Rebased 4 commits; conflict in `pkg/rt/generated.sums` resolved by regenerating (merge driver wrote a stale digest during rebase; `make build` / lgbgen recomputed → `919416ec…`, committed as `1143337a`).
- Baseline green: focused `pkg/rt` tests, `TestRunner/entry_frame`, `examples/aot/native-entry` (`55` / `5702887`).
- Next: Phase 1 — structured lowerer metadata as source of truth (#657).

mparrett:
**Phase 1 checkpoint** — lowering metadata is now the source of truth (#657).

Head: `87e8b7b0` (on base `1f281df4`)

- Added `lower-ns-to-go-result` / `lower-all-ns-to-go-result`; string-returning wrappers preserved.
- `lg-compile` selects the entry by source name from `:fns` and passes `{:fn …}` to `analyze-entry`.
- Text scanning (`go-src-has-func?` / `go-func-needs-error?`) is compat-only — off the critical path.
- Confirmed: `main` + `-main` → frame calls `prog.Main__main` (summary: `Main__main ✓`).
- Tests cover collision, identical-arity, and metadata-over-text.

Next: Phase 2 (gogen AST frame) or Phase 3 (arity rejection / args order / ns-qualified fallback).

mparrett:
**Phase 3 checkpoint** — entry/runtime semantics corrected.

Head: `e6eedb70`

- **3.1** Accept only `[]`, `[argv]`, `[& args]`. `[first & rest]` fails lg-compile with a clear diagnostic before writing output.
- **3.2** `SetCommandLineArgs` runs immediately after `BootCore`, before decode/load.
- **3.3** `InvokeProgramEntry(ec, namespace, name, argv)` resolves via `LookupNS(ns).LookupLocal(name)` only.
- **3.4** Frame owns `defer rt.ShutdownAllPods()` (documented; matches lg / lg-runtime).

Verified: entry_frame + InvokeProgramEntry tests, fib.native `55`/`5702887`, arity-reject exit 1 with no output dir.

Next: Phase 2 (gogen AST frame) or Phase 5 (generic E2E matrix). Phase 4 still needs the Gloat ownership decision.

mparrett:
**Follow-up** on Phase 1 test hardening + return-hinted entry gap.

Head: `5a9e95c4`

- `find-program-entry` now unwraps `(with-meta …)` name/arity via the same `ir.build` return-hint helpers the lowering pipeline uses. Both `(defn -main ^long [] …)` and `(defn ^long -main [] …)` emit a frame (`Main ✓`).
- Added `lower-ns-to-go-result-contract`: promised keys, `Main__main` collision record, and `lower-ns-to-go` = `(:source result)`.

Ready for Phase 5 (E2E matrix should include the return-hint cases).

mparrett:
**Phase 4 (partial) — frame emission is opt-in.**

Head: `b05e7cd0`

```text
./lg scripts/lg-compile [--entry-frame] <out-dir> <import-prefix> <file.lg>...
```

- Without `--entry-frame`: packages only (historical behavior). Summary still prints.
- With `--entry-frame`: exactly one `main.go` frame (+ private bridge when needed).
- Example/CI pass the flag. `TestLgCompileEntryFrameOptIn` covers no-flag / flag / package parity.
- Unsupported entry arities are rejected only when `--entry-frame` is set.

**Correction** to the earlier Gloat note ([comment](https://github.com/nooga/let-go/pull/628#issuecomment-5069872390)): gloat does not need a new gate to avoid a duplicate `main`. Omit `--entry-frame` and `lg-compile` will not emit one. Gloat keeps owning its executable and shared-library templates; let-go owns only explicitly requested standalone frames.

Draft note for Ingy is staged locally for a quick prose pass before posting.

mparrett:
Status checkpoint on `wt/425-entry`. nnunley's central ask — replace string emission with gogen — is still open; re-review waits on that.

Since the last round:

- `--entry-frame` opt-in: without the flag, `lg-compile` writes packages only (no `main.go`, no private bridge). The summary line still reports when a native entry is available.
- Bridge contract e2e: no flag + private → no frame/bridge; flag + public → one `main.go`; flag + private → frame + bridge; package output identical excluding the intentional bridge file.
- Native-entry matrix in CI (`TestLgCompileEntryFrameOptIn` / `TestNativeEntryMatrix`): return-hint spellings, typed/no-error call shape, unsupported-arity rejection (flag-only), argv / variadic, collision, namespace-qualified VM fallback, fib spike inputs.

Next on this branch:

1. Phase 2 — rewrite frame emission onto gogen AST, with the matrix as the regression fence.
2. Phase 6 — full verification / generated-artifact checks.
3. Phase 7 — reply on each review thread, update #657 bookkeeping, then mark ready and ask for re-review.

.github/workflows/go.yml
mparrett: Replaced the fib-only CI step with a table-driven e2e matrix (`TestLgCompileEntryFrameOptIn` / `TestNativeEntryMatrix`) and trimmed the workflow commentary. Fib stays as `examples/aot/native-entry/`; CI exercises the failure modes (arity rejection, collision, private bridge, return hints, fallback) rather than one happy-path binary.

examples/aot/native-entry/build.sh
mparrett: Kept this script as the fib example builder (comments now match that). The generic harness lives in `test/e2e/native_entry_matrix_test.go`. An lg-written dogfood harness is a good follow-up; not in this PR.

pkg/rt/core/ir/passes/entry_frame.lg
mparrett: Agreed — private `main`/`-main` stays a valid entry. With `--entry-frame`, a same-package `NativeMain` bridge is emitted so package main can call it; the matrix covers private `-main []` end to end.

pkg/rt/core/ir/passes/entry_frame.lg
mparrett: Addressed via structured lowerer metadata instead (#657 / Phase 1): `lower-ns-to-go-result` returns collision-resolved `:go-name`, `:needs-error?`, and arity, and `analyze-entry` consumes that. The text scanners remain only as a compatibility fallback. Exposing a Lisp go-parser surface is still useful longer term; it isn't needed for this frame path.

pkg/rt/core/ir/passes/entry_frame.lg
mparrett: Done in `957652e5`: frame and private bridge are built with gogen AST and rendered through `go/format`. Added the small constructors the frame needed (`defer-stmt`, `slice-expr`, `call-variadic`, `range-stmt`, `with-go-directive` for `//go:embed`). Collision-resolved names still come from `ir.lower-go` metadata, not gogen path munging.

mparrett:
Phase 6/7 complete — ready for re-review.

**Phase 6 (local):** `make build` / `check-generated` / `gogen-diff` green; `go test ./pkg/rt/ ./test/ ./test/e2e/` and `go test ./...` green; fib example prints `55` / `5702887` once each. Gloat `RUN_SLOW_TESTS=1 prove -v test/lg-native-bin.t` green against this tip. `test/lg-shared-lib.t` still fails building the shared lib (`rt.LoadProgramNamespaces` undefined in Gloat's go.mod pin) — same on sibling `let-go` main; tracked on gloathub/gloat#7, not a #628 regression.

**Phase 7:** replied + resolved the five review threads; closed #657 as done here. Lint CI failure on `957652e5` fixed in `e21014a7` (staticcheck on the new error string).

@nnunley — requesting re-review.

mparrett:
CI update on `e21014a7`: **lint is green**. The remaining `tinygo-wasi-build` failure is `wasmtime: command not found` on the runner (exit 127) — unrelated to this PR; same workflow is green on `main`.

mparrett:
@ingydotnet follow-up on the earlier note about #628 and gloat's templates.

Short version of the boundary I think we want: **Gloat owns Gloat's frames; Let-Go owns explicitly requested standalone frames.**

`lg-compile` can emit a standalone native-entry `main.go`, but emission is now explicit opt-in:

```text
./lg scripts/lg-compile --entry-frame <out-dir> <import-prefix> <file.lg>...
```

Without `--entry-frame`, `lg-compile` writes only lowered packages — the historical behavior. The Item 3 summary line still prints either way.

So:

- Standalone let-go example / CI passes `--entry-frame`; let-go owns that frame.
- Gloat omits the flag and keeps owning `template/lg-main.go` and the shared-library template. Those frames have different semantics (skip the driver `MainChunk` while draining native overrides; shared-lib support), so they should not be forced onto let-go's standalone shape.

Earlier I suggested gating one of the two frames if gloat pointed `lg-compile` at its build dir. With opt-in, that concern goes away: gloat needs no new flag to keep today's package-only path — just don't pass `--entry-frame`.

Does that match how you want the boundary drawn for now? Longer term I'm curious whether gloat would ever want to consume a structured let-go manifest/frame, or prefer to keep its specialized drivers.

nooga (approved):
LGTM. This is a well-executed, well-tested feature — checked out the branch and ran it for real: built and ran `examples/aot/native-entry` end to end (`fib.native` → 55, `fib.native x` → 5702887), inspected the generated `main.go` (clean, correct — boot → args → decode → load namespaces → argv-box → `prog.Main` → error-checked exit). `test/e2e/native_entry_matrix_test.go` is genuine E2E across 17 cases (public/private `-main`/`main`, argv/variadic adaptation, arity rejection, main+`-main` collision, error propagation and exit codes, both return-hint spellings) — actually builds and executes each binary, not just structural checks on generated source.

Verified the structural concern from nnunley's review is properly addressed: frame emission now goes through gogen AST constructors + `go/format`, not string templates, and entry selection is driven by structured `:fns` metadata rather than text-scanning; the old text-scan predicates are an explicit off-critical-path fallback.

On the `tinygo-wasi-build` CI failure: pulled the actual job log rather than taking the "unrelated infra" claim on faith — confirmed it's a wasmtime-installer flake (`curl` choking on an unresolved `{}` version placeholder in the download URL), and confirmed the identical step succeeded cleanly on the next `main` CI run the same day. Genuinely unrelated to this PR.

Two minor, non-blocking notes:
- Generated `main.go`'s import block isn't goimports-grouped (stdlib/third-party/local all together) — compiles fine, cosmetic.
- A multi-arity `(defn -main ([] …) ([x] …))` silently isn't recognized as an entry at all (falls through with no diagnostic) rather than erroring — probably fine given multi-arity `-main` is unusual, but worth a doc note if anyone hits it.

One mechanical thing before merge: `mergeable: CONFLICTING` is, same as the other PRs in this batch, just the `generated.sums` digest — `main` has moved ~9 commits since the last rebase. Zero real conflicts locally via the `merge=sums` driver; needs a `git rebase origin/main && make generate` push to clear it on GitHub's side.

Approving.

mparrett:
@nooga Thanks for the approval. FYI: addressed the two non-blocking suggestions in #692

PR #629, opened 2026-07-24
fix(tinygo): stub SameFile for js/wasm in pkg/bundle

mparrett:
## Summary
- TinyGo's `os` omits `SameFile` under `js`/`wasm_unknown` (`types_anyos.go` is `!js && !wasm_unknown`), so `pkg/bundle.CollectResources` failed to compile for the TinyGo browser target after #522.
- Route the call through a build-tagged helper: real `os.SameFile` everywhere TinyGo provides it (stock Go, TinyGo wasi/native); `false` on js/wasm where host-side `lg -b` isn't the payload path.

Fast-follow to #522.

## Test plan
- [x] `go test ./pkg/bundle/` green
- [x] `tinygo build -target=wasm` of a probe that imports `pkg/bundle` succeeds
- [x] `tinygo build -target=wasi` of the same probe succeeds (uses real SameFile)

### DIFF
--- a/pkg/bundle/bundle.go
+++ b/pkg/bundle/bundle.go
@@ -390,7 +390,7 @@ func collectResourceDir(dir, rootReal, relPrefix string, files map[string][]byte
 		if !info.Mode().IsRegular() {
 			continue // FIFO, device, socket
 		}
-		if exclude != nil && os.SameFile(info, exclude) {
+		if exclude != nil && sameFile(info, exclude) {
 			continue // never embed the bundle's own output file
 		}
 		if _, exists := files[key]; exists {
--- a/pkg/bundle/samefile.go
+++ b/pkg/bundle/samefile.go
@@ -1,0 +1,17 @@
+//go:build !tinygo || (!js && !wasm_unknown)
+
+/*
+ * Copyright (c) 2021-2026 Marcin Gasperowicz <xnooga@gmail.com>
+ * SPDX-License-Identifier: MIT
+ */
+
+package bundle
+
+import "os"
+
+// sameFile reports whether fi1 and fi2 describe the same file. Thin wrapper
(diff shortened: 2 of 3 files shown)


mparrett:
Self-review: **No findings.**

[#629](https://github.com/nooga/let-go/pull/629) is a tight, correct fix. CI is all green; waiting on review.

### Assessment

Build tags are mutually exclusive and match TinyGo’s own `SameFile` availability (`types_anyos.go` is `!js && !wasm_unknown`):

| Build | File used |
|---|---|
| stock Go | `samefile.go` → `os.SameFile` |
| TinyGo wasi/native | `samefile.go` → `os.SameFile` |
| TinyGo js/wasm | `samefile_tinygo.go` → `false` |

Call site change is one line; existing `TestCollectResourcesExcludesOutputBinary` still covers the real path.

The stub returning `false` is fine in practice: `CollectResources` is only used from host `lg -b` (`lg.go`). TinyGo wasm/runtime pulls in `pkg/bundle` for `ReadBundledSelf` (`cmd/lg-runtime`), not resource collection — so the stub is a linkability shim, not a behavior change on a live path.

### Residual notes (not blockers)

- Branch is **1 behind** `main` (the ir-compile-strict commit) — should merge clean; rebase optional.
- No TinyGo-wasm CI leg for `pkg/bundle` (only wasi/`lg-runtime`); we already smoke-tested wasm locally.

nnunley (approved):
Reviewed — clean. The build-tag split is complementary and exhaustive (`!tinygo || (!js && !wasm_unknown)` vs `tinygo && (js || wasm_unknown)`), and the js/wasm stub returning false is correctly scoped to the host-only `lg -b` resource-collection path, not the wasm payload. LGTM.

PR #639, opened 2026-07-27
feat(rt): hoist 222 clojure.core primitives to named //lg:native decls

nnunley:
**Stack:** #627 (fix) → **this** → per-package-registrars → provenance-manifest. Bottom of this PR's new work sits on #627; until #627 merges, the diff shows that fix commit too — it drops out once #627 lands.

Lifts the inline anonymous closures registered as `clojure.core` primitives in `lang.go` into named, `//lg:native`-annotated top-level functions, via a new codemod (`cmd/hoist-natives`). Three wins: real stack-trace frames (`rt.CorePlus`, not `installLangNS.func42`); declarative registration; and closures narrow toward typed/`ec`-threaded signatures.

### Commits (this feature)
- `feat(rt)`: non-terminating diagnostics for generated-primitive registration (audit + `LG_REGPRIM_DEBUG` + regression-guard test)
- `feat(cmd)`: the `hoist-natives` codemod
- `refactor(rt)`: lift shared native helper closures to package level
- `feat(build)`: runtime-free primitive registrar generator (`lgprimgen`) — decouples the generator from the runtime it produces (the p0/p1 bootstrap problem)
- `feat(cmd/hoist-natives)`: safety analysis (5 skip categories: multi-name aliases, `//lg:native` dups, `.Lookup`-by-name, stdlib-`.lg` redefinitions, surviving-reference captures)
- `feat(vm,rt)`: guard-deviation diagnostics (`LG_GUARD_DEBUG`)
- `feat(rt)`: **the 222-primitive hoist** + regenerated artifacts

### Verification
`make check-generated` OK (bundle + lowered tree); `clojure.core` surface **unchanged at 807 publics**; `native-prims-intact?` true at boot and after `(require 'clojure.string)`; full `go test ./...` green.

222 of 292 closures were classified safe; the rest were correctly skipped (see the safety-analysis commit). Reviewed per-commit + a whole-branch pass.


mparrett:
The hoist holds up under measurement:

| lane | result on `a0778dbf` |
| --- | --- |
| bundled (`go test -short ./...`) | 0 failures |
| source bootstrap (`-tags bootstrap -short ./... -skip TestClojureTestSuite`) | 37 failures |

Those 37 are byte-identical to the set #627 produces on its own, so this PR adds none of its own. Your `807 publics` figure reproduces — `(count (ns-publics 'clojure.core))` is 807 both pre- and post-hoist. Bodies come across unchanged, `//lg:name` carries the Lisp names, and the per-body arity checks survive the `[]vm.Value` → `...vm.Value` signature change.

**The audit guard doesn't cover the case its commit message claims.** `405af75d` describes `TestAuditGeneratedPrimitivesCleanAfterInit` as the permanent regression guard that "would have caught the alias bug." On this branch it passes while the bootstrap lane is 37 red:

```
--- PASS: TestAuditGeneratedPrimitivesCleanAfterInit (0.00s)
--- PASS: TestAuditGeneratedPrimitivesFlagsMisregistration (0.00s)
```

`AuditGeneratedPrimitives` (`native_prims_lifecycle.go:163`) walks `genPrimBindings` and asserts each *generated* primitive resolves in its canonical namespace. They all do. What breaks is the namespace's source-defined siblings: `clojure.string/upper-case` is generated and guarded, `clojure.string/join` comes from `string.lg` and is gone, and nothing in the audit looks at it. Mechanism is in my #627 review.

Cheapest way to close the gap is to have the guard also assert one known source-only name resolves after a `require`. `clojure.string/join` is a good canary because `upper-case` is the only generated primitive in that namespace, so the two sit side by side.

**Every emitted adapter carries a dead arity check.** All 224 open with:

```go
if len(vs) < 0 {
    return vm.NIL, fmt.Errorf("wrong number of args (%d), expected at least 0", len(vs))
}
```

Always false. No correctness impact, since the real checks live in the bodies (`if len(vs) != 1 { ... }`), but the emitter could skip it when the minimum is zero rather than shipping 224 copies.

**Question on `cmd/hoist-natives`.** It's 1060 lines with no tests, and nothing references it — not the `Makefile`, not `scripts/generate.lg`, not a workflow. If more hoist batches are coming it earns its place and wants a smoke test on a small fixture. If it did its job once, a branch may suit it better than a permanent entry in `cmd/`. Either way works for me; I'd like to know which it is so I review it as a tool or as an artifact.

The `pkg/vm` additions (`AllVars`, `GuardDeviated`) read fine: snapshot under `RLock`, diagnostic-only, fast path untouched.

Nothing here needs to change before #627 is settled, which is where the blocker sits. Tell me which way you're taking `hoist-natives` and I'll re-run both lanes once the alias fix is in.

nnunley:
### Rebased onto current `main` — now supersedes #613's direct-call natives

Rebasing revealed that #613 ("direct-call natives for the hot `clojure.core` surface") landed while this stack was in flight, building the **same** `//lg:native` surface — and shipping an **arity-lossy subset** that was harmless only because it registered into the dead `clojure.core` namespace (the bug #627 fixes). Once #627 canonicalizes that namespace, the arity loss becomes live and breaks core fns.

**Reconciliation (this PR now owns the direct-call surface at full arity):**
- **#627 folded in** as the base commit (circular dependency — see #627).
- Removed #613's `native_prims.go` partials that this hoist supersedes: the 8 overlap natives (`assoc assoc! atom map? range re-find reset! symbol`) plus `not=` and `swap!` — each was a fixed-arity shadow of a variadic definition. The hoist's variadic `//lg:native` versions (and, for `not=`/`swap!`, the intact `lang.go` closures) reclaim them.
- Kept the 14 arity-complete `#613`-only natives (`conj deref get int name namespace nth push-binding! pop-binding! reduce some str subs upper-case`) for their direct-call perf; the full test suite confirms they behave correctly now that they're live bindings (they were previously dead code in the dead namespace).

**Verification (reconciled tree):** both build tags compile; 1287 tests pass; `check-generated` OK; `ir-stress` ratchet 2456/2467 native (11-failure baseline held); boot clean; the previously-broken forms `(symbol "a" "b")`, `(not= 1 2 3)`, `(assoc! (transient {}) :a 1 :b 2)`, `(swap! a f 1 2 3 4)` all evaluate correctly.

mparrett (changes requested):
Requesting changes for one source-bootstrap blocker.

The ordinary lane passes:

```text
go test -short -count=1 ./...
```

But the bootstrap lane fails on current head `c823f74e`. Focused repro:

```text
go test -tags bootstrap -run 'TestLGBParity/ns-require-string' -count=1 ./pkg/bytecode
```

This fails with `Can't resolve s/join in this context`; the full bootstrap lane cascades through other source namespaces. It also matches the currently failing GitHub Actions `build` check.

**[P1] Preserve the pending source load for registrar-created namespaces** (`internal/primgen/prims_emit.go:707`).

The emitter now calls `LookupOrRegisterNSNoLoad("clojure.string")`. During `rt` init that creates the canonical `string` placeholder, but it does not set `nsNeedsLoad`. Later `RequireNS("clojure.string")` reaches `LookupOrRegisterNS`, sees an existing namespace with `needsLoad == false`, and returns it without executing `string.lg`. Only the generated `upper-case` binding exists, so source-only siblings such as `join`, `capitalize`, and `blank?` never become resolvable.

Please mark a namespace created solely for generated bindings as still needing its real source/bundle load, without re-marking an already-loaded namespace, and add a bootstrap regression covering a source-only sibling such as `clojure.string/join`.

`make check-generated` reached its consistency check successfully. I did not find another correctness blocker in the hoisted bodies or registrar generation.

mparrett (approved):
Rechecked the updated head. The pending-load fix addresses the source-bootstrap blocker: the focused `clojure.string/join` repro passes, both the ordinary and bootstrap full test lanes pass locally, `make check-generated` passes, and all current GitHub checks are green. No remaining findings.

Issue #637, opened 2026-07-27
Nothing enforces `GENERATED-TRACKED` against `scripts/generate.lg`


mparrett:
- **Depends on:** #634 (introduces the list)
- **Epic:** #528

## Problem

#634 adds `GENERATED-TRACKED` to the Makefile: the committed generated
artifacts that `make check-generated` stashes, regenerates, and compares. The comment says "keep this list in step with `scripts/generate.lg`" and nothing checks that it is.

If `generate.lg` grows a sixth committed output and the Makefile list is not updated, the gate silently under-covers it. That is the failure #634 exists to fix, reintroduced one level up. Before #634 the gate compared only `core_compiled.lgb`, which is how `zz_primitives_generated.go` sat drifted from its generator without anything noticing.

## Where the outputs are declared

Four of the five outputs are already named at their call site:

```
(gen-go   "…/op_generated.head"        "pkg/ir/ir_ops.lg"    "pkg/ir/op_generated.go")
(gen-go   "…/ir_bridge_generated.head" "pkg/ir/ir_bridge.lg" "pkg/rt/ir_bridge_generated.go")
(gen-lisp "pkg/ir/ir_data.lg" "pkg/rt/core/ir/data/generated.lg")
(run! … "./cmd/lginterop" … "-primitives-out" "pkg/rt/zz_primitives_generated.go")
```

Only `core_compiled.lgb` is implicit, written inside `lgbgen --target=both` along with `generated.sums` and the gitignored lowered tree.

## Proposal

Have the pipeline record what it wrote, and have the gate audit its own list against that record.

`gen-go` and `gen-lisp` already funnel through helpers, so they can append their `out` path to an atom. `generate.lg` writes the collected list beside `generated.provenance`, which it already emits. `check-generated` regenerates before it compares, so the record is fresh by the time it reads it: the gate fails if any recorded path is missing from `GENERATED-TRACKED`.

No git, no filename heuristics, and it works under jj — the record comes from the pipeline rather than from the VCS.

**A missing or empty record must be an error, not a pass.** Otherwise the check degrades into exactly the silent under-coverage it exists to catch.

Residual gap: `core_compiled.lgb` still has to be named by hand, because `lgbgen` does not report its outputs. That shrinks a five-entry list in the Makefile to a one-entry declaration next to the `lgbgen` call, in the file that invokes it. Closing it fully means teaching `lgbgen` to report what it wrote, which is a separate change.

[...]

PR #640, opened 2026-07-27
feat: per-package //lg:native registrar surface (migrate corefns)

nnunley:
**Stack:** #627 → #639 → **this** → provenance-manifest. Sits on #639; until #627 and #639 merge, the diff includes their commits too (drops out as they land bottom-up).

Generalizes the primitive-registrar generator from a single `pkg/rt` target to a reusable **per-package surface**: any Go package can expose native primitives by annotating plain funcs (`//lg:native` / `//lg:ns` / optional `//lg:name`) and generating a committed, self-registering registrar in that package. Proven by migrating `pkg/rt/corefns` off its hand-written `NativeModule`.

### Commits (this feature)
- `feat(primgen)`: parameterize emitter target package name + import path
- `feat(primgen)`: package-level contribute/own mode via `//lg:bind` marker
- `feat(primgen)`: self-registering `init()` (rt → installer queue / external → direct) + shadow-warn wrap
- `feat(rt)`: exported `BindGeneratedPrimitive` (own-mode external binding API; designed, not yet wired)
- `feat(rt)`: rt's registrar self-registers via the installer queue (drops the explicit `lang.go:419` call) — **ordering-gated**: `native-prims-intact?` true, 807 publics, full suite green
- `feat(primgen)`: qualify `rt.*` runtime symbols + mode-aware imports for non-rt targets (rt output byte-identical)
- `feat(corefns)`: generate the native registrar from annotations (contribute mode) — deletes the hand-written `NativeModule`
- `build`: content-gate the corefns generated registrar in `check-generated`

### Design
Two modes: **contribute** (metadata-only `rt.RegisterNativeModule`, e.g. corefns → `clojure.core`) vs **own** (also binds vars). Own-mode external wiring (xxh3) and the non-annotated metadata path are captured as future work in the design spec.

### Verification
`make check-generated` OK; `go test ./...` green (1774 tests); reviewed per-commit + a whole-branch pass (READY TO MERGE).


mparrett:
Measured the same way as #639:

| lane | result on `1fe41f84` |
| --- | --- |
| bundled (`go test -short ./...`) | 0 failures |
| source bootstrap (`-tags bootstrap -short ./... -skip TestClojureTestSuite`) | 37 failures |

The 37 are the same set #627 and #639 produce, so this PR adds none of its own. All three carry exactly #627's breakage.

Init ordering is where this shape of change usually goes wrong, so that's what I checked, and the corefns migration comes out behavior-preserving. Before: `corefns.go`'s `init()` called `rt.RegisterNativeModule(...)`. After: the generated `init()` calls `RegisterGeneratedPrimitives()`, which makes the same call from the same package at the same point. `corefns` imports `rt`, so rt's init — `runInstallers()` included — is guaranteed complete first. The asymmetry with rt's own registrar going through `RegisterInstaller` is correct: rt can't call into itself before its namespaces exist, and corefns can because it's downstream. `TestCorefnsSelfRegistersDirectCallMetadata` covers the result.

**The #634 overlap lives here, not on #641.** I named `1fe41f84` in my #641 comment, which is true by nesting but points at the wrong PR to act on. The commit adds `pkg/rt/corefns/zz_primitives_generated.go` as a sixth committed generated output, gated by a third hand-written `cmp` block in `check-generated`. #634 replaces those blocks with a loop over a `GENERATED-TRACKED` list, so folding corefns into that list covers it and closes #637 at the same time. The ordering question from #641 stands unchanged; this is the commit to change once it's settled.

**Both generated files carry a stale banner.** They open with `Code generated by lginterop -primitives`, but `lginterop` stopped generating them in #639 — `lgprimgen` does now. The string is emitted at `internal/primgen/prims_emit.go:36` and `:70`. Cheapest to fix in this PR, since the artifacts are already being regenerated here and correcting the banner costs no additional digest churn.

**One commit-message correction.** `46856119` reads as though the migration decouples corefns from rt. It drops `pkg/rt` from `corefns.go`'s imports, but `zz_primitives_generated.go` adds it back, so the dependency moves into the generated file rather than going away. The real win — registration generated from annotations instead of hand-maintained — holds up on its own and is worth stating as the claim.

Nothing here needs to change before #627 is settled. Two asks when you get to it: fix the banner while the artifacts are already in flight, and say which way you want the corefns gate resolved so I know whether to fold it into #634's list or leave the third block alone.

nnunley:
Rebased onto current `main`. **Stack is now #639 → this → #641** (#627 was folded into #639 and closed). This PR also carries a small integration fix surfaced by the rebase: the core primitive registrar now self-registers via the installer queue (drained last by `zz_run_installers.go`), so `registerBuiltinsModule()` moved into that drain site to run **after** the primitives — preserving the EPIC-012 seam ordering (builtins override generated primitives for `vector`/`cons`/`contains?`/`array-map`/`nth`). Verified: 1289 tests pass, both build tags, `check-generated` OK.

internal/primgen/generate.go
```diff
@@ -0,0 +1,183 @@
...
+			allSpecs[i].GoPkg = "github.com/nooga/let-go/pkg/rt/builtins"
+		}
+	}
+
+	// Derive the target package name and path for emitFile
+	targetPkgPath := goPkg
+	if targetPkgPath == "" {
+		targetPkgPath = "github.com/nooga/let-go/pkg/rt/builtins"
+	}
+	pkgName := targetPkgPath[strings.LastIndex(targetPkgPath, "/")+1:]
```
mparrett: [P1] Preserve the source package name instead of deriving it from the import path. Go does not require a package clause to match the final import-path component; versioned module paths commonly look like `example.com/primitives/v2` while the files declare `package primitives`. With that valid input, lgprimgen currently emits `package v2`, and compiling the directory fails with `found packages primitives ... and v2`. `scanSource` already records `primSpec.Package`; derive and validate the emitted package name from the parsed source (and handle the empty-stub case explicitly) rather than using the `-go-pkg` basename.

mparrett (changes requested):
Requesting changes for one portability issue in the new per-package generator surface. I reviewed the PR-specific range against #639, ran the focused primgen/rt/corefns tests, and reproduced the inline failure through lgprimgen's public CLI. The current Go workflow failure is inherited from the #639 stack's namespace-load cascade, so I have not filed that again as a #640-specific issue.

internal/primgen/generate.go
nnunley: Fixed. The emitted package clause now comes from the parsed source (`primSpec.Package`, which `scanSource` already records) instead of the `-go-pkg` import-path basename. All scanned files must agree on the package name (hard error otherwise), and the empty-stub case reads the clause directly via a new `readPackageName` helper.

Regression test `TestGeneratePreservesSourcePackageName` covers your exact repro: import path `example.com/primitives/v2` with `package primitives` now emits `package primitives`, never `package v2`. For the real `rt`/`corefns` targets the source package name equals the basename, so the committed registrar output is byte-unchanged (verified: regeneration produced only a 1-line refresh in `pkg/rt/zz_primitives_generated.go` unrelated to this).

mparrett: @nnunley Reviewed fresh this morning — the original P1 is fixed and no longer blocking. One non-blocking issue remains: `TestGeneratePreservesSourcePackageName` currently uses `//lg:native clojure.core noop`, but the scanner recognizes only a standalone `//lg:native` (with separate `//lg:ns` / `//lg:name` directives). The test therefore discovers 0 specs and exercises only the empty-stub branch. The implementation itself passes the original valid-directive `/v2` repro, and the focused primgen/corefns/rt tests are green; it would just be good to correct the fixture so the regression test also covers the non-empty-spec path.

mparrett (approved):
Both asks are fixed. Re-reviewed at `6ec535b0` against current `main` (`cb99d7c5`), now that #639 has merged and the range is this PR's own work.

The P1 is resolved at the source: the emitted package clause comes from `primSpec.Package` with an agreement check across scanned files, and `readPackageName` handles the empty-stub case explicitly. The `/v2` repro that produced `package v2` now produces `package primitives`.

The fixture correction from my 30 July note is in too, and it moved coverage rather than just wording: the generator reports `1 specs` where it previously reported `0`, so the spec-based path is exercised.

Verification on this head: `make check-generated` OK across all four lockstep checks, `go test ./...` green.

Clearing my CHANGES_REQUESTED. Three follow-ups, none blocking.

**The regression test does not guard its own premise.** I reverted the fixture to the old `//lg:native clojure.core noop` spelling and re-ran `TestGeneratePreservesSourcePackageName`: it still passes, on `0 specs`, through the empty-stub branch. Both assertions hold on the stub output, so the regression we just fixed can return silently — a scanner change or a fixture edit puts it back with a green test. An assertion on the spec count, or on the emitted `noop` binding, pins the premise.

**The ordering comment in `zz_run_installers.go` claims a guarantee I cannot observe.** Moving `registerBuiltinsModule()` out of `lang.go`'s straight-line init converts the EPIC-012 seam from sequential code into a dependency on Go's filename-sort init order. That order is correct today — `zz_primitives_generated.go` sorts before `zz_run_installers.go` and nothing in `pkg/rt` sorts after it — so this is not a bug. But I inverted the two calls to see what enforces the precedence, and nothing does: `go test ./pkg/... ./test/...` stays green, and `vector`, `cons`, `contains?`, `array-map` and `nth` all resolve to the same functions at identical relative offsets under both orderings. The overlap the comment describes is real for four of them; `contains?` is not in the generated registrar at all. So either the precedence deserves a test that fails when the order is wrong, or the comment is describing something another mechanism already guarantees and should say so.

**Two new file walks, two different rules.** `hasBindMarker` scans build-constrained files; `readPackageName` skips them via `hasBuildConstraint`. Same PR, same shape of walk, opposite treatment. A `//lg:bind` in a `_wasm.go` file would select own mode for every build. Worth making consistent while both are new.

Two modes with a marker selecting between them reads clearly, and proving the generalization by migrating `corefns` off the hand-written `NativeModule` is what made it reviewable — `rt`'s generated registrar staying byte-identical is the evidence that carried it.

PR #642, opened 2026-07-27
fix(ir): give function-level recur its own op and lower it to OP_RECUR_FN

### LINKED ISSUE #638: IR→bytecode: function-level `recur` lowers to OP_TAIL_CALL instead of OP_RECUR_FN — `*ir-compile*` crashes when the recur is taken
Under `*ir-compile*`, a function-level `recur` that is actually taken crashes at runtime. The IR path lowers the recur terminator to `OP_TAIL_CALL`, which expects a callee on the stack that the terminator never carries. `OP_RECUR_FN` — the opcode for exactly this shape — is never emitted from the IR path.

Verified on `e5d560d1`, darwin/arm64.

## Repro

```clojure
(require 'ir.passes.pipeline)
(set! *ir-compile* true)
(defn sumr [i s n] (if (< i n) (recur (inc i) (+ s i) n) s))
(defn sumloop [n] (loop [i 0 s 0] (if (< i n) (recur (inc i) (+ s i)) s)))
(set! *ir-compile* false)

(println "loop/recur:" (sumloop 10))
(println "no-recur-taken:" (sumr 5 0 5))
(println "fn-level recur taken:" (sumr 0 0 10))
```

```
loop/recur: 45
no-recur-taken: 0
error: TypeError: let-go.lang.Int is not a function
```

`loop`/`recur` is fine — it lowers to a block-arg branch back to the loop header. Only the function-level spelling breaks, and only when the back-edge is taken.

## Mechanism

`build-recur` emits a `:tail-call` terminator whose refs are the replacement arguments alone, with no callee (`build.lg:1194` — the comment there says so: "`:tail-call` refs are just the new arg values (self-recursive, no fn ref needed)").

`lower.lg:526` emits that terminator as `:tail-call`, which maps to `vm.OP_TAIL_CALL` (`op_generated.go:202`). At runtime `OP_TAIL_CALL` (`vm.go:743`) resolves its callee with `f.nth(arity)` and calls `AsFn` on it. With no callee pushed, that reaches an argument instead — hence `Int is not a function`.

The VM already has the right opcode. `OP_RECUR_FN` (`vm.go:1010`) pops `arity` arguments, copies them into `f.args`, and resets `ip`/`sp` — which is precisely the function-level recur semantics. The bytecode compiler emits it; the IR path does not.

## Why nothing catches it

The `defn` macro's fallback to the bytecode compiler covers *unlowerable* forms. This form lowers successfully, to bad bytecode, so neither the silent fallback nor `*ir-compile-strict*` sees anything wrong. There is also no test for function-level `recur` on this path — `test/ir_compile_loop_test.lg` covers the `loop` spelling only.

## Fix

Give the IR a distinct terminator for function-level recur and map it to `OP_RECUR_FN`. Today `:tail-call` names two different shapes: the source-built recur back-edge (arguments only) and the bytecode-lifted `OP_TAIL_CALL` (callee plus arguments). Splitting them fixes this and removes the ambiguity that #636 runs into from the Go-lowering side —
#636's design comment proposes the same split for that reason.

Tests worth adding alongside: a taken function-level `recur` under `*ir-compile*` matches the VM, at several arities, and with the recur in both arms of an `if`.

Filing separately from #636 because it is a different backend and a wrong-code path on its own. Happy to take it.

mparrett:
Closes #638.

A function-level `recur` — `(defn f [..] .. (recur ..))` with no enclosing `loop*` — crashed under `*ir-compile*` the moment the back-edge was taken. Arity 3 gave `Int is not a function`; arities 1 and 2 gave `index out of range [-1]`.

`build-recur` emitted a `:tail-call` terminator carrying only the replacement arguments, and `lower.lg` maps `:tail-call` to `OP_TAIL_CALL`, which resolves its callee at `f.nth(argc)`. With no callee pushed, that read an argument or indexed below the frame base, which is also why the failure mode moved with the arity.

`OP_RECUR_FN` is the opcode for this shape and has been there all along: it pops `argc` arguments, copies them into `f.args`, and resets `ip`/`sp`. `recurCompiler` emits it for the same source. Only the IR path diverged.

## The fix

Split the two spellings that were sharing one keyword. `:tail-call` is a call: `Refs[0]` is a callee, it can dispatch anywhere, and it is what bytecode lifting decodes `OP_TAIL_CALL` into. `:recur-fn` carries only arguments and statically re-enters the executing function. Conflating them is what let a callee-less terminator reach a callee-resolving opcode.

This is not the silent bytecode fallback. Lowering reported success and emitted bad bytecode, so neither the fallback nor `*ir-compile-strict*` saw anything wrong.

## Consumers that move with the terminator

**structurize.** `structure-straight` matched `:tail-call`, so without updating it every function-level recur stops structuring and drops to the goto+label emitter — `sumr` goes from 1 goto/1 label to 4/4 — and under `LG_STRICT_STRUCTURED` the function is refused outright and falls back to bytecode. Updated, it structures as before and the rendered Go for these shapes is unchanged.

**liveness.** `:recur-fn` refs have to be seeded, or the back-edge arguments are dead-code eliminated.

**lower_go.** Its two back-edge emitters key on the new op, and the private helpers follow the name (`function-needs-recur-fn?`, `recur-fn-arg-expr`).

## A second defect with the same root

`inline-eligible?` recognized only recursion that goes through a `:load-var` of the callee's own name. A function-level recur re-enters through the terminator and never loads its own name, so such a callee read as inlineable.

Splicing it is a miscompile: the cloned back-edge rebinds parameter slots by index against the *caller's* frame, so a 1-arg callee inlined into a 2-arg caller writes the caller's `a0` while the loop body reads `a1`, and the caller spins forever. Latent behind `*enable-inline*`, which is opt-in, but wrong whenever it is on.

## Test gaps that kept this alive

`test/ir_compile_loop_test.lg` covered only the `loop*` spelling. `structurize_test.lg`'s corpus describes itself as "every shape build.lg can emit" while omitting function-level recur. An arm matching the old terminator kept passing in both. Both corpora now carry it.

## Validation

- `make test`, `go test ./pkg/...`, `make check-generated`: green
- gofmt clean; `GOOS=linux`, `GOOS=js GOARCH=wasm`, and `GOOS=plan9 GOARCH=amd64` all build
- New coverage: arities 1 through 3, permuted arguments, recur in either `if` arm, an untaken back-edge, and a million-deep run, each asserted against a bytecode-compiled twin as well as expected values
- Red-checked: reverting the terminator alone makes the `*ir-compile*` tests error, and reverting the structurize arm alone makes the two new corpus rows report 1 goto fallback each
- The inline miscompile was reproduced on a pre-guard build before the guard was written

### DIFF
--- a/pkg/ir/ir_ops.lg
+++ b/pkg/ir/ir_ops.lg
@@ -125,7 +125,18 @@
     ;; the bytecode defCompiler does), EXCEPT gogen must INTERN the var at
     ;; runtime (rt.InternVar) rather than look it up. .Refs[0] = var (Const
     ;; aux=*vm.Var), optional .Refs[1] = value. stack-in -1: 1 or 2 refs.
-    ["Def"       -1  1 false false nil                false true  false false "Def"        ".Refs[0] = var (Const aux=*vm.Var), optional .Refs[1] = value; interns + returns the var (2-ref also sets root)"]])
+    ["Def"       -1  1 false false nil                false true  false false "Def"        ".Refs[0] = var (Const aux=*vm.Var), optional .Refs[1] = value; interns + returns the var (2-ref also sets root)"]
+
+    ;; Function-level `recur`: rebind this function's parameters and
+    ;; re-enter it. Distinct from TailCall, which is a CALL — it carries a
+    ;; callee in Refs[0] and can dispatch anywhere. RecurFn carries only the
+    ;; replacement arguments and statically re-enters the executing function,
+    ;; which is why it maps to OP_RECUR_FN (pops argc args, resets ip/sp) and
+    ;; not OP_TAIL_CALL (resolves a callee via f.nth(argc), which a
+    ;; callee-less terminator does not supply). infer? is false for the same
+    ;; reason as TailCall: it infers :unknown via the default rather than
+    ;; joining typeinfer's terminator-ops set.
+    ["RecurFn"   -1  0 false true  OP_RECUR_FN        false false false false "RecurFn"    "function-level recur back-edge; .Refs = replacement args (no callee), .Aux = argc int"]])
 
 (g/check-spec ops)
 
--- a/pkg/ir/lisp_inline_test.go
+++ b/pkg/ir/lisp_inline_test.go
@@ -301,3 +301,39 @@ func TestDevirtualizesMultiBlockClosureCall(t *testing.T) {
 		t.Fatalf("multi-block capturing closure call should be devirtualized (no InvokeValueEC/BoxNativeFn):\n%s", src)
 	}
 }
+
+// TestInlineRejectsFunctionLevelRecurCallee pins the second spelling of
+// self-recursion. self-recursive? only spots a :load-var of the callee's own
+// name; a function-level `recur` re-enters through the terminator instead, so
+// without has-recur-fn? the callee looks inlineable.
+//
+// Splicing it is a miscompile, and the arity-mismatched case shows why: the
+// cloned :recur-fn rebinds parameter slots BY INDEX, so a 1-arg callee inlined
+// into a 2-arg caller writes the caller's a0 while the loop body reads a1. The
+// counter never advances and Usec2 spins forever. Guarded, the call survives as
+// a call (nooga/let-go#638).
+func TestInlineRejectsFunctionLevelRecurCallee(t *testing.T) {
+	ensureLoader()
+	v := runLispExpr(t, `(do (create-ns (quote inlinerecurz))
+	     (binding [*ns* (the-ns (quote inlinerecurz))]
+	       (eval (quote (defn cnt [n] (if (< n 0) 0 (recur (dec n))))))
+	       (eval (quote (defn usec2 [p q] (cnt q)))))
+	     (binding [ir.passes.inline/*enable-inline* true]
+	       (ir.passes.pipeline/lower-ns-to-go "inlinerecurz" (quote inlinerecurz)
+	         [(quote (defn cnt [n] (if (< n 0) 0 (recur (dec n)))))
+	          (quote (defn usec2 [p q] (cnt q)))])))`)
+	rendered, ok := v.Unbox().(string)
+	if !ok {
+		t.Fatalf("expected rendered Go string, got %T", v.Unbox())
+	}
+	// Definition plus a surviving call site.
+	if got := strings.Count(rendered, "Cnt("); got != 2 {
+		t.Errorf("expected Cnt to appear twice (definition + un-inlined call), got %d:\n%s",
+			got, rendered)
+	}
+	// The back-edge must stay inside Cnt: Usec2 gets no loop of its own.
+	usec := rendered[strings.Index(rendered, "func Usec2("):]
+	if strings.Contains(usec, "func_entry") {
+		t.Errorf("recur back-edge was spliced into the caller:\n%s", usec)
+	}
+}
--- a/pkg/ir/lisp_inline_test.go
+++ b/pkg/ir/lisp_inline_test.go
@@ -332,7 +332,11 @@ func TestInlineRejectsFunctionLevelRecurCallee(t *testing.T) {
 			got, rendered)
 	}
 	// The back-edge must stay inside Cnt: Usec2 gets no loop of its own.
-	usec := rendered[strings.Index(rendered, "func Usec2("):]
+	at := strings.Index(rendered, "func Usec2(")
+	if at < 0 {
+		t.Fatalf("no func Usec2( in rendered output:\n%s", rendered)
+	}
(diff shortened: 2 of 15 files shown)


nnunley (commented):
Reviewed against #638. The `:tail-call` → `:recur-fn` split is the right fix and is wired through build/lower/lower_go/structurize/liveness/inline/typeinfer, matching `recurCompiler`'s function-level branch; the tests assert against bytecode twins across arities 1–3, both `if` arms, an untaken edge, and a deep run.

One worth addressing:

**`:recur-fn` is missing from `pure-terminator-ops`** (`pkg/rt/core/ir/passes/purity.lg`, ~L70 — still `#{:branch :branch-if :return}`). A function-level recur is a side-effect-free back-edge, so blocks ending in `:recur-fn` should read as pure; without it, LICM/pure-block passes won't apply to recur-fn loops. Not a correctness bug — a missed optimization, and a catalog-coherence gap (the new terminator isn't reflected in every op-classifying set).

Nits: the new test in `pkg/ir/lisp_inline_test.go` does `rendered[strings.Index(rendered, "func Usec2("):]` with no guard — a `-1` from a future regression panics with a confusing message instead of failing cleanly; no variadic function-level-recur test (`(defn f [a & rest] (recur …))` is handled but untested); and `:tail-call` still sits in `ir-effectful-ops` (`lower_go.lg` ~L538) though build.lg no longer emits it — if it reappeared, structurize would silently `:goto`-fallback rather than erroring.

mparrett:
Thanks — all four addressed, plus the rebase. Now at `41aebd17`, MERGEABLE.

**`:recur-fn` in `pure-terminator-ops`** (`7ada4776`). You were right that this was a catalog-coherence gap: the terminator went into build/lower/lower_go/structurize/liveness/inline/typeinfer and stopped short of purity. Added with a comment tying it to `:branch` — a back-edge that rebinds args and jumps has no effect of its own.

**Rebase.** Eight behind, now zero. One data point for the `generated.sums` thread: the `merge=sums` driver did fire during this rebase and rewrote the digest itself, so this one needed no hand-resolve. I ran a full `--target=both` regen afterwards regardless, and `check-generated` reports the bundle and lowered tree in sync.

**The unguarded `strings.Index`.** Guarded, with a `t.Fatalf` naming the missing function.

**Variadic function-level recur.** Added `ir-fnrecur-var`/`bc-fnrecur-var` twins in the existing pattern, plus `ir-compile-fn-level-recur-variadic`. The rest slot changes on every back-edge — `(recur (dec n) (cons n xs))` — rather than being passed through, so it exercises the packed-rest rebinding and not only the fixed params. Untaken case included.

**`:tail-call`: I left the effectful sets alone.** It's still in `lower.lg:335` and `lower_go.lg:538`. Removing it looked worse than keeping it: if a `:tail-call` ever did appear, marking it non-effectful would let `effect-strictly-between?` move loads across it, which fails harder than the case we're guarding. Conservative seems right for those sets.

The exposure was narrower than the set membership. This branch replaced structurize's `(= op :tail-call) [:tail b]` arm with `:recur-fn`, so a stray `:tail-call` now falls to the `:else` arm and degrades to an unstructured `:goto` with a bumped fallback counter — silent, as you said. I added an explicit arm there that throws instead.

A throw on a reachable path would be its own regression, so I checked it isn't one. `build.lg` has no `:tail-call` emission left, the full suite is green, and the new throw fires in neither census: `ir-stress-gate` holds the ratchet at 2476/2487 with all 11 failures inside the known baseline, and `ir-stress-bytecode` runs the same corpus through the bytecode path. Zero occurrences in either per-defn log.

Happy to add the effectful-set cleanup on top if you'd prefer it — I'd want it as a deliberate "`:tail-call` is retired from the IR vocabulary" change rather than a one-line deletion, since `dump.lg` and `liveness.lg` still name it as a valid terminator.

nooga (approved):
LGTM. Verified the 07-30 review feedback (missing `:recur-fn` in `pure-terminator-ops`, unguarded slice in `lisp_inline_test.go`, missing variadic-recur test coverage) is all addressed by the follow-up commit, plus a bonus loud-fail guard against stray `:tail-call` terminators.

Verified:
- `go build ./...` clean
- `go test ./pkg/vm/... ./pkg/compiler/... ./pkg/ir/...` and the `.lg` integration suite (`ir-compile-fn-level-recur-*`, `structurize_test.lg`) all pass
- `generated.sums` regen is clean, no drift

One non-blocking observation: `:tail-call` looks to be dead as an IR terminator now — nothing under `pkg/rt/core/ir/` emits it anymore, though the plumbing for it (`ir-effectful-ops`, `dump.lg`'s `terminator-ops`) is left in place. Might be worth a follow-up to just delete it, now that structurize throws loudly if it ever resurfaces. Not blocking this PR.

Issue #644, opened 2026-07-29
vm: `Frame.Run` is Go-recursive — deep lg→lg recursion aborts the process


mparrett:
## Summary

Every non-tail call that dispatches directly to a bytecode function consumes a Go stack frame, because `Frame.Run` reaches a nested `Frame.Run` through `ec.Invoke`. Past roughly 350k levels the Go runtime hits its 1 GB goroutine stack limit and the process dies with `fatal error: stack overflow`: an unrecoverable runtime abort, not a let-go error. No `try` catches it, `RunProtected` never sees it, and an embedding host loses the process.

This is the non-tail case #620 puts out of scope. #620 defers it to #464, but #464 is *Epic: Runtime performance & concurrency* and says nothing about a trampoline or about `Frame.Run`. There is no issue tracking this work. That's what this one is for.

## Measurements

`dc316a0c`, darwin/arm64, Go 1.26.3, default 1 GB max stack.

```clojure
;; non-tail recursion
(defn cd [n] (if (= n 0) 0 (+ 1 (cd (- n 1)))))
(println (cd 375000))
```

```clojure
;; tail call through a closure with a real capture
(def ev? nil) (def od? nil)
(let [zero 0
      e (fn [n] (if (= n zero) true  (od? (- n 1))))
      o (fn [n] (if (= n zero) false (ev? (- n 1))))]
  (def ev? e) (def od? o))
(println (ev? 500000))
```

| Shape | Depth | Result |
|---|---|---|
| top-level `*Func` mutual tail call | 5M | fine — `OP_TAIL_CALL` reuses the frame |
| capturing `*Closure` tail call | 100k | fine |
| capturing `*Closure` tail call | 500k | **`fatal error: stack overflow`** |
| non-tail self recursion | 300k | fine |
| non-tail self recursion | 375k | **`fatal error: stack overflow`** |

That works out to roughly 2.9 KB of Go stack per recursive invocation path (1 GB / ~350k). `Frame.Run` is a single large switch and appears to dominate that path, but the measurement includes its `ec.Invoke`/`invokeIn` wrappers too. The ceiling is therefore sensitive to otherwise unrelated changes along the call path.

Two notes for anyone reproducing this.

[...]

Issue #620, opened 2026-07-22
Design: constant-space tail calls on the explicit-frame VM (closures, multi-arity, apply)


nnunley:
## Summary

#645 changes the starting point for #620.

The compiler already emits `OP_TAIL_CALL` for ordinary calls in tail position. Before #645, a tail call to a `*Closure` or `*MultiArityFn` fell through `ExecContext.Invoke` into a nested `Frame.Run`, so deep recursion grew the Go stack. #645 makes those direct bytecode spans Go-stack-safe by descending through an explicit parent-linked frame chain.

That is not yet general tail-call elimination. A closure or multi-arity tail call still allocates and retains one child `Frame` per call. The remaining #620 goal is therefore:

> Reuse the current VM frame for every resolvable bytecode tail call, including closures and multi-arity functions, and make `apply*` capable of handing its target back to the VM tail loop without exposing a continuation as a let-go `Value`.

The implementation should be stacked on #645. It must not introduce a second callee resolver or a second frame-acquisition path.

## State after #645

| Call path | #645 behavior | Remaining #620 work |
|---|---|---|
| Non-tail direct bytecode call | Descend to a child heap frame; flat Go stack | None for Go-stack safety; non-tail depth remains linear heap growth |
| Tail `*Func` | Existing same-frame reuse | Move onto the shared resolver/transition path |
| Tail `*Closure` | Descend to a child frame | Reuse the current frame while installing captures |
| Tail `*MultiArityFn` | Resolve the variant, then descend | Reuse the current frame after variant selection |
| Tail metadata-wrapped bytecode fn | Unwrap, then descend | Reuse through the same target preparation path |
| Tail native leaf | Invoke and return its value | Preserve terminal behavior and error attribution |
| Tail `apply*` | Native calls public `ExecContext.Invoke`, hiding the target behind a new invocation segment | Return a private pending-call outcome to the tail loop |
| Tail `ProtocolFn` | Protocol dispatch calls public `ExecContext.Invoke` | Resolve the implementation into the shared raw path, if included in v1 |
| Tail `MultiFn` | Dispatch function and selected method are nested invocations | Requires a resumable dispatcher; explicitly include or defer |

### Goals for v1

- Constant Go **and VM-frame** space for tail calls to `*Func`, `*Closure`, and `*MultiArityFn`, including metadata wrappers and closures over multi-arity functions.
- Tail-transparent `apply*` without allowing an internal continuation to escape through `Value`, variables, collections, or public host APIs.
[...]

PR #645, opened 2026-07-29
vm: make direct bytecode calls non-recursive with an explicit frame chain

mparrett:
## Summary

This PR makes contiguous direct bytecode-call spans non-recursive in Go by running child `Frame`s inside one interpreter dispatch loop.

- Replaces the implicit `Frame.Run` → `ec.Invoke` → `Frame.Run` call chain with explicit parent links between live heap frames.
- Descends directly through `MetaFn`, `Func`, `Closure`, and `MultiArityFn` bytecode callees while preserving arity checks, variadic packing, and closure captures.
- Routes normal returns and errors through the explicit chain, releasing completed and failed child frames.
- Preserves logical frame entry/exit behavior for `*lg-trace*`, opcode and opcode-pair profiling, and allocation-attribution shadow stacks.
- Adds focused regressions for multi-arity selection, captures, lifecycle ordering, profiler boundaries, and error-path frame release.

This is the implementation prototype discussed in #644. It intentionally does **not** close that issue yet; the resource-budget and native-callback boundaries below remain design work.

## Why

Every ordinary non-tail bytecode call currently reaches a nested `Frame.Run` through `ec.Invoke`. Deep recursion therefore grows the Go stack until the runtime aborts the process with `fatal error: stack overflow`. That abort is not a let-go `ExecutionError`, cannot be caught by `try`, and takes down an embedding host.

Frames are already heap objects. The missing state is the parent-child link currently represented by the Go call stack. This patch makes that link explicit and lets one dispatch loop descend and return without recursively entering Go.

## Observed depth

- The original single-arity non-tail and capturing-closure repros run to 10M levels without Go-stack growth.
- The corrected multi-arity repro now completes at 1M levels; the first prototype still overflowed because `MultiArityFn` fell back to `ec.Invoke`.
- Nested catch/finally propagation remains intact.

The claim is deliberately scoped to contiguous bytecode-call spans. A native builtin that calls back into let-go can still start another Go-stack segment.

## Corrected throughput measurements

Fresh stripped binaries from `dc316a0c` and this branch, using the preserved interleaved ABBA harness:

| workload | rounds | A/A median (range) | branch vs baseline |
|---|---:|---:|---:|
| `fib(32)` | 25 | -0.25% (-12.78, +6.77) | **-4.25%** (-13.22, +0.45) |
| `reduce`/`map` | 25 | +0.35% (-6.98, +5.92) | **+4.22%** (-2.70, +11.61) |
| `loop`/`recur` control | 12 | +1.15% (-3.43, +9.11) | -5.08% (-18.22, +7.89) |

Lower is faster. `fib` favored the branch in 23/25 paired rounds; `reduce`/`map` favored baseline in 24/25. The call-free control is too noisy to interpret. This is therefore workload-dependent, not a blanket throughput win.

## Validation

- [x] `go test -count=1 -timeout 60s ./...`
- [x] `go test -race ./pkg/vm -count=1`
- [x] Existing `TestLgTraceOutputAndPropagation`
- [x] Existing transduce allocation ratchet
- [x] 1M-level multi-arity repro
- [x] Nested catch/finally repro
- [x] Focused lifecycle, profiler, capture, and failed-frame-release tests

## Why this remains draft

- **Resource budget:** non-tail depth becomes linear heap growth. A configurable live-frame or execution budget is still needed to turn runaway recursion into a catchable `ExecutionError` rather than a possible heap OOM.
- **Native callback boundary:** builtins such as `reduce` that synchronously call back into let-go can still accumulate Go-stack segments under alternating recursion. Removing that boundary requires a resumable native-call protocol, CPS across the builtin surface, or an equivalent continuation mechanism.
- **Shared resolver:** #620 tail-frame reuse and this non-tail descent need one callee-preparation/frame-transition seam. This prototype proves the runtime shape but still duplicates some preparation logic from `Func`/`Closure` invocation.
- **Performance:** the corrected collection workload regresses 4.22% and needs profiling before a merge-ready disposition.
- **#618 interaction:** #618's current `OP_CALL_SELF` path enters `child.Run()` recursively, so it does not compose with this design as written. Its remaining static-callee shortcut should be evaluated inside this explicit descent path rather than merged unchanged.

## Relationship to CPS

CPS remains a legitimate broader compiler/backend strategy. Go would still need continuation applications lowered to jumps, a trampoline, or an explicit abstract machine; after defunctionalization that runtime state resembles this frame chain. This draft is the smaller runtime experiment needed to address `Frame.Run` recursion without first replacing the compiler IR.

### DIFF
--- a/pkg/vm/explicit_frame_test.go
+++ b/pkg/vm/explicit_frame_test.go
@@ -1,6 +1,9 @@
 package vm
 
-import "testing"
+import (
+	"strings"
+	"testing"
+)
 
 func testBytecodeFnReturningArg(consts *Consts, arity int) *Func {
 	chunk := NewCodeChunk(consts)
@@ -22,7 +25,18 @@ func testInvokeZeroChunk(consts *Consts, fn Fn) *CodeChunk {
 	return chunk
 }
 
-func TestChildFrameForSelectsMultiArityAndPreservesCaptures(t *testing.T) {
+func testTailCallZeroChunk(consts *Consts, fn Fn) *CodeChunk {
+	chunk := NewCodeChunk(consts)
+	chunk.Append(OP_LOAD_CONST)
+	chunk.Append32(consts.Intern(fn))
+	chunk.Append(OP_TAIL_CALL)
+	chunk.Append32(0)
+	chunk.Append(OP_RETURN)
+	chunk.SetMaxStack(4)
+	return chunk
+}
+
+func TestResolveBytecodeCallSelectsMultiArityAndPreservesCaptures(t *testing.T) {
 	consts := NewConsts()
 	oneArg := testBytecodeFnReturningArg(consts, 1)
 	twoArg := testBytecodeFnReturningArg(consts, 2)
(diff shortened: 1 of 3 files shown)


nnunley (commented):
Reviewed with a focus on the frame-chain error paths (multi-agent pass + manual verification). The non-recursive frame-chain model looks sound. I checked the "stack corruption when a caught `childFrameFor` error is handled" theory and it does **not** hold — `handleError` resets `f.sp` to the handler snapshot (`vm.go:569`), so a caught arity error unwinds the peeked callee/args cleanly. So the two items below are the real ones.

**Leftover PROBE scaffolding.** The OP_RETURN unwind path has `panic(fmt.Sprintf("PROBE: parent ip %d holds %s, not a call opcode", …))` plus several `// PROBE:` comments. A `panic()` on a frame invariant in the hot dispatch loop crashes the process (bypassing try/catch) instead of surfacing an error — worth removing or converting to a real error before merge.

**OP_TAIL_CALL closure arity handling is inconsistent with OP_INVOKE.** When `childFrameFor` returns `cerr != nil` for a `*Closure` tail call, the code drops `cerr` and falls through to `ec.Invoke(fn, a)`. The error still surfaces (Invoke re-validates arity), so it isn't lost — but the message/path differs from OP_INVOKE (which wraps `cerr` + `handleError`) and does redundant work. Worth making the two consistent.

Nits: missing bounds checks before `f.code.code[f.ip]`/`[f.ip+1]` in the unwind path and `wrapCallSite()` (only reachable on an already-corrupt IP); `childFrameFor` duplicates the arity/variadic-packing logic in `Func.invokeIn` (drift risk, since EPIC-007 assumes bytecode/runtime parity); and `enterFrame` sets `f.prevOp = 0` while the comment says the parent's prev-op is preserved (profiler cross-frame transition is lost).

mparrett:
Addressed the review feedback in `160fb23`.

- Replaced the `PROBE` panic with `suspendedCallArity`, which validates the frame/code, both `ip` reads, opcode, stack depth, and arity before unwind. `wrapCallSite` now uses the same checked path, and continuation failures surface as catchable `ExecutionError`s instead of panics or unchecked drops.
- Extracted `resolveBytecodeCall` as the prepared-target resolver described in #620. It owns metadata unwrapping, multi-arity selection, closure captures, arity validation, and variadic packing; frame allocation and same-frame installation remain separate transitions.
- Made `OP_TAIL_CALL` resolve errors through the same wrapped/`handleError` path as `OP_INVOKE`.
- Routed zero-arity closure and multi-arity tail calls through direct bytecode descent, closing the remaining recursive `Frame.Run` path.
- Made argument ownership explicit for same-frame tail replacement: fixed-arity slices borrowed from `f.stack` are copied into a reusable frame-owned buffer before stack reset, and variadic packing always uses fresh storage. This removes the old `append(sargs, restlist)` aliasing hazard.
- Clarified the profiler comment: `enterFrame` resets the newly entered child, while the suspended parent retains its own `prevOp`; cross-frame pairs remain intentionally omitted, matching the existing regression.

Added regressions for continuation validation, variadic source immutability, frame-owned tail arguments, tail/invoke arity parity and catchability, and zero-arity closure descent.

Validation:
- `go test -count=1 -timeout 60s ./...`
- `go test -race -count=1 ./pkg/vm`
- `go vet ./pkg/vm`
- `git diff --check`

The repository's configured `golangci-lint` binary was not available locally, so CI will provide that remaining check.

nooga (approved):
LGTM. The follow-up commit addresses both items from the prior review — the PROBE-panic on frame-invariant violation is now a catchable `*ExecutionError` via `suspendedCallArity`, and `OP_TAIL_CALL` now routes through the same `resolveBytecodeCall` as `OP_INVOKE` so the arity-error handling is provably consistent between them. It goes further than asked: `installBytecodeCall` now copies borrowed-stack args into `f.argbuf` before resetting the frame stack for same-frame tail transitions, with a dedicated test (`TestInstallBytecodeCallOwnsBorrowedArgs`) pinning that invariant.

Verified:
- `go build ./...`, `gofmt -l pkg/vm/*.go` clean
- `go test ./pkg/vm/ -race -count=1` clean
- Full suite (`go test ./...`) passes; `.lg` integration suite passes
- Frame release/pool-poisoning behavior on failed descents checked explicitly (`TestDescendedErrorReleasesChildFrame`)

Two things I'm treating as accepted, disclosed scope rather than blockers — flagging for the record:
1. Unbounded non-tail recursion into a direct bytecode callee now grows the heap toward OOM instead of hitting a Go stack overflow — still uncatchable either way, no resource budget yet. Fine to land as a step, but let's not lose track of the budget follow-up.
2. Self-reported perf: `reduce`/`map` +4.22%, `loop`/`recur` control -5.08%, `fib` -4.25% (faster). Workload-dependent; worth keeping an eye on in the ratchet once #564/#655 land.

PR #653, opened 2026-07-30
fix(aot): register a boxed override for typed-return fns (#554)

### LINKED ISSUE #554: AOT lowering: a fully-typed fn signature gets no boxed wrapper or override registration
## What

When a defn lowers to a fully-typed native signature — return type inferred to a concrete Go type instead of `vm.Value` — the emitter generates neither a boxed wrapper nor a `RegisterGoOverrides` entry for it, so the native fn can't be drained as an override. Only fns that lower to the boxed shape `func(ec, args []vm.Value) (vm.Value, error)` get registered.

## Repro

`benchmark/aot/src/bench/loop-recur.lg` (from #552) hits this. The hot loop lowers cleanly to a typed leaf:

```go
func SumLoop(ec *vm.ExecContext) int {
    ...
    for { ...; acc = acc + i; i = i + 1; continue }
}
```

but `RegisterGoOverrides("bench.loop-recur", ...)` only registers `"run"` — the boxed 0-arg wrapper `vm.Int(SumLoop(ec))`. There's no `"sum-loop"` entry, so `SumLoop` is unreachable as an override on its own. The fixture works around it by keeping the typed loop in its own defn behind a boxed `run`, which is why #552's README documents the split.

## Why it matters

A hot, fully-typed leaf fn is the case you'd most want to dispatch natively, and today it's unreachable as an override without a hand-written boxed trampoline. `build.lg`'s native-dispatch check would also reject such a fn as a silent bytecode fallback, so a fixture can't expose one directly.

## Suggestion

Auto-generate a boxed wrapper plus a `RegisterGoOverrides` entry for typed-signature fns — box the return, adapt the arg vector — so a typed leaf drains like any other override without the hand-split.

Adjacent issues: #425 (native-entry / silent perf-cliff) and #531 (consolidate the registration + interop mechanisms). This could fold into either, but the specific typed-return case isn't captured in either today.

mparrett:
## Summary

A fn that lowers to a fully typed signature registered no override at all, so the lowered Go was emitted and then never reached — the var kept its bytecode fn and every dynamic call ran the VM. `override-uniform-value?` demanded `vm.Value` in *and* out; a typed return failed it.

This admits the case where the params are already uniform `vm.Value` and only the return is typed, and boxes the result in the wrapper:

```go
return vm.Int(SumLoop(ec)), nil
```

Closes #554.

## Why it matters

The canonical shape is a loop whose result infers to `int`, which is what `benchmark/gloat/loop-recur.clj` is — the loop sits directly in `-main`, so there is no inner fn to register and nothing dispatches. Measured on darwin/arm64, gloat v0.1.62 `-E lglvm`, let-go v1.12.2, hyperfine 3 warmup / 10 runs:

| leg | mean |
|---|---|
| VM (`lg loop-recur.clj`) | 60.5 ms ± 0.7 |
| AOT, as gloat builds it today | 60.0 ms ± 0.7 |
| AOT, same program with a boxed return | 6.9 ms ± 1.5 |

8.7x, from lowering that was already happening.

fib doesn't hit this because `fib` is its own registerable fn: the VM runs `-main` as bytecode, calls `fib`, and that first call enters the native island where the recursion stays. loop-recur has no inner fn, so the entry's registration is the only door in.

This is also the shape external comparisons compile. cljgo's published AOT table has let-go's loop-recur at 37.3 ms against a 36.6 ms VM leg — the same null result on their hardware.

Multi-arity fns need the same treatment and get it here: `multi-arity-override-groups` required *every* arity to be uniform `vm.Value`, so `(defn choose ([x] (= x 1)) ([x y] (= x y)))` emitted `Choose_1`/`Choose_2` returning `bool` and no `init()` at all. Each arm boxes independently, since one arm may return `vm.Value` while another returns a primitive.

## The two judgement calls

**Params are deliberately not coerced.** Only the return side is lifted. Unboxing an arg needs a proof that the runtime value really is that primitive; the direct-call path gets that at the call site (`coerce-arg-to-type` only direct-calls a typed param when the arg is proven that type, else falls back), but a dynamic override boundary has no such proof. A typed *param* therefore still stays on bytecode, and `TestTypedParamStillSkipsOverride` pins that.

**Typed return plus error looks unreachable, so the wrapper stays single-branch.** A body containing a call that can error carries its result as `vm.Value`, so it never presents a typed return; I found no shape that produces the pair. `override-entries` gates on `(not (:needs-error? r))` to state that rather than emit a two-value branch nothing exercises.

## Fixture cleanup

`benchmark/aot/src/bench/loop-recur.lg` kept the loop in a separate `defn` behind a boxed `run` purely to work around this — its header comment said so. The loop moves back into `run`, and `build.lg`'s native-dispatch check (which fails the build if any fixture silently falls back to bytecode) still passes on all 7 fixtures. The row also got faster, 10.15x over the VM against 6.58x with the indirection.

## Test plan

- [x] `go test -short -timeout 60s ./... -skip TestClojureTestSuite`
- [x] `make test` and `make build`
- [x] `bash benchmark/run.sh loop-recur` — `aot-build: OK (all 7 fixtures dispatch native)`
- [x] `gofmt -l`; `GOOS=linux`, `GOOS=js/wasm`, `GOOS=plan9/amd64` builds
- [x] CI green

`TestLowerNsInlinesTypedSibling` needed a follow-up commit. Its callee returns `bool`, so it now registers a boxed override, and the test proved inlining by counting `Callee(` across the whole file — the new wrapper made that 2. The inlining was still correct; the assertion moved into `Caller`'s body, where the invariant actually lives.

### DIFF
--- a/benchmark/aot/src/bench/loop-recur.lg
+++ b/benchmark/aot/src/bench/loop-recur.lg
@@ -1,15 +1,14 @@
 ;; Tight loop with tail-call recursion — AOT fixture (same workload as ../loop-recur.clj)
 ;;
-;; The loop lives in its own defn: lowered it becomes a fully typed native
-;; (func(ec) int), which the emitter does not register as an override on its
-;; own; `run` wraps it with a boxed signature so the override registers and
-;; dispatches into the native loop via an intra-ns direct call.
+;; The loop lives directly in `run`: it lowers to a fully typed native
+;; (func(ec) int), which registers as an override because the wrapper boxes the
+;; typed return (#554). Before that, a typed signature registered nothing, and
+;; the fixture had to keep the loop in a separate defn behind a boxed `run` to
+;; get native dispatch at all.
 (ns bench.loop-recur)
 
-(defn sum-loop []
+(defn run []
   (loop [i 0 acc 0]
     (if (< i 1000000)
       (recur (inc i) (+ acc i))
       acc)))
-
-(defn run [] (sum-loop))
--- a/pkg/ir/lisp_typed_directcall_test.go
+++ b/pkg/ir/lisp_typed_directcall_test.go
@@ -14,7 +14,10 @@ package ir_test
 
 import (
 	"regexp"
+	"strings"
 	"testing"
+
+	"github.com/nooga/let-go/pkg/vm"
 )
 
 func TestLowerNsInlinesTypedSibling(t *testing.T) {
@@ -29,7 +32,11 @@ func TestLowerNsInlinesTypedSibling(t *testing.T) {
 		       (ir.passes.pipeline/lower-ns-to-go "typedres" (quote typedres)
 		         [(quote (defn callee [x] (= x 1)))
 		          (quote (defn caller [y] (callee y)))])))`)
-	src := v.String()
+	vs, ok := v.(vm.String)
+	if !ok {
+		t.Fatalf("expected rendered Go source string, got %T", v)
+	}
+	src := string(vs)
 
 	// callee still lowers as its own bool-returning fn (public → PascalCase per #371).
 	if !regexp.MustCompile(`func Callee\(ec \*vm\.ExecContext, [a-z0-9_]+ vm\.Value\) bool`).MatchString(src) {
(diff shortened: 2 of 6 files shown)


nooga (approved):
LGTM. Verified the fix and its gating logic empirically — the four new tests in `pkg/ir/lisp_typed_return_override_test.go` cover the single-arity fix, the typed-param exclusion boundary, and multi-arity per-arm boxing, and I confirmed `override-eligible?`/`override-boxable-return?` are mutually exclusive by construction so the multi-arity gating can't leak a fallible arm through.

Verified: `go build ./...`, `go test ./... -count=1` (skip clojure-test-suite) all pass; `gofmt -l` clean on touched files.

One thing worth fixing, not because it's a bug today but because the comment is actively misleading: the docstrings on `override-wrapper-fn-lit`/`override-entries` (`pipeline.lg:900-904`, ~1058) claim "typed return + needs-error is unreachable." I built a counterexample — `(defn root2 [] (clojure.math/sqrt 4.0))` — where `clojure.math/sqrt` is typed `:float` but isn't recognized as error-free, so the pair *does* occur; it's correctly excluded by the explicit `(not (:needs-error? r))` guard, not by structural impossibility. If someone later reads "unreachable" and "simplifies" by dropping that guard as redundant, they'd reintroduce a real bug (silently dropping a Go-level error). Please reword the comment to say the guard is load-bearing.

(Unrelated aside, filed for awareness not action here: the generated code for that same `root2` example doesn't actually compile today — `ec.Invoke(...)` returns `vm.Value` assigned directly to a `float64` — pre-existing bug in the general typed-return + fallible-call path, worth its own issue.)

PR #654, opened 2026-07-30
refactor(primgen): format registrar in-process with go/format; fix false "pure go/ast" docs

nnunley:
`internal/primgen`'s package doc (and `scripts/generate.lg`) claimed the primitive-registrar generator *"depends only on go/ast + text templates."* An audit found that inaccurate: it scans with `go/ast`, but **emits by string concatenation** (there is no `text/template` anywhere) and **formatted by shelling out to the external `gofmt` binary** (`os/exec`).

This PR:

1. **Removes the `gofmt` subprocess.** `gofmtCode` (renamed `formatGo`) now formats in-process with `go/format.Source()` — gofmt's own parser + printer, so the output is **byte-identical** (the regenerated 241-spec `pkg/rt/zz_primitives_generated.go` is unchanged) — dropping the `os/exec` dependency on a `gofmt` executable being on `PATH`.
2. **Corrects the two false doc claims** to describe what the generator actually does: scans annotated signatures with `go/ast`, builds the registrar by string concatenation, and canonicalizes it in-process with `go/format` — keeping the true and meaningful invariant (runtime-free: it does not import `pkg/compiler`/`pkg/rt`, so it runs mid-migration when the runtime can't boot).

No behavior or output change — purely removing an external-tool dependency and making the docs honest. `make check-generated` is clean; the registrar is byte-for-byte identical. The deeper cleanup — the string-concatenation emitter → a `.lg` generator on the compiler surface — is a separate, larger change and is out of scope here.


mparrett (approved):
Approved — no blocking findings.

**Optional / non-blocking:** If this PR is intended to complete the documentation cleanup, the old “pure go/ast” description still appears in `Makefile:74` and `cmd/lgprimgen/main.go:16`. Consider aligning those comments with the new “runtime-free; go/ast scan, string emission, in-process go/format” wording here or in a follow-up. This does not affect behavior and is not a merge blocker.

Verified locally:
- `go test ./internal/primgen`
- regenerated the 241-spec registrar; the committed output remained byte-for-byte clean
- `make check-generated`
- `git diff --check`

All current GitHub checks are passing.

PR #655, opened 2026-07-30
bench: compare against a pinned-release baseline

mparrett:
## Summary

The VM and AOT legs are both built from the working tree. A regression in shared runtime code moves both, the ratio between them stays flat, and the harness reports parity — so that whole class of regression is invisible to it by construction.

It just hid one. `reduce` is 1.66x slower than v1.12.2 on current main, while `run.sh` reported let-go 69.1 ms / let-go AOT 69.1 ms and flagged nothing.

This adds a third leg built from a pinned release, running the identical driver and fixture, plus a per-workload regression check.

## What the check reports

Running on current main, all seven workloads:

```
=== Regression check vs v1.12.2 ===
  fib                2560.7 ms vs   2416.8 ms   1.06x  ok
  loop-recur           74.9 ms vs     71.6 ms   1.05x  ok
  map-filter            8.1 ms vs      8.2 ms   0.99x  NOISY (sigma 41% — rerun on an idle machine)
  persistent-map       18.1 ms vs     18.5 ms   0.97x  ok
  reduce               68.4 ms vs     41.3 ms   1.66x  REGRESSION
  tak                2717.0 ms vs   2925.7 ms   0.93x  faster
  transducers          89.3 ms vs    112.1 ms   0.80x  faster
```

One workload trips, six don't, and the noisy one is suppressed rather than mis-asserted. `results.md` grows a `baseline <ref>` column, so the same information survives into the committed table.

`BASELINE_STRICT=1` turns the check into a non-zero exit for CI. `BASELINE_REF=none` skips the leg entirely. A baseline that fails to build warns that regressions won't be visible in that run, and continues.

## Three important details

**The ref is pinned in a file, not derived from the newest tag.** `git tag --sort=-v:refname` puts the stale `v2.0.x` line on top, which is two months older than the actual latest release. Auto-detection would silently pick the wrong reference and the whole check would mean nothing.

**The baseline builds in a detached worktree, never in-tree.** `make`'s regeneration rules are timestamp-driven and `git checkout` rewrites mtimes, so building an old ref over the current tree can pair a stale `core_compiled.lgb` with freshly compiled Go. Those binaries mismeasure by ~1.8x, and they do it *consistently* — the error survives repeated runs and survives reversed-order A/B, so it reads as a real regression rather than as noise. A fresh worktree has clean mtimes and no stale artifacts.

**The check compares medians and refuses to assert on a noisy run.** If either leg's relative sigma exceeds 20% it prints `NOISY` and doesn't trip the gate. A first cut using means reported 1.90x on a machine that turned out to have an unrelated process pegging a core; medians plus the sigma guard make one scheduling hiccup unable to cry wolf.

Binaries are cached under `benchmark/.baseline/<sha>/` (gitignored), so repeated runs don't rebuild.

## Test plan

- [x] `bash benchmark/run.sh reduce` — flags the known regression
- [x] `bash benchmark/run.sh` (full) — completes, `results.md` renders with the baseline column, check runs across all seven workloads
- [x] Decision logic exercised against synthetic inputs for all four branches: ok (1.02x, exit 0), REGRESSION (1.77x, exit 3), faster (0.85x, exit 0), NOISY (1.77x but sigma 31%, exit 0 — correctly declines)
- [x] `bash -n benchmark/run.sh`
- [x] CI green

`benchmark/results.md` is deliberately **not** updated here: every run I had available was on a machine under external load, and the committed table should be regenerated somewhere quiet. The numbers quoted above are directionally right but shouldn't be pasted into the file.

## Note

The `reduce` regression this surfaces bisects to #639 and is not fixed here — this PR is only the instrument that makes it visible. Filing that separately.


benchmark/run.sh
```diff
@@ -182,18 +237,65 @@ for bench in "${BENCHMARKS[@]}"; do
...
+# This is the part that catches a regression in shared runtime code. Comparing
+# the VM leg to the AOT leg cannot: both are built from the working tree, so a
+# shared regression moves both and the ratio between them stays flat.
+REGRESSION_FOUND=0
+if [ -n "$BASELINE" ]; then
+    echo ""
+    echo "=== Regression check vs $BASELINE_REF ==="
+    THRESHOLD="${REGRESSION_THRESHOLD:-1.10}"
+    for bench in "${BENCHMARKS[@]}"; do
+        name="$(basename "$bench" .clj)"
```
nnunley: Make sure to sanitize the bench name from BENCHMARKS to avoid having a exploitable vector here.

nnunley (dismissed):
Just realized that the benchmark base names are inside of an evaluated shell escape, so we should try to make sure that we sanitize the strings to avoid a side-channel attack.

benchmark/run.sh
mparrett: Thanks — addressed in f23e755.

mparrett:
> Just realized that the benchmark base names are inside of an evaluated shell escape, so we should try to make sure that we sanitize the strings to avoid a side-channel attack.

Thanks, good catch. Addressed in f23e755.

nooga (approved):
LGTM. Verified nnunley's shell-injection concern is genuinely closed, not just narrowly patched — reproduced the pre-fix vulnerability (a crafted benchmark filename reaching `sh -c` via the interpolated command string → arbitrary code execution), then reproduced against the actual post-fix `validate_benchmark_name`/`benchmark_name` guard and confirmed it rejects the same input before any shell-evaluated string is built, for both the filesystem-glob path and the `FILTER_BENCHES` CLI-arg path. Ran the script live end to end (`BASELINE_REF=none bash benchmark/run.sh fib`) and it works correctly.

One non-blocking gap for a follow-up: `BASELINE_REF` isn't run through the same validator. I traced every place it's used (git rev-parse argv, a hyperfine label, echo/markdown text, Python argv comparison) and none of them reach a shell-evaluated string, so it's not currently exploitable, and nothing in CI invokes this script yet. Still, since it's the same class of value, worth sanitizing it too as cheap defense-in-depth before this script gets wired into CI.

Approving.

Issue #599, opened 2026-07-21
AOT lowering: returns stay boxed (`vm.Value` + `rt.AddValue`) even for recursive int fns — ~3× off pure Go


mparrett:
A self-recursive integer fn lowers with unboxed params but a boxed return. fib on main (9c9a3d6) generates:

```go
func Fib(ec *vm.ExecContext, arg0 int) (vm.Value, error) {
    ...
    v2 = arg0 < 2
    if v2 {
        v15 = vm.Int(arg0)                  // box on the base case
    } else {
        arg__4, callErr = Fib(ec, arg__3)   // vm.Value comes back
        ...
        v15 = rt.AddValue(arg__4, arg__7)   // generic boxed add
    }
    return v15, nil
}
```

The param side is right (`arg0 int`, from the #357/#530 work), but every return round-trips through `vm.Value`, and the addition dispatches through `rt.AddValue` instead of `int + int`.

Measured on fib(32), darwin/arm64, main 9c9a3d6:

- lowered (above): ~26 ms
- hand-written pure-Go int fib: ~8.5 ms
- VM bytecode: ~590 ms

So lowering already wins ~20× over the VM, but boxing + generic arithmetic leave ~3× on the table for the flagship recursive-int case.

Return hints don't change this. All the accepted hint spellings (#484 / #489) still emit `(vm.Value, error)` + `rt.AddValue` — the hints parse but nothing in lowering consumes them for the return type. (The idiomatic mid-position spelling `(defn f ^long [args] …)` currently breaks lowering outright — filed as #598.)

Related surface:

- #551 is the param-side version of this ask (infer float params without `^double`); this is the return side.
- #554 shows the typed-return machinery partially exists — a loop-recur leaf infers an `int` return and gets a fully-typed signature. It's returns through self-recursion (and hinted returns generally) that stay boxed, plus #554's registration gap for whichever fns do get there.

What closing it could look like, in increasing ambition:

1. Consume an accepted return hint to pin the fn's native return type, and box only at the trampoline/override boundary — Clojure's model (declaration on the fn, coercion at the edge).
2. Fixpoint return-type inference through self/mutual recursion — fib's return type is join(int, int + int) = int with no hint at all.

fib is the canonical downstream benchmark (gloat's engine matrix runs it per engine); either path should close most of that 3× gap — what remains is the `callErr` threading (#281) and the `ec` parameter. Happy to prototype the hint-driven variant (1) if that shape looks right.

PR #661, opened 2026-07-31
fix(ir): treat a return-hinted arity vector as one arity (#598)

### LINKED ISSUE #598: AOT lowering: `(defn f ^long [args] …)` — the idiomatic return-hint position — silently drops the fn from lowering (misparsed as multi-arity)

mparrett:
Closes #598.

## The bug

The reader reifies `(defn f ^long [n] …)` — the idiomatic Clojure position for a return hint — as `(defn f (with-meta [n] {:tag (quote long)}) …)`. The lowering pipeline split single- from multi-arity on a bare `vector?` test, so that list fell through to the multi-arity arm, where `with-meta` got taken for an args vector and threw "seq expected Seqable" inside `expand-fn-args`.

The throw is caught per-fn and demoted to a `WARNING`, so the function vanished from the lowered package and the program kept running on the VM. Adding the hint that is supposed to make code faster turned native lowering off.

Repro from the issue, on `cb99d7c5`:

```
$ ./lg scripts/lg-compile out fibv fib.lg
WARNING: lowering failed for fib: … "seq expected Seqable" …
```

`fib` is absent from the generated Go. After this change the same input emits `func Fib(ec *vm.ExecContext, arg0 int) (vm.Value, error)`.

## The fix

`ir/build.lg` already had the right predicate for this shape — #484/#489 fixed the same syntax at the compiler layer, which is why the hinted form compiles and runs correctly on the VM. It was private. This promotes `return-hinted-arity-vector?` and `unwrap-arity-vector` and uses them at the three places in `pipeline.lg` that tell arities apart by `vector?`:

- **The `compile-form*` arity branch** — the one that dropped the function.
`args-vec` keeps the hinted form rather than unwrapping it here, because `build-fn` already unwraps it. To be precise about what this does not do: the `:tag` is dropped by `unwrap-arity-vector` and read by nothing today, so the hint still has no effect on the return type. Consuming it is #599.
- **`defn-key`** — a hinted single-arity defn was keyed `:multi`, so it could
dedupe against an unhinted same-name defn of the same real arity.
- **`single-arity-defn-form?`** — the inline-registry gate, where the
consequence was milder: hinted fns were permanently un-spliceable.

## Tests

`pkg/ir/lisp_return_hint_arity_test.go` asserts the hinted form lowers and keeps its unboxed `int` param. It also pins the three spellings the issue reported as already working, so a narrower fix that special-cased only the broken one fails. A third test pins that a real multi-arity defn still routes as multi-arity, since the widened predicate must not swallow it.

One correction to the issue while writing those: I listed `(defn fib (^long [^long n] …))` as lowering correctly, which it does, but it lowers as `Fib_1` via the multi-arity path rather than as a bare `Fib`. The test expects that name.

## Scope

The lowering half only. Returns still come back boxed, which is #599. I checked whether the wrapped-arity spelling already unboxes its return; it does not. Recursive `fib` yields `(vm.Value, error)` in both spellings, so there is nothing in that path to reuse for #599.

## Checks

`gofmt` and `go vet` clean. `go test ./pkg/...` and `./test/...` both pass. Cross-builds pass for linux/amd64, js/wasm, plan9/amd64, and wasip1/wasm. `pkg/rt/generated.sums` is updated from `make generate`; `core_compiled.lgb` is byte-identical, since the IR namespaces load from embedded source rather than the bundle.

### DIFF
--- a/pkg/ir/lisp_return_hint_arity_test.go
+++ b/pkg/ir/lisp_return_hint_arity_test.go
@@ -1,0 +1,98 @@
+/*
+ * Copyright (c) 2026 Marcin Gasperowicz <xnooga@gmail.com>
+ * Part of the let-go project; see CONTRIBUTORS for full list of authors.
+ * SPDX-License-Identifier: MIT
+ */
+
+package ir_test
+
+// #598: `(defn f ^long [n] …)` — the idiomatic return-hint position — reads as
+// `(defn f (with-meta [n] {:tag …}) …)`. The lowering pipeline split single- from
+// multi-arity on a bare `vector?` test, so the hinted form fell through to the
+// multi-arity arm, threw, and was demoted to a warning: the fn silently vanished
+// from the lowered package and the program ran on the VM instead. These tests
+// assert the hinted form lowers, that the spellings which already worked still
+// do, and that real multi-arity is still routed as multi-arity.
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/nooga/let-go/pkg/vm"
+)
+
+// Intern the defn before lowering it: build resolves same-ns call targets
+// through the namespace, so a self-recursive fn is an unresolved symbol
+// otherwise. (lisp_defmulti_native_test.go does the same for its multimethod.)
+func lowerForms(t *testing.T, pkg string, forms string) string {
+	t.Helper()
+	runLispExpr(t, `(do `+forms+`)`)
+	v := runLispExpr(t, `(ir.passes.pipeline/lower-ns-to-go "`+pkg+`" (quote core) (quote [`+forms+`]))`)
+	s, ok := v.(vm.String)
+	if !ok {
+		t.Fatalf("expected rendered Go string, got %T: %v", v, v)
+	}
+	return string(s)
+}
+
+func TestReturnHintedArityLowers(t *testing.T) {
+	ensureLoader()
+
+	rendered := lowerForms(t, "rhpkg",
+		`(defn rhfib ^long [^long n] (if (< n 2) n (+ (rhfib (- n 1)) (rhfib (- n 2)))))`)
+
+	if !strings.Contains(rendered, "func Rhfib(") {
+		t.Fatalf("return-hinted defn did not lower to a native fn; rendered:\n%s", rendered)
+	}
+	// The param hint still has to survive the unwrap — a lowered fn taking
+	// vm.Value would mean the arity vector reached build-fn stripped of its
+	// element metadata.
+	if !strings.Contains(rendered, "arg0 int") {
+		t.Errorf("expected an unboxed int param on the lowered fn; rendered:\n%s", rendered)
+	}
+}
+
+// The spellings the issue reported as already working. They route through the
+// same branch, so a fix that only special-cased the broken one would show up here.
+func TestUnhintedAndAlternateHintSpellingsStillLower(t *testing.T) {
+	ensureLoader()
+
+	cases := []struct {
+		name  string
+		form  string
+		wantF string
+	}{
+		{"no hint", `(defn rhplain [^long n] (+ n 1))`, "func Rhplain("},
+		{"param hint only", `(defn rhparam [^long n] (* n 2))`, "func Rhparam("},
+		{"hint on the name", `(defn ^long rhname [^long n] (- n 1))`, "func Rhname("},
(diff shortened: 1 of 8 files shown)


mparrett:
Self-review verified. No findings.

- Full `go test -count=1 -timeout 60s ./...` passes.
- `go vet ./...` passes.
- Additional deduplication and inline-registry probes pass.
- Generated-artifact, lint, race, CodeQL, WASI, and gogen-diff checks pass.
- The main `build` CI job passed.

pkg/ir/lisp_return_hint_arity_test.go
```diff
@@ -0,0 +1,98 @@
...
+		t.Fatalf("expected rendered Go string, got %T: %v", v, v)
+	}
+	return string(s)
+}
+
+func TestReturnHintedArityLowers(t *testing.T) {
+	ensureLoader()
+
+	rendered := lowerForms(t, "rhpkg",
+		`(defn rhfib ^long [^long n] (if (< n 2) n (+ (rhfib (- n 1)) (rhfib (- n 2)))))`)
```
nnunley, line 42: Can you add a metadata oracle test using clj to validate this form?

It strikes me that having a post lowering to bytecode/go test that validates that the lowered version can return the attached metadata would be good.

nnunley (approved):

pkg/ir/lisp_return_hint_arity_test.go
mparrett: Thanks! Added a Clojure-derived metadata oracle plus a generated-Go trampoline fixture, and confirmed the bytecode and native paths return the expected metadata with native dispatch.

PR #662, opened 2026-07-31
feat(ir): seed self-recursive calls from a declared return type (#599)

### LINKED ISSUE #599: AOT lowering: returns stay boxed (`vm.Value` + `rt.AddValue`) even for recursive int fns — ~3× off pure Go

mparrett:
Closes #599. **Stacked on #661** (`wt/fix-598-return-hint`) — review that first; the idiomatic `^long` spelling does not reach lowering without it.

Because it targets a branch rather than `main`, the workflows do not fire and the check list here is empty apart from skipped jobs. Everything in Checks below was run locally; CI will cover it once #661 merges and this retargets.

## The cause is upstream of codegen

A self-recursive integer fn lowered with unboxed params but a boxed return: `func Fib(ec, arg0 int) (vm.Value, error)`, with the addition dispatching through `rt.AddValue`. That looked like a codegen gap, but it is an inference one. The self-call node types as `:unknown`, so `(+ (fib …) (fib …))` sees two unknown operands, stays boxed, and the function's own return type never resolves either. Nothing downstream can recover from that.

A return hint breaks the cycle. `(defn fib ^long [^long n] …)` seeds the self-call with the declared type, and the typed-result machinery that already exists carries it the rest of the way:

```go
func Fib(ec *vm.ExecContext, arg0 int) (int, error) {
	...
	if v2 {
		v15 = arg0
	} else {
		tr8, te8 = Fib(ec, arg__3)
		...
		v15 = arg__4 + arg__7
	}
	return v15, nil
}
```

## Numbers

In-process, darwin/arm64, fib(27), all three arms in one process, 4 repeats:

| arm | median ns/op | |
|---|---|---|
| boxed (before) | 1,998,000 | |
| typed (after) | 725,000 | 2.76× faster |
| hand-written Go | 624,000 | the typed lowering is 1.16× off, was ~3× |

The residual over pure Go is the `ec` param and the `(T, error)` result — #281's `callErr` threading, as the issue anticipated.

One caveat on the method: the machine was not idle (load ~5), so treat the absolute figures as soft. The three arms ran back to back in one process and the ratios held across all four repeats, which is the claim I would stand on. Process-level timing on this machine produced nonsense — "lowered beats hand-written Go" — which is a good reminder that a result contradicting the mechanism means the harness is wrong, not the code.

## Three pieces

- **`ir/build.lg`** gains `declared-return-type`, decoding the tag the reader
leaves on the arity vector's metadata. `unwrap-arity-vector` drops that metadata, so before this nothing read it — #598 only stopped the hinted form from vanishing, it did not give the hint meaning.
- **`typeinfer`** types a `:call` from that declaration, ahead of the
structural guesses, since an explicit hint outranks them.
- **`lower_go`** boxes a typed call's result only when the receiving temp is a
`vm.Value`. It boxed unconditionally before, which was right while a callee's result could only ever land in a `vm.Value` temp. Now the temp can itself be natively typed, and boxing into it emits `arg__4 = vm.Int(tr8)` against `var arg__4 int`, which does not compile.

## Why self-calls only

This restriction is load-bearing, not conservatism. Seeding calls to *other* functions lets a declaration the callee's body contradicts corrupt the caller. `(defn liar ^long [n] "not-an-int")` lowers honestly to `string`, but a caller seeded with `:int` emitted:

```go
func Caller(ec *vm.ExecContext, arg0 vm.Value) (int, error) {
	var v2 int
	v2 = vm.String(Liar(ec, arg0))   // does not compile
```

I hit that while building this. Restricted to self-calls, an inconsistent hint can only produce a type the declaring function's own body already joins, so the fallout stays local and shows up as that function declining to type its return. There is a test for it.

A non-recursive function's return type already falls out of its body, so the declaration adds nothing there — self-recursion is exactly the gap.

## Blast radius

Opt-in. No defn in `pkg/rt/core` carries a return hint, so the core lowering is untouched: `make check-selfhost` reports the native re-lowering fixpoint still byte-identical, and `make gogen-diff` passes.

## Checks

`gofmt` and `go vet` clean. `go test ./pkg/...` and `./test/...` pass. Cross-builds pass for linux/amd64, js/wasm, plan9/amd64, and wasip1/wasm. Four new tests cover the typed self-recursion, the unhinted case still boxing, the contradictory-hint containment, and a `^double` variant so the mechanism is not int-specific.

### DIFF
--- a/pkg/ir/lisp_declared_return_test.go
+++ b/pkg/ir/lisp_declared_return_test.go
@@ -1,0 +1,82 @@
+/*
+ * Copyright (c) 2026 Marcin Gasperowicz <xnooga@gmail.com>
+ * Part of the let-go project; see CONTRIBUTORS for full list of authors.
+ * SPDX-License-Identifier: MIT
+ */
+
+package ir_test
+
+// #599: a self-recursive integer fn used to lower with unboxed params but a
+// boxed return — `(vm.Value, error)` plus `rt.AddValue` for the addition —
+// because the self-call typed as :unknown and poisoned the numeric chain. A
+// declared return type (`(defn f ^long [n] …)`) seeds the self-call, after
+// which the existing typed-result machinery carries native ints all the way
+// through. These tests pin the resulting shape, that it stays opt-in, and that
+// the seed does not leak across function boundaries.
+
+import (
+	"strings"
+	"testing"
+)
+
+func TestDeclaredReturnTypeUnboxesSelfRecursion(t *testing.T) {
+	ensureLoader()
+
+	rendered := lowerForms(t, "drpkg",
+		`(defn drfib ^long [^long n] (if (< n 2) n (+ (drfib (- n 1)) (drfib (- n 2)))))`)
+
+	if !strings.Contains(rendered, "func Drfib(ec *vm.ExecContext, arg0 int) (int, error)") {
+		t.Fatalf("expected a natively-typed int return; rendered:\n%s", rendered)
+	}
+	// The payoff: generic boxed addition is gone in favour of int + int.
+	if strings.Contains(rendered, "rt.AddValue") {
+		t.Errorf("addition still dispatches through rt.AddValue; rendered:\n%s", rendered)
+	}
+}
+
+// Opt-in: without a hint nothing changes, so the whole corpus of unhinted code
+// lowers exactly as before.
+func TestUnhintedRecursionStillBoxes(t *testing.T) {
+	ensureLoader()
+
+	rendered := lowerForms(t, "drpkg",
+		`(defn drplain [^long n] (if (< n 2) n (+ (drplain (- n 1)) (drplain (- n 2)))))`)
+
+	if strings.Contains(rendered, "func Drplain(ec *vm.ExecContext, arg0 int) (int, error)") {
+		t.Fatalf("unhinted fn typed its return without a declaration; rendered:\n%s", rendered)
+	}
+}
+
+// The seed is restricted to self-calls. A declaration the body contradicts must
+// not reach a *caller's* inference: before that restriction, drcaller returned
+// (int, error) and assigned vm.String(...) into an int temp, which does not
+// compile. drliar itself is free to lower honestly as string.
+func TestContradictoryHintDoesNotCorruptCallers(t *testing.T) {
+	ensureLoader()
+
+	rendered := lowerForms(t, "drpkg",
+		`(defn drliar ^long [n] "not-an-int") (defn drcaller [n] (drliar n))`)
+
+	if strings.Contains(rendered, "func Drcaller(ec *vm.ExecContext, arg0 vm.Value) (int, error)") {
+		t.Fatalf("callee's contradictory hint leaked into the caller's return type; rendered:\n%s", rendered)
+	}
+	// The give-away statement from the broken version: a string boxed into an
+	// int-typed temp.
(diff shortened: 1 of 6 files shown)


nooga (approved):
LGTM. Did a full correctness pass on the new `*declared-return-types*`/`call-declared-type` machinery — it's a clean structural mirror of the existing `*deftype-ctor-types*`/`call-dtype` pattern, and I traced the self-call containment (bare-name string comparison) and the multi-arity exclusion to confirm they hold up as claimed.

Verified: `go build ./...`, `go test ./... -count=1` pass (one unrelated `test/e2e` artifact-generation gap in a fresh worktree, not a regression); `make check-generated` confirms the `.lgb`/lowered-tree regen is byte-identical to what's committed.

Two small non-blocking gaps, both worth a follow-up issue rather than holding this PR:
1. Return-type seeding only fires for the `(defn f ^long [n] …)` hint spelling — the equally-legal `(defn ^long f [n] …)` (hint on the name) isn't picked up by `declared-return-type` (`build.lg:1486-1495`), so a self-recursive fn hinted that way silently stays boxed.
2. That same function silently returns `nil` on an unsupported tag (e.g. `^String`) rather than warning, unlike the sibling param-hint path which does print a warning for the same situation.

PR #667, opened 2026-08-01
refactor(ir): catalog-driven form-head dispatch with load-time coherence check

nnunley:
Moves the source-form special-form facts out of three hand-maintained places
(the build-list cond chain and the free-vars / captures-of member sets) into
one catalog, consulted everywhere and enforced at build time.

## What changed
- form_heads.lg: catalog of the 14 special form heads + their handler bindings
- build-list dispatches via a catalog-keyed handler lookup, not a hardcoded chain
- free-vars / captures-of derive scope-handling membership from catalog queries
- Load-time coherence check, bidirectional: a form missing from the catalog OR
  a catalogued form without a handler fails generation, naming the form
- Tests: catalog coverage, coherence, and a catalog-edit-changes-behaviour demo

## Evidence
- IR corpus: 2458/2469 (99.55%) — matches pre-change baseline; check-generated green
- Parity (bytecode vs gogen_ir): 6311 assertions identical
- Red-build drills exercised both directions (failure names the missing form)
- Perf at par from raw bench records (times at-par-to-improved, allocs unchanged)

Built against main dc316a0c7.

Related: #268


mparrett (changes requested):
Reviewed at `2301c20e`, rebased onto `b170a08e`. Built it, ran the suite, and ran both red-build drills. The suite is green, the corpus numbers hold, and the catalog↔handler-map direction of the coherence check does what the body says. The catalog↔dispatch direction does not, and that's the gate in the title.

**[P1] The coherence check validates a map that dispatch never reads — `pkg/rt/core/ir/form_heads.lg:136`**

The check compares the catalog against `form-head-handlers`, the symbol map in `form_heads.lg`. `form-head-handler` resolves through `form-head-handlers-fn`, the thunk `reset!` at the bottom of `build.lg`. I deleted `'try build-try` from the thunk and left the check untouched: it still printed "bidirectional coherence check passed", `make generate` exited 0, and 63 functions fell back to bytecode with `WARNING: lowering failed … unresolved symbol try`. A catalogued form whose handler is missing doesn't fail generation; it silently degrades the lowered tree.

The fix is small. Point the check at the same source dispatch uses, or drop the symbol map and let the thunk be the single handler binding.

**[P2] The catalog's boolean columns aren't checked against the code, and one is wrong — `form_heads.lg:42`**

`[quote true false …]` says `quote` has no explicit handler in `captures-of`, but `build.lg:238` reads `(= head 'quote) #{}`. The force-push turned this green by deleting `'quote` from the expectation in `test/ac3-coupling-demo.lg:17` instead of correcting the row, so the assertion no longer checks and the catalog still misdescribes the code. `[var true true "var reference: doesn't evaluate, no captures"]` contradicts itself the same way. Nothing validates these two columns against `build.lg` in either direction; the coherence check covers only the form-head column, which is how the mismatch survived.

`quote` and `var` are also the two heads whose `cond` arms stayed ungated, so their flags have no effect today. The rows are free to drift, and that's the part to settle before more forms land.

**[P2] Adding a form takes four edits, not the two the file documents — `form_heads.lg:135`**

`required` is a hardcoded 14-symbol literal twenty lines below the catalog, and `extra-in-catalog` throws on anything absent from it. Following the file's own recipe ("adding a form = edit catalog + edit handler map, nothing else"), I added a row plus a handler entry; `make generate` died with `unexpected form heads in catalog (not in build-list): #{monitor-enter}`. That message also misattributes its source, since it never consults `build-list` — which now derives its dispatch from the catalog. Counting the non-test copies, the three hand-maintained places the body names became four: catalog, `form-head-handlers`, `required`, and the `reset!` thunk.

**[P2] Three of the nine new tests are `(is true)` — `test/form_heads_catalog.lg:46`**

`form-heads-dispatcher-equivalence` ("TODO: implement once generator exists"), `red-build-drill-ac1-missing-form` ("to actually see the failure, manually remove a form"), and `ac3-demonstration-catalog-edit-changes-behavior` all assert `(is true)`. Three more compare the catalog against literals in the test file labelled "ground truth", so they can't detect divergence from `build.lg`. `current-captures-of-heads` at line 32 omits `quote` and agrees with the wrong catalog row. The body lists the catalog-edit-changes-behaviour demo as delivered; it's a comment block.

**[P2] Comments describe a generator this PR deletes — `form_heads.lg:7`, `build.lg:720`**

The header points at `pkg/rt/core/ir/form_heads_gen.lg`, dropped in `d14c5a50`. `build.lg` twice calls `form-heads-dispatcher` "the generated dispatcher … emitted from the catalog during `make generate`"; it's a hand-written `defn` twenty lines above. `test/ac3-coupling-demo.lg:36` still says "Full implementation pending: refactoring the cond chains to query the catalog", which this PR does.

**Verdict: not correct.** The load-time gate accepts a build it should reject, and the failure mode is a silently degraded lowered tree rather than an error.

One thing I checked and didn't find: I expected the per-call set reconstruction in `has-free-vars-handler?` to cost something, since it rebuilds the set on every gated `cond` arm inside a recursive walk. Interleaved A/B over `lgbgen --target=go`, gated against gates-removed, came out 65.2/67.2s gated and 69.4/69.2s ungated. No measurable regression, so it isn't a finding.

Two decisions and the rest follows: whether the check reads the thunk or the two handler maps collapse into one, and whether `quote`/`var` get corrected rows or gated `cond` arms like the other twelve heads.

nnunley:
Addressed all five findings; re-review requested. Reviewed-at base unchanged (`b170a08e`).

**The two decisions:** collapsed the two handler maps into one — the `form-head-handlers-fn` thunk is now the single binding, and the coherence check reads it directly (`@form-head-handlers-fn`); and gated the `quote`/`var` cond arms through catalog queries like the other twelve, with their rows corrected — the boolean columns are now load-bearing.

**Your drill, both directions, same pass:** removing `'try` from the handler thunk → `make generate` exit 2, "form heads missing from handler map: #{try}"; removing the `try` catalog row → exit 2, "unexpected form heads in handler map (not in catalog): #{try}"; restored → exit 0. The catalog direction was genuinely dead in the reviewed revision — your drill caught it; fixed in `dfed55e0`.

**Also fixed:** `required` literal removed (catalog is the single source; adding a form is now catalog row + handler entry), the misattributed error message, the three `(is true)` placeholders replaced with real assertions, the narrowed `ac3-coupling-demo` expectation restored, and the stale generator prose corrected.

**Perf:** your interleaved A/B replicated — branch-minus-base is within noise on `bench-ratchet` (we also found our recorded local baseline is stale; base itself fails it by 3-6x, tracked separately). Catalog query sets are now precomputed at load (`3030c8e`) regardless.

Corpus: `make ir-stress` all-ok; parity-quick 6311/6311; `make check-generated` green.

mparrett (changes requested):
The design objection from round 1 is resolved. The symbol map is gone, `form-head-handlers-fn` is the single binding, and the coherence check reads the thunk the dispatcher uses. I re-ran both drills against `3030c8e7`:

| drill | result |
| --- | --- |
| delete `'try build-try` from the thunk | `exit 1` — `coherence check FAILED: form heads missing from handler map: #{try}` |
| delete the `[try …]` catalog row | `exit 1` — `unexpected form heads in handler map (not in catalog): #{try}` |

Drill through `go run -tags bootstrap ./cmd/lgbgen --target=go` rather than the bundled `lg`: the bundled binary ignores the source edit, so both drills pass whatever you do to the catalog.

The two-edit claim holds too. Adding a `[monitor-enter false false …]` row and one handler-map entry, and nothing else, completes with exit 0 and writes `pkg/rt/generated.sums`.

## Blocking: `generated.sums` is stale at the head

`build` and `generated-artifacts` both fail on the same gate:

```
check-generated: STALE — .lg/lgbgen sources changed without regeneration.
  recorded: 3d140e9c922d7b851f48d72726c522e5799ebf42300920135f0f39cb78a6d0ea
  computed: 8652877800ec15c4c828918dc5c6a1a8987fd446e0d9e3fe1ba128a52858bdd0
```

This is intrinsic to the head, not base drift: `main` did move to `1f281df4` while the PR sat, but I get byte-identical recorded and computed digests on the detached head with no merge.

The commit order explains it. `8e66b2d4` refreshed `generated.sums`, then `3030c8e7` edited `pkg/rt/core/ir/form_heads.lg`, a bundle input, with no regen after. So "`make check-generated` green" in the re-review note was accurate at `8e66b2d4` and not at the head. `make generate` and commit clears both jobs.

## Residual on the boolean columns

The columns are corrected and both `cond` arms now go through catalog queries, so the round-1 finding is addressed. What is left is that they are load-bearing in one direction only: flipping a `true` to `false` disables a gated arm, but flipping a `false` to `true` does nothing for `if`, `do`, `def` and `recur`, because those have no arm to gate.

That matters for one comment. `test/form_heads_catalog.lg:120-128` tells the reader to change `[if false …]` to `[if true …]`, rebuild, and "observe: behavior changed WITHOUT editing free-vars". It will not: `free-vars` has no `if` arm, so the flip only changes what `has-free-vars-handler?` returns, and `if` takes the generic walk either way. The demo reads as if all 14 heads were symmetric. Either point it at a head with a gated arm, or say outright that those four rows carry metadata and nothing else.

## Minor

**`form-head-names` is the one query still rebuilt per call.** `form_heads.lg:51-54` is still `(into #{} (map first form-heads-catalog))`, and `build.lg:753` calls it on every list form, the hottest of the three. `form-head-handler` also re-invokes the thunk per lookup. I A/B'd it, three interleaved pairs of `lgbgen --target=go`: head 109.6/113.5/112.5s, memoized 120.8/103.9/133.5s. No measurable win, the same null result you got on the gated sets. Raising it for consistency, since the commit that precomputed the two sibling sets skipped the third.

**Two unconditional load-time `println`s.** `form_heads.lg:128` and `build.lg:1655` mean every `(require 'ir.build)` writes two lines to stdout: `scripts/lg-compile`, `scripts/ir-stress.lg`, `scripts/ti-baseline.lg`, `examples/aot/yamlstar/ys-lower-report.lg`, and the `make test` output all carry them. No other file under `pkg/rt/core/` has a top-level `println`; `purity.lg` and `trace.lg` gate theirs inside functions. A normal `lg` boot stays clean, so the cost is noise in the tooling paths, not a user-facing regression. The `build.lg` line also labels itself `ir.form-heads` while being emitted from `ir.build`.

**Leftover prose.** `test/form_heads_catalog.lg:7` still lists "dispatcher equivalence: generated dispatcher ≡ current cond chain", which is the old generated-dispatcher framing. The PR body says "built against main dc316a0c7"; the merge base is `b170a08e`.

**`build-let*` and `build-loop*`.** Both exist only to forward to `build-let` / `build-loop` from the handler map, which could bind `'let* build-let` directly. Their docstrings say the dispatcher "aliases both to the same handler"; the wrapper is what stops that from being true.

Requesting changes for the regen alone. Everything else here is P3 and none of it needs to block: one `make generate` and this is an approve.

nnunley:
The generated.sums staleness blocker is now fixed.

**What changed:** Commit 3030c8e7 modified `pkg/rt/core/ir/form_heads.lg` to precompute form-head handler sets at load time, but did not regenerate the manifest. This commit ran `make generate` to refresh `generated.sums`, bringing it back into sync with the current .lg sources.

**Verification:** 
- \`make check-generated\` passes locally (bundle + lowered tree verified in lockstep)
- Pre-push gate passed (ir-stress + test suite green)
- CI build and generated-artifacts lanes will now pass

Ready for re-review.

mparrett (approved):
Re-reviewed at `72ff2f3`. The prior `generated.sums` blocker is fixed: `make check-generated` and `make test` both pass locally, and I found no remaining blocking code issues. Approved.

Before merge, please rebase onto current `main` and regenerate `pkg/rt/generated.sums`; GitHub currently reports the PR as conflicting, and the generated manifest is the only merge conflict I found. The current head also has no GitHub Actions check runs attached.

nnunley:
Rebased onto current main and regenerated pkg/rt/generated.sums via `make generate` per approval note. CI checks now attached and building.

PR #668, opened 2026-08-02
feat(clojure): add clojure.repl / clojure.pprint, fix Var metadata bundle round-trip

nooga:
## Summary

- Adds `clojure.repl` (`doc`, `source`, `find-doc`, `apropos`, `dir`) and `clojure.pprint` (`pprint`, `cl-format` directives, `print-table`) as source-loaded `.lg` namespaces, backed by a new native `raw-source` primitive for reader-based `source-fn`.
- Threads `:arglists` metadata through `defn`/`defn-`, and records `:file`/`:line`/`:column` at def time from `FormSource`.
- Adds namespace introspection helpers (`AliasesSnapshot`, `ReferredVars`, `Unmap`) backing `ns-aliases`/`ns-refers`/`ns-unmap`.
- **Fixes a real, confirmed gap**: Var metadata (`:doc`, `:private`, `:dynamic`, `:arglists`, `:file`/`:line`/`:column`) did not survive the `pkg/rt/core_compiled.lgb` precompiled-bundle round-trip, so a normal build showed none of it for `core.lg`'s own ~800 functions or any other bundled namespace. `defCompiler` now also emits a runtime call to the existing `apply-def-meta!` builtin so the metadata is reapplied when the chunk's bytecode actually runs in the decoding process — mirroring how `set-macro!` already survives this exact round-trip for `:macro`. No `.lgb` wire-format change needed.

## Test plan

- [x] `go build ./...`
- [x] `go test ./... -count=1` (including `test/e2e`)
- [x] New Go test (`pkg/compiler/def_meta_bundle_roundtrip_test.go`) compiles a `def`, encodes it, decodes it against a Var the compiling process never touched, runs the decoded bytecode, and asserts metadata is present — verified to fail without the fix
- [x] New `.lg` integration tests: `test/clojure_repl_test.lg`, `test/clojure_pprint_test.lg`
- [x] Manually verified `(clojure.repl/doc comp)` now shows `:arglists` for bundled core functions; `clojure.repl/source-fn` still works for `clojure.repl`'s own functions (kept source-loaded on purpose — see code comments)


pkg/vm/namespace.go
```diff
@@ -529,6 +529,73 @@ func (n *Namespace) AllVars() map[Symbol]*Var {
...
+}
+
+// Unmap removes a single LOCALLY INTERNED symbol from this namespace's
+// registry. Backs the ns-unmap core fn. Scope note: this only removes a
+// var this namespace itself defined — a symbol visible purely through a
+// refer (see ReferredVars) has no registry entry to remove, matching how
+// refers are resolved dynamically rather than copied into the registry.
+func (n *Namespace) Unmap(name Symbol) {
+	n.mu.Lock()
+	delete(n.registry, name)
```
mparrett, line 639: **[P2] `ns-unmap` leaves referred symbols mapped.** When `ns-unmap` targets a referred symbol, `Unmap` only deletes from the local registry, so `(ns-unmap *ns* 'map)` returns successfully but `(resolve 'map)` remains truthy. Clojure removes the namespace mapping whether it is an intern or a refer; this needs per-symbol removal or exclusion from the refer tables too.

pkg/rt/lang.go
```diff
@@ -7616,6 +7615,115 @@ func CoreNsPublics(vs ...vm.Value) (vm.Value, error) {
...
+//lg:name remove-ns
+func CoreRemoveNs(vs ...vm.Value) (vm.Value, error) {
+	if len(vs) != 1 {
+		return vm.NIL, fmt.Errorf("wrong number of arguments %d", len(vs))
+	}
+	sym, ok := vs[0].(vm.Symbol)
+	if !ok {
+		return vm.NIL, fmt.Errorf("remove-ns expected Symbol")
+	}
+	RemoveNS(string(sym))
```
mparrett, line 7726: **[P2] Prevent removal of `clojure.core`.** The newly public `remove-ns` exposes the test-isolation helper without guarding the canonical core namespace. `(remove-ns 'clojure.core)` currently succeeds and makes `(find-ns 'clojure.core)` return nil, whereas Clojure rejects the operation; guard `core` after alias resolution.

pkg/rt/core/clojure/pprint.lg
```diff
@@ -0,0 +1,323 @@
...
+        at? (:at? node) colon? (:colon? node)]
+    (case ch
+      \A (let [[v st] (take-cl-arg st)] [(cl-pad (str v) p0 p1 p2 p3 at?) st false])
+      \S (let [[v st] (take-cl-arg st)] [(cl-pad (pr-str v) p0 p1 p2 p3 at?) st false])
+      \D (let [[v st] (take-cl-arg st)] [(render-radix v 10 p0 p1 colon? at?) st false])
+      \B (let [[v st] (take-cl-arg st)] [(render-radix v 2 p0 p1 colon? at?) st false])
+      \O (let [[v st] (take-cl-arg st)] [(render-radix v 8 p0 p1 colon? at?) st false])
+      \X (let [[v st] (take-cl-arg st)] [(render-radix v 16 p0 p1 colon? at?) st false])
+      \% [(apply str (repeat (or p0 1) \newline)) st false]
+      \& [(apply str (repeat (or p0 1) \newline)) st false]
```
mparrett: **[P2] Implement fresh-line semantics for `~&`.** `~&` is currently identical to `~%`, but fresh-line must emit nothing when output is already at the start of a line. For example, `cl-format nil "x\n~&y"` returns `"x\n\ny"` here instead of `"x\ny"`; rendering needs to track the current output column or preceding output.

pkg/rt/core/clojure/pprint.lg
```diff
@@ -0,0 +1,323 @@
...
+              else-clause (nth clauses 0 [])
+              then-clause (nth clauses 1 [])]
+          (render-nodes (if v then-clause else-clause) st2))
+
+        :else
+        (let [[v st2] (take-cl-arg st)
+              n (count clauses)]
+          (if (and (integer? v) (>= v 0) (< v n))
+            (render-nodes (nth clauses v) st2)
+            (if (pos? n) (render-nodes (nth clauses (dec n)) st2) ["" st2 false]))))
```
mparrett: **[P2] Require `~:;` for a conditional default.** For indexed `~[...]`, an out-of-range selector must produce no output unless the final clause is introduced by `~:;`. This branch always treats the last clause as the default, so `cl-format nil "~[zero~;one~;many~]" 5` returns `"many"` instead of `""`; the parser needs to retain whether the separator carried the colon modifier.

mparrett (changes requested):
Requesting changes for four reproducible compatibility defects in the newly exposed APIs. The existing and new test suites pass, but each case below diverges from Clojure's public behavior.

nooga:
Thanks for the thorough review — all four fixed in 6881553.

- **`ns-unmap` leaves referred symbols mapped**: added a per-namespace `unmapped` set that `lookupViaRefers`/`ReferredVars` now check, so unmapping a referred symbol actually clears its resolution (`(resolve 'map)` → nil), not just the local registry entry. Kept it separate from the existing `:refer-clojure :exclude` set rather than reusing it — reusing it broke `string.lg`'s compile order (it excludes names like `subs`/`upper-case` from core auto-refer while still relying on the core fallback until its own same-named `def` runs later in the file).
- **`remove-ns` on `clojure.core`**: now rejected with an error, guarded after alias resolution (blocks both `'clojure.core` and `'core`).
- **`~&` fresh-line**: `cl-format`'s render state now tracks whether output is at the start of a line and only emits `n-1` newlines (instead of `n`) when already there.
- **`~[...~]` default clause**: the parser now tracks whether the separator introducing the last clause was `~:;`; an out-of-range index falls through to that clause only when it was, otherwise it emits nothing — matching your repro.

Added `test/ns_introspection_test.lg` and extended `test/clojure_pprint_test.lg` to cover all four (including the previously-passing-by-accident `~[...~]` case, which now correctly expects `""` for out-of-range with plain `~;`). Full `go test ./... -count=1` and `go generate ./pkg/rt/` both pass.

pkg/vm/namespace.go
```diff
@@ -529,6 +541,87 @@ func (n *Namespace) AllVars() map[Symbol]*Var {
...
+// :exclude): that one only suppresses the warn-on-shadow message and must
+// NOT block refer resolution — some core namespaces (e.g. string.lg) declare
+// :exclude for a name while still relying on it resolving to clojure.core's
+// version until their own same-named def further down the file runs. A
+// later (def name ...) still works fine here regardless: Lookup checks the
+// local registry before ever consulting refers.
+func (n *Namespace) Unmap(name Symbol) {
+	n.mu.Lock()
+	delete(n.registry, name)
+	n.unmapped[name] = true
```
mparrett, line 640: **[P2] Let an explicit refer restore an unmapped symbol.** The new tombstone is never cleared by `Refer` or `ReferList`, despite the comment promising that the symbol remains hidden only until re-referred. After unmapping `map`, an explicit `(refer 'clojure.core :only '[map])` still leaves `(resolve 'map)` nil: the PR produces `true → false → false`, while Clojure produces `true → false → true`. Please clear the corresponding `unmapped` entries when symbols are explicitly referred again, including the `:refer :all` path.

mparrett (changes requested):
The original four cases are fixed. One follow-up compatibility issue remains in the `ns-unmap` tombstone behavior; details inline.

nooga:
Good catch — fixed in b1fecd2.

`Refer`/`ReferList` now clear the corresponding `unmapped` tombstone(s) for any symbol they bring into scope: `ReferList` clears exactly the symbols it's given, and `Refer` with `all=true` clears the target namespace's current public vars (snapshotted before taking the lock, per the file's no-nested-locks invariant). So `ns-unmap` → `resolve` nil → explicit re-refer → `resolve` truthy again, matching Clojure.

Extended `test/ns_introspection_test.lg` to cover both the `refer-list` (`:only`) and `refer` (`:all`) restoration paths. Full `go test ./... -count=1` passes.

mparrett (approved):
All requested compatibility issues are resolved, including both re-refer paths after `ns-unmap`. Targeted probes and the full test suite pass; CI is green.

pkg/rt/core/clojure/repl.lg
```diff
@@ -0,0 +1,132 @@
+;; lgbgen:skip — Var metadata now survives the .lgb round-trip (defCompiler
```
nnunley, line 1: Can you add a tracking issue for including this in the compile?

nnunley (approved):
LGTM. I love seeing the lginterop working, though... the codegen needs work still.

pkg/rt/core/clojure/repl.lg
nnunley: nm. Non-blocking. I'm going to go ahead and consider this part of the large optimization goals.

Issue #672, opened 2026-08-02
register-test! registers the same var twice when a deftest is re-evaluated


abogoyavlensky:
`register-test!` in `pkg/rt/core/test.lg` appends unconditionally:

```clojure
(defn register-test! [v]
  (let [n *ns*
        xs (get *registered-tests* n [])]
    (set! *registered-tests* (assoc *registered-tests* n (conj xs v)))))
```

Evaluating a `deftest` twice hands it the same var both times, so the var is registered twice and `run-tests` then runs and counts that test twice.

This is the ordinary REPL/editor flow - edit a test, send the form again, run the suite:

```clojure
(deftest edited-test (is (= 1 1)))
(deftest edited-test (is (= 1 1)))   ; re-eval after an edit
(run-tests)
;; Tests: 2 Pass: 2   <- one test, counted twice
```

It compounds across a session: N re-evaluations mean N runs of that test, inflated pass counts, and fixtures re-running per duplicate. Batch runs (`lg`, a `.lg` file, CI) never hit it because each namespace is loaded once, so it only shows up interactively - where it's most confusing.

Vars compare by identity, so skipping the `conj` when the var is already present makes registration idempotent while leaving distinct vars and their registration order untouched.

mparrett:
@abogoyavlensky this looks fixed and safe to close — flagging rather than closing it myself since it is yours.

PR #673 fixed it and merged on 2026-08-06. It never auto-closed because its body reads `Resoves: #672`, so GitHub's keyword parser did not match the reference.

It is fixed twice over now. #863 ported `clojure.test` outright, so the `register-test!` in your report no longer exists in `pkg/rt/core/test.lg` at all. The behaviour holds on `827ef18b`:

```clojure
(require '[test :refer [deftest is]])
(def runs (atom 0))
(deftest t (swap! runs inc) (is true))
(deftest t (swap! runs inc) (is true))   ;; same name, re-evaluated
(test/run-tests)
```

```
Ran 1 tests containing 1 assertions.
0 failures, 0 errors.
```

One registration, one run. Found while auditing open issues against merged PRs for the v1.13.0 cut (#718).

abogoyavlensky:
@mparrett Yes, thank you, the issue is fixed now, closing the issue.

PR #673, opened 2026-08-02
fix(test): register each test var once per namespace

abogoyavlensky:
Resoves: #672


mparrett (commented):
Thank you for the PR! Semi-automated PR review below, for consideration.

Reviewed `1ad26c1` against merge-base `b170a08e`. One P3, no blockers.

The fix hinges on `=` being identity for vars, and it is. `*Var` has no `Equals` method, so `=` falls through `identityEqShortCircuit`, which decides same-pointer early, into `valueEqualsCtx`, which has no `Var` case and terminates at `return a == b` (`pkg/rt/lang.go:1263`), a Go interface comparison and so pointer identity. Two distinct vars aliased to the same fn object therefore still register separately, because the comparison never derefs. `distinct-vars-still-register-separately` would not catch that case on its own, since its two probes hold different fns, but the semantics make it moot.

## [P3] `Resoves: #672` will not close the issue

The PR body reads `Resoves: #672`. GitHub's closing keywords are `resolve`/`resolves`/`resolved`, so the misspelling means #672 will not auto-close or cross-link on merge.

## Verdict

Correct. Minimal and targeted, and the semantics it relies on hold. The tests cover idempotent re-registration, distinct vars registering separately, and order preservation, and the harness note explains why the file binds `*registered-tests*` rather than calling `clear-registered-tests!`, which would wipe the registry the enclosing runner is iterating over. `generated.sums` and `core_compiled.lgb` were both regenerated.

Checked and not raised as findings: the linear `some` scan makes registration O(n²) in tests per namespace, which is negligible at realistic namespace sizes since the comparisons are pointer-equality; and the guard does not help the rename case (rename a `deftest`, re-evaluate, and the stale var stays registered), which is pre-existing and not fixable without keying the registry by name rather than by var.

## Merge-order note

Same conflict I flagged on #671. Both PRs edit `pkg/rt/core/test.lg` from the same base, #673 in `register-test!` and #671 in `deftest` directly above it, and both regenerate `core_compiled.lgb` and `generated.sums`, so they conflict either way. Landing this one first is the cheaper order: 6 source lines against #671's ~130, so #671 is the one to rebase.

abogoyavlensky:
@mparrett fyi, conflicts are resolved.

mparrett (approved):
Rebase verified: the net diff against `main` is identical to what I reviewed at `1ad26c1`, so the earlier review stands. The one P3 (the `Resoves: #672` typo, which will stop GitHub closing the issue on merge) is not blocking — worth a one-character fix to the description before this lands, but it does not gate the merge.

Approving. Note this still conflicts with #671: both edit `pkg/rt/core/test.lg` and regenerate the bundle, so landing this one first and rebasing #671 onto it is the cheaper order.

PR #676, opened 2026-08-04
fix(test): isolate Clojure compat const pools per case

nnunley:
The pre-push compatibility gate could leave a timed-out worker goroutine running while the next case reused the same `vm.Consts` pool. That overlap raced `Consts.Intern` against reads from the following case.

## What changed
- allocate a private `vm.NewConsts()` pool inside each `runCompatTest` case
- retain a separate `bootstrapConsts` pool only for suite setup
- remove `-parallel 1`; no compatibility subtest calls `t.Parallel()`, so the flag did not serialize the workers

A timed-out worker can now outlive its case without sharing mutable const-pool state with later cases.

## Evidence
- `go test -race ./test/ -count=1 -run TestClojureTestSuite`: 243 passed
- `go test ./test/ -count=1 -run TestClojureTestSuite`: 243 passed
- pre-push short Go suite: 1725 passed
- `make ir-stress-gate`: passed

The branch is rebased onto current `main`.


mparrett (commented):
Reviewed `67157f3`. One P1.

## [P1] `-parallel 1` is a no-op for this package, so the race is unfixed

`go test -parallel N` only limits tests that call `t.Parallel()`, and there is no `t.Parallel()` anywhere under `test/` — the compat subtests at `test/zz_compat_test.go:152` already run sequentially, so the flag changes nothing.

The concurrency is a raw goroutine rather than test parallelism. `runCompatTest` spawns `go func()` at `test/zz_compat_test.go:216` specifically so the harness can abandon it on timeout, and an abandoned goroutine outlives its subtest while still holding the shared `c *vm.Consts` created at `:83`. That is what races the next subtest's goroutine, and `-parallel` has no authority over raw goroutines.

The two measurements in the description are also not comparable: the ~6/10 failure rate was taken under `-race`, while the gate — and the 16 green runs offered as evidence — run without it. A failure rate that high, if it had really persisted, would make 16 consecutive green runs essentially impossible, which is itself a sign the before and after were measured in different configurations.

The fix you already name as follow-up (per-test `Consts` pools) addresses the real cause, as would bounding the abandoned goroutine so it cannot outlive its subtest.

Secondary, not a separate finding: the hook description says it "mirrors CI compat step", and `.github/workflows/go.yml:87` and `:93` run the suite with no `-parallel` flag, so this edit makes the text diverge from CI without changing behavior on either side.

## Verdict

Nothing breaks — a no-op flag cannot regress anything. The concern is that it does not do what the title and body describe, so merging it leaves the next flake to be debugged against a gate that appears to already carry the fix. Worth holding until the goroutine lifetime or the shared pool is addressed.

The load-bearing checks are `rg 't\.Parallel\(\)' test/` returning nothing and the `go func()` at `test/zz_compat_test.go:216` — both quick to re-run if you read them differently.

nnunley:
The P1 is fixed at `7570d850191b`, and the PR title/body now describe the actual fix. `-parallel 1` was removed because no test calls `t.Parallel()`. Each compatibility case now owns a private `vm.NewConsts()` pool, so a timed-out worker cannot race the next case. Final normal and race-configured compatibility runs each passed 243 tests; the short Go suite and IR-stress gate also pass.

mparrett (approved):
@nnunley Thanks for addressing the P1. The per-case vm.NewConsts() genuinely closes it; the -parallel 1 no-op is gone, and the PR body describes what actually landed.

(follow-up) One open item is a residual: pkg/resolver/resolver.go:153 still compiles every required namespace through bootstrapConsts, so a leaked worker can still race Consts.Intern via the loader path. I'd call that a P2 follow-up, not a blocker, because it's pre-existing structure the PR didn't introduce.

PR #677, opened 2026-08-04
build: single-source Go toolchain via go.mod toolchain directive

nnunley:
Make Go 1.26.5 the repository authority across CI, local tooling, and bootstrap paths, preventing version skew and malformed fallback downloads.

## What changed

- Pin `toolchain go1.26.5` in `go.mod` as the version authority.
- Remove the competing hard-coded Go pin from `mise.toml`.
- Derive Make bootstrap version from `go.mod`.
- Fall back only to a full-patch `go X.Y.Z` directive.
- Fail loudly when neither directive provides an exact version.

## Evidence

- `make test`: 242 Clojure files, 6,311 assertions, zero failures/skips.
- `make check-generated`, `go vet ./...`, and pinned lint passed.
- No-host bootstrap selected `go1.26.5.darwin-arm64.tar.gz`.
- Full-patch fallback passed; minor-only and missing-version mutants failed loudly.
- Restoring the prior Makefile reproduced an empty version and malformed archive URL.

## Migration note

A machine that ran golangci-lint under the old toolchain may carry a stale analysis cache. Run `golangci-lint cache clean` once if untouched files report a Go version mismatch.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -24,7 +24,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: actions/setup-go@v5
+      - uses: actions/setup-go@v6
         with:
           go-version-file: go.mod
       # Use the official action: it installs a cached, prebuilt golangci-lint
@@ -56,7 +56,7 @@ jobs:
           submodules: recursive
 
       - name: Set up Go
-        uses: actions/setup-go@v5
+        uses: actions/setup-go@v6
         with:
           go-version-file: go.mod
 
@@ -123,7 +123,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Set up Go
-        uses: actions/setup-go@v5
+        uses: actions/setup-go@v6
         with:
           go-version-file: go.mod
       - name: Race-detector tests (pkg/vm, pkg/rt)
@@ -144,7 +144,7 @@ jobs:
         with:
           submodules: recursive
       - name: Set up Go
-        uses: actions/setup-go@v5
+        uses: actions/setup-go@v6
         with:
           go-version-file: go.mod
       - name: Build for GOOS=wasip1
@@ -161,7 +161,7 @@ jobs:
         with:
           submodules: recursive
       - name: Set up Go
-        uses: actions/setup-go@v5
+        uses: actions/setup-go@v6
         with:
           go-version-file: go.mod
       - name: Build runtime-only without net/http
@@ -190,7 +190,7 @@ jobs:
         with:
           submodules: recursive
       - name: Set up Go
-        uses: actions/setup-go@v5
+        uses: actions/setup-go@v6
         with:
           go-version-file: go.mod
       - name: Install TinyGo
@@ -222,7 +222,7 @@ jobs:
           submodules: recursive
 
       - name: Set up Go
-        uses: actions/setup-go@v5
+        uses: actions/setup-go@v6
         with:
           go-version-file: go.mod
 
@@ -287,7 +287,7 @@ jobs:
         with:
           submodules: recursive
       - name: Set up Go
-        uses: actions/setup-go@v5
+        uses: actions/setup-go@v6
         with:
           go-version-file: go.mod
       - name: make gogen-diff
@@ -304,7 +304,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Set up Go
-        uses: actions/setup-go@v5
+        uses: actions/setup-go@v6
         with:
           go-version-file: go.mod
       - name: make check-generated
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -43,7 +43,7 @@ jobs:
             echo "perf-data branch does not exist yet; charts will be empty."
           fi
       - name: Set up Go
-        uses: actions/setup-go@v5
+        uses: actions/setup-go@v6
         with:
           go-version-file: go.mod
       - name: Build WASM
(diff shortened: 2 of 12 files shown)


mparrett (commented):
Reviewed `e08c243`. One P2, non-blocking.

## [P2] A third Go pin survives at the stale version

`Makefile:20` still reads `GO-VERSION ?= 1.26.3`. It feeds the makeplus/makes auto-install path (`Makefile:5-23`) that provisions Go into `.cache/local/` when `go` is not already on PATH, so it is the same stale 1.26.3 this PR removes from `mise.toml`, left in a third place. Nothing breaks: a contributor with no system Go gets 1.26.3 installed and `GOTOOLCHAIN=auto` pulls 1.26.5 on top for the build. What does not hold is the invariant the description states, that one line in go.mod moves CI, mise, and bare-go devs together. The next bump still needs a second edit, in the file least likely to be remembered. Pointing `GO-VERSION` at the go.mod toolchain line, or dropping the default so `makes` derives it, would finish the job in the same spirit as the rest of the change.

## Verdict

Correct, non-blocking. The skew it removes is real and nothing regresses.

Checked and not raised as findings: `Dockerfile:1` uses floating `golang:alpine`, a fourth version source, but it is unpinned and pre-existing rather than part of this change. The `toolchain` directive itself is right — all 15 `setup-go` call sites already pass `go-version-file: go.mod`, and `GOTOOLCHAIN=auto` covers the case where setup-go resolves only the `go 1.26` line. Dropping the mise pin is safe in practice because the Makefile fallback catches a contributor who would otherwise be left without Go.

mparrett (changes requested):
Blocking [P2]: CI needs to consume the exact toolchain pin. The workflows still use actions/setup-go@v5 with go-version-file: go.mod; support for preferring the toolchain directive was added in v6. The latest head run confirms the gap: setup-go logged Setup go version spec 1.26 and only happened to resolve 1.26.5. Once 1.26.6 is available, CI can select it, and GOTOOLCHAIN=auto will not downgrade that newer bundled toolchain back to 1.26.5. Please upgrade the setup-go call sites to v6+ or otherwise pass the exact version so CI actually follows the repository-owned source.

nnunley:
Rebased onto `main@8a8222d0` and added the guard this PR was missing.

Deleting `mise.toml`'s `go` entry single-sources the version, but nothing stopped a second pin from reappearing and drifting again — which is how `Makefile` (1.26.3) and `mise.toml` (1.26.5) came to disagree in the first place, with `go.mod` pinning neither.

`scripts/check_mise_go_sync.py` + a `pre-commit`-stage prek hook on `mise.toml`/`go.mod`:

- no `go` entry in `mise.toml` (the state this PR establishes) → pass, silently
- entry agrees with `go.mod` → pass
- entry disagrees → fail, naming both values and pointing at `go.mod`
- `go.mod` states only major.minor → any patch of that minor is accepted, since it pins no patch to match

Falsifier, the exact drift this PR removes:

```
$ printf '[tools]\ngo = "1.26.3"\n' > mise.toml && python3 scripts/check_mise_go_sync.py
Go version pins disagree:
  mise.toml pins go = "1.26.3" but go.mod's toolchain directive says 1.26.5.
  ...
exit=1
```

Six shape cases checked (toolchain vs stale patch, minor-only go.mod with a patch pin, wrong minor, no pin, go.mod stating nothing, single-quoted value). `prek run --all-files` is clean apart from `checkmake`, which fails identically on unmodified `main` — verified by running it on a fresh commit at `main@8a8222d0`.

Net diff is 5 files: `Makefile`, `go.mod`, `mise.toml`, `.pre-commit-config.yaml`, `scripts/check_mise_go_sync.py`.

---
Delivery verified by local `pr-claim-gate`: exact PR/validated head `0df165c2c1574cf206873a1e427ed5f9f2499200`; relevance claims `scripts/check_mise_go_sync.py`, `.pre-commit-config.yaml` occur in both the delivered delta and the PR feature delta; receipt `/Users/ndn/.local/state/pr-claim-gate/nooga-let-go/pr-677.json`.

nnunley:
Addressed the blocking [P2]. You were right that CI was not following the repository-owned pin — and the run logs show it directly.

All 15 `actions/setup-go` call sites across 7 workflows move `v5` → `v6`. v6.0.0's breaking change is exactly the toolchain handling you cited ("Improve toolchain handling to ensure more reliable and consistent toolchain selection and management", actions/setup-go#460).

Before/after from the actual job logs, same `build` job:

| head | setup-go | log line | GOTOOLCHAIN |
|---|---|---|---|
| `312aabc5` | v5 | `Setup go version spec 1.26` | `auto` |
| `f8e11da8` | v6 | `Setup go version spec 1.26.5` | `local` |

So CI now resolves the `toolchain go1.26.5` directive itself rather than landing on 1.26.5 by coincidence via the `go 1.26` line, and `GOTOOLCHAIN=local` means the resolved toolchain is used as-is. When 1.26.6 ships, CI follows go.mod instead of drifting ahead of it.

Also in this push, from the earlier thread: the mise/go.mod drift guard (`scripts/check_mise_go_sync.py` + a pre-commit-stage prek hook). Absent `go` entry passes silently; a present one must agree with go.mod or the hook fails naming both values.

14/14 checks green on `f8e11da8`.

---
Delivery verified by local `pr-claim-gate`: exact PR/validated head `f8e11da8589a4c92c4bace2d1b965a8ff686aa4a`; relevance claims `.github/workflows/go.yml` occur in both the delivered delta and the PR feature delta; receipt `/Users/ndn/.local/state/pr-claim-gate/nooga-let-go/pr-677.json`.

mparrett (approved):
Re-reviewed f8e11da. The blocker is resolved: all 15 setup-go call sites now use v6, and the current build log shows setup-go parsed 1.26.5 and set GOTOOLCHAIN=local. I also reviewed the mise/go.mod drift guard. All 14 required checks are green; no remaining blockers.

Issue #679, opened 2026-08-05
pre-push gate: compat-suite patch overlay breaks under hook env (GIT_DIR leakage) in worktrees


nnunley:
The compat-suite patch overlay (test/fanout_ratchet_test.go: ensureJankSuite / syncJankSubmodule / applyPatchOverlay) shells out to `git -C <submodule> ...` without scrubbing hook-context environment. During a real `git push`, git exports GIT_DIR (and friends) for the pre-push hook; those exec.Command calls inherit it, repo discovery resolves to the parent worktree's gitdir instead of the submodule, the overlay silently no-ops, and the gate fails deterministically:

- `TestJankSuiteCoversLetGoUnicodeScalar`: "patched fixture lacks the let-go Unicode scalar expectation"
- `TestClojureTestSuite/char`: 8/9 (the patched :lg expectation missing)

Reproduction: from a git worktree, `git push` (pre-push gate) reds on the two tests above; the identical suite run manually (`go test ./test -count=1`, or `prek run --hook-stage pre-push --all-files`) is green in the same checkout. Manual `git apply --check` / `--reverse --check` of test/patches/lg-char-unicode-scalar.patch behaves correctly outside hook context.

Impact: any push from a worktree reds the gate on an unrelated branch; historically read as flakiness (some of the recent gate reds attributed to the vm.Consts race were likely this, deterministic).

Fix shape: scrub GIT_DIR / GIT_WORK_TREE / GIT_INDEX_FILE from the env of the git subprocess calls in the overlay/sync helpers (exec.Command Env filtering), so submodule repo discovery is path-based as intended.

mparrett:
@nnunley this looks fixed and safe to close — flagging rather than closing it myself since it is yours.

PR #680 fixed it and merged on 2026-08-14. It never auto-closed because its body reads `Fixes issue #679`: the word between the keyword and the reference stops GitHub matching it.

The scrub is on `main` as `filterHookEnv` in `test/fanout_ratchet_test.go:146`, which strips `GIT_DIR`, `GIT_WORK_TREE`, `GIT_INDEX_FILE` and `GIT_PREFIX`, and names this issue in its doc comment alongside the reason: git honours the hook's `GIT_*` bindings over `-C`'s path-based discovery, which is what broke overlay patch application in worktrees.

Found while auditing open issues against merged PRs for the v1.13.0 cut (#718).

PR #680, opened 2026-08-05
fix(test): scrub hook env (GIT_DIR) from compat-suite overlay git calls

nnunley:
## Summary

Scrub GIT_DIR and related environment variables from git subprocess calls in the compat-suite overlay helpers. The pre-push hook environment leaks these variables, breaking submodule gitlink resolution in worktree contexts.

## What changed

- Added `filterHookEnv()` helper to scrub GIT_DIR, GIT_WORK_TREE, GIT_INDEX_FILE, and GIT_PREFIX from the process environment
- Applied the filter to all git exec.Command calls in syncJankSubmodule, applyPatchOverlay, and revertPatchOverlay
- These functions now use path-based -C repo discovery as intended

## Verification

- Pre-push gate passes locally when pushing this branch from the worktree (end-to-end proof of the fix)
- No behavior change outside hook context

Fixes issue #679

### DIFF
--- a/test/fanout_ratchet_test.go
+++ b/test/fanout_ratchet_test.go
@@ -114,6 +114,31 @@ func ensureJankSuite() {
 	})
 }
 
+// filterHookEnv removes GIT_* variables that the pre-push hook environment
+// sets. These variables break git -C repo discovery in subprocess calls (see
+// issue #679): the hook passes GIT_DIR, GIT_WORK_TREE, GIT_INDEX_FILE, and
+// GIT_PREFIX to its subprocesses; when those processes exec git with -C,
+// git honors the hook's GIT_* bindings over -C's path-based discovery,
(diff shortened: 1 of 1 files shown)


mparrett:
Review finding (non-blocking): could we add an automated regression test for the poisoned hook environment? The current overlay test runs outside hook context, so it stays green if these `cmd.Env` assignments are later removed. A focused test can inject a valid foreign/worktree `GIT_DIR` (plus the other scrubbed variables), exercise an affected submodule/overlay command, and assert path-based `-C` discovery still wins. That would turn the manual end-to-end verification for #679 into durable coverage.

I found no blocking correctness issues in the implementation. I also verified the targeted overlay test with an injected worktree `GIT_DIR`; all 14 GitHub checks are green.

mparrett (approved):

Issue #678, opened 2026-08-05
strict-mode AOT: 6/16 gold-aot fixtures only run via silent fallback — binding conveyance, @f deref rot, closure capture untracked


nnunley:
The new Phase-1 parity gate (branch `backend-parity-gate`, lands with the strict-mode audit) classifies the 16 `test/gold-aot` fixtures under `*ir-compile-strict*`. Result: **10 genuinely AOT-lower (Tier 1, all in cross-backend parity), 6 compile only via silent trampoline/bytecode fallback (Tier 2)** — i.e. they *run* under `-tags gogen_ir` but their lowering never actually happens, the #660 failure mode made visible per-fixture:

| Fixture | Failure class under strict | Existing tracking |
|---|---|---|
| `binding.lg` | dynamic binding / push-binding conveyance does not lower | — (this issue) |
| `ir_pipeline.lg` | `@f` deref / push-binding rot | — (this issue) |
| `pipeline_dump_ir.lg` | same `@f` deref rot (nil deref under strict; runtime error under default gogen) | — (this issue) |
| `closure_capture.lg` | closure capture shape does not lower strict | — (this issue) |
| `math_float.lg` | float arithmetic lowering | #551, #562 |
| `typed_cross_fn.lg` | cross-fn typed signature | #554 |

Why it matters: these six sit on the highest-blast-radius semantics for AOT (dynamic binding conveyance, closure capture) — exactly where a silent miscompile would be catastrophic. Until they lower strict, they are excluded from parity coverage (honest Tier-2 frontier) rather than counted as green.

Ask: close the untracked classes (binding conveyance / `@f` deref rot / closure capture) so these fixtures graduate to Tier 1 and the parity gate's coverage ratchets up. Related: #660 (the silent-fallback disease generally), #262 (gogen gap-closing epic).

Issue #660, opened 2026-07-31
Lowering falls back to bytecode silently, and only the bench harness notices


mparrett:
When a function does not lower, it is simply absent from the rendered Go and the program keeps working on the VM. That is the intended design — gloat's `lower-lg-to-go` docstring says so directly ("empty means nothing lowered (the program still runs, pure bytecode)"). The problem is that nothing in the shipping toolchain reports it, so "I compiled this AOT" and "this is running interpreted" are indistinguishable without measuring.

## Where the check exists today

`benchmark/aot/build.lg:124`:

```
(die (str ns "/run did NOT lower to a native fn — refusing to benchmark a silent bytecode fallback"))
```

The bench harness builds the binary, then probes the resulting var to confirm it dispatches to a native fn, and fails the build if it does not. That check is the only thing standing between a lowering regression and a benchmark number that quietly measures the interpreter.

Nothing outside `benchmark/aot/` does this. `ir.passes.pipeline/lower-ns-to-go` returns rendered Go source; a function that was dropped leaves no trace in the return value beyond not being there.

## Why it's worth fixing

Both #554 and #598 are instances of this failure mode:

- #554 — a fully-typed signature gets no boxed wrapper or override
registration, so the lowered body is never reachable.
- #598 — `(defn f ^long [args] …)`, the idiomatic return-hint position, is
misparsed as multi-arity and the function is dropped from lowering entirely.

Neither was caught by the toolchain. Both were found by someone noticing a benchmark number that did not move, and #554 sat open long enough to make our own `benchmark/aot/src/bench/loop-recur.lg` carry a hand-split workaround. A silently-dropped function is the failure mode this pipeline is most prone to, and right now it is invisible.

## The ask

Two pieces, either useful alone:

1. **A report of what lowered.** Have the lowering entry point return, or
write out, the set of functions it lowered and the ones it considered and dropped, with a reason where one is known. Machine-readable, so a build can assert on it.
2. **An opt-in strict mode** that fails when a function the caller asked to
lower did not. The bench harness's post-build probe is the after-the-fact version of this; doing it in the compiler is cheaper and catches it earlier.

#596's compile command looks like the natural home for both — a `lg compile` that can say "lowered 6 of 7, dropped `foo` (unsupported return hint)" and take a flag to make that an error.

Filing this separately from #596 because the reporting is useful even without a new CLI surface: `lower-ns-to-go` returning the manifest would let `benchmark/aot/build.lg` drop its post-build probe today.

[...]

PR #681, opened 2026-08-05
test(parity): gate complete-corpus strict capability and engine output

nnunley:
Backend miscompiles can pass human review when a differential run is optional, and a successful `gogen_ir` run does not by itself show that strict IR lowering succeeded. This PR turns engine-output parity and strict-capability coverage into a complete-corpus, measured, committed contract.

## What changed

- `make engine-parity-gate` discovers every `test/gold-aot/*.lg` fixture instead of maintaining hand-curated tiers.
- Each fixture is measured under four runs: bytecode control, `gogen_ir` control, bytecode strict, and `gogen_ir` strict.
- Classification is derived from those results:
  - **full strict parity**: both strict runs succeed and match their controls and each other;
  - **partial strict capability**: at least one strict run reaches the owned IR fallback seam;
  - **strict output divergence**: a successful strict run differs from its control or from the other engine;
  - **fatal**: setup, control, or unrelated strict execution fails.
- Partial and divergence are independent facts, so a fixture can appear in both.
- Strict rejection is attributed by the harness, not by matching process output: the strict wrapper binds its own fallback log and exits with a dedicated code only when that log records a real IR fallback. A runtime error that merely contains `ir-compile-strict:` stays fatal.
- Control and strict wrappers are symmetric — same prerequisites, same sequential top-level evaluation — so a difference is semantic, never setup noise.

## Reporting: what is and isn't working

The gate now prints a coverage report and writes it to `LETGO_PARITY_REPORT` when set. Non-full fixtures are measured a second time with strict off, so the report names the exact `defn` and the raw lowering reason behind every gap, plus a bucket census of distinct reasons and their sites. Full-parity fixtures need no extra run: strict success already proves zero fallback.

## Committed case ledger

`test/parity-ledger.txt` records one line per case, so a commit diff shows exactly what changed:

- the gate fails on `ADDED`, `REMOVED`, or `STATUS-CHANGED` cases;
- a removed case must be explained by a `TOMBSTONE` entry, and an `UNEXPLAINED` tombstone fails;
- a tombstoned fixture that reappears is `ADDED`, so the tombstone must be removed deliberately;
- `LETGO_PARITY_LEDGER_CLEANUP=1` clears tombstones inherited from an earlier revision.

Two shrink-only baselines still encode direction (partial capability, unattributed divergence). The ledger, both baselines, and any rederive are written as one staged, validated, rollback-safe set.

## Measured result

- 15 full strict parity
- 1 partial strict capability — `pipeline_dump_ir.lg`, `defn type-display`: `ir/lower: junk-below mismatch for block 15 (0 vs 2); unsupported shape`
- 0 strict output divergences
- 0 fatal results
- 16 total cases, 0 tombstones

That single gap is one instance of the largest structural class on the bytecode lowering path, so the ratchet has a real target rather than an aspiration.

## Scope

This proves strict capability plus normalized full-output agreement across the bytecode and `gogen_ir` engines. It does **not** claim that fixture source executed through generated Go or a native entry frame. Native-entry proof using `lg-compile --entry-frame` and the scoped Go AST oracle is the immediate follow-up.

## Validation

- Complete-corpus rederive is byte-identical to the committed ledger and baselines
- Tombstone lifecycle exercised end to end: removal fails, explanation passes, reappearance fails, prior-revision cleanup drops, restore returns byte-identical
- Ambient `GOFLAGS=-short` falsifier
- Classification, provenance, overlap, corpus-completeness, ledger-parser, and three-file rollback regressions
- `make gogen-diff`
- Full Go suite: 1,724 tests across 72 packages
- Repository pre-push hooks
- Independent engineering re-review approved this revision

---
[ap](https://github.com/nnunley/let-go-spacedock/blob/main/op-catalog/backend-parity-gate.md)
Related: #660, #678

This branch still contains the identical hook-environment scrub from #680; it resolves as overlap whichever lands first. The earlier #676 overlap is no longer part of this PR diff.


Makefile
```diff
@@ -222,6 +226,24 @@ check-selfhost: lowered $(GO)
 gogen-diff: lowered $(GO)
 	go test -run TestGogenAOTDiff -count=1 -v ./test/e2e/
 
+# Parity gate: Phase 1 — Tier-1 fixtures (genuine AOT, both backends pass strict mode and agree).
+# This is the baseline gate; all 10 Tier-1 fixtures must remain in parity.
+# Two-sided shrink-only ratchet: new divergence fails, stale allowlist entries fail.
+# Re-seed allowlists (after a reviewed change) with:
+#   LETGO_PARITY_REDERIVE=1 go test -run TestParityGatePhase1 -count=1 ./test/e2e/
+.PHONY: parity-gate-phase1
+parity-gate-phase1: lowered $(GO)
```
mparrett: [P1] Could we wire this target into an enforced path, probably the existing `gogen-diff` CI job or `make gogen-diff`? Right now neither CI nor the pre-push configuration invokes it, while the ordinary test lanes use `-short` and skip `TestParityGatePhase1`. As written, the new strict parity gate can fail without blocking a merge — the exact gap this PR is aiming to close.

test/e2e/parity_gate_phase1_test.go
```diff
@@ -0,0 +1,195 @@
...
+
+	// Run tier-1 fixtures on both backends
+	diverged := map[string]bool{}
+	for _, name := range tier1Fixtures {
+		wrappedPath := wrappedFixtures[name]
+		bcOut, bcOk := runFixture(bc, wrappedPath)
+		aotOut, aotOk := runFixture(aot, wrappedPath)
+
+		// Both should succeed and agree
+		if !bcOk || !aotOk || bcOut != aotOut {
```
mparrett: [P1] Could we keep strict-AOT failures separate from output divergence here? This collapses both-backend strict failures, one-sided failures, and actual output differences into the same `diverged` map, after which an entry in either allowlist accepts any of them. In an adversarial probe, `pipeline_dump_ir.lg` failed on both sides; adding it only to the gogen allowlist made the gate pass. A Tier-1 strict failure should probably remain an unconditional failure, with allowlists reserved for precisely attributed parity deviations.

test/e2e/parity_gate_phase1_test.go
```diff
@@ -0,0 +1,195 @@
...
+			diverged[name] = true
+			t.Logf("DIVERGE %s: bc_ok=%v bc_out=%q aot_ok=%v aot_out=%q",
+				name, bcOk, bcOut, aotOk, aotOut)
+		} else {
+			t.Logf("AGREE %s: output=%q", name, bcOut)
+		}
+	}
+
+	// Load per-backend allowlists
```
mparrett: [P2] The documented `LETGO_PARITY_REDERIVE` flow is not connected: the environment variable is never read and `writeParityXfail` is never called. I ran the documented command and it exited successfully without modifying either allowlist. Could we add the re-derive branch before normal validation, or remove the command and unused writer?

mparrett (changes requested):
Thank you for putting this together. The strict-mode audit and backend-parity direction are valuable + the careful evidence and the prerequisite hook fixes included here are appreciated. Ran both new targets locally: the happy-path gate passes all 10 Tier-1 fixtures, and the audit reproduces the reported 10/6/0 classification.

I found three issues to tighten up before merging, mainly so the gate provides the enforcement and strict-AOT guarantees it promises. I’ve left the details inline.

test/parity-divergence.txt
```diff
@@ -0,0 +1,5 @@
+# Machine-generated shrink-only baseline of unattributed strict engine-output divergences.
+# The target state is empty. Rederive only from a complete valid census.
+# Partial and divergence are independent facts; a fixture may appear in both.
+# LETGO_PARITY_REDERIVE=1 make engine-parity-gate
+
```
mparrett: Non-blocking nit: `git diff --check` reports this as a new blank line at EOF. The rederive path reproduces it, so the durable cleanup would be in `baselineBytes` (avoid the second trailing newline when the generated baseline has no entries), rather than editing this generated file alone.

mparrett (approved):
Thanks for the substantial follow-up here. I rechecked the current implementation and reran `make gogen-diff`: both mandatory legs pass, strict failures are now classified independently from allowlisted capability gaps, and the documented rederive path is connected and reproduces the committed ledger and baselines byte-for-byte. All three requested changes are addressed.

Approving with one non-blocking whitespace nit inline.

PR #683, opened 2026-08-05
build: content-gate every committed generated artifact, not just three

mparrett:
`make check-generated` verified three of the six committed generated artifacts. It regenerated through `lgprimgen` and `lgbgen` directly, and three artifacts — `pkg/ir/op_generated.go`, `pkg/rt/ir_bridge_generated.go` and `pkg/rt/core/ir/data/generated.lg` — are produced only by `scripts/generate.lg`, so a gate that never ran it could not see them drift. That matters because a stale committed artifact means the bytecode and `gogen_ir` engines run different versions of the IR pipeline, which is the failure `parity-full` exists to catch and the gate exists to prevent.

This replaces #634, which patched `cmd/lginterop/prims_emit.go` — a file #639/#640 moved to `internal/primgen/`, where the import-grouping bug it fixed has since been fixed independently. The thesis survived the move; the diff did not. Closing that one in favor of this.

## Why the widening is sound

Widening the gate to drive the whole pipeline is only correct if `make generate` regenerates every artifact in the list. It did not: the corefns registrar was reachable solely through the Makefile's `core_compiled.lgb` prerequisite, an mtime rule that stops firing once the committed registrar is newer than its sources. A gate regenerating through `make generate` alone would have passed corefns drift silently.

#641 has since landed that fix on `main`, gated on the provenance manifest rather than on mtime, so `generate.lg` and the tracked list agree. This PR depends on it rather than duplicating it.

## Change

- `scripts/check-generated.sh` content-checks the six committed artifacts: it preserves them, regenerates through the full `make generate`, and `cmp`s each. Driving the whole pipeline rather than the two generators is what widens the coverage.
- The gate is non-mutating on every path. It restores the working-tree bytes on pass, on failure, on generator error, and on interrupt, so a failure no longer strands a half-regenerated multi-file tree with no printed way back.
- It reports every stale artifact instead of exiting at the first, and a failing generator surfaces the underlying error rather than `make[1]: *** [generate] Error 1`.
- The gate lives in `scripts/` rather than in a Makefile recipe so it stays VCS-agnostic: the repository is also used from `jj` workspaces with no `.git` directory, so it preserves and restores file bytes directly rather than reaching for `git stash`.
- `docs/contribution-policy.md` described the gate as bundle-and-lowered-tree only, which was already stale against `main`.

## On the `check-generated-manifest` prerequisite

Dropping it does not change whether a stale digest fails CI. It still does, in the `build` job (which runs `check-generated-manifest` directly) and in `TestGeneratedArtifactsAreFresh`. Neither is touched here.

What changes is that a stale digest no longer aborts `check-generated` before the content gate can run. The digest is a proxy; the `cmp` is the check that binds sources to artifacts, and the proxy should not suppress the answer.

`pkg/rt/generated.sums` is preserved and restored rather than compared, because lgbgen's `writeBundle` refreshes it on every regeneration — without that, a `check-*` target would mutate a tracked file.

## Verification

- Passes on current `main`, reporting all six artifacts in lockstep, with the tree byte-identical afterwards.
- Catches injected drift in each of the five text artifacts, including the three the previous gate could not see, and reports all five in a single run.
- Restores the exact pre-gate bytes on failure, confirmed by digest.
- `make check-generated-manifest` still passes standalone.

## Shell portability

The script targets plain `/bin/sh` with no `-e`. The makeplus branch that sets `SHELL := bash` and `-e -o pipefail` is skipped whenever `go` is already on PATH, which is every CI run. `mktemp` and `cp` failures are checked explicitly and a `trap` removes the temp dir on interrupt.


nnunley:
Ran the gate rather than reading it. On this head all six `GENERATED-TRACKED` artifacts
report OK and the tree is clean afterward; injecting drift into `pkg/ir/op_generated.go`
reports exactly that file, still reports the other five, exits non-zero, and leaves
`op_generated.go`, `core_compiled.lgb` and `generated.sums` byte-identical to what the gate
found. The non-mutating claim holds on the failure path.

The coverage gap is real: `op_generated.go`, `ir_bridge_generated.go` and
`ir/data/generated.lg` come only from `generate.lg`, so the old gate — which drove lgprimgen
and lgbgen directly — could not see them drift. The corefns defect is real too, and the fix
is the right one: the printed remediation could not fix the failure it printed.

**One bug.** The comment at `Makefile:318-320` says the tree is restored on every path, but
the trap at `:356` is `rm -rf "$tmp"` on `EXIT INT TERM` — it deletes the stash without
restoring. Ctrl-C during the ~2min regeneration therefore leaves a half-regenerated tree
*and* destroys the only copy of the committed bytes, which is the one case where you most
want it back. Restore-then-remove fixes it. Related: the restore loop at `:362`/`:375` uses
bare `cp` with no `|| exit` under a shell without `-e`, so a failed restore is silent, while
the stash loop does check.

**Ordering, and this one is on me.** #641 is mine and it makes `make generate` selective —
`staleOutputs` in `pkg/genmanifest/depmanifest.go` hashes *inputs* and only `os.Stat`s
outputs for readiness, never their content. Compose that with this gate and a stale committed
output whose inputs are unchanged means generate skips the stage, so the `cmp` compares the
file to itself and passes — blind to the drift the gate exists to catch. Your injected-drift
check passes today only because generation is unconditional.

The two also both add a corefns line to `scripts/generate.lg`, and merging them auto-merges
*cleanly* into a tree containing both stages. I would rather land this first and take the
burden on #641: it rebases, resolves the duplicate stage, and has to force a full
regeneration under the gate — or teach the staleness check to hash output content.

**One bug I found by running the gate rather than reading it, and it is caused by this PR.**
Dropping the `check-generated-manifest` prerequisite leaves `$(GO)` as the only one, and
`check-generated` is not `.PHONY` in either version. On `main` the prerequisite has no matching
file, so make always rebuilds and the omission is invisible; without it, a file named
`check-generated` in the repo root makes the target up to date:

```
$ make check-generated          # on this head, with a stray ./check-generated present
make: `check-generated' is up to date.
EXIT=0

$ mv check-generated /tmp/ && make check-generated
OK: core_compiled.lgb in lockstep with source.
EXIT=0
```

Both exit 0; one verified six artifacts and the other verified nothing. `go build ./cmd/check-generated`
produces exactly that filename, and my working copy had one in it, which is how I hit it.
Adding `check-generated` and `check-generated-manifest` to a `.PHONY` fixes it.

mparrett:
@nnunley Both bugs fixed in 8bc08883.

The EXIT/INT/TERM trap now restores from the stash before removing it, and each restore `cp` fails loud (`|| exit 1`) the same way the stash loop does. `check-generated` and `check-generated-manifest` are `.PHONY` — confirmed a stray `./check-generated` no longer makes the gate a silent no-op.

Agreed on landing this ahead of #641; the gate needs unconditional generate until that side forces a full regen (or hashes output content). Happy to take the duplicate `generate.lg` corefns line as a clean rebase conflict when #641 moves.

nnunley (changes requested):
So, broadly looks good. That said, I'm finding the amount of scripting inside the makefile a bit of a code smell. Can you factor out your check generated logic into a script under either tools/ or scripts/ ?

mparrett:
@nnunley Thanks for the review. Agreed. Factored the gate into `scripts/check-generated.sh` in 1b90840. `make check-generated` is now a thin entry point; the script owns the artifact list, stash/restore handling, full-regeneration byte comparisons, and the lowered-tree compile/native-dispatch checks. I also updated the contribution-policy references to point at the script.

Verified with:
- `make check-generated`
- `make check-generated-manifest`
- `shellcheck scripts/check-generated.sh`
- `git diff --check`

mparrett:
@nnunley — the gate logic is out of the Makefile. It lives in `scripts/check-generated.sh` now, and the `check-generated` target is a single `@scripts/check-generated.sh` line: Makefile −89/+5, script +99, plus a matching update to `docs/contribution-policy.md`. That is `refactor(build): extract generated gate script`, which went up after your review, so it will not have been in view when you read this.

The branch is also rebased onto current `main`. The red merge box was `pkg/rt/generated.sums`, which is the merge-driver gap rather than a content conflict, same as on #734 and #781; it is green now and CI is running.

Ready for another look when you have a moment.

nooga (approved):
LGTM. nnunley's request (factor the gate out of the Makefile) is addressed by `scripts/check-generated.sh`, the earlier trap/.PHONY issue is fixed, and CI's check-generated job is green on the rebased head. Only overlap with #734 is the generated manifest. nnunley, shout if you want another look; otherwise I'll merge.

mparrett:
Rebased onto `a6763e77`; head is now `fe27a1e4`. The branch is 6 commits down to 3, and no longer touches any generated file.

The body predicted this: #641 also added corefns generation to `generate.lg`, and whichever landed second would drop that line. #641 landed first, so three commits here were dead weight:

- `fix(generate): emit the corefns registrar from generate.lg` — superseded. #641's version is gated on the provenance manifest rather than run unconditionally, and calls `run-command!`; the copy here still called `run!`, which is now core's two-arity `run!`, so the merged script failed with "expected 2 args, got 9".
- `build: refresh generated provenance for the generate.lg change` — `generated.manifest` only, refreshing provenance for the commit above.
- `chore(generated): refresh the manifest digest after rebasing onto main` — `generated.sums` only.

What remains is the `Makefile`, `docs/contribution-policy.md` and `scripts/check-generated.sh`. With those three dropped the rebase replayed with no conflicts, and `scripts/check-generated.sh` reports all six artifacts in lockstep with the lowered tree compiling and dispatching natively, regenerating nothing and leaving the tree clean. So the branch never needed to carry derived files; those commits were the residue of an earlier rebase, and carrying them is what made each subsequent rebase conflict.

I refreshed the body to match. All 15 required checks green, and the branch is mergeable again.

nooga (approved):
@mparrett Good fixes — the trap/cleanup ordering bug, the silent `cp` failure, and the missing `.PHONY` guard are all addressed, and pulling the gate logic into `scripts/check-generated.sh` makes it much easier to reason about.

One gap I want to flag explicitly rather than let it ride silently: I reproduced nnunley's predicted hole — for the 3 artifacts gated behind #641's selective regen (`zz_primitives_generated.go`, `corefns/zz_primitives_generated.go`, `core_compiled.lgb`), if a committed artifact and its recorded manifest hash are already self-consistently wrong with no input/generator change, `make generate` skips the stage and the gate ends up comparing the stash to itself and passing. It's narrow — needs the manifest to already agree with the wrong content, not just a forgotten regen — but the PR's claim of catching injected drift across all five text artifacts overstates it for these three. Worth a follow-up issue rather than blocking this PR on it. Green light to merge.

mparrett:
@nnunley I saw you have changes requested on this PR. Can you let me know if anything else is needed on the current head, or whether that can be cleared? Thanks!

nnunley (approved):

Issue #682, opened 2026-08-05
ci(tinygo): failed wasmtime install passes and breaks the boot step


mparrett:
`tinygo-wasi-build` can go red on a PR that changed nothing related to TinyGo, and the reported failure points at the wrong step.

Seen on [run 30875299785](https://github.com/nooga/let-go/actions/runs/30875299785/job/91885451485) (2026-08-04, on #628 at `e21014a7`). The `Install wasmtime` step logged a failure and still passed:

```
Installing latest version of Wasmtime ({)
  Checking for existing Wasmtime installation
  Fetching archive for Linux, version {
curl: (3) nested brace in URL position 75:
https://github.com/bytecodealliance/wasmtime/releases/download/{/wasmtime-{-x86_64-linux.tar.xz
Error: Could not download Wasmtime version '{'.
```

`install.sh` failed to resolve "latest" and interpolated a literal `{` into the download URL. It printed `Error:` and exited 0, so the step was green. Two steps later:

```
wasmtime: command not found
##[error]Process completed with exit code 127
```

The job is attributed to `Boot past the reflect shim (must not trap)`, so it reads as "this PR broke the TinyGo boot" — the exact regression that check exists to catch. The TinyGo build itself succeeded; only the wasm runtime was missing.

It passed on a re-run with no changes, and `.github/workflows/go.yml` is byte-identical between the two bases I compared (`1f281df4` and `6f75229d`), so the failure is transient and upstream, not in the repo. The cost is the misattribution: it lands on whichever PR is in flight.

Two fixes, either of which is enough:

1. **Pin the version.** The `Install TinyGo` step directly above already does this (`v0.41.1` via `wget` + `dpkg`), so pinning wasmtime the same way makes the job consistent and removes the "latest" resolution that failed here.
2. **Verify after install.** A `wasmtime --version` line in the install step makes a failed install fail where it happened, rather than two steps later.

I can send the pin as a PR if that shape sounds acceptable.

PR #685, opened 2026-08-05
ci(tinygo): pin wasmtime and verify it landed on PATH

mparrett:
`tinygo-wasi-build` could fail on a PR that touched nothing related to TinyGo, and blame the wrong step while doing it. #682 has the full trace; the short version is that `install.sh` resolves "latest" by sed-ing `tag_name` out of an unauthenticated GitHub API response. When that call is rate-limited the sed yields a literal `{`, the download 404s, and the script still exits 0. The install step stayed green, and the job died two steps later at `wasmtime run` with `command not found` — attributed to `Boot past the reflect shim (must not trap)`, which reads as a real TinyGo regression.

Two changes, both in the `tinygo-wasi-build` job:

- **Pin the version.** `--version v47.0.3` sends the installer down its specific-version branch, which never calls the GitHub API, so the failure mode is gone rather than reported better. This also matches `Install TinyGo` directly above, which has always pinned (`v0.41.1`).
- **Verify after install.** `wasmtime --version` as its own step, since `GITHUB_PATH` only affects later steps. A bad install now fails at the install, not two steps downstream.

Verified the pinned path end to end: the installer takes `--version` and downloads that archive with no API call, and the `v47.0.3` linux asset the workflow fetches returns 200.

v47.0.3 is current as of 2026-07-31. Pinning does mean someone bumps it eventually; that seems right for a job whose purpose is proving our TinyGo build boots past the reflect shim, not tracking wasmtime releases. If you want tip coverage it belongs in a scheduled job, not a per-PR gate that can go red on an unrelated PR.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -203,10 +203,17 @@ jobs:
         run: |
           wget -q https://github.com/tinygo-org/tinygo/releases/download/v0.41.1/tinygo_0.41.1_amd64.deb
           sudo dpkg -i tinygo_0.41.1_amd64.deb
+      # Pinned like TinyGo above. Unpinned, install.sh resolves "latest" by
+      # sed-ing tag_name out of an unauthenticated GitHub API response; when
+      # that call is rate-limited the sed yields a literal `{`, the download
+      # 404s, and the script still exits 0 — so the job died two steps later
+      # at `wasmtime run` with `command not found` (#682).
       - name: Install wasmtime
         run: |
-          curl -sSf https://wasmtime.dev/install.sh | bash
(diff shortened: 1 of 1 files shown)


nnunley:
Confirmed the mechanism rather than assuming it: `install.sh` does take `--version`, the
v47.0.3 tag resolves, and the verify step is a separate step so the `$GITHUB_PATH` append
from the previous one has actually taken effect — it is a real check, not a tautology.
Turning a rate-limited API into an attributable failure at the install step instead of a
`command not found` two steps later is worth the eight lines.

Only nit: the version is a literal in the workflow, like the TinyGo pin above it, so a future
bump touches two places. Fine at this size.

nnunley (approved):

Issue #656, opened 2026-07-30
`reduce` is 1.75x slower than v1.12.2 — bisected to #639


mparrett:
## What

`(reduce + 0 (range 1000000))` runs 1.75x slower on `main` than before `428ce9e3` (#639, "hoist 222 clojure.core primitives to named `//lg:native` decls"). darwin/arm64, go1.26.3, hyperfine 20 runs:

| build | mean | GC cycles | CPU / wall |
|---|---|---|---|
| `862c7cec` (#622, the parent) | 33.1 ms ± 0.9 | 6 | 33.8 / 33.1 |
| `428ce9e3` (#639) | 57.7 ms ± 0.8 | 30 | 72.5 / 57.7 |

Allocation rather than compute: five times the GC cycles for identical work, and CPU exceeding wall time where the parent's was flat. Confirmed in both benchmark orders, so it isn't thermal drift.

## Blast radius: `reduce` only

Same two binaries across the whole suite:

| workload | ratio |
|---|---|
| fib | 1.03x |
| tak | 1.07x |
| loop-recur | 0.96x |
| map-filter | 0.95x |
| persistent-map | 1.04x |
| transducers | 1.06x |
| **reduce** | **1.77x** |

Everything but `reduce` is inside run-to-run noise.

## Plausible mechanism

`reduce` invokes its reducing fn once per element, a million times in this fixture, through `ec.Invoke(mfn, []vm.Value{acc, elem})` in `reduceColl` (`pkg/rt/native_prims.go`). `+` is one of the 222 prims #639 hoisted, so anything the hoist added to a core fn's per-call path gets multiplied by the element count. That matches both the allocation signature and why only `reduce` moved.

`map-filter` also calls `reduce` and is unaffected, which fits: its reduce walks 100 elements, not 1,000,000.

I have the commit, not the line: which of the hoisted prims regressed, and what allocates on each call, are both still open.

## Repro

```bash
cat > reduce-in-main.clj <<'EOF'
(defn -main [] (reduce + 0 (range 1000000)))
(-main)
EOF

for sha in 862c7cec 428ce9e3; do
  git checkout -f -q $sha && git checkout -q -- . && git clean -xqfd
  make build && cp lg /tmp/lg-$sha
done
hyperfine --warmup 5 --runs 20 "/tmp/lg-862c7cec reduce-in-main.clj" \
                               "/tmp/lg-428ce9e3 reduce-in-main.clj"
[...]

Issue #531, opened 2026-07-16
Consolidate the spread-out native-behavior mechanisms (registration + interop emitters)


nnunley:
## What "spread out" means concretely

Defining one piece of runtime behavior — "make `upper-case` callable from let-go" —
currently has **six mechanisms**, and which one a given function uses is an accident
of when it was written:

| # | Mechanism | Where | Scale | Namespace/name/privacy decided |
|---|---|---|---|---|
| 1 | Imperative `ns.Def` + anonymous `Wrap` closure | `pkg/rt/lang.go` | ~384 Def sites, ~5,800 lines | at the call site, imperatively |
| 2 | `//lg:native`-annotated named Go funcs → generated registrar | `pkg/rt/native_prims.go` (+`_lifecycle`), emitted by `cmd/lginterop -primitives` | 14 directives so far | at the definition, declaratively |
| 3 | Direct-call-eligible builtins (dual-use: interpreter registration AND AOT direct-call target) | `pkg/rt/builtins/` | growing | in lang.go, still imperatively |
| 4 | Pure `.lg` core, compiled to bundle + lowered tree | `pkg/rt/core/*.lg` | most of the stdlib | in the `.lg` source |
| 5 | External-package interop: go/types scan → gogen emit | `cmd/lginterop` main.go + `scripts/lginterop.lg` → `pkg/rt/interop_<pkg>.go` | xxh3 today, glplat pending | generated installer |
| 6 | Reflection-based host interop (boxed values, method dispatch) | `pkg/rt/host_interop.go`, `vm.MustBox` | ambient | n/a (dynamic) |

The cost is exactly the #411 diagnosis: re-homing `clojure.string` primitives,
adding an `io` ns, or privatizing the ~82 compiler internals means hand-editing
hundreds of imperative call sites in mechanism 1, because that's where most
behavior still lives. Meanwhile mechanisms 2 and 3 already demonstrate the fix
but cover a sliver of the surface.

## Correction to the epic text first

#528 lists `scripts/lginterop.lg` under "Superseded but still alive." **It is not
superseded — it is a live runtime dependency of mechanism 5.** `cmd/lginterop`'s
external-package mode reads `scripts/lginterop.lg` from the repo root on every
run (`writeGenScript`, cmd/lginterop/main.go:419), concatenates the exports onto
it, and shells out to `lg` to render the interop file. Deleting it today breaks
`lginterop -packages`. (The #467 corpus exclusion was about ir-stress coverage,
[...]

PR #686, opened 2026-08-05
perf(rt): restore reduce's ArrayVector and Range fast paths

mparrett:
`reduce` is 1.88x slower than v1.12.2 on `main` (#656, measured 2026-08-05). The cause is a duplicate registration, not anything in the primitives #639 hoisted.

## Root cause

`reduce` has had two implementations for months:

- a `vm.NewCtxNativeFn` closure in `pkg/rt/lang.go`, carrying an ArrayVector fast path, a Range fast path (direct arithmetic, zero seq/chunk allocation), and a reused two-element argument buffer;
- the `//lg:native` `Reduce`/`Reduce3` decls in `pkg/rt/native_prims.go`, delegating to `reduceColl`, which has none of that.

Both predate #639. The generated registration used to land in a distinct `"clojure.core"` namespace instead of canonical `"core"` — the alias bug #639's first commit fixed — so it was invisible and the closure won the name. Once the alias resolved, the generated registrar took the var root, because it drains after `installLangNS`. `reduce` switched to the naive body with no diff to `reduce` itself and no failing test.

Reducing `(range 1000000)` through that body allocates 66.5 MB against 17.5 MB: a fresh `[]vm.Value{acc, elem}` per element, plus an `ArrayChunk` per 32 elements via `ChunkedFirst`. 33 GC cycles against 6.

`TestReduceRangeFastPath` stayed green throughout. It asserts results, not which path produced them.

## What changed

Both fast paths and the reused buffer move into `reduceColl`, and the lang.go closure is deleted, leaving one implementation. The general seq path gets the buffer too, which it never had: `ec.Invoke` does not retain its argument slice, so one buffer per `reduce` call beats one per element.

The rest of the change adds the missing guard. A name registered in both places is a silent hazard: the closure never runs, and nothing stops the two bodies drifting. Every generated registration that lands on a var an earlier registration already interned is now recorded, and a test asserts that set against an accepted list, so adding a `//lg:native` decl for an already-hand-registered name fails in CI.

The recorder ignores names an earlier *generated* registration installed. Registration state is process-global, so without that check a repeat registration records its own predecessor as a collision, making the set depend on how many times the registrar ran: `go test -count=2 ./pkg/rt` reported phantom collisions for the two probe primitives that `pkg/rt`'s own tests bind through the registration path.

Twelve names are in the accepted set: `conj`, `deref`, `get`, `int`, `name`, `namespace`, `nth`, `pop-binding!`, `push-binding!`, `some`, `str`, `subs`. Each is a duplicate implementation awaiting the same port-then-delete treatment. All twelve pairs were compared body for body: the `//lg:native` decls are faithful ports, several carrying the closure's comments verbatim, so `reduce` is the only one that had drifted. The duplication is still worth retiring — a second copy nothing executes will drift again — but no other name is silently losing behavior today.

## Verification

`(reduce + 0 (range 1000000))` on darwin/arm64, hyperfine 3 warmup / 14 runs, confirmed in reversed order (2026-08-05):

| build | mean | GC cycles |
|---|---|---|
| pre-#639 (`862c7cec`) | 34.6 ms | 6 |
| `main` | 64.9 ms | 33 |
| this branch | 41.3 ms | 9 |

Every reduce shape improves, not only the two with fast paths: over a list 37.1 → 30.1 ms, over a vector 29.8 → 25.5 ms, over a map 97.1 → 90.1 ms (pre-#639 parity).

Semantics spot-checked across 2-arg and 3-arg forms, nil and empty collections, vectors, lists, maps, and `reduced` short-circuit; `pkg/rt`'s existing `TestReduceRangeFastPath` and `TestReduceArrayVectorFastPath` cover the ported paths.

Two new tests cover the guard. `TestNoNewShadowedHandRegistrations` ratchets the accepted set. `TestReRegistrationIsNotAHandShadow` pins that a repeat generated registration is not a collision; it fails when that check is removed. `go test -count=2 ./pkg/rt` is green.

`make check-generated` passes; `pkg/rt/native_prims.go` feeds the manifest digest, hence the `generated.sums` bump.

## Residual

This closes most of the gap to pre-#639, not all of it, and the remainder is shared with other primitives. On `main`, `some` (1.49x), `get` (1.18x), `nth` (1.17x), `str` (1.11x), `conj` and `deref` (1.06x) all sit above pre-#639 with no fast-path story of their own. `some` is clean at #639 and bisects to #645, so at least one further regression is in flight. Measured separately, not in this PR.


mparrett:
> ## Perf (pkg/vm micros) — base vs head, same runner
> Base `6f75229d77ff` vs head `2220c61d9fc1`. Anchor-normalized; informational, not a gate.
> 
> **bench-ratchet** — 0 regression(s) > 12.0% budget, 0 missing, 0 new
> 
> * baseline: `AMD EPYC 9V74 80-Core Processor` / `go1.26.5` / `linux-amd64`
> * current:  `AMD EPYC 9V74 80-Core Processor` / `go1.26.5` / `linux-amd64`
> * anchor: baseline `1.091 ns/op`, current `1.093 ns/op` (`+0.2%`)
> 
> Benchmark	Baseline×	Current×	Δ%	Wall	Best since	Status
> _nothing past the budget threshold_						
> _Showing 0 of 53 benchmarks (past budget). [Full table in the run summary →](https://github.com/nooga/let-go/actions/runs/31055573152)_

Might be nice to surface wins here, too. 🤷🏻

mparrett:
**Single-shot (EPYC 9V74):** 0 of 53 benchmarks past the 12% budget, anchor stable at +0.2%.

**Repeat A/B (EPYC 7763, N=7 interleaved ABBA):** all 53 families comparable, 0 missing / 0 new / 0 flaky, would-gate empty at 6%, 8% and 10%. Largest median 2.70% (VectorConj/ArrayVector/10); median of all medians −0.42%. The two families that would have shown trouble are flat:

FuncInvoke/Closure +1.80%, FrameDispatch +0.70%.

So the fix is neutral on the VM micros, which is what it should be — it touches pkg/rt, not pkg/vm. And "0 new" confirms the prediction: no reduce family exists, so this run says nothing about the actual win.

One thing worth noting from the raw cycles: single-cycle deltas reach ±10% while every median sits under 3%. Cycle 3 is visibly contaminated —  its anchor dropped 1.246 → 1.177 ns and most families spike in that one cycle. The interleaving plus median absorbed it exactly as #445  predicted, which is a decent argument for the repeat lane over the single-shot one.

pkg/rt/shadowed_registration_test.go
```diff
@@ -0,0 +1,93 @@
...
+	"clojure.core/int",
+	"clojure.core/name",
+	"clojure.core/namespace",
+	"clojure.core/nth",
+	"clojure.core/pop-binding!",
+	"clojure.core/push-binding!",
+	"clojure.core/some",
+	"clojure.core/str",
+	"clojure.core/subs",
+}
```
nnunley, line 36: I think the issue is that the conversion to lg native isn't complete yet, so these weren't covered as needed. This should be tracked by https://github.com/nooga/let-go/issues/531

nnunley (approved):
Note: I think the shadowed hand registered methods are a gap in the migration to lg:native. A fast follow would be to create tracking issues under https://github.com/nooga/let-go/issues/531 to complete the migration.

mparrett:
Filed both under #531:

- **#696** — retire the twelve names currently in `knownShadowedHandRegistrations`. All twelve pairs were compared body for body while working on this PR, and the `//lg:native` decls are faithful ports, so it's deletion work rather than porting. `reduce` was the only one that had drifted.
- **#697** — drain the rest of `pkg/rt`'s hand registrations family by family, with #531's S4 ratchet as the exit criterion. It carries the current counts: `lang.go` is already past halfway, and the untouched clusters (`math`, `async`, `syscall`, `term`, `os`) are coherent enough to be one PR each.

Also rebased onto main, which cleared the merge conflict. Checks are green.

PR #688, opened 2026-08-05
feat(rt): add os/unzip

### LINKED ISSUE #687: Add native os/unzip for archive extraction
### What's missing

let-go has no way to extract a zip archive. Anything that needs to expand a jar, a wheel, or a downloaded release archive has to shell out to `unzip` via `os/sh`.

### Why it matters

That makes an external binary a hard runtime requirement for otherwise pure let-go tooling. It is the last host-binary dependency in a Maven dependency resolver I am building on let-go, and `unzip` is not present by default everywhere.

### Proposed shape

`(os/unzip zip-path dest-dir)` returning `dest-dir`, defined in `installOsNS` alongside `os/sh` and inheriting that file's `!tinygo` gate. Go's `archive/zip` does the work.

The part worth getting right is path containment: an entry named `../evil.txt`, or one writing through a symlink that already exists in the destination, must not escape. `os.Root` (go1.24, and go.mod already requires 1.26) handles both — it opens each path component under the destination and refuses any that escapes, which also closes the check-then-write race a manual guard leaves open. For archives arriving over a network, skipping symlink entries rather than recreating them seems the right trade.

abogoyavlensky:
Resolves: #687 

`(os/unzip zip-path dest-dir)` extracts a zip archive into `dest-dir` and returns it. Motivation: dependency tooling built on let-go currently shells out to `unzip` to expand jars, and this removes that host-binary requirement.

Extraction is confined by `os.Root`, which opens each path component under the destination and refuses any that escapes. That covers both the lexical case (`../evil.txt`) and the symbolic one (`link/x` where `dest/link` already points elsewhere), and unlike a check-then-write guard it leaves no window for a process sharing the destination to swap a validated directory for a symlink after the check. Absolute entry names are refused rather than silently re-rooted.

Safety over fidelity otherwise: symlink, device, fifo and socket entries are skipped rather than recreated. Permissions follow the `unzip(1)` contract — as recorded, masked by the umask — with a 0644 fallback for DOS/FAT entries that record no unix mode. A failed extraction is never worse than no extraction: a directory entry colliding with an existing file is an error rather than a swallowed `EEXIST`, and each entry is opened before its destination is cleared.

Tests: `pkg/rt/os_unzip_test.go` builds archives programmatically, including the adversarial ones, plus a `test/os_unzip_test.lg` smoke test.

### DIFF
--- a/pkg/rt/os.go
+++ b/pkg/rt/os.go
@@ -8,12 +8,15 @@
 package rt
 
 import (
+	"archive/zip"
 	"bytes"
+	"errors"
 	"fmt"
 	"io"
 	"net"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"runtime"
 
 	"github.com/nooga/let-go/pkg/vm"
@@ -262,6 +265,28 @@ func installOsNS() {
 		return vm.Int(port), nil
 	}))
 
+	// os/unzip — (os/unzip zip-path dest-dir) → dest-dir
+	// Extracts a zip archive into dest-dir, creating it if missing and
+	// overwriting existing files. Entries that would land outside dest-dir
+	// are refused (see unzipEntryTarget); symlink entries are skipped.
+	ns.Def("unzip", mustWrap(func(vs []vm.Value) (vm.Value, error) {
+		if len(vs) != 2 {
+			return vm.NIL, fmt.Errorf("os/unzip expects 2 args")
+		}
+		src, ok := vs[0].(vm.String)
+		if !ok {
+			return vm.NIL, fmt.Errorf("os/unzip expected String path")
+		}
+		dest, ok := vs[1].(vm.String)
+		if !ok {
+			return vm.NIL, fmt.Errorf("os/unzip expected String destination")
+		}
+		if err := unzipTo(string(src), string(dest)); err != nil {
+			return vm.NIL, err
+		}
+		return vs[1], nil
+	}))
+
 	// os/os-name — (os/os-name) → "linux", "darwin", "windows", ...
 	ns.Def("os-name", mustWrap(func(vs []vm.Value) (vm.Value, error) {
 		return vm.String(runtime.GOOS), nil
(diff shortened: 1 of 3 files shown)


mparrett (approved):
Approving. Built it and ran the suite: 22 unzip subtests pass, and it builds for `js/wasm`, `plan9/amd64`, `linux/amd64`, and `wasip1/wasm`. The `windows/amd64` failure is pre-existing in `term.go` and reproduces on the merge base.

Routing every write through an `os.Root` confined to the destination closes the check-then-write race that a lexical containment guard cannot, and the tests cover zip-slip, absolute entry names, and escape through a pre-existing symlink. Stripping setuid via `.Perm()` and separating a recorded mode from `archive/zip`'s synthesized `0666` are both easy to miss.

No findings.

Not a change request: `io.Copy` is unbounded, so a crafted archive can expand without limit. That matches `io/slurp` and `http/get`, and there is no in-tree caller yet, so it isn't a defect here. It becomes worth revisiting when something starts unzipping artifacts fetched over the network, where an optional caller-supplied byte limit — the shape `net/read!` already uses — would fit better than a hidden default.

mparrett:
Now these have landed: between them they cover half of what let-go needs for Grenadine's host contract.

Checked all 17 slots against `lg` on current main. Most already map onto `os`, `io`, `http`, and `edn`: `:http-get`, `:read-edn`, `:getenv`, `:home-dir`, `:run-process`, `:mkdirs`, `:exists`, `:find-files`, `:read-bytes` / `:write-bytes`, plus the `os/stat` fields behind `:directory` and `:regular-file`. With `hash/sha1` and `os/unzip` merged, that leaves:

- `:atomic-move` — no rename in `os` or `clojure.core`
- `:delete-tree` — `delete-file` only, so no recursive form
- `:canonical-path` / `:absolute-path` — no `path` or `filepath` namespace
- `:bytes->utf8` / `:byte-count` — depends on how bytes get represented, so more a design question than a missing function

Are you working on any of these? If not I'd pick up the first two, which look like small additions in the same shape as `os/unzip`. Asking first since #690 and #688 landing together read like a batch, and you may have these queued already.

The complete contract is `src/grenadine/host/jvm.clj` in `clojurestar/grenadine`, with `glojure.clj` beside it as the non-JVM implementation. let-go already passes Grenadine's portable resolver suite (52 tests, 180 assertions); the effect host is what's missing.

abogoyavlensky:
@mparrett No, I'm not working on any of these, please, feel free to pick it up.

PR #690, opened 2026-08-05
feat(rt): add sha1/sha256 to hash ns

### LINKED ISSUE #689: Add sha1/sha256 to the hash namespace
### What's missing

The `hash` namespace exposes xxh3 (non-cryptographic) only. There is no way to compute a SHA digest, so verifying a downloaded artifact means shelling out to `sha1sum` / `sha256sum` — or to `shasum` on macOS, where those two do not exist.

### Why it matters

Maven and most package registries publish `.sha1` / `.sha256` checksums next to each artifact, so checksum verification is table stakes for dependency tooling written in let-go. Depending on a host binary whose name differs per platform makes that awkward and environment-dependent.

### Proposed shape

`(hash/sha1 s)` and `(hash/sha256 s)`, returning the lowercase hex digest of the string's bytes — the same hex `sha1sum` / `sha256sum` print.

let-go strings are raw Go byte strings, so these hash binary content correctly: `(hash/sha256 (slurp "some.jar"))` is exactly the use case.

`crypto/sha1` and `crypto/sha256` are pure Go, so unlike xxh3 (which needs the murmur3 fallback on TinyGo) these can be registered on every build with no build tags. A Go-side installer can seed the `hash` namespace and `pkg/rt/core/hash.lg`'s `(ns hash)` then extends it rather than replacing it — I verified that ordering holds.

abogoyavlensky:
Resolves: #689 

`(hash/sha1 s)` and `(hash/sha256 s)` return the lowercase hex digest of the string's bytes — the same hex `sha1sum` / `sha256sum` print. let-go strings are raw Go byte strings, so binary content (a jar slurped from disk) hashes correctly.

Motivation: dependency tooling built on let-go currently shells out to `sha1sum`/`shasum` to verify Maven artifact checksums.

`pkg/rt/hash_sha.go` carries no build tags — `crypto/sha1` and `crypto/sha256` are pure Go, so TinyGo/wasm builds get these too (unlike xxh3, which needs the murmur3 fallback). The installer seeds the `hash` namespace from Go and `pkg/rt/core/hash.lg`'s `(ns hash)` extends it.

Tests in `test/hash_sha_test.lg`: empty and `"abc"` vectors for both, a multibyte literal pinned to real `sha1sum`/`sha256sum` output (proving byte- rather than rune-semantics), a spit/slurp round-trip, and arity/type rejection.


mparrett (approved):
Approving. Built it and confirmed that `(ns hash)` in `hash.lg` extends the Go-seeded namespace rather than replacing it: `hash/sha1` and `hash/xxh3-64` both resolve in the same build. Full suite green, and `TestGeneratedArtifactsAreFresh` passes with the regenerated `.lgb` and `generated.sums`. The no-build-tags claim holds — it builds for `js/wasm`, `plan9/amd64`, and `linux/amd64`.

No findings.

nnunley:
Heads up: this PR now conflicts with `main` after today's merges (#564, #647, #648, #650, #666, #688).

The conflict is entirely in generated artifacts — `pkg/rt/core_compiled.lgb` and `pkg/rt/generated.sums`. Your source changes (`pkg/rt/core/hash.lg`, `pkg/rt/hash_sha.go`, `test/hash_sha_test.lg`) don't conflict with anything.

Those two files can't be merged textually; they need regenerating on the new base:

```
git fetch upstream && git rebase upstream/main   # take either side for .lgb/.sums
make generate                                     # regenerates both from source
make check-generated                              # verifies lockstep
```

CI was green on your previous head, so I expect this to be mechanical. I have `maintainerCanModify` and could push the rebase for you, but I'd rather not rewrite your branch without asking — say the word if you'd prefer I do it.

nnunley:
I went ahead and pushed the rebase (`maintainerCanModify` is on) — say the word if you'd rather I hadn't and I'll leave the next one to you.

`87b3afa` → `e9a3e57`, rebased onto `main@5ac4d62`. Your commit and authorship are unchanged; the only thing I touched was the conflict:

- `pkg/rt/generated.sums` conflicted, so I regenerated it with `make generate` rather than merging it textually. `pkg/rt/core_compiled.lgb` was regenerated in the same pass.
- Your source diff (`hash.lg`, `hash_sha.go`, `hash_sha_test.lg`) is byte-identical to what you pushed — verified by diffing the old and new patches.
- `make check-generated` passes, and the repo's pre-push hooks (go test, Clojure compat suite, ir-stress ratchet) pass on the exact rebased commit.

CI is running now; I'll merge once it's green.

PR #699, opened 2026-08-08
feat(rt): add os/absolute-path and os/canonical-path

mparrett:
Stacked on #698; review that first. The delta here is the two path functions.

let-go has no path namespace, so nothing resolves a relative path and nothing collapses two names for one file into a single string. Both are slots in Grenadine's host contract (#688).

## Why two functions and not one normalize

They answer different questions, and merging them would force every caller to accept the stricter one.

**`os/absolute-path` is lexical with respect to its argument.** It resolves against the working directory and cleans traversal without looking the path up, so it answers for a path that does not exist yet. That is what a caller naming a file it is about to create needs, and a symlink in the path stays a symlink. It does read the process cwd, so it is not free of filesystem access and can fail when the working directory has been removed.

**`os/canonical-path` also resolves symlinks.** Following a link can only be done by looking, so this one requires the path to exist. In exchange it gives a string that compares and keys correctly: two names for one file produce one answer, which is the property that makes it the right form for a cache key or an identity check.

A missing path is an error from `canonical-path` rather than a cleaned-up guess, since there is no honest answer. Both refuse an empty path, matching `os/delete-tree` on the parent branch: `filepath.Abs("")` answers with the cwd, so an unset variable would produce a plausible path instead of an error.

## Resolution order

`canonical-path` resolves symlinks *before* absolutizing, and the order is load-bearing. `filepath.Abs` cleans `..` lexically, so absolutizing first collapses `..` against the link's parent rather than the target's. With `a/link` → `b/c`, the input `a/link/../target.txt` becomes `a/target.txt` and errors, where the answer is `b/target.txt`. The first draft of this PR had it the wrong way round and reported `ENOENT` for a path the kernel opens fine; there is a regression test, and it builds its path by concatenation because `filepath.Join` would clean the case away before the native saw it.

## Placement

Both go in `os` beside `rename` and `delete-tree` rather than opening a `path` namespace for two functions. If the surface grows to `join`, `basename`, and `dirname`, a `path` namespace becomes the better home and these should move with it. Happy to start that namespace here instead if you'd rather set the shape now.

## Verification

- `pkg/rt/os_paths_test.go` — 12 tests: resolution against cwd, traversal cleaning, an already-absolute path, symlink resolution, two names collapsing to one string, the missing-path divergence between the two, argument guards, the `..`-after-symlink regression, and that `absolute-path` leaves a symlink unresolved where `canonical-path` resolves it. The symlink cases skip where the platform refuses symlinks, and compare against an `EvalSymlinks` of the target rather than the raw path, since `t.TempDir` itself sits under a symlink on darwin.
- `test/os_paths_test.lg` — the same surface from lg, including the lexical-vs-filesystem distinction.
- Builds clean for `linux/amd64`, `js/wasm`, `wasip1/wasm`, and `plan9/amd64`.

One caveat on browser `js/wasm`: Go's `wasm_exec.js` shim throws `ENOSYS` from `process.cwd()` and `fs.lstat()`, so `absolute-path` fails on relative input there and `canonical-path` fails on any input. That matches `os/cwd`, `os/ls` and `os/stat`, which are already dead on that target, so it is not a regression — but unlike them these two carry doc comments making unconditional promises, and I can add a caveat if you'd rather they said so. `wasip1` and `plan9` are both fine: wasip1 tracks a real cwd and has working `Lstat`, and plan9 implements `evalSymlinks` as `Lstat` plus `Clean`.

## Still open from the contract survey

`:bytes->utf8` and `:byte-count` are left out because they need a decision rather than a function. Bytes are `vm.String` today (`read-bytes` returns one), so `bytes->utf8` is identity and the real question is whether an invalid-UTF-8 check belongs on it. `byte-count` is a genuine gap — `count` is rune-based, so `(count "héllo")` is 5 where the byte length is 6, and lg has no way to ask for the latter. Where it belongs is the open part: `count`'s neighbours are in core, which there is active work to shrink.

### DIFF
--- a/pkg/rt/generated.manifest
+++ b/pkg/rt/generated.manifest
@@ -145,7 +145,7 @@ pkg/rt/core_compiled.lgb pkg/rt/native_prims.go generator eb0246a1c8565285c8b3fb
 pkg/rt/core_compiled.lgb pkg/rt/native_prims_lifecycle.go generator 62e02df9b4ca29ed1c65a82d500f83b6a89eb2a69c741f95bbc35693a85f60e6
 pkg/rt/core_compiled.lgb pkg/rt/net.go generator 6b117f6b365173941f01996d3813113bf58604c975a2f8ec2776ca538f987452
 pkg/rt/core_compiled.lgb pkg/rt/net_wasm.go generator e7780099c634a41c9a4a26792bb3a86710a2148431af8c6853e5d034682bbe42
-pkg/rt/core_compiled.lgb pkg/rt/os.go generator 720d5964ae9cf671770d56b7a62050fdc1e2115fba419458333093f3cdff31d7
+pkg/rt/core_compiled.lgb pkg/rt/os.go generator 444077d3727b8a18f5c78672883953d090c4b56d381064158e27aae370dec71d
 pkg/rt/core_compiled.lgb pkg/rt/os_tinygo.go generator 9edb8783e742feb2645643e9a07114caf0f5c11ca775f7c1cb994a8983e7ede9
 pkg/rt/core_compiled.lgb pkg/rt/parallel.go generator 520cd217f93da579ee05d3419d980505baa8d45a5e068fc1e54f61f8a90744b8
 pkg/rt/core_compiled.lgb pkg/rt/pods.go generator 2945d0fe60eedba514418518782746bbbe84a1859b3c7baf13d3ef46062d38b3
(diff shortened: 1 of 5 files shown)


nnunley:
The split is right, and the reasoning for not merging them into one `normalize` is the part
worth keeping in the docstrings: one answers for a path that does not exist yet, the other
requires existence because a symlink can only be resolved by looking. Erroring on a missing
path in `canonical-path` rather than returning a cleaned-up guess is the honest choice.

mparrett:
Follow-up on this os series: please update the README

nnunley (approved):

PR #708, opened 2026-08-09
ci(perf-repeat): post the shadow-check report as a sticky PR comment

mparrett:
`perf-pr-repeat.yml` writes a job summary and uploads its evidence, but has no comment step, so a 65-minute run leaves nothing on the PR. On #706 I had to download the run artifact and read `aggregate.json` to find out what the lane had concluded. The shadow phase is supposed to be collecting false-positive and provenance data, and output nobody sees does not get compared.

This adds the comment job, mirroring the one already in `perf-pr.yml` rather than inventing a second pattern: separate job so the elevated token is scoped away from the one that checks out and builds PR code, sticky upsert by marker so repeated pushes update in place, `continue-on-error` because fork PRs get a read-only token and that should not redden the run.

The comment carries the aggregate — provenance header, family table, would-gate verdict — and points at the job summary for the per-cycle logs. The first self-test posted the whole report instead: 120 lines with seven literal `::group::` markers, which are Actions log-folding directives that do not render in a comment. Trimming to the text after the last `::endgroup::` takes it to 36 lines.

## Two choices worth surfacing

**The marker is `perf-pr-repeat-report`, deliberately not `perf-pr-report`.** Both lanes run on the same PR during the shadow phase, and the header of this workflow says comparing them is the point, so neither may overwrite the other.

**The comment is gated on bench success.** The bench job exits nonzero when families are not comparable across cycles. Posting a delta table built from a measurement already known to be broken seems worse than posting nothing, and the job summary still carries the report in that case. Easy to relax if you would rather see the integrity failure on the PR too.

## Verification

`actionlint` clean, including its shellcheck pass — the file was clean before and still is. The change is inert unless the `perf-repeat` label is applied, and since `.github/workflows/perf-pr-repeat.yml` is inside the workflow's own `paths` filter and same-repo PRs run it from the PR branch, labeling this PR exercises the new job on itself.

### DIFF
--- a/.github/workflows/perf-pr-repeat.yml
+++ b/.github/workflows/perf-pr-repeat.yml
@@ -54,6 +54,8 @@ jobs:
     # for a noisier runner, where the memory-bound families (and so the whole
     # suite) run slower. Right-size from observed runtime during the shadow phase.
     timeout-minutes: 90
+    outputs:
+      pr: ${{ steps.refs.outputs.pr }}
     steps:
       - uses: actions/checkout@v4
         with:
@@ -86,8 +88,11 @@ jobs:
           git fetch --no-tags origin "$base_ref" "refs/pull/${PR}/head:pr-head"
           head_sha="$(git rev-parse pr-head)"
           base_sha="$(git merge-base "origin/${base_ref}" pr-head)"
-          echo "base=${base_sha}" >> "$GITHUB_OUTPUT"
-          echo "head=${head_sha}" >> "$GITHUB_OUTPUT"
+          {
+            echo "pr=${PR}"
+            echo "base=${base_sha}"
+            echo "head=${head_sha}"
+          } >> "$GITHUB_OUTPUT"
           echo "Base (merge-base): ${base_sha}"
           echo "Head:              ${head_sha}"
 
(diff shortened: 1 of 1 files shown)


nnunley:
It self-tested, which is the nicest kind of evidence: the sticky comment is already rendered on the PR with the aggregate table, the would-gate verdict, the artifact link, and no leaked `::group::` noise.

Shell holds under `set -euo pipefail` — the `grep -n … | tail -1 | cut … || true` guard covers grep's exit 1, the no-groups path falls back to the whole report, and `rc` is captured and re-exited after `comment.md` is written, so an integrity failure stays red while the summary and artifact survive. The distinct marker (`perf-pr-repeat-report` vs `perf-pr-report`) means the two shadow lanes cannot overwrite each other, and keeping the elevated token in a separate job mirrors `perf-pr.yml:173-210`.

**One nit that is worth fixing in both lanes rather than just this one.** `github.rest.issues.listComments({owner, repo, issue_number})` is unpaginated — 30 per page by default — so on a PR past 30 comments the marker is not found and the "sticky" comment silently duplicates on every run. I checked: `perf-pr.yml:207` has the same call, so this is inherited rather than introduced. `github.paginate(github.rest.issues.listComments, …)` fixes both. This repo's PRs do get long enough to hit it.

Two smaller ones, neither blocking: the report text originates from PR-branch code and is interpolated inside a ``` fence, so a head branch could break out of the fence — same exposure as the existing lane, and the reason the token is in a separate job, but the artifact content is still head-controlled. And a stale sticky comment persists if the `perf-repeat` label is later removed; no cleanup path.

Worth noting what this unblocks: #706's author had to hand-post benchmark results because this step did not exist yet.

nooga (approved):
Reviewed. Mirrors perf-pr.yml's comment job correctly (separate job, scoped token, continue-on-error for forks, sticky upsert by marker) and self-tested cleanly on this PR.

Two choices both look right: the distinct marker is necessary since both shadow lanes run on the same PR by design, and gating the comment on bench success is the right call — a red bench already surfaces in checks/job-summary, so nothing's hidden, and it avoids posting a delta table built from data the workflow itself flagged as non-comparable.

nnunley's pagination catch is real and correctly non-blocking here since #715 fixes it (and the same bug in perf-pr.yml) right after. All checks green.

PR #710, opened 2026-08-10
fix(vm,repl): stop tab completion crashing on the cyclic refer graph

mparrett:
Pressing Tab in the REPL kills the process:

```
lg 1.12.2 (9c9a3d6)
user=> de<Tab>
runtime: goroutine stack exceeds 1000000000-byte limit
fatal error: stack overflow
```

`FuzzySymbolLookup` walks refers recursively with no visited set, and the refer graph is cyclic. `clojure.core` requires `let-go.types` for `array?`/`bigint?`, and `RegisterNS` auto-refers `clojure.core` back into it, so the walk ping-pongs between those two namespaces until the stack is exhausted:

```
core         -> [let-go.types]
let-go.core  -> [core let-go.types]
let-go.types -> [core let-go.core]
user         -> [core let-go.core let-go.types]
```

Both edges are intended, so the walk has to tolerate them. The recursion is old, but it was harmless until the cycle existed — `v1.11.1` completes fine, and the first tag carrying the `clojure.core` -> `let-go.types` edge from #410 is `v1.12.0`. nREPL completion calls the same function, so editors on `lg` 1.12.x hit it too.

Two smaller bugs sit on the same path:

- A prefix naming an unregistered namespace (`nosuchns/x<Tab>`) passes a nil `*Namespace` into the walk and segfaults. The visited-set rewrite skips nil targets.
- readline *inserts* a candidate at the cursor, so a candidate has to be the part not typed yet. The completer returned whole symbols, so `ns-unm<Tab>` produced `ns-unmns-unmap`. Candidates come back unqualified, because the lookup resolves the namespace segment of the prefix and matches on the name; the segment to strip is therefore the name, which is also what readline echoes ahead of each candidate when it lists them. This one predates 1.12 and makes completion useless rather than fatal, which is how it stayed hidden behind the crash.

Still not handled, and out of scope here: `FuzzyNamespacedSymbolLookup` resolves a qualified prefix against the namespace registry directly, so an alias (`str/rep<Tab>` after `[clojure.string :as str]`) still yields nothing.

## Verification

New tests, each confirmed to fail against the unpatched code:

- `pkg/vm`: a two-namespace refer cycle terminates and yields each symbol once (stack overflow before).
- `pkg/complete`: `Do` returns the untyped remainder and never repeats the typed prefix; an unknown-namespace prefix returns no candidates instead of crashing. The completer moved out of package `main` to get there — test files are not allowed at the repo root, and it depends only on `rt` and `vm` (readline's `AutoCompleter` is satisfied structurally, so `pkg/complete` does not import readline).

Driven manually in a pty at 1.12.2 and at this branch: `de<Tab>`, `defre<Tab>` -> `defrecord`, `ns-unm<Tab>` -> `ns-unmap`, `nosuchns/x<Tab>` -> no candidates, no crash.

### DIFF
--- a/lg_repl.go
+++ b/lg_repl.go
@@ -48,10 +48,19 @@ func (c *completer) Do(line []rune, pos int) (newLine [][]rune, length int) {
 	prefix := string(head[start:])
 
 	symbols := rt.FuzzyNamespacedSymbolLookup(c.ctx.CurrentNS(), vm.Symbol(prefix))
+
+	// readline INSERTS the candidate at the cursor, so a candidate has to be the
+	// part not typed yet — returning whole symbols turns "ns-unm<tab>" into
+	// "ns-unmns-unmap". Candidates come back unqualified (the lookup resolves the
+	// ns segment of the prefix and matches on the name), so the typed part to
+	// strip is the name segment, not the whole word. length is the same segment's
+	// rune count: readline echoes that many runes ahead of each candidate when it
+	// lists them.
+	_, namePrefix, _ := vm.Symbol(prefix).NamespacedRaw()
 	for _, s := range symbols {
-		newLine = append(newLine, []rune(string(s)+" "))
+		newLine = append(newLine, []rune(strings.TrimPrefix(string(s), string(namePrefix))+" "))
 	}
-	length = pos - start
+	length = len([]rune(string(namePrefix)))
 	return
 }
 
--- a/lg_repl.go
+++ b/lg_repl.go
@@ -12,57 +12,12 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
-	"unicode"
 
 	"github.com/chzyer/readline"
 	"github.com/nooga/let-go/pkg/compiler"
-	"github.com/nooga/let-go/pkg/rt"
+	"github.com/nooga/let-go/pkg/complete"
 	"github.com/nooga/let-go/pkg/vm"
 )
-
-func isCompletionTerminator(r rune) bool {
-	switch r {
-	case '(', ')', '[', ']', '{', '}', '"', '\\', '\'', '@', '`', '~', ';', '#':
-		return true
-	}
-	return unicode.IsSpace(r)
-}
-
-// completer implements readline's AutoCompleter interface.
-// readline passes line as []rune and pos as a rune index; we stay in runes
-// throughout so non-ASCII input and mid-line cursors are handled correctly.
-type completer struct {
-	ctx *compiler.Context
-}
-
-func (c *completer) Do(line []rune, pos int) (newLine [][]rune, length int) {
-	if pos > len(line) {
-		pos = len(line)
-	}
-	head := line[:pos]
-
-	start := pos
-	for start > 0 && !isCompletionTerminator(head[start-1]) {
-		start--
-	}
-	prefix := string(head[start:])
-
-	symbols := rt.FuzzyNamespacedSymbolLookup(c.ctx.CurrentNS(), vm.Symbol(prefix))
-
-	// readline INSERTS the candidate at the cursor, so a candidate has to be the
-	// part not typed yet — returning whole symbols turns "ns-unm<tab>" into
-	// "ns-unmns-unmap". Candidates come back unqualified (the lookup resolves the
-	// ns segment of the prefix and matches on the name), so the typed part to
-	// strip is the name segment, not the whole word. length is the same segment's
-	// rune count: readline echoes that many runes ahead of each candidate when it
-	// lists them.
-	_, namePrefix, _ := vm.Symbol(prefix).NamespacedRaw()
-	for _, s := range symbols {
-		newLine = append(newLine, []rune(strings.TrimPrefix(string(s), string(namePrefix))+" "))
-	}
-	length = len([]rune(string(namePrefix)))
-	return
-}
 
 // ANSI color codes for syntax highlighting. ansiReset and ansiBold are
 // declared in lg_ansi.go; the rest are repl-only.
@@ -135,7 +90,7 @@ func repl(ctx *compiler.Context) {
 	config := &readline.Config{
 		Prompt:          prompt,
 		EOFPrompt:       "exit",
-		AutoComplete:    &completer{ctx: ctx},
+		AutoComplete:    &complete.Completer{Ctx: ctx},
 		InterruptPrompt: "^C",
 		HistoryFile:     historyFile(),
 		Painter:         &syntaxHighlighter{},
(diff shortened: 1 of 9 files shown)


mparrett:
**Self-review: No findings.**

Visited-set rewrite on `FuzzySymbolLookup` is the right fix for the `clojure.core` ↔ `let-go.types` refer cycle; nil targets are skipped; completer remainder/`length` match [readline’s AutoCompleter contract](https://github.com/chzyer/readline/blob/master/complete.go) (`Do("gi", 2) => ["t", "t-shell"], 2`); nREPL correctly still gets full symbol names. New tests fail closed on the old behavior; CI is green on `56ee65e2`.

**Verdict: correct** — existing callers keep working, Tab no longer stack-overflows, and the insert-vs-whole-symbol bug is fixed. Residual: alias-qualified completion (`str/rep`) still empty, as the PR calls out; no test for a qualified prefix remainder (`clojure.core/ns-unm` → `ap `), but that’s the same `TrimPrefix` path as the unqualified case.

nooga:
## Code review (Claude)

This PR fixes tab-completion crashing on the cyclic refer graph, but the fix surfaces a couple of pre-existing correctness gaps that were previously unreachable (the crash masked them), plus a few smaller nits.

### Correctness

1. **`pkg/vm/namespace.go:659`** — `fuzzySymbolLookup` recurses into a referred namespace's full registry, ignoring the `Refer`'s `only` set. A restricted `(:require [lib :refer [foo]])` still surfaces every other public symbol in `lib` as a completion candidate. Verified: `FuzzySymbolLookup(user, "f", true)` returns `[fred flip foo]` instead of `[foo]` when `lib` also defines public `fred`/`flip`. Picking one of the extra candidates yields an unresolvable symbol. This also contradicts the new doc comment claiming the function "collects every symbol visible from ns."

2. **`pkg/vm/namespace.go:662`** — same function never checks the calling namespace's `unmapped` tombstone set before recursing into a referred namespace, so a symbol hidden via `ns-unmap` still appears as a completion candidate. `lookupViaRefers`/`ReferredVars` both honor `unmapped`; `fuzzySymbolLookup` doesn't. Verified: `Refer` then `Unmap("foo")`, and `foo` still shows up in fuzzy results.

3. **`pkg/complete/complete.go:19`** — `isTerminator()` hand-rolls symbol-boundary characters instead of reusing the reader's actual rules, and has already drifted: `pkg/compiler/reader.go`'s `isWhitespace` treats `,` as whitespace; `isTerminator` does not. `(foo,bar<TAB>` will scan back past the comma and use `foo,bar` as the prefix instead of stopping at the comma, silently producing no completions.

### Efficiency / reuse (minor)

4. **`pkg/vm/namespace.go:524`** — `registrySnapshot()` unconditionally deep-copies a namespace's entire registry before `fuzzySymbolLookup` filters by prefix, and this now runs on every reachable namespace on every completion request. Filtering under the RLock (appending only matches) would make it O(matches) instead of O(registry size). Previously the cyclic crash meant this cost was never paid in production — this PR is what makes it real.

5. **`pkg/complete/complete.go:46`** — the namespace/name split for prefix trimming is derived independently via `vm.Symbol(prefix).NamespacedRaw()`, duplicating the identical split `rt.FuzzyNamespacedSymbolLookup` already does internally. If `NamespacedRaw`'s splitting rule ever changes, the two derivations can silently disagree.

6. **`pkg/complete/complete.go:58`** — `len([]rune(string(namePrefix)))` is a redundant `Symbol -> string -> []rune` conversion; `vm.Symbol`'s underlying type is `string`, so `[]rune(namePrefix)` converts directly. No functional effect, just noise on a per-keystroke path.

Nothing here blocks the fix itself (the crash-on-cycle bug is correctly fixed), but 1–3 are real completion-correctness bugs worth a follow-up.

nnunley:
Real crash fix, and the root cause checks out in-tree: `pkg/vm/namespace.go:644-647` recursed over `refersSnapshot()` with no visited set, and the refer graph genuinely is cyclic. The fix is minimal and right — one shared `seen` map gives set-union semantics, the root is visited first so `lookupPrivate=true` is not lost, and each symbol comes back once per defining namespace.

**The second fix is understated in the body.** `rt.FuzzyNamespacedSymbolLookup` (`pkg/rt/lang.go:482-493`) hands a nil `*Namespace` to `FuzzySymbolLookup` for an unregistered ns prefix — a nil-deref panic — and `pkg/nrepl/server.go:348` calls the same function. So this also fixes an **nREPL server crash**, not only the REPL one. Worth saying, since it widens who benefits.

**One finding: the new comment asserts an invariant the function does not hold.** It says `FuzzySymbolLookup` "collects every symbol visible from ns". It does not:

- `Refer.only` is ignored — the walk recurses into the whole target registry regardless of `ref.all`/`ref.only`, where `isShadowingCoreRefer` (`namespace.go:213-225`) does check.
- the `unmapped` tombstone set that `lookupViaRefers`/`ReferredVars` honor is ignored.
- refers are followed transitively, though refer visibility is not transitive.

All three are **pre-existing** — the old recursion had the same shape — but this PR is what makes them reachable, since the crash was masking them. I would soften the comment or file the follow-up rather than let it state a guarantee the code does not give. This overlaps items 1-2 of nooga's review.

Smaller, none blocking:

- `fuzzySymbolLookup` nil-checks the target ns but not the `*Refer` itself; `r.ns` on a nil `r` panics, and sibling code at `namespace.go:215` does defend against that. Cheap symmetry.
- `registrySnapshot()` (`namespace.go:522`) deep-copies each reachable namespace's whole registry per keystroke — pre-existing cost, but newly *paid*, because the walk now completes. Filtering under the RLock would make it O(matches).
- `isTerminator` (`pkg/complete/complete.go:19`) has drifted from the reader: `pkg/compiler/reader.go` treats `,` as whitespace and `isTerminator` does not, so `(foo,bar<TAB>` scans past the comma. Pre-existing, carried over verbatim by the move.
- Packaging scope the body does not mention: `pkg/complete` has no build tag while `lg_repl.go` is `//go:build !plan9 && !js && !wasip1`, so the package now compiles on all targets. It imports only compiler/rt/vm/strings/unicode and the cross-target builds are green, so this looks fine — flagging it as a scope change rather than a problem.

Tests are good: the cycle test asserts termination *and* single-occurrence, and the completer test pins the remainder contract and the unknown-ns non-crash. I did not drive a real pty, so the qualified-prefix `length` semantics are a code-reading conclusion plus your transcript, not something I reproduced.

mparrett:
@nooga @nnunley Thanks for the reviews. Solid findings worth doing. Happy to take those as a follow-up.

mparrett:
@nooga @nnunley Thanks again for the reviews. Follow-ups from your notes are on the tip (`:only` / `unmapped` / comma boundary / registry filter); the one-level refer walk is stacked as draft #713. Ready for a final look whenever you have a moment.

nooga (approved):

nnunley (approved):

Issue #665, opened 2026-07-31
IR API: zero-alloc inst iteration for .lg pipeline walks (recover the accepted contribution-model alloc cost)


nnunley:
## Problem

Walking a function's instructions from `.lg` IR-pipeline code forces allocations that a hot loop cannot avoid: `ir/block-insts` hands back a (possibly lazy) sequence, and predicate-style walks (`some` + closure) allocate a closure per block. The `:try` contribution-model refactor (op-catalog workflow, #268) measured the cost directly: consolidating the scattered `function-needs-error?` / `function-needs-rt?` conditionals into per-op contribution predicates costs **+20.9% (bytecode) / +11.6% (gogen_ir) allocs/op on BenchmarkIRCompile**, with no zero-alloc expression available at the call sites. Alternatives tested: loop/recur variant was *worse* (+69%); the closure form is the local minimum under the current API.

Captain decision (2026-07-31): the regression is **accepted conditionally** — the consolidation lands now, and this improved IR API is the follow-up that recovers the cost. Until it exists, the workflow's performance-metrics requirement (bench-ratchet at par) remains the standing gate for further IR-pipeline work.

## What this issue asks for

An inst-iteration surface the IR pipeline can use without per-block/per-call allocation:

- Eager inst access (indexed vector or preallocated array view) or an iterator that does not realize lazy seqs per block.
- A walk/fold primitive that takes a named top-level fn (no closure capture) or is specialized at lowering time.
- Longer-term alignment with the ops-as-types direction: contribution facets as native catalog properties with direct dispatch, so whole-function property folds are table lookups, not interpreted predicate calls.

## Acceptance

BenchmarkIRCompile allocs/op for the contribution-model code path returns to (or below) the pre-refactor baseline, with the consolidation retained.

Related: #268 (op catalog — where the cost was taken), #464 (runtime performance epic), #663 (performance measurement status).

nnunley:
[...]

PR #711, opened 2026-08-10
feat(data): port clojure.data's protocols, and diff IR functions structurally

nnunley:
`data.lg` carried clojure.data's public `diff` but not the two protocols it is built on, so the signature matched Clojure while the contract underneath did not — nothing outside `map?`/`set?`/`sequential?` could take part. An `ir` Function, being an atom, fell through to `atom-diff` and reported two functions as *wholly* different rather than structurally so.

## What changed

`EqualityPartition` and `Diff` are now defined and `diff` dispatches through them, matching upstream's shape: compare partitions, `diff-similar` when they agree, `atom-diff` when they do not.

One deliberate divergence, documented in the code: upstream reaches the collection families through Java interfaces, but let-go's families are defined by predicates with no shared supertype, so the `Object` default consults `map?`/`set?`/`sequential?` before falling back to `:atom`. A new built-in collection therefore classifies correctly without enumerating its concrete type, while any type that extends `EqualityPartition` still takes precedence.

**Behaviour is unchanged.** Every pre-existing case returns exactly what it did before — I ran all eleven against a build of the old file first, then against the new one.

## The extension point, which is the point

```clojure
(extend-type DiffBox
  data/EqualityPartition (equality-partition [_] :map)
  data/Diff              (diff-similar [a b] (data/diff (.-m a) (.-m b))))

(data/diff (DiffBox. {:a 1 :b 2}) (DiffBox. {:a 1 :b 3}))
;; before: [whole-a whole-b nil]      after: [{:b 2} {:b 3} {:a 1}]
```

## ir.diff

`ir.diff` projects a Function into comparable data and diffs it, so a difference reports as data rather than as a failed substring match over `ir.dump` text:

```
;; two fns differing only in + vs -
only in a: {:blocks [{:insts [nil nil {:op :add}]}]}
only in b: {:blocks [{:insts [nil nil {:op :sub}]}]}
```

versus the 43 `strings.Contains(dump, "Add v0 v1 : int")`-style assertions in `pkg/ir` today.

Three things it deliberately normalizes, each because including them reports differences that are not differences: the `:uses-cache`/`:uses-dirty?` memo (two identical fns otherwise disagree on whether anything has asked for uses yet — pinned by a test), the `:types` transient (read through `ir/type-of` instead), and the flat `:insts`/`:source-info` side tables (blocks already name the nids they own).

The projection is explicit rather than a `Diff` impl because a Function is a bare `let-go.lang.Atom` — the same type as every other atom — so extending that type would hijack diffing for all of them. Giving Functions their own deftype would fix that, and is tracked separately against #665/EPIC-017.

## Verification

`test/data_diff_test.lg` 9 tests / 26 assertions, `test/ir_diff_test.lg` 6 / 22, both green. Full `.lg` suite, `make generate` and `make check-generated` green. One pre-existing divergence from Clojure is pinned in the test with a comment rather than silently changed: `(diff [1 2] [1 2 3])` reports `both` as `[1 2 nil]` here versus Clojure's `[1 2]`, because upstream rebuilds each side from an index map while let-go builds fixed-length vectors.


mparrett (approved):
No findings.

Checked the protocol rewrite against the old predicate `cond` (nil, strings vs sequentials, cross-partition), the DiffBox extension path, and `fn->data` normalization (uses-cache, types via `type-of`, const values in `:aux`). Behavior matches what was already there; the Object-default divergence is documented and matches let-go's predicate-shaped collections. CI green on `53653b22`.

mparrett:
@nnunley Agreed on the literal; same shape as the TinyGo pin, and I’ll leave both as-is at this size. If you’re good with the rest, could you drop an approval on this one? Thanks!

PR #712, opened 2026-08-10
refactor(ir): reach the fn shape through accessors, not past the data layer

nnunley:
`ir.data` owns the Function representation and publishes a nid/bid-shaped accessor surface, but thirteen sites in `validate`, `lattice` and three passes went around it into `(:insts @f)` and `(:blocks @f)` — reaching for the flat tables because no accessor named them, not because the accessors were wrong.

## What changed

Adds the three that were missing — `all-insts`, `raw-blocks`, `all-types` — documented as representation-shaped and therefore a last resort, since every other accessor here is phrased per-nid or per-bid. Then moves every out-of-layer deref onto them or onto an accessor that already existed:

- `(count (:insts @f))` → `ir/inst-count`
- `lattice`'s hand-rolled types-table read → `ir/all-types`, which is what `ir/type-of` already reads underneath

**No `@f` remains outside the data layer.**

## Why now

This is groundwork for the representation work tracked against #665 / EPIC-017. A representation change is only safe where callers go through accessors, and these thirteen sites were the ones that would have broken in place. EPIC-017 ranks packed inst records as the top measured cost, so the seam has to exist before that layout can move underneath it. Nothing here changes behaviour — it only narrows how the shape is reached.

## Note for anything that extends this surface

A `defn` in `ir/data.lg` is invisible to callers until it is *also* listed in the intern block at the bottom of the file, which republishes `ir.data` under `ir`. Omitting it does not fail at the call site: it fails at namespace load with `namespace failed to load: ir.passes.typeinfer`, which is a long way from the cause. I confirmed the mechanism by reverting just the intern-list line and watching the same panic return.

Related, and fixed here: `ir.passes.legalize` made 30 `ir/*` calls with no `(:require [ir.data])` at all, so it was resolving on incidental load order. Now declared.

## Verification

`pkg/ir` green, full `.lg` suite green, `make generate` and `make check-generated` green — the last two matter because this is on the bootstrap path (`core_compiled.lgb` and the sums move).


mparrett (approved):
Reviewed [nooga/let-go#712](https://github.com/nooga/let-go/pull/712) (`481965cd` vs merge-base `8a8222d0`).

No findings.

This is a mechanical encapsulation pass: three last-resort accessors (`all-insts`, `raw-blocks`, `all-types`) plus intern-list entries, then every out-of-layer `(:insts/@blocks/@types @f)` in validate/lattice/legalize/licm/mutability moved onto them or onto existing `ir/inst-count`. Grep confirms no remaining Function-atom field peeks outside `ir.data` (the stale `data/generated.lg` twin is still dead/unrequired, as on main). The legalize `(:require [ir.data])` fix is real load-order hygiene, not drive-by.

CI is green on that head (`build`, `generated-artifacts`, `gold-differential`, etc.). No material test gap — the wrappers are identity-preserving, and the suite already covers the call sites.

PR #713, opened 2026-08-11
fix(vm): make FuzzySymbolLookup refer walk one-level

mparrett:
## Summary

Stacked on #710. Most of the review follow-ups already landed on that PR tip (`:refer [only]`, `ns-unmap`, comma token boundary, registry prefix filter, ns-split dedupe). This layer takes the remaining visibility gap nnunley called out:

- Refer is not transitive — `lookupViaRefers` / `ReferredVars` only read each refer target's locals, but fuzzy still recursed into a `:refer :all` target's own refers and could offer candidates that do not resolve in the completing ns.
- Nil-guard `*Refer` / `r.ns` the same way `isShadowingCoreRefer` already does.
- Doc comment updated to state the one-level rule (and why the cyclic refer graph can no longer make this recurse).

## Test plan

- [x] `go test ./pkg/vm/ -run FuzzySymbolLookup` — cycle still terminates; new `ReferNotTransitive` fails against the #710 tip
- [x] `go test ./pkg/complete/ ./pkg/nrepl/`
- [ ] Tab-complete in a ns that `:refer :all`s a mid ns which itself refers a leaf — leaf symbols should not appear

### DIFF
--- a/pkg/vm/bench_test.go
+++ b/pkg/vm/bench_test.go
@@ -388,6 +388,26 @@ func BenchmarkLazySeqRealize(b *testing.B) {
 	}
 }
 
+// BenchmarkFuzzySymbolLookup_NarrowPrefix simulates interactive completion
+// against a namespace with a clojure.core-sized :refer :all target (300+
+// public vars) and a narrow prefix that only a handful of them match. Guards
+// against FuzzySymbolLookup regressing to copying (and mostly discarding)
+// every referred namespace's entire public registry per call — see
+// referredSymbolsPrefix's doc comment.
+func BenchmarkFuzzySymbolLookup_NarrowPrefix(b *testing.B) {
+	lib := NewNamespace("bench-fuzzy-lib")
+	for i := 0; i < 400; i++ {
+		lib.Def("sym-"+itoa(i), TRUE)
+	}
+	ns := NewNamespace("bench-fuzzy-user")
+	ns.Refer(lib, "", true) // :refer :all
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		FuzzySymbolLookup(ns, Symbol("sym-1"), true)
+	}
+}
+
 func itoa(i int) string {
 	switch {
 	case i < 10:
--- a/pkg/vm/namespace.go
+++ b/pkg/vm/namespace.go
@@ -649,44 +649,37 @@ func (n *Namespace) Unmap(name Symbol) {
 }
 
 // FuzzySymbolLookup collects every symbol visible from ns whose name starts
-// with s. Visibility matches lookupViaRefers / ReferredVars: only the ns's own
-// registry and its *direct* refers contribute — refer is not transitive — and
-// a `:refer [only]` refer contributes just the listed names. Symbols ns has
-// ns-unmap'd are excluded even when a refer would otherwise bring them in.
+// with s. Visibility matches Lookup: the ns's own registry, then the referred
+// set from ReferredVars (baseline auto-refers, explicit :refer / :refer [only],
+// private filtering, and the ns-unmap tombstone). Refer is not transitive —
+// ReferredVars only reads each refer target's locals — which is also why a
+// cyclic refer graph cannot make this recurse.
 //
-// Direct-only is also why a cyclic refer graph (clojure.core requires
-// let-go.types; RegisterNS auto-refers clojure.core back into it) cannot make
-// this recurse: each refer contributes the target's locals, never the
-// target's own refers.
+// Locals are merged on top of ReferredVars so a re-Def after ns-unmap still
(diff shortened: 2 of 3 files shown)


nooga:
## Code review (Claude)

@mparrett Nice follow-up on the transitive-refer gap. Found 4 more issues, all in `FuzzySymbolLookup` (`pkg/vm/namespace.go`):

### Correctness

1. **`pkg/vm/namespace.go:685`** — the final `ns-unmap` filter applies to `ns`'s own registry symbols too, not just refer-contributed ones. So `ns.Def("foo")` → `ns.Unmap("foo")` → `ns.Def("foo")` again still gets filtered out of completion, even though `Lookup` resolves it fine. This directly contradicts `Unmap`'s own doc comment, which promises "a later `(def name ...)` still works fine here regardless." Verified empirically: `Lookup` finds the var, `FuzzySymbolLookup` omits it.

2. **`pkg/vm/namespace.go:671`** — the updated doc comment claims visibility "matches `lookupViaRefers` / `ReferredVars`" but omits the `isBaselineNS` precedence rule both of those apply. `(ns foo (:require [clojure.core :refer [inc]]))` overwrites the `clojure.core` refers-map entry that `RegisterNS` auto-referred as `{all:true}` with `{all:false, only:{inc}}`. `lookupViaRefers` ignores the only-restriction for baseline namespaces, so `Lookup(foo, 'dec')` still resolves — but `FuzzySymbolLookup(foo, "d", true)` omits `dec` since it walks `r.only` literally with no baseline exemption. Verified empirically: completion misses symbols that actually resolve. Pre-existing gap, but the comment now asserts a parity that doesn't hold.

3. **`pkg/vm/namespace.go:684`** — refer-contributed names and `ns`'s own registry names are concatenated with no de-dup, unlike `ReferredVars` which merges into a map keyed by symbol. A locally-`def`'d symbol that shadows a referred one (the exact case `Def`'s warn-on-shadow check exists to flag) shows up twice in the completion candidates.

### Reuse (contributing factor to #2)

4. **`pkg/vm/namespace.go:661`** — this is now the third independent hand-rolled copy of refer-visibility rules in this file (alongside `lookupViaRefers` and `ReferredVars`), which is exactly how #2 happened. Worth considering whether `FuzzySymbolLookup` can share the same visibility helper as one of the other two instead of re-deriving all/only/private/baseline filtering itself, so a future refer-semantics change doesn't need three hand-applied edits again.

mparrett:
@nooga Thanks — all four landed in c303141.

`FuzzySymbolLookup` now composes local `registryPrefixSymbols` over `ReferredVars` instead of walking refers itself. That picks up baseline `:all` precedence, keeps the unmap tombstone off re-Def'd locals, dedups shadowed names, and drops the third hand-rolled visibility copy. Regression tests cover the three correctness cases.

nooga:
## Re-review (Claude)

@mparrett Confirmed — all four correctness/dedup findings from before are fixed in c303141. Differential-tested the `ns-unmap` → re-`def` tombstone case against the pre-#713 code too: the old implementation genuinely returned the wrong (empty) result there, so that's a real bug fix, not just new test coverage. Full suite green, race-stress on concurrent `Def`/`Unmap`/`Refer` vs `FuzzySymbolLookup` clean.

One new finding, an efficiency regression:

**`pkg/vm/namespace.go`** — `FuzzySymbolLookup` now gets its referred set via `ns.ReferredVars()`, which unconditionally copies the full `PublicVars()` of `clojure.core`, every baseline namespace, and every `:refer :all` target into a map *before* filtering by prefix. That replaces the old lock-scoped prefix-filtered scan. Since this runs on every keystroke of interactive tab-completion / nREPL completion, a narrow prefix like `"map-in"` now pays for copying 300+ `clojure.core` vars per call just to throw nearly all of them away.

I pushed a fix for this on top of your branch: `referredSymbolsPrefix`, which mirrors `ReferredVars`'s exact visibility rules (baseline precedence, `:refer [only]`, private filtering, unmap tombstone) but filters by prefix inline via `registryPrefixSymbols` under each source namespace's own read lock, instead of building the full map first. `FuzzySymbolLookup` now composes `registryPrefixSymbols` (locals) + `referredSymbolsPrefix` (refers) the same way it already did for locals.

Added `BenchmarkFuzzySymbolLookup_NarrowPrefix` (400-var `:refer :all` target, narrow prefix) to guard the regression: **~44µs/52KB → ~18µs/27KB per call** (≈2.5x faster, ~48% less memory). Full `go test ./... -count=1` green.

mparrett:
> I pushed a fix for this on top of your branch

@nooga nice performance fix- thanks! I rebased the stack and will do one more sanity check but this should be good to merge when CI comes back green.

nooga (approved):

PR #719, opened 2026-08-11
perf(vm): keep defer scaffolding out of the dispatch hot path

mparrett:
Since Go 1.21, `deferreturn` walks the same `_panic.nextDefer` machinery the panic path uses, so a function that merely *contains* a defer statement pays return-path scaffolding on every call — even when the defer is never registered at runtime. CPU profiles of the #700 workload show `runtime.(*_panic).nextDefer` and `runtime.(*_panic).start` attributed to `Frame.Run`'s normal returns, on both current main and a #645 revert build.

#645 put two such defers on every host-to-VM entry: `Run`'s panic-cleanup defer and `runLoop`'s `leaveFrame` defer. Both exist only to keep allocation-attribution balance and frame-pool hygiene when a Go panic unwinds through the VM; the panic itself is re-raised either way. Native functions that invoke a bytecode callback per element (`some`, `reduce` with a closure) pay the entry cost once per element, which is where most of #700's consumer-visible regression comes from.

This change hoists both defers into wrappers that are entered only when allocation attribution is on (`runChainProtected`, `runLoopAttr`), leaving `Run`, `runChain`, and `runLoopInner` with no defer statements at all. `leaveFrame` ignores its argument (attribution keeps its own frame stack), so the wrapper can leave on the loop's behalf without tracking which frame the loop ends on.

Behavior note: without attribution enabled, a Go panic now leaks the chain's pooled frames to GC instead of returning them to the pool. That matches pre-#645 behavior exactly — a panic skipped `runBytecodeCallTarget`'s `ReleaseFrame` the same way. The attribution tests and #645's frame-lifecycle tests cover the gated path.

Measured on the #700 shape, `(some (fn [x] (< 9999999 x)) (range 1000000))`, darwin/arm64, interleaved runs: main 89.9 ms → this branch 72.4 ms, against v1.12.2's 60.1 ms — about 59% of the gap on that workload.

Draft on purpose: the local box could not adjudicate the micro families. `FuncInvoke/Direct` and `Closure` improved consistently across runs, but `FrameDispatch`/`SeqIteration` swung both ways inside local noise, and the #706 investigation showed `runLoop` restructures can shift register allocation in either direction. The `perf-repeat` label is the point of this draft — if those families hold, this is ready for review.

Verified: `pkg/vm`, `pkg/rt`, and `test/` suites; linux, js/wasm, and plan9 builds.

### DIFF
--- a/pkg/vm/vm.go
+++ b/pkg/vm/vm.go
@@ -801,36 +801,76 @@ func releasePanickedFrames(state *frameRunState) {
 func (f *Frame) Run() (result Value, resultErr error) {
 	f.parent = nil
 	state := frameRunState{root: f, current: f}
-	entering := true
+	// A defer statement in this function would cost deferreturn scaffolding on
+	// every host->VM entry even when never registered (the compiler emits the
+	// return-path walk unconditionally), so the panic-cleanup defer lives in a
+	// separate wrapper taken only when allocation attribution is on. Without
+	// attribution, a Go panic leaks the chain's pooled frames to GC — exactly
+	// what pre-#645 did when a panic skipped runBytecodeCallTarget's
+	// ReleaseFrame; the panic itself is preserved either way.
+	if allocAttrEnabled {
+		return runChainProtected(&state)
+	}
+	return runChain(&state)
+}
+
+// runChainProtected owns the panic-path cleanup that keeps allocation
+// attribution balanced and pooled frames released. Kept out of Run and
+// runChain so their fast paths carry no defer scaffolding.
+func runChainProtected(state *frameRunState) (result Value, resultErr error) {
 	defer func() {
 		if r := recover(); r != nil {
-			// runLoop's defer already left the current frame. Balance and release
-			// every suspended child-owned frame before preserving the panic.
-			releasePanickedFrames(&state)
+			// runLoop's attr wrapper already left the current frame. Balance and
+			// release every suspended child-owned frame before preserving the panic.
+			releasePanickedFrames(state)
 			panic(r)
 		}
 	}()
+	return runChain(state)
+}
+
+// runChain drives the dispatch loop and the error-resume protocol for the
+// frame chain rooted at state.root.
+func runChain(state *frameRunState) (result Value, resultErr error) {
+	entering := true
 	for {
-		result, resultErr = state.current.runLoop(&state, entering)
+		result, resultErr = state.current.runLoop(state, entering)
 		entering = false
 		if resultErr == nil {
 			return result, nil
 		}
 		var resumed *Frame
-		resumed, resultErr = releaseFailedFrames(&state, resultErr)
+		resumed, resultErr = releaseFailedFrames(state, resultErr)
 		if resumed == nil {
 			return NIL, resultErr
 		}
 	}
 }
 
+// runLoop dispatches to the attr-tracking wrapper or the bare loop. leaveFrame
+// is a no-op unless allocation attribution is on, and a defer statement inside
+// the dispatch loop itself would cost deferreturn scaffolding on every VM
+// entry, so the deferring variant is a separate function.
 func (f *Frame) runLoop(state *frameRunState, entering bool) (Value, error) {
+	if allocAttrEnabled {
+		return f.runLoopAttr(state, entering)
+	}
+	return f.runLoopInner(state, entering)
+}
+
+// runLoopAttr balances attrPopFrame for whichever logical frame is current
+// when the loop unwinds; suspended parents remain active. leaveFrame ignores
+// its argument (attribution keeps its own frame stack), so the wrapper can
+// leave on behalf of the loop without tracking its final frame.
+func (f *Frame) runLoopAttr(state *frameRunState, entering bool) (Value, error) {
+	defer func() { leaveFrame(state.current) }()
+	return f.runLoopInner(state, entering)
+}
+
+func (f *Frame) runLoopInner(state *frameRunState, entering bool) (Value, error) {
 	if entering {
 		enterFrame(f)
 	}
-	// f changes as the loop descends and returns. On a final return or error,
-	// leave whichever logical frame is current; suspended parents remain active.
-	defer func() { leaveFrame(f) }()
 	for {
 		inst := f.code.code[f.ip]
 		if f.debug {


mparrett:
### Self-review

No actionable findings.

The defer-wrapper split preserves normal, error, and panic cleanup ordering. Verified with:

- `go test ./pkg/vm ./pkg/rt`
- A repeated descended-frame panic probe with allocation attribution enabled; shadow-stack and frame-pool cleanup remained balanced
- `git diff --check`

All required GitHub checks pass; only the repeat benchmark remains pending.

Non-blocking suggestion: commit a regression test for the attribution-enabled panic path, since existing tests do not exercise it directly.

mparrett:
The repeat A/B came back (run 31540713790, N=7 interleaved, sign verified against the raw base/head samples in `aggregate.json`; positive = head slower):

| family | median delta | worst |
|---|---:|---:|
| FuncInvoke/Closure | −26.5% | −28.2% |
| FuncInvoke/Direct | −26.3% | −28.5% |
| MultiArity/1Arg, /2Args | −24.8%, −24.9% | −27.0%, −25.1% |
| VariadicInvoke/1Arg, /5Args | −12.9%, −6.9% | −14.2%, −8.3% |
| SeqIteration (all variants) | ±0.8% | — |
| FrameDispatch | **+5.8%** | +7.1% |

`SeqIteration` held — the regalloc worry from the local runs was noise. The one real cost is `FrameDispatch`: all seven rounds positive, so it is the #706 mechanism (any restructure of the large dispatch function re-solves register allocation), not noise. It stayed under the 6% gate budget; `would_gate` is empty at every threshold.

The trade also shows end to end: a pure `loop`/`recur` counting loop runs ~6–8% slower on this branch, while invoke-heavy shapes (`fib`, the #700 `some`/`reduce`-with-closure workloads) gain 20%+. Callback-heavy code wins big; tight non-calling bytecode loops pay a little.

Taking that trade and marking ready for review: the shapes #700 tracks are the consumer-visible ones, and they are the ones that improve. If the dispatch loss matters more than expected, the follow-up lever is shrinking `runLoopInner` by moving cold error arms out of the hot switch — a change that would need its own `perf-repeat` pass either way.

mparrett:
Consumer-shaped follow-up on the dispatch-vs-invoke trade: xsofy's headless native engine bench (`bench/native.lg`, per-turn timing, no render/bridge), three interleaved rounds per binary on a loaded darwin/arm64 box — p50 with min-across-rounds in parentheses:

| fixture | main | this branch | + PreparedCall (#720 branch, reduce+some adopted) |
|---|---:|---:|---:|
| `:wait` | 35 (30) | 40 (22) | 23 (18) |
| `:autoex` | 30 (26) | 28 (20) | 23 (19) |
| `:descend` | 32 (31) | 23 (18) | 25 (15) |
| `:busy` | 30 (24) | 27 (19) | 28 (22) |

Noise is high on this box (one main-leg round caught a load spike), but the interleaved pattern is consistent: the branch stack is never worse and typically 15–30% better per game turn. The `FrameDispatch` +5.8% from the repeat A/B does not surface in this workload — the game loop is invoke/callback-dominated, so the trade documented above holds on the real consumer shape, not just the synthetic battery.

pkg/vm/vm.go
```diff
@@ -801,36 +801,76 @@ func releasePanickedFrames(state *frameRunState) {
...
 			return NIL, resultErr
 		}
 	}
 }
 
+// runLoop dispatches to the attr-tracking wrapper or the bare loop. leaveFrame
+// is a no-op unless allocation attribution is on, and a defer statement inside
+// the dispatch loop itself would cost deferreturn scaffolding on every VM
+// entry, so the deferring variant is a separate function.
 func (f *Frame) runLoop(state *frameRunState, entering bool) (Value, error) {
```
nooga, line 854: **Low severity** (perf, on-topic for going further): this gate doesn't inline. `go build -gcflags=-m` reports `cannot inline (*Frame).runLoop: function too complex: cost 144 exceeds budget 80`, so the entry path is now `Run → runChain → runLoop → runLoopInner` — one non-inlined call more than main's `Run → runLoop`, paid on every host→VM entry and every error-resume iteration, i.e. exactly the cost this PR is shrinking.

Folding the gate into `runChain`'s loop (hoist `attr := allocAttrEnabled` out of the loop, then call `runLoopAttr`/`runLoopInner` directly) deletes the hop and lets `runLoop` go away entirely. I built and benched that variant: `FuncInvoke/Closure` 44.6 → 44.2 ns/op median, `Direct` unchanged at ~43.4 — within noise on this box, but structurally strictly better and free.

pkg/vm/vm.go
```diff
@@ -801,36 +801,76 @@ func releasePanickedFrames(state *frameRunState) {
...
+	}
+	return f.runLoopInner(state, entering)
+}
+
+// runLoopAttr balances attrPopFrame for whichever logical frame is current
+// when the loop unwinds; suspended parents remain active. leaveFrame ignores
+// its argument (attribution keeps its own frame stack), so the wrapper can
+// leave on behalf of the loop without tracking its final frame.
+func (f *Frame) runLoopAttr(state *frameRunState, entering bool) (Value, error) {
+	defer func() { leaveFrame(state.current) }()
```
nooga, line 866: **Low severity** (latent, nothing broken today): this passes an argument the callee ignores, and the doc comment above leans on that fact. But `state.current` isn't always the loop's live frame — in `OP_RETURN` (vm.go:919-925) there's a window after `leaveFrame(child)`/`ReleaseFrame(child)` where `state.current` still points at the frame that was just handed back to the pool, before `state.current = f` runs. The old `leaveFrame(f)` was correct by construction at every instant.

Nothing misbehaves now because `leaveFrame` takes `_ *Frame`, but the first time that parameter gains a real use (per-frame attribution accounting, tracing, a debug assert) a panic in that window becomes a use-after-pool-release. Passing `nil`, or dropping the parameter from `leaveFrame` altogether, makes the invariant structural instead of documented.

pkg/vm/vm.go
```diff
@@ -801,36 +801,76 @@ func releasePanickedFrames(state *frameRunState) {
...
 	state := frameRunState{root: f, current: f}
-	entering := true
+	// A defer statement in this function would cost deferreturn scaffolding on
+	// every host->VM entry even when never registered (the compiler emits the
+	// return-path walk unconditionally), so the panic-cleanup defer lives in a
+	// separate wrapper taken only when allocation attribution is on. Without
+	// attribution, a Go panic leaks the chain's pooled frames to GC — exactly
+	// what pre-#645 did when a panic skipped runBytecodeCallTarget's
+	// ReleaseFrame; the panic itself is preserved either way.
+	if allocAttrEnabled {
```
nooga, line 811: **Low severity** (coverage gap, not a bug): gating `releasePanickedFrames` behind `allocAttrEnabled` is a deliberate documented trade-off that matches pre-#645, and in practice `NativeFn.Invoke`/`invokeCtx` recover `thrownPanic` at the native boundary, so panics rarely unwind through an active `runLoopInner`. Not flagging the leak itself.

The concern is that the **default** configuration's panic path now has no test at all — the PR note says the attribution tests cover "the gated path," and `TestDescendedErrorReleasesChildFrame` (explicit_frame_test.go:344) exercises the error path, not the panic path. With `LG_ALLOC_ATTR` unset `Run` is defer-free, so a future change that reintroduces a must-release-on-panic invariant would pass the whole suite silently. A test that panics out of a native inside a descended frame with attribution off, asserting the currently-intended pool behavior, would pin this.

nooga (commented):
Reviewed the dispatch-path changes. **No correctness bugs found** — three **low-severity** notes inline, none blocking.

Verification: `go build`, `go vet`, `go test ./... -count=1`, `go test -race ./pkg/vm/`, and `LG_ALLOC_ATTR=1 go test ./pkg/vm/ ./pkg/rt/` all pass. Traced attribution push/pop balance by hand across every exit of `runLoopInner` (root return, descend via `OP_INVOKE`/`OP_TAIL_CALL`, `OP_RETURN` unwind, error return, resume-after-handler, panic, nested `Run` re-entry) — the `runChainProtected` + `runLoopAttr` split is balance-equivalent to the old unconditional defers in every case. `frameRunState` still stays on `Run`'s stack in both paths, so no new heap alloc from passing `&state` into the wrappers.

The perf win reproduces: `(some (fn [x] (< 9999999 x)) (range 1000000))`, 3 interleaved reps on darwin/arm64 — main 277/324/254 ms vs PR 212/212/202 ms.

pkg/vm/vm.go
mparrett: Agreed, and taken as-is in #730: the gate is hoisted into `runChain`'s loop, `runLoopAttr`/`runLoopInner` are called directly, and `runLoop` is deleted. Interleaved go-bench medians there match your read — within noise, one call shorter. Kept off this PR so it merges as reviewed.

pkg/vm/vm.go
mparrett: Took the stronger option in #730: the parameter is gone, so the `OP_RETURN` window can't ever matter — the invariant is structural, and `runLoopAttr`'s defer sheds its closure as a side effect.

pkg/vm/vm.go
mparrett: #730 adds the test: a raw (unwrapped) callable panics inside a descended frame, asserting the panic value survives and the pool sees the intended behavior in both configurations — leak-to-GC with attribution off, exactly-once release via `runChainProtected` with it on. A future change that reintroduces a must-release-on-panic invariant on the default path now has to update that test consciously.

mparrett:
All three review notes are addressed in #730, stacked on this branch — the runChain fold (`runLoop` deleted), a parameter-free `leaveFrame`, and the default-config panic-path test. Cut as a draft ahead of the merge queue on purpose: the threads get a live diff to point at instead of a promise, and this PR merges exactly as you reviewed it. #730 rebases onto main once the stack lands and picks up `perf-repeat` when the queue is quiet. Nothing here blocks #719 from my side.

nooga (approved):

PR #723, opened 2026-08-12
perf(vm): move cold error arms out of the dispatch loop

mparrett:
Stacked on #719. The repeat A/B there found one real cost: `FrameDispatch` +5.8% median, the register-allocation sensitivity of restructuring the large dispatch function (#706's mechanism notes). This change shrinks that function's hot text so the allocator has a smaller region to solve: eight `//go:noinline` helpers absorb the cold arms — error construction and wrapping (`execErr`/`execWrap`/`wrapCallErr`), the not-a-function, integer overflow, divide-by-zero, and bit-op type errors — across 98 sites, plus one hoist of the per-instruction debug print. `runLoopInner` text drops from 25,872 to 20,384 bytes (−21%). Error output is byte-identical; behavior is unchanged.

Draft, `perf-repeat` label, because the claim needs the amd64 lane to decide it: on darwin/arm64 the +5.8% `FrameDispatch` regression does not reproduce at all — the base branch already runs the dispatch loop at parity with main there — so local measurement can only establish no harm, which it does (dispatch and `some` flat vs base across interleaved batches; the base branch's invoke-family wins survive intact at 1.22–1.26× vs main). If the repeat A/B shows `FrameDispatch` reclaimed, this erases the one cost on #719's trade; if flat, it is a hot-function shrink with no perf claim and can be judged on maintainability alone.

The two commits are separately bisectable (helpers, then the debug-print hoist).

Verified: `pkg/vm`, `pkg/rt`, and `test/` suites; linux, js/wasm, and plan9 builds; error-message byte-parity checked against the base branch.

### DIFF
--- a/pkg/rt/generated.manifest
+++ b/pkg/rt/generated.manifest
@@ -266,7 +266,7 @@ pkg/rt/core_compiled.lgb pkg/vm/uuid.go generator 249cc824c4af570c1693f968fdbab2
 pkg/rt/core_compiled.lgb pkg/vm/value.go generator 9c00df4c29b70dc9fea057628152cf139bdc73abc75d8666877fb7bf94be8c96
 pkg/rt/core_compiled.lgb pkg/vm/var.go generator 816e670512bc08f78a6d8a39bd8280dea82b171454ac99a54be2595942aee39d
 pkg/rt/core_compiled.lgb pkg/vm/vector.go generator 3695aa8424bbc2f34ae3b100331af59925fbf35ce6875c950c92e562f47bbc10
-pkg/rt/core_compiled.lgb pkg/vm/vm.go generator c9eac4f6738ae3267a18e771575da28531cceff76afa7e89ef31e95534612df6
+pkg/rt/core_compiled.lgb pkg/vm/vm.go generator 9dd0383eeabc14f34f3cea1e1bb40e5525bebf11b64aff0b56a98bd6664b4cc6
 pkg/rt/core_compiled.lgb pkg/vm/void.go generator bea0103c574dad2dd702302b7611a9947ac917e73fb8eba82e28c57768b71627
 pkg/rt/core_compiled.lgb pkg/vm/volatile.go generator dd18155eda1c0fb376f9934fea08a1a4389c9efbbc1d56685f04839a99f280ef
 pkg/rt/core_compiled.lgb pkg/rt/core/async.lg input f1ae4c5801422270587d811529d1735f9882945b0a380e7cc0b783ecca15795a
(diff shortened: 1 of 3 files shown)


mparrett:
The repeat A/B came back (run 31556152262, N=7 interleaved, sign verified against the raw samples; negative = this branch faster than base #719):

| family | median delta | worst |
|---|---:|---:|
| FrameDispatch | **−4.76%** | −6.73% |
| FuncInvoke/* (all) | flat | — |
| everything else | within ±3% / per-family floors | — |

`would_gate` empty at every budget. The one `confirm_hits` entry (`VectorConj/PersistentVector/10`, +1.6% median with a noisy worst-case) is a small family well inside its A/A floor.

So the shrink reclaims most of the +5.8% `FrameDispatch` cost the base branch's repeat A/B measured: net across #719 + this branch vs main, the invoke families keep their −25% to −26% and dispatch lands ~+1%. The one cost on #719's trade is essentially erased on the platform that had it (darwin/arm64 never reproduced it; both measurements are the amd64 lane).

Marking ready for review. The two commits are separately bisectable if the debug-print hoist wants independent scrutiny.

mparrett:
## Self-review

No findings on the incremental #719 → #723 diff.

Reviewed the cold-helper extraction and debug-trace hoist for behavioral parity. Error types, message text, wrapping order, source attribution, handler routing, and debug output ordering are preserved. I also checked the sibling #726 → #727 stack for overlapping assumptions/conflicts and found none.

Validation:
- `go test -count=1 -timeout 60s ./pkg/vm ./pkg/rt ./test` passes.
- The PR’s repeat A/B reports `FrameDispatch` −4.76% median versus #719, reclaiming most of #719’s measured dispatch regression.
- A broader `go test ./...` run passed the affected packages, then failed in unrelated e2e coverage because this worktree lacks generated lowering artifacts (`.lgbgen-tree.sum`) and the determinism test exceeded the 60s suite timeout.

Verdict: clean; ready to merge once the stack’s normal requirements are satisfied.

nooga:
Reviewed the incremental #719 → #723 diff (vm.go). The noinline-helper extraction is a clean 1:1 substitution at every site I checked — spot-checked all of the bit-op/overflow/div-by-zero arms plus the invoke/return paths; error type, message text, and wrap order are all preserved exactly. Matches the description and your own self-review.

The repeat A/B you already posted (-4.76% median FrameDispatch vs #719) settles the perf claim, and CI is green. No concerns from me — looks ready to merge.

mparrett:
@nooga — your 08-29 read went up as a comment rather than an approving review, so it never counted against the required approval and this is still blocked. Mind converting it? Nothing has moved since: still green, still 0 behind `main`.

nooga (approved):
Approving formally this time (my 08-29 read was a comment, hence the lingering REVIEW_REQUIRED). Re-checked the extraction against main line by line — every cold arm keeps the same error value, handler-stack and frame-release behaviour. perf-repeat's -4.76% FrameDispatch median reclaims most of #719's regression. Merging; #730 goes on top.

Issue #720, opened 2026-08-11
vm: PreparedCall — resolve once, reuse one frame for per-element callback invokes


mparrett:
## Why

A native function that invokes a bytecode callback per element — `some`, `reduce` with a closure, `every?`, the lazy-seq step functions — pays three costs on every `ec.Invoke` that are constants of the call site, not of the call:

- `resolveBytecodeCall`: unwrap Meta/Closure/MultiArity and arity-check, producing the same target every element
- frame pool traffic: two `framePoolMu` lock/unlock pairs per call
- `NewFrame` field initialization and `ReleaseFrame` clearing

This predates #645 — v1.12.2 pays all of it too. Post-#645 these shapes also carry the per-entry cost tracked in #700, which makes the per-element boundary the dominant cost of callback-heavy seq operations.

## Design

`PreparedCall` resolves the callable once, owns one frame, and per call resets only what a callee can have changed, then runs the dispatch loop:

```go
pc := ec.PrepareCall(fn, 1)   // nil => caller keeps the ec.Invoke path
if pc != nil {
    defer pc.Release()
}
...
v, err := pc.Call1(elem)
```

Reset details the branch pins with tests:

- The per-call reset must restore `code`/`consts`/`closedOvers`, not just `args`/`ip`/`sp`: a tail call in the callee body rebinds the owned frame to the tail target through `installBytecodeCall`.
- `handlers` and `debug` need clearing too — an error unwind can leave stale handlers, and tracing can arm `debug` mid-walk.
- `PrepareCall` returns nil for variadic, native, and other non-bytecode targets, so every adopter keeps its generic fallback and the change stays incremental.
- A `PreparedCall` is single-owner: one native call activation, no sharing, no reentrant use. Nested uses of the same lg function each prepare their own.

## Status and numbers

Branch `perf/vm-prepared-call` (stacked on #719's `perf/vm-defer-free-dispatch`) implements this and adopts it in `rt.Some`. On `(some (fn [x] (< 9999999 x)) (range 1000000))`, darwin/arm64, interleaved runs:

| build | median |
|---|---:|
| main (f3ca5f9b) | 89.9 ms |
| `perf/vm-defer-free-dispatch` | 72.4 ms |
| `perf/vm-prepared-call` | 51.0 ms |
| v1.12.2 | 60.1 ms |

The #700 regression on this shape is fully recovered and the result is ~15% below the release baseline, with #644's stack safety kept.

## Adopters

Each is a small, independently measurable change:

[...]

PR #727, opened 2026-08-12
perf(rt): adopt PreparedCall in reduce's fold loops

mparrett:
Second adopter from #720, stacked on the PreparedCall PR. `reduce`'s four fold loops — ArrayVector, Range, chunked seq, and linear seq — each invoked the reducing fn through `ec.Invoke`, paying per-element resolution, frame-pool mutex traffic, and frame init. Resolve the reducer once per fold with `PrepareCall` and reuse its frame; non-bytecode and variadic reducers keep the reused-`fargs` fallback in all four loops.

Adds `PreparedCall.Call2` for binary callables, introduced with its first consumer rather than ahead of one. `PrepareCall`'s arity guard widens to admit 2, and each `CallN` validates against the prepared arity — a mismatch returns an error instead of panicking or invoking with unpopulated argument slots.

Measured on `(reduce (fn [a x] (+ a x)) 0 (range 1000000))`, darwin/arm64, interleaved mins: 50.5 ms vs 67.3 on the base branch and ~58.7 on v1.12.2 — the second adopter to land below the release baseline. Same `perf-repeat` caveat as the base PR: the micro families don't route through `rt` reduce, so the lane is a regression guard; the e2e number is the claim.

New tests: `Call2` repeated invocation, `CallN`/arity mismatch errors, unsupported-arity rejection. Suites green: `pkg/vm`, `pkg/rt`, `test/`.


nooga:
Reviewed prepared_call.go and the four reduce fold-loop call sites. Call2 mirrors Call1's arity-guard pattern correctly, PrepareCall's widened arity check (1..2) matches, and all four loops (ArrayVector/Range/chunked/linear) follow the same pc != nil / fargs-fallback shape consistently, with defer pc.Release() placed right after acquisition. Matches the description; CI green. No concerns.

mparrett:
@nooga — same as #723: your 08-29 read is a comment rather than an approving review, so the required approval is still outstanding and the PR sits blocked. Mind converting it? Green and 0 behind `main`.

nooga (approved):
Approving formally (my 08-29 read was a comment). Verified `reduced` short-circuit, arity errors and reducing-fn panics are preserved across all four fold loops; perf-repeat clean with a real win on the reduce family. Independent of the vm.go pair, merging.

PR #730, opened 2026-08-12
perf(vm): remove the runLoop hop; make leaveFrame parameter-free

mparrett:
Draft until the stack ahead of it settles; I'll add `perf-repeat` for an EPYC adjudication once the queue is quiet.

Carries the three low-severity notes from #719's review as a stacked PR, so #719 merges as reviewed and its perf-repeat baseline stays put.

- **Fold the attribution gate into `runChain`.** `runLoop` was a dispatcher too big to inline (`cost 144 exceeds budget 80`), so the entry path paid `Run → runChain → runLoop → runLoopInner` — one non-inlined call more than needed on every host→VM entry and every error-resume iteration. The gate is now hoisted out of `runChain`'s loop and calls `runLoopAttr`/`runLoopInner` directly; `runLoop` is gone.
- **Drop `leaveFrame`'s parameter.** It took a `*Frame` it never read, and in `OP_RETURN` there is a window after `ReleaseFrame(child)` where `state.current` still points at the pooled frame — the first real use of the parameter would have been a use-after-release. With no parameter the invariant is structural, and `runLoopAttr`'s defer sheds its closure.
- **Pin the default-config panic path.** With `LG_ALLOC_ATTR` unset, `Run` is deliberately defer-free and a Go panic leaks the chain's pooled frames to GC (pre-#645 parity) — previously documented but untested. The new test panics a raw (unwrapped) callable inside a descended frame and asserts the panic value survives and the pool sees the intended behavior in both configurations: leak-to-GC with attribution off, exactly-once release via `runChainProtected` with it on.

Measured (interleaved go-bench medians, N=6, vs #719's head): `FrameDispatch` 0.94×, `FuncInvoke/Closure` 0.97×, `Direct` 1.00× — flat to slightly better, matching the review's own bench of the fold. 

Verified: `pkg/vm`, `pkg/rt`, `pkg/genmanifest`, `test/` suites; `go test -race ./pkg/vm`; both suites again with `LG_ALLOC_ATTR=1`; the new test passes in both configurations; linux, plan9, js/wasm builds.

### DIFF
--- a/pkg/rt/generated.manifest
+++ b/pkg/rt/generated.manifest
@@ -266,7 +266,7 @@ pkg/rt/core_compiled.lgb pkg/vm/uuid.go generator 249cc824c4af570c1693f968fdbab2
 pkg/rt/core_compiled.lgb pkg/vm/value.go generator 9c00df4c29b70dc9fea057628152cf139bdc73abc75d8666877fb7bf94be8c96
 pkg/rt/core_compiled.lgb pkg/vm/var.go generator 816e670512bc08f78a6d8a39bd8280dea82b171454ac99a54be2595942aee39d
 pkg/rt/core_compiled.lgb pkg/vm/vector.go generator 3695aa8424bbc2f34ae3b100331af59925fbf35ce6875c950c92e562f47bbc10
-pkg/rt/core_compiled.lgb pkg/vm/vm.go generator 9dd0383eeabc14f34f3cea1e1bb40e5525bebf11b64aff0b56a98bd6664b4cc6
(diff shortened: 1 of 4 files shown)


mparrett:
@nooga @nnunley — this is the only one of the dispatch-family PRs with no review on it at all. Green, 0 behind `main`. Either of you able to take it?

nooga (approved):
LGTM. Mechanical hoist plus the parameter removal, both call sites updated; the new pool-pinning test passes in both `LG_ALLOC_ATTR` configs and under `-race`; perf-repeat clean. Merges cleanly onto #723 apart from the generated manifest. Please rebase once #723 lands and I'll merge.

PR #733, opened 2026-08-12
build: move build outputs into build/, promote lg into bin/

nnunley:
## Summary

`make` wrote `lg` at the repository root, alongside every ad-hoc `go build` output. Outputs now go to `build/`, and `make build` promotes `lg` into `bin/`.

Stacked on #732, which added the ignore rules and the push-stage hook.

## Why

Two consequences of building at the root:

The ignore list was a per-binary list maintained by noticing — `/lgbgen`, `/lginterop`, `/lgprimgen`, `/bench-ratchet`, `/perf-page`, `/let-go`, `*.test`. `check-generated` was never noticed.

`make clean` removed `$(LG)` only. `lgbgen`, `lginterop`, `lgprimgen`, `bench-ratchet`, `perf-page`, every `*.test` and `check-generated` survived it. With directory-scoped outputs, `clean` removes `build/` and `bin/` and is complete by construction.

## Shape

```make
BUILD-DIR := build
BIN-DIR   := bin
LG        := $(BUILD-DIR)/lg
build: $(LG-PROMOTED)
$(LG-PROMOTED): $(LG)
	install -m 0755 $(LG) $@
clean: clean-lowered
	$(RM) -r $(BUILD-DIR) $(BIN-DIR)
```

Promotion is part of the build rule rather than a separate target, so `bin/lg` cannot silently lag `build/lg`.

**Copy rather than symlink.** Symlink creation on Windows needs Developer Mode or elevation, and the project ships Windows builds (`.goreleaser.yml`). The staleness argument for symlinks is answered by promoting inside the build rule.

## Consumers

- `scripts/generate.lg` and `scripts/gogen-trampoline.lg`: `--lg` defaults, and the Makefile's own `generate` invocation now passes `$(LG)`.
- `test/benches/ysbench.sh` and `test/namespace_shadow_warning_test/run.sh`: both honour `$LG` and default to `build/lg`.
- `.selfhost-lgbgen` moves to `$(BUILD-DIR)/selfhost-lgbgen`.

Unaffected: `.github/workflows/` contains no reference to the built binary by path — workflows use `go build -o` with explicit temp paths, `go run`, or `make`. goreleaser names its own output and writes into `dist/`, so released artifacts are still `lg`.

## Policy

`docs/contribution-policy.md` gains a Build outputs section: binaries never live at the repository root, build ad-hoc tools with `go build -o build/<tool> ./cmd/<tool>`, and the rule is enforced by `forbid-compiled-binaries` at commit and push rather than requested. `AGENTS.md` is gitignored in this repository, so the policy document is the durable home for it.

## Validation

- `make clean && make build` produces `build/lg` and `bin/lg`; `jj status` reports neither.
- `make clean` removes both directories.
- `go test ./test/... ./pkg/rt/... ./pkg/ir/...` passes, e2e included (248s).


Makefile
```diff
@@ -33,8 +33,14 @@ include $M/shell.mk
...
 
 # Prefer - to _ for make var names (won't conflict with env vars):
-LG := lg
-LG-PROFILE ?= lg-profile
+# Build outputs live under build/; bin/ holds the promoted current copy that a
+# PATH entry or a script can point at. Both are gitignored as directories, so a
+# new binary needs no new ignore entry, and `make clean` removes both.
+BUILD-DIR := build
+BIN-DIR := bin
+LG := $(BUILD-DIR)/lg
```
mparrett, line 44: [P1] Update every Make consumer to use `$(LG)`

Changing `LG` here stops `build` from producing `./lg`, but the recipes for `fanout-ratchet{,-update,-show}`, `ir-stress`, `jank-stress`, all the `ir-stress-*` gates/rebaseline targets, and `ratchets{,-update}` still execute `./lg`. On a clean checkout their `build` prerequisite produces only `build/lg` and `bin/lg`, so each recipe then fails with `./lg: No such file or directory`; notably the configured pre-push `make ir-stress-gate` is among them. `make -n fanout-ratchet` ends with `./lg ...` after building/promoting the two new paths. Please replace these recipe invocations with `$(LG)` and sweep the remaining source-build documentation/examples as appropriate.

.pre-commit-config.yaml
```diff
@@ -31,12 +31,16 @@ repos:
...
+      # itself: `git commit` never runs, so the pre-commit stage never fires for
+      # a jj-based workflow. pre-push is the only gate those commits pass
+      # through, and it is where prek runs under jj-pre-push.
       - id: forbid-compiled-binaries
         name: forbid committing compiled executables
         description: Reject staged Mach-O / ELF / PE binaries before they bloat history
         entry: python3 scripts/reject_compiled_binaries.py
         language: system
-        stages: [pre-commit]
+        stages: [pre-commit, pre-push]
```
mparrett, line 46: [P2] Inspect all pushed commits, not only the net diff

At `pre-push`, prek/pre-commit supplies filenames from the `from_ref...to_ref` diff and this hook opens those paths from the current working tree. A binary added in one new commit and deleted in a later new commit is absent from that net diff (`git diff --name-only from...to`), yet its blob and commit are still pushed and permanently bloat history. The same current-tree lookup can also disagree with the exact `to_ref` under a jj workflow. Please make the pre-push path walk the new commits/blobs via Git using the supplied from/to refs, while retaining the current filename-based behavior for pre-commit.

mparrett (changes requested):
Commented inline with some findings to address before merge.

nnunley:
P2 addressed at `87c46d80ee7c`.

At pre-push the hook now ignores prek’s net-diff filenames and enumerates every Git object reachable from `PRE_COMMIT_TO_REF` but not `PRE_COMMIT_FROM_REF`. It classifies each newly reachable blob with the existing ELF, Mach-O, universal Mach-O, and PE/DOS magic checks. The pre-commit path retains filename-based working-tree inspection.

`always_run: true` is required because a range that adds a binary in one commit and deletes it later can have an empty net diff. A real prek drill for that exact shape rejects the push and reports the intermediate path/blob. Regression tests also cover every retained magic, safe blobs, invalid refs (fail closed), inherited pre-push environment, and hook configuration.

During the gated push, the existing 60s Go-package hook timeout consistently killed the unchanged e2e package; the hook now matches the Makefile’s 600s default, with a synchronization test. All pre-push hooks passed.

.pre-commit-config.yaml
nnunley: Addressed in `87c46d80ee7c`.

The pre-push path now ignores the net-diff filename argv and enumerates the object set reachable from `PRE_COMMIT_TO_REF` but not `PRE_COMMIT_FROM_REF`; every newly reachable blob is checked for the retained ELF/Mach-O/universal/PE magic signatures. The pre-commit path still checks supplied working-tree filenames.

The hook is `always_run: true`, so an add-then-delete range with an empty net diff still executes. Regression coverage builds that exact two-commit shape in a scratch repository, plus safe-range, invalid-ref/fail-closed, all magic signatures, and inherited pre-push-environment cases. A real prek drill also rejected the intermediate ELF blob and reported its path/blob ID. All pre-push hooks passed on the pushed head.

Makefile
```diff
@@ -75,7 +89,33 @@ default:: run
...
+# booted fast, and not obviously broken" — NOT "verified correct". Deep
+# correctness is make test / ir-stress-gate / gogen-diff, each ~2 min, which is
+# too slow to run on every promotion.
+$(LG-PROMOTED): $(LG) $(BOOTPROBE)
+	@$(MAKE) --no-print-directory smoke
+	@mkdir -p $(BIN-DIR)
+	install -m 0755 $(LG) $@
+	@echo "promoted $(LG) -> $@"
+
+$(BOOTPROBE): $(GO) $(ROOT-GO-FILES) pkg/**/* pkg/rt/core_compiled.lgb
```
mparrett: [P2] Track the promotion gate inputs

This target does not depend on `cmd/bootprobe/*.go`, and `$(LG-PROMOTED)` does not depend on `scripts/smoke.lg` or `scripts/smoke-boot.sh`. After a successful build, each of `make -n -W cmd/bootprobe/main.go build`, `make -n -W scripts/smoke.lg build`, and `make -n -W scripts/smoke-boot.sh build` reports that nothing needs doing. Thus a later change to the probe or either gate can leave the existing `bin/lg` marked as promoted without rebuilding the probe or rerunning the updated smoke checks, contradicting the stated meaning of `bin/lg`. Please add the bootprobe sources and smoke scripts to the relevant prerequisite lists, or otherwise make `build` rerun the gate when those inputs change.

mparrett:
@nnunley Thanks for the updates. Latest review turned up one new finding: https://github.com/nooga/let-go/pull/733#discussion_r3785583638

Also left a note on https://github.com/nooga/let-go/pull/732

Makefile
nnunley: Fixed at `f3555c44c925`. `$(BOOTPROBE)` depends on `cmd/bootprobe/*.go`, and `$(LG-PROMOTED)` depends on the candidate, bootprobe, and both smoke scripts. `test/build_promotion_prereqs_test.go` uses future-dated artifacts plus `make -n -W` to assert each input reruns both smoke commands and the final install. Clean exclusive `make build` passed with a 3.362 ms median under the 8 ms budget.

mparrett (approved):
Re-reviewed at f3555c4 after the rebase and promotion-prerequisite fix. The earlier findings remain addressed: Make consumers use the build output, the pre-push guard scans newly reachable blobs, and changes to the bootprobe or either smoke script now rerun the promotion gates and install. Focused Go tests, a clean make build, and the three incremental make -W checks pass locally; current Go and CodeQL runs are green. No blocking findings.

nnunley:
Remediation published at `a9664decbf71a2ea9e6e38bf89f9ce0dd12db8d8`.

Push mode is now selected when either pre-push ref variable is present. A missing TO ref fails closed; a missing FROM ref scans the complete TO ancestry from the null boundary, so orphan-history pushes cannot bypass compiled-binary scanning. `TestRejectCompiledBinariesChecksOrphanHistoryWithoutFromRef` locks the case down.

Evidence: 14 focused tests, Python compilation, full short suite (1,744 tests), check-generated, clean exclusive build, and independent re-review passed. The published head now has 14 passing GitHub checks and no failures.

mparrett:
Re-checked at new head `a9664de`.

- One new commit fixes orphan-history push scanning.
- The implementation and regression coverage look correct.
- Focused tests and Python compilation pass locally.
- All 14 GitHub checks pass.
- PR is mergeable and the aggregate review decision remains `APPROVED`.

No new findings and no additional review action is needed. The older addressed threads remain mechanically unresolved, but they are not blocking.

Issue #596, opened 2026-07-20
CLI: a first-class AOT compile command


mparrett:
The AOT pipeline is proven end to end (#425: lower `.lg` → typed Go → standalone binary, 16.6× on the fib spike), but driving it requires either `scripts/lg-compile` with hand-set `LG_SOURCE_PATHS` plus manual `go.mod`/`go build` steps, or gloat's external orchestration. There's no `lg`-native way to say "compile this program to a binary."

Proposal: wrap the #425 pipeline behind one command — working name `lg compile` (or an `-native` output mode beside `-c`/`-w`):

```
lg compile [-o out] [-tags ...] main.lg
```

which would lower the program through gogen, emit the Go package, scaffold the module (the same generated-module machinery `lg -w` uses, including the released-binary version pinning from #594), and drive `go build` to a standalone binary. Failures land in one place instead of three tools.

Prerequisite already in flight: #557 self-contains `gogen` as embedded core source, so a released binary can run the lowering without a let-go checkout — that's what makes this turnkey rather than dev-tree-only. The lowering capability gaps (#358, #551, #554, #562) bound what compiles today, but they're orthogonal: the command is worth having for the kernel-shaped programs that already lower.

Open questions I'd rather not presume on: subcommand vs flag (`lg compile` vs `lg -native`), whether partial lowering falls back to embedding bytecode for non-lowerable fns or errors out, and whether gloat should grow on top of this command or stay a parallel path. Happy to take a first cut once the shape is settled.

mparrett:
#557 merged: the prerequisite noted above (\`gogen\` self-contained as embedded core source, closing #425 Item 2) is done. \`lg-compile\` now resolves \`gogen\` from shipped binaries with no \`LG_SOURCE_PATHS\` or cwd plumbing.

Remaining gap toward this command: #425 Item 1 (the native-entry frame generator), still unstarted per that issue. Item 3 there (signal the silent perf-cliff) is described as largely subsumed by Item 1. The lowering-coverage issues (#358, #551, #554, #562) stay orthogonal, as noted above.

mparrett:
Claiming this — picking it up now.

@nnunley I floated this on #729 earlier today; putting it here so it's visible from the tracker. Shout if you're already partway in and I'll back off.

[...]

PR #735, opened 2026-08-12
feat(aot): embed the AOT compile driver as the lg.compiler namespace

mparrett:
Driving AOT lowering required a let-go checkout on disk. Everything the driver *requires* has shipped inside the binary since #557 (gogen) and #628 (the entry frame); the driver itself was the last piece living on the filesystem, as `scripts/lg-compile`. #596 asks for a first-class `lg compile`, and that cannot be built on a file released binaries do not carry. This is the groundwork, and it stands on its own: a shipped `lg` drives lowering from any directory.

## What changed

`scripts/lg-compile`'s body becomes the `lg.compiler` namespace at `pkg/rt/core/lg/compiler.lg` — an ordinary embedded core namespace, so it is enrolled in `EmbeddedNSNames` and swept by genmanifest's `sourceSpecs` like every other core source, with no auxiliary-embed exception. Editing it churns `pkg/rt/generated.sums`, which is the cost of the driver being part of the compiler rather than an auxiliary attached to it. `scripts/lg-compile` stays as the CLI shim, unchanged in path, argv, and exit codes. The layering follows the architecture in #786.

The driver returns a result map (`{:emit-fails n :lowered-funcs n :entry …}`) rather than printing and calling `os/exit`, so a Go caller can consume the outcome directly. `lg.compiler/exit-code` owns the exit policy as a pure function of that map, and the shim applies it.

## Enrolled is not bundled

A core namespace is also a bundled namespace, and `lg.compiler` requires `gogen`, `ir.data`, `ir.passes.pipeline`, and `ir.passes.entry-frame`. A require drags its transitive closure into the const pool, so enrolling the driver pulls the whole IR pipeline into `core_compiled.lgb`. Measured on one machine on 2026-09-02, that took the bundle from 308,175 to 1,068,039 bytes and the smoke-boot median from 4.060 ms to 23.385 ms, against an 8 ms budget that `make generate` enforces.

`cmd/lgbgen` already had the mechanism: `isIRBundleSkipped` keeps `ir.*` out of the bytecode bundle while leaving it in Go lowering, added for this same regression. It is generalized here to `isBundleSkippedTool` and now covers `lg.compiler` and its descendants. The family prefix rather than the exact name is deliberate: #786 moves `ir.*` under `lg.compiler.ir.*`, and an exact-name check would then match neither name and restore the regression silently. The driver loads from embedded source on demand exactly as `ir.*` does, so a plain `lg` script pays nothing at startup, and `pkg/rt/core_compiled.lgb` on this branch is byte-identical to `main`'s.

`pkg/rt/gogen_src.go` also loses its `//go:build !bootstrap` gate: lgbgen builds with `-tags bootstrap`, and under that tag `gogen` had no embedded source, so `lg.compiler` would not compile. That leaves `gogen` auxiliary — absent from `EmbeddedNSNames`, never precompiled, never lowered itself.

## The single-top-level-form constraint

The script ran its whole body inside one top-level form on purpose: top-level forms compile and run one at a time, and the driver `eval`s user `(ns …)` forms, which switch the global current namespace, so any later top-level form in the script resolved against the loaded file's namespace instead of the script's.

A namespace has no such problem. Every `defn` is compiled when the namespace loads, and the user-code `eval` happens inside `load-specs!`, a function body invoked later. The sequencing hazard is structural, so it is gone rather than worked around.

Related: `(def --script-ns *ns*)` was captured for a restore that never existed — `git log -S` finds one commit touching it (#348), where it was already unused. It is dropped rather than carried over.

## Compatibility

External orchestrators invoke the driver by path and read its output, so the shim preserves that contract:

- `scripts/lg-compile` keeps its path and positional argv (`[--entry-frame] <out-dir> <import-prefix> <file.lg>…`).
- The `EMIT-FAIL <path> pkg=<pkg> returned <type>` line still goes to stdout. Callers grep for it, so the move to return values keeps printing it as well as counting it.
- An EMIT-FAIL is still not an exit failure. A function that does not lower stays on the trampoline and the program runs, so partial lowering is a normal outcome the orchestrator judges for itself; only a fatal error exits non-zero.
- `--entry-frame` stays opt-in, so package-only callers that own their own executable frames see no difference.

## Documentation

`docs/design/go-aot-backend.md` gains a "Driving lowering" section. The doc mentioned neither the driver nor `scripts/lg-compile` before; it now carries where the driver lives, why core-embedded is not the same as bundled, and the stdout/exit contract above.

## Verification

- New: `TestLgCompileEmitFailContract` pins both halves of the stdout/exit contract. It injects a synthetic non-lowered result rather than using a `.lg` fixture, because no construct probed (try/catch, atom, reify, deftype, dynamic `set!`, lazy-seq, macro-expanded body, multimethod dispatch) fails to lower any more, and a fixture picked for today's gaps would stop testing the path the moment lowering covered it. The exit-status half runs the real shim against a stubbed `main-from-args`, so it fails if "exit 1 when `:emit-fails` is nonzero" is reintroduced.
- New: `TestLgCompilerResolvesFromEmbeddedSource` in `pkg/resolver` resolves the driver with an empty search-path slice, so only embedded resolution can satisfy it, and calls a driver fn to prove the source compiled and its transitive requires resolved embed-only.
- `TestLgCompileEntryFrameOptIn`, `TestNativeEntryASTGate`, `TestLoweringDeterminism`, and `TestDeftypeSkeletonNativeLowering` pass unchanged. `TestLoweringDeterminism` is the check #425 named as the reason to keep `gogen` out of core.
- `make check-generated` clean, and it leaves the tree unmodified.
- Builds under the default and `bootstrap` tags, and for `linux/amd64`, `js/wasm`, `plan9/amd64`, and `wasip1/wasm`.

Part of #596, under #258.


mparrett:
Self-review turned up one contract change that shouldn't be in here: the shim had picked up an `os/exit 1` on a non-zero EMIT-FAIL count. The original script never did — it printed the line, kept going, and exited 0 — so exit codes were the one part of the compatibility note above that wasn't true.

It reaches two callers, both in package-only mode. gloat checks the exit code before it greps stdout, so a non-zero exit sends it down the generic `lg-compile failed` branch and the specific EMIT-FAIL line never prints; that branch also appends a "your lg may be stale, run `make lg-dev`" hint, which would be the wrong diagnosis. legmacs' `build.sh` runs under `set -e` and would abort a build that previously finished, since an unlowered fn stays on the trampoline and the program still runs.

Fixed in d3ae481 by dropping the exit. `:emit-fails` stays in the driver's return map, which is where the `lg compile` Go caller should read it if it wants to decide differently. 5b11a84 is a separate docstring-position nit in the same file.

Verified: the entry-frame opt-in and native-entry matrix e2e tests pass, the resolver embed test passes, `examples/aot/native-entry/build.sh` still gives 55 / 5702887, and the shim's exit codes match the pre-PR script again — 0 on success and on EMIT-FAIL, 1 on an entry-arity error.

One test gap stays open: nothing asserts the EMIT-FAIL exit code, which is how this got past a green run. Producing a genuine EMIT-FAIL isn't cheap, since the per-ns catch only fires on an internal lowering exception, so I checked the shim's control flow against synthetic result maps instead.

nnunley (changes requested):
The extraction is useful, but I don't want `lg.aotdriver` to become a separate permanent compiler architecture.

The existing `ir.*` tree is where let-go's AOT compilation, optimization, and lowering architecture belongs. This namespace does substantially more than adapt a command: `parse-spec`, dependency ordering, lowering through the IR pipeline, package emission, and entry-frame generation are compilation services. Please put that reusable orchestration behind an appropriate `ir.*` API rather than a new `lg.aotdriver` seam.

The placement rationale also treats `gogen`'s current auxiliary `!bootstrap` embedding as the architectural precedent. That is a transitional constraint, not the destination: `gogen` is intended to become the bootstrapped Go-emission backend for the IR pipeline. This PR does not need to complete that bootstrap work, but it should preserve that direction instead of copying the exception into another compiler namespace.

For the eventual CLI, I suggest a thin namespace such as `lg.commands.compile` that only parses arguments, invokes the IR compilation API, and maps its result to diagnostics and exit status. `scripts/lg-compile` can call the same API as a compatibility shim.

Please revise the namespace/layering accordingly. If the work needs to be split, keep the complete bootstrap and command architecture explicit as follow-up work rather than dropping it from the roadmap.

nnunley:
Tracking the agreed compiler namespace and bootstrap destination in #786.

Please use that issue as the architectural reference for this PR. The relevant boundary for #735 is: reusable program compilation belongs behind `lg.compiler`; IR implementation belongs under `lg.compiler.ir.*`; the eventual `lg.commands.compile` namespace remains a thin CLI adapter. The broader namespace migration and `gogen` bootstrap can be split into follow-ups, but this PR should not establish `lg.aotdriver` as the permanent seam or treat the current auxiliary `!bootstrap` embedding as the destination.

mparrett:
@nnunley Agreed on the layering, and I'll restructure. `lg.aotdriver` was the wrong seam: the orchestration in it is compilation, not command adaptation, and it belongs behind an `ir.*` API.

The split falls out cleanly. Compilation services move to `ir.*`: `parse-spec`, `spec-dep-syms`, `dep-ns-of-entry`, `topo-order-specs`, `load-specs!`, `lowerable?`, `definitional?`, `count-lowered-funcs`, `emit-frame!`, and `compile-program`. What's left is CLI: argument parsing, diagnostics, and exit status. That's the thin `lg.commands.compile` you described, with `scripts/lg-compile` as a shim over the same API.

One question before I move it, because it decides how much this PR carries.

`ir.*` lives under `pkg/rt/core/ir/`, so putting the orchestration there enrolls it in `EmbeddedNSNames` and in genmanifest's `sourceSpecs`. The bootstrap then has to build the driver, and every edit to it churns `pkg/rt/generated.sums`. That mechanical consequence is what drove the original placement, and it's separate from the precedent argument. I take your point that `gogen`'s auxiliary embedding is transitional and shouldn't be copied forward.

Two ways to read "preserve that direction without completing the bootstrap work here":

1. Put the services under `pkg/rt/core/ir/` now and accept both consequences. It's the real destination, and there's no interim state to unwind later.
2. Ship them under the `ir.*` API but keep them auxiliary-embedded for now, with the move into the bootstrapped universe tracked as follow-up.

I lean toward (1) if you're comfortable with the sums churn, since (2) spreads `ir.*` across two embedding universes, which seems like a worse seam than the one you're asking me to remove. Tell me which you want and I'll build it that way.

One constraint has to survive the refactor: gloat invokes `scripts/lg-compile` by path and reads its stdout, so the `EMIT-FAIL <path> pkg=<pkg> returned <type>` line and the process exit code are an external contract. The shim keeps them working, and I'll pin both with a test so a later cleanup can't drop them silently.

Proposed shape, to keep the diffs reviewable:

- **B1** — compilation services behind the `ir.*` API, `scripts/lg-compile` working through the shim, no `lg.aotdriver`.
- **B2** — `lg.commands.compile` and the CLI wiring.
- **Follow-up issue** — `gogen` as the bootstrapped Go-emission backend for the IR pipeline, so the bootstrap and command architecture stays on the roadmap rather than being absorbed into this PR.

I'll also update #596, which still frames the layering as an open question.

mparrett:
@nnunley My previous comment names the wrong destination — I wrote it before reading #786.

The services I listed — `parse-spec`, `spec-dep-syms`, `dep-ns-of-entry`, `topo-order-specs`, `load-specs!`, `lowerable?`, `definitional?`, `count-lowered-funcs`, `emit-frame!`, and `compile-program` — go behind `lg.compiler`, not `ir.*`. `ir.*` is itself scheduled to move to `lg.compiler.ir.*`, so pointing new orchestration at it would have added to the migration rather than followed it.

Withdrawing the follow-up issue I offered as well: #786 covers that ground, and I'll work against it as the architectural reference.

The split I proposed lines up with your migration steps, so I'll frame it that way instead:

- **B1** — step 2: define the `lg.compiler` API boundary and move program-level AOT orchestration behind it. `scripts/lg-compile` keeps working as a shim over that API. No `lg.aotdriver`.
- **B2** — step 5: `lg.commands.compile` as the thin CLI adapter, with the existing script and one-liner invocation left intact.

Steps 3, 4, and 6 are the broader migration, and I'm not proposing to start them in this PR.

One question left, narrower than the one I asked before. Step 4 rules out enrolling `gogen` in the bootstrap universe before the #557 cold/hot fixpoint and determinism constraints are resolved. I read that as scoped to the Go backend rather than to `lg.compiler` itself, which leaves open where the new orchestration namespace sits in the meantime: enrolled in `EmbeddedNSNames` and genmanifest's `sourceSpecs` from the start, or auxiliary-embedded until the migration reaches it.

My default is auxiliary for B1, with tracked removal criteria under your last acceptance bullet, since enrolling now makes every edit churn `pkg/rt/generated.sums` and puts the bootstrap on the hook for a namespace that step 3 will move anyway. That does lean on the mechanism you've said shouldn't become the destination, so it's your call rather than mine to make quietly. Tell me which and I'll build B1 that way.

Unchanged from before: gloat invokes `scripts/lg-compile` by path and reads its stdout, so the `EMIT-FAIL <path> pkg=<pkg> returned <type>` line and the process exit code are an external contract. I'll pin both with a test so a later cleanup can't drop them silently.

nnunley:
Architecture decision: enroll `lg.compiler` immediately in the standard embedded/generated namespace set. Please revise B1 so program-level orchestration lives behind `lg.compiler` (not a standalone `lg.aotdriver` subsystem), and add it to `EmbeddedNSNames` plus genmanifest `sourceSpecs`; no temporary auxiliary-embed exception. Preserve `scripts/lg-compile` as the compatibility shim. Also add the promised contract test pinning `EMIT-FAIL` stdout and its non-fatal exit status, then update onto current `main` and rerun default/bootstrap plus AOT/native-entry checks. I will re-review the exact refreshed head.

mparrett:
> Architecture decision: enroll `lg.compiler` immediately in the standard embedded/generated namespace set. Please revise B1 so program-level orchestration lives behind `lg.compiler` (not a standalone `lg.aotdriver` subsystem), and add it to `EmbeddedNSNames` plus genmanifest `sourceSpecs`; no temporary auxiliary-embed exception. Preserve `scripts/lg-compile` as the compatibility shim. Also add the promised contract test pinning `EMIT-FAIL` stdout and its non-fatal exit status, then update onto current `main` and rerun default/bootstrap plus AOT/native-entry checks. I will re-review the exact refreshed head.

Done as specified, with a caveat.

**Delivered** The driver is now `lg.compiler`, living at `pkg/rt/core/lg/compiler.lg` as an ordinary core namespace. `pkg/rt/aotdriver_src.go` and the `pkg/rt/aotdriver/` tree are gone, so there is no auxiliary-embed exception for the driver (`gogen` remains auxiliary; see below). Both registrations you named come for free from the location: `EmbeddedNSNames` walks `core/`, and genmanifest's `sourceSpecs` already sweeps `pkg/rt/core` for `.lg`, so `generated.sums` and `generated.manifest` now churn on driver edits. This accepts the provenance churn you asked for. `scripts/lg-compile` stays as the CLI shim, unchanged in path, argv, and exit codes.

**The caveat** A core namespace is also a *bundled* namespace, and `lg.compiler` requires `gogen`, `ir.data`, `ir.passes.pipeline` and `ir.passes.entry-frame`. A require drags its transitive closure into the const pool, so enrolling the driver pulled the whole IR pipeline into `core_compiled.lgb`, whose namespace table on `main` lists no `ir.*` at all.

Measured on one machine on 2026-09-02:

| | `core_compiled.lgb` | smoke-boot median |
|---|---|---|
| `main` | 308,175 B | — |
| `lg.compiler` bundled | 1,068,039 B | 23.385 ms |
| with the bundle skip | 308,175 B | 4.060 ms |

`make generate` fails outright at 23.385 ms against the 8 ms budget, so this is not a judgment call about taste. The driver itself is 30 chunks; the 760 KB is entirely its requires.

I resolved it with the mechanism already in the tree rather than a new one. `cmd/lgbgen` has `isIRBundleSkipped`, which keeps `ir.*` out of the bytecode bundle while leaving it in Go lowering, added for this same regression, per its comment: "the bundle had grown 109K→650K". I generalized it to `isBundleSkippedTool` and added `lg.compiler` and its `lg.compiler.*` descendants — the family prefix rather than the exact name, because #786 moves `ir.*` under `lg.compiler.ir.*`, and an exact-name check would then match neither the old name nor the new one and restore this regression silently. That keeps your two requirements intact: enrolled in `EmbeddedNSNames`, swept by `sourceSpecs`, and costs a plain `lg` script nothing at startup, because the driver loads from embedded source on demand exactly as `ir.*` does. **If you would rather** the driver be eagerly bundled and the boot budget be raised to match, say so and I will change it; I didn't think you were asking for it, but could have been wrong.

**One change not specified** `pkg/rt/gogen_src.go` had to lose its `//go:build !bootstrap` gate. lgbgen builds with `-tags bootstrap`, so under that tag `gogen` had no embedded source and `lg.compiler` would not compile: `Can't resolve gogen/ns->go-pkg`. Being resolvable is not being enrolled: gogen is still auxiliary, still absent from `EmbeddedNSNames`, still never precompiled or lowered. `TestLoweringDeterminism`, which #425 named as the reason to keep gogen out of core, passes.

**The contract test** — corrected since I first posted this. As written it did not prove what I claimed. It checked `write-packages!` directly, then separately ran the shim on a *successful* compile, so a nonzero `:emit-fails` never went through the shim at all: reintroducing "exit 1 when `:emit-fails` is nonzero" left the test passing. Thanks for pushing on it.

`TestLgCompileEmitFailContract` now injects a synthetic result through the real shim — stubbing `lg.compiler/main-from-args` and evaluating `scripts/lg-compile` against the stub — so a result carrying three EMIT-FAILs and no `:error` must still exit 0 and print `done.`. It fails against that reintroduced regression and passes without it, which is the only reason to trust it. The `:error` case is pinned the same way, so "exits 0" is not vacuous. The exit policy also moved into a pure `lg.compiler/exit-code`, so the shim has no exit decision of its own.

No `.lg` fixture drives any of this: try/catch, atom, reify, deftype, dynamic `set!`, lazy-seq, a macro-expanded body, and multimethod dispatch all lower now, and a fixture picked for today's gaps would stop testing the path the moment lowering covered it while still looking like it did.

**Checks.** Rebased onto current `main`. `go build` under default and `-tags bootstrap`, plus linux/amd64, js/wasm and plan9/amd64. `pkg/resolver` passes, including the renamed embed test — the driver still resolves with an empty search path, so a released binary needs no checkout. The AOT/native-entry e2e suite passes: `NativeEntry*`, `LgCompile*`, `GogenAotDiff`, `LoweringDeterminism`, `DeftypeSkeleton`. `make check-generated` is green.

mparrett:
@nnunley Refreshed head is `d2d26524`, rebased onto `a6763e77`, with all 15 required checks green.

The rebase carried one real conflict, in `cmd/lgbgen/main.go`: #502 added a `compress` argument to `writeBundle` on the line above the `compileIRForLowering` to `compileToolsForLowering` rename. Both sides are kept.

I also dropped the two provenance-refresh commits and regenerated once against current `main` rather than replaying stale digests. The manifest enrolls `pkg/rt/core/lg/compiler.lg` as generator and input, and picks up `pkg/rt/gogen/gogen.lg`, which the sweep sees now that `gogen_src.go` carries no `!bootstrap` gate.

## On the bundle skip

`pkg/rt/core_compiled.lgb` on this branch is byte-identical to `main`'s: same 311,302 bytes, `cmp` clean.

One further argument for the skip. `ir.*` is already bundle-skipped by this same function, and #786 moves `ir.*` under `lg.compiler.ir.*`, so skipping the `lg.compiler` family puts the driver where the namespaces it will absorb already sit rather than carving a new exception. Enrollment is unaffected: `lg.compiler` is in `EmbeddedNSNames` and in the genmanifest source specs, as you asked.

The offer from that comment stands. If you would rather have the driver eagerly bundled and the boot budget raised to match, say so and I will build it that way.

## Checks

Beyond CI, run locally on the refreshed head: default and `-tags bootstrap` builds, the `pkg/resolver` embed test, and the AOT suite: `TestLgCompileEmitFailContract`, `TestLgCompileEntryFrameOptIn`, `TestNativeEntryASTGate`, `TestLoweringDeterminism` and `TestDeftypeSkeletonNativeLowering`. `make check-generated` is clean and leaves the tree unmodified.

nooga:
@nnunley Everything asked for in the architecture negotiation is delivered, and I verified the load-bearing piece directly: `lg.compiler` is enrolled in `EmbeddedNSNames`/genmanifest with no auxiliary-embed exception, `scripts/lg-compile` is preserved as a compatibility shim, and the EMIT-FAIL contract test (`lg_compile_emit_fail_contract_test.go`) passes all 4 subtests including the exact stdout-shape pin.

The one open item is mparrett's disclosed tradeoff — skipping `lg.compiler.*` from the eager bytecode bundle via `isIRBundleSkipped` to keep boot time under the 8ms gate (bundling it eagerly takes boot from ~4ms to ~23.4ms). He's asked you to weigh in on this twice, most recently on today's refreshed head (`d2d26524`). That's the only thing between this and merge — flagging it here so it doesn't get lost.

pkg/rt/core/lg/compiler.lg
```diff
@@ -0,0 +1,280 @@
+;; lg.compiler — the AOT compile driver, as a namespace.
+;;
+;; #596: this is scripts/lg-compile's body moved into a proper ns so a released
```
nnunley: We need a linter that checks for breadcrumbs and dev logs in comments. This needs to be cleaned up

nnunley (approved):
It needs some tightening for durable comments. I see a lot of devlog style comments throughout, which should be captured outside the code being changed.

mparrett:
> It needs some tightening for durable comments. I see a lot of devlog style comments throughout, which should be captured outside the code being changed. 

Working on this now.

pkg/rt/core/lg/compiler.lg
mparrett: Agreed, and the header you anchored on is the worst of it.

The shape: 178 of the 602 added lines are comments, and two rationales are restated at every site they touch — the bundle-skip/startup-cost argument in five files, "EMIT-FAIL is not fatal" in three. That repetition is most of what reads as devlog.

Cleanup, pushing shortly:

- Change narration moves out of the code and into the PR body, which becomes the squash commit message: "moved from the script", "why this can be a plain ns when the script could not", "an earlier version of this test", and the bundle measurement (308,175 → 1,068,039 bytes, 4.060 → 23.385 ms, one machine, 2026-09-02).
- Architecture goes to `docs/design/go-aot-backend.md`, which mentions neither `lg-compile` nor the driver today: where the driver lives, that core-embedded is not the same as bundled, and the stdout/exit contract external orchestrators depend on.
- The roadmap notes ("TRANSITIONAL", and the argument for matching the `lg.compiler.*` prefix rather than the exact name) go to #786, which is where they stay current as the tree moves.
- One owner per surviving rationale, a pointer at the other sites: the skip policy at `cmd/lgbgen.isBundleSkippedTool`, the exit policy at `lg.compiler/exit-code`, the checkout-free property in the resolver test.

The mechanism comments stay: `lowerable?`'s multi-arity note, `definitional-heads`, `topo-order-specs`' cycle termination, the `:ns` stamp, and the gloat stdout contract. I am also keeping one dated line of the measurement at the skip site, because `isIRBundleSkipped` already carries "~6ms (the bundle had grown 109K→650K)" and cutting mine entirely would leave the new skip less documented than the one beside it. Say so if you would rather that went too.

On the linter: I have `lg lint` on a branch — a report-only static scanner with a pattern library, `--format json` and `--strict` — but its patterns read form trees for expensive shapes, so it cannot see comments at all. A breadcrumb/devlog check needs raw source and would want to cover `.go` as well, so it is a new rule on an existing frame rather than something already written. I will open the scanner as its own PR and take the comment rule as the first addition to it, kept out of this one either way.

mparrett:
Regarding this issue:
> keep boot time under the 8ms gate (bundling it eagerly takes boot from ~4ms to ~23.4ms).

Not flagged as a blocking concern, but wanted to note for posterity, since I think it was a bit of an agent slip, failing to wait for a quiet machine and grabbing numbers that are noisy (not confirmed suspected). This also shouldn't be read as a dismissal of a possible issue.

mparrett:
Correcting my own numbers. The boot measurement I posted on 2026-09-02, and which now sits in `docs/design/go-aot-backend.md`, does not reproduce, and the conclusion I drew from it was wrong.

I rebuilt both arms from the current head and measured them interleaved, 25 pairs each, under two host conditions. Interleaved because sequential arms let host drift masquerade as a difference.

| | bundle | boot median | over the 8 ms budget | median-of-5 gate fails |
|---|---|---|---|---|
| loaded host, skip on | 311,302 B | 2.716 ms | 0/25 | 0.0% |
| loaded host, bundled | 1,097,639 B | 9.243 ms | 22/25 | 99.5% |
| quiet host, skip on | 311,302 B | 1.861 ms | 0/25 | 0.0% |
| quiet host, bundled | 1,097,639 B | 5.974 ms | 1/25 | 0.0% |

Gate-failure rates are the real median-of-5 gate resampled 2000 times from each run's samples.

Three things follow.

**The size claim holds.** 311,302 to 1,097,639 bytes, close to the 308,175 to 1,068,039 I reported; both bundles have grown a little with `main`.

**The relative cost holds.** Bundling costs 3.2 to 3.4 times the boot, stable across both host conditions. That part was never load-dependent.

**The absolute claim does not hold, and it was the load-bearing one.** I wrote that bundling "fails outright at 23.385 ms against the 8 ms budget, so this is not a judgment call about taste." On a quiet host the bundled arm boots in 5.974 ms and passes the budget, failing the simulated gate 0% of the time. I never saw 23.385 ms again under either condition; the worst single sample across 100 runs was 21.0 ms, on the loaded host. Part of the gap is real improvement in `main` since 2026-09-02 — the skip-on arm is byte-identical to `main` and moved from 4.060 ms to 1.861 ms over the same period, which #781 and #624 would explain — and the rest is that the original number came off a host I cannot show was quiet.

So the honest form of the argument is narrower than what I posted. Bundling the driver costs roughly 3.2x the boot of every `lg` invocation, for namespaces a plain script never touches, and at a 5.974 ms median with a 2.6x spread it would flake an 8 ms gate on any loaded host, CI included. That still justifies the skip. It does not justify the claim that eager bundling is infeasible: on a quiet host it fits the budget as it stands, with no raise needed.

Which makes your open question cheaper than I made it sound. If you want the driver eagerly bundled, the cost is a boot regression of a few milliseconds and a gate that needs either a raised budget or a quiescence guarantee, not a categorical failure. My preference is still the skip, on the "pure cost for something never loaded" argument rather than the budget one, but the choice is yours and the earlier number should not weigh on it.

I will correct `docs/design/go-aot-backend.md` once you have said which way you want it, since what that page should claim depends on the answer.

## What bundling would actually pull in

Checking the two artifacts rather than reasoning about them, with `cmd/lgbstat` on the current head:

| | bytes | chunks | consts | `*vm.Func` |
|---|---|---|---|---|
| skip on | 311,302 | 937 | 2,611 | 924 |
| bundled | 1,097,639 | 2,164 | 7,037 | 2,150 |

The driver is 30 chunks. The bundle gains 1,227, so what arrives with it is the require closure: the `ir.*` pipeline and `gogen`.

That reframes your question. It is not whether the driver should be bundled, it is whether the IR pipeline should be — which the `ir.*` skip already answered when that bundle grew 109K to 650K. Eager bundling reverses that decision as a side effect rather than on its merits.

It also puts `gogen` in the bundle. #786 step 4 holds gogen back until the #557 cold/hot fixpoint and determinism constraints are resolved, and I do not want to assume that "precompiled into the bundle" and "enrolled in the bootstrap universe" are the same line — they may not be. Your read on that decides whether this is a real conflict with step 4 or only an adjacent one.

mparrett:
@nnunley This PR is approved, but I think worth a check here. When you have a moment, could you take a look at [my last comment](https://github.com/nooga/let-go/pull/735#issuecomment-5590247176) and let me know if it implies any changes in this PR, a follow-up, or something else entirely? Thanks!

pkg/rt/core/lg/compiler.lg
mparrett: Opened #835 to track the comment linter, rather than folding it into the scanner branch I mentioned.

mparrett: Cleanup landed in `ab699061`. Architecture rationale moved to `docs/design/go-aot-backend.md`, change narration to the PR body, roadmap notes to #786, and each surviving rationale now has one owner with pointers at the other sites. The mechanism comments listed above stayed, including the dated measurement line at the skip site. #835 tracks the linter.

nnunley:
I'm fine with the gown and compiler being in the bundle, and definitely in the broader bootstrap universe. I had removed it because of the overall size bump at the time

mparrett:
Thanks. Since either is acceptable, I am keeping the skip as the PR has it: the cost is the 3.2x boot measured above, for namespaces a plain script never loads. The design-doc paragraph now carries those numbers in place of the withdrawn ones.

Separately, #838 trims the `ir.data` chunks that were in that baseline bundle regardless of this PR.

mparrett:
@nooga — rebased onto main, conflicts resolved, ready for a final look.

Two of them, both introduced by #838:

- `cmd/lgbgen/main.go`: the substantive one. This branch renames `isIRBundleSkipped` → `isBundleSkippedTool`, widening the skip to cover `lg.compiler.*` alongside `ir.*`; #838 rewrote the adjacent Phase-0 block to hoist the ir.data bootstrap into a closure. Kept both: `compileToolsForLowering` now calls `bootstrapIRData()` before compiling the skipped namespaces, so ir consts still land after `writeBundle`.
- `pkg/rt/generated.manifest`: the generator digest for `cmd/lgbgen/main.go`. Refreshed with `make generate` rather than picking a side, since neither side's hash describes the merged file.

The regenerated artifacts are byte-identical; only the provenance digests moved. `make check-generated` is clean and the `lg-compile` contract e2e tests pass.

nooga (approved):

PR #736, opened 2026-08-12
fix(reader): set literals read as sets, not as (hash-set …)

nnunley:
## Summary

`(read-string "#{1}")` returned the form `(hash-set 1)` where Clojure returns the set `#{1}`. Maps and vectors already read as data; sets were the only collection literal that did not.

Stacked on #734.

## Before

```clojure
(read-string "#{1 2}")        ;=> (hash-set 2 1)
(set? (read-string "#{1}"))   ;=> false
(read-string "{:a 1}")        ;=> {:a 1}        ; already data
(read-string "[1 2]")         ;=> [1 2]         ; already data
```

Genuine reader macros were never the issue — `'sym`, `@x` and `#'foo` expand to forms in Clojure too. `#{}` was the outlier.

## Change

`readSet` builds the set value instead of consing `hash-set` onto a list. Evaluation is unchanged: `compileForm`'s `vm.SetType` case already emitted the same `hash-set` invocation and compiled each element, so `#{x (f y)}` still evaluates its elements.

## What that surfaced

Three sites had assumed a set literal always arrives as a list, because before this it always did.

**`ir/build.lg`** — `build-form` threw on the first quoted set it saw:

```
ERROR in test: build-form: unrecognized form #{entry}
```

from `test/gogen_nth_non_int_test.lg`, whose trigger form contains `visited #{entry}`. Adds `build-set`, modelled on `build-vector`, plus the `set?` cases in `captures-of` and `free-vars` so a value captured or free only inside a set literal is still walked — the same reason the map cases exist.

**`ir/lower_go.lg`** — a set reaching `:const` had no Go emitter, so the lowered tree stopped compiling:

```
pkg/rt/core_go_lowered/ir/build/build.go:1872:6: declared and not used: name_sym
pkg/rt/core_go_lowered/ir/build/build.go:1999:1: missing return
```

That is `BuildFn`, because `build-fn*` uses `#{}` as a `reduce` init (build.lg:1533). Adds `boxed-set-expr` → `vm.NewSet`, alongside the existing vector, list and map emitters.

**`core.lg`** needed nothing: the `case` macro already tested `(or (set? test-val) set-literal?)`, so it accepted both shapes.

## Tests

`test/set_literal_reader_test.lg`, 15 assertions across three deftests, pinning the data/eval split in both directions:

- `read-string` returns a set, is `set?`, equals `#{1 2}`, and is not a list; `#{}` reads as an empty set.
- Maps and vectors still read as data; `'sym` and `@x` still expand to forms.
- Unquoted set literals still evaluate their elements (`#{1 2 x}` with `x` bound, `#{(+ 1 1)}`).
- Quoted set literals hold unevaluated elements — `'#{a b}` equals `#{'a 'b}`, and `'#{entry}` contains the symbol, which is the IR case that was previously unreachable.

## Validation

- `make generate` + `make check-generated` clean; lowered tree compiles and dispatches natively under `-tags gogen_ir`.
- `go test ./test/ ./pkg/...` green.
- Bundle and lowered tree regenerated.


mparrett (changes requested):
The representation change exposes three additional traversal gaps:

1. **Deftype/defrecord field references inside sets no longer compile.** `-dt-rewrite` in `pkg/rt/core/core.lg` treats non-sequential forms as atomic. Sets previously arrived as `(hash-set …)` lists, so their elements were recursively rewritten. On this branch:

```clojure
(defprotocol SetFieldP (field-set [this]))
(deftype SetFieldBox [x] SetFieldP (field-set [this] #{x}))
(field-set (->SetFieldBox 7))
```

fails with `Can't resolve x in this context`; the base branch returns `#{7}`. Please add set traversal there and in the native `dt-rewrite-field` equivalent.

2. **Native IR lowering does not macroexpand set elements.** `expand-all` in `pkg/rt/core/ir/passes/pipeline.lg` descends into vectors and maps but not sets. A gogen fixture with `(defn run [] #{(when true 1)})` returns `#{1}` under bytecode but `#{(if true (do 1) nil)}` under native dispatch. The gogen trampoline passes on `main` and fails on this PR. `expand-all` needs a set branch before `build-set` receives the elements.

3. **Syntax-quoted sets bypass unquote and symbol processing.** `syntaxQuote` in `pkg/compiler/reader.go` handles vectors, maps, and lists, but the new `SetType` falls into the default quote branch. ``(let [x 1] `#{~x})`` now yields `#{(unquote x)}` instead of `#{1}`; gensym and namespace qualification inside syntax-quoted sets are likewise skipped.

The regular `go test -count=1 ./test/ ./pkg/...` suite passes and `make check-generated` is clean, so these need focused regression coverage.

nnunley:
## Performance: branch vs main

The ratchet's baseline (`docs/perf/baseline.json`) was captured on go1.26.3 and the current toolchain is go1.26.5, so `make bench-ratchet` reports regressions on `main` itself. A control run on unmodified `main` (af4b7f44) reproduces them with byte-identical allocation counts, which scopes them out of this change:

| benchmark | control = main | this branch |
|---|---|---|
| `pkg/compiler.BenchmarkInitFromLGB` | +42.6% | +33.2% |
| `pkg/ir.BenchmarkIRCompile [bytecode]` | +9.9% | +10.4% |
| `pkg/ir.BenchmarkIRCompile [gogen_ir]` | −13.7% | −23.2% |

Branch measured against `main` on the same toolchain and machine, anchor-normalised:

| benchmark | main × | branch × | Δ vs main | wall |
|---|---|---|---|---|
| `pkg/compiler.BenchmarkInitFromLGB` | 1 307 085 | 1 220 929 | **−6.6%** | 1.36 ms |
| `pkg/ir.BenchmarkIRCompile [bytecode]` | 14 537 718 | 14 607 273 | **+0.5%** | 16.2 ms |
| `pkg/ir.BenchmarkIRCompile [gogen_ir]` | 10 624 979 | 9 461 319 | **−11.0%** | 10.5 ms |

Allocation counts are identical between control and branch — `7028 → 26118`, `100159 → 114256`, `111864 → 151751` — so none of the allocation growth in the ratchet output originates here.

### Interleaved A/B

`scripts/ab_repeat.py --n 7`, profile `pr-fast`, order BH HB BH HB BH HB BH, 53 benchmarks on Apple M3 / go1.26.5:

```
  budget  6%: clean
  budget  8%: clean
  budget 10%: clean
```

Every family median lands within ±0.8% (largest: `VectorCreation/ArrayVector/100` +0.75%, `VectorConj/ArrayVector/1000` +0.71%, `FuncInvoke/Closure` +0.22%). The harness classifies the run as a no-op.

### Limits of the above

`pr-fast` covers `pkg/vm` under `-tags gogen_ir` only. It does **not** include `BenchmarkInitFromLGB` or `BenchmarkIRCompile`, so the two rows above are single-run measurements. Both point in the improvement direction, and neither is confirmed by repetition; I would not claim them as results.

Separately, `docs/perf/baseline.json` needs re-derivation against the current toolchain before the ratchet can gate anything — related to #663.

nnunley:
Addressed in `96647a94`, but the fix is wider than reviewed and I want the shape re-reviewed rather than just the branches.

## Your three items — all three reproduced

| item | site | confirmed |
|---|---|---|
| 1 | `-dt-rewrite` (core.lg) | yes |
| 2 | `expand-all` (pipeline.lg) | yes |
| 3 | `syntaxQuote` (reader.go) | yes — `` `#{~x} `` gave `#{(unquote x)}` |

## Two of them predate this PR

Probing each site with map/set/vector/list literals turned up more than set gaps:

| literal in a deftype method | main | this branch, before the fix |
|---|---|---|
| `[x]` | `[7]` | `[7]` |
| `{:k x}` | **COMPILE-ERROR** | **COMPILE-ERROR** |
| `#{x}` | `#{7}` | COMPILE-ERROR |

`{:k field}` in a deftype/defrecord method has **never** compiled — `Can't resolve x in this context` on `main` today. And `expand-all` returns `#{(when true 1)}` unexpanded on `main` as well; it was simply unreachable while sets read as `(hash-set ..)` lists and took the seq arm.

So item 1's map half and item 2 are pre-existing defects this PR *surfaces*, not causes. Item 3 is genuinely new.

## A fourth site

`dt-rewrite-field` in `pipeline.lg` — the native-lowering twin you referred to as "the native `dt-rewrite-field` equivalent". Its docstring says it mirrors `core.lg`'s `-dt-rewrite`; it mirrored the defect too, with the same `vector?`/`seq?`-only dispatch.

## Why not just add `set?` branches

All four sites enumerate concrete types and treat anything unmatched as an atom. Adding `set?` fixes today's symptom and leaves the mechanism that produced it — the map hole proves it has already failed silently once, for as long as the code has existed.

Each site now dispatches on the collection lattice:

```clojure
(cond
  (symbol? form)     ...
  (map? form)        ... ; associative arms FIRST
  (set? form)        ...
  (vector? form)     ...
  (not (seq? form))  (if (coll? form) (throw ...) form)   ; loud, not silent
  ...)
```

Two details worth checking in review:

- **Associative arms must precede `vector?`**, because a map entry *is* a vector (`(vector? (first (seq {:a 1})))` => `true`). Descending into an entry as a 2-vector rewrites the key as though it were a value. The new tests cover map KEYS specifically.
- **Unknown `coll?` throws.** Same rule as the ops catalog in #268 — a missing facet should be a loud failure, not an invisible nil.

`sequential?` would have been the wrong guard here, incidentally: it is `false` for sets, so a `sequential?`-keyed traversal still skips `#{…}`.

## Regression coverage

Four falsifiers, each failing on `main`:

- `defrecord_field_scope_test.lg` — map values, map keys, set elements, collections nested in a map
- `special_forms_test.lg` — unquote / unquote-splicing / nesting inside a syntax-quoted set
- `expand_all_collections_test.lg` — macroexpansion inside every collection kind

`special_forms_test.lg` on `main`: 25 pass / 3 fail. On this branch: 28 pass / 0 fail.

## Validation

- `make check-generated` clean — bundle, lowered tree, both registrars in lockstep
- `go test ./test/ ./pkg/...` — green, 0 failures
- `make gogen-diff` — `TestGogenAOTDiff` PASS (the native/bytecode differential your item 2 cited)
- `make ir-stress-gate` — 2503/2514 lower natively, 11 failures all within the known baseline bucket

Also corrected two comments in `core.lg` and `pipeline.lg` still asserting that set literals read as `(hash-set ..)` forms.

Kept as a separate commit from the reader change so the representation change and the traversal repair review independently.

mparrett (commented):
All three items from the first round reproduce as fixed. I built `96647a94` and its stack base `ca44b6d7` and ran ~40 probes across both:

| probe | base `ca44b6d7` | branch `96647a94` |
|---|---|---|
| `#{x}` in a deftype method | `#{7}` | `#{7}` |
| `{:k x}` in a deftype method | `Can't resolve x` | `{:k 7}` |
| `` `#{~x} `` | `(hash-set 1)` | `#{1}` |

The claim that the map half of item 1 predates this PR checks out: the base binary errors on `{:k x}` too. `go test -count=1 ./test/ ./pkg/...` is green, `make check-generated` is clean, `make gogen-diff` passes, and every probe returns identical output under `-tags gogen_ir`.

On the wider reshape: the `expand-all` set arm closes a divergence where bytecode expanded a set element and native lowering did not, which is worth having on its own. Three things need another pass — a new regression in the reader, and two in the lattice dispatch.

## Duplicate element forms now evaluate once

`reader.go:819` builds a set of *forms*, so two `=`-equal element forms collapse before compilation and only one of them is ever evaluated:

```
#{(gensym) (gensym)}           base: #{G__1 G__2}   branch: #{G__1}
#{(swap! c inc) (swap! c inc)} base: counter 2      branch: counter 1
```

Clojure's reader rejects this outright with `Duplicate key`, so silently evaluating one of the two is a behavior neither old let-go nor Clojure has. No test covers it. Either throw the way Clojure does, or carry the elements as a slice through the read and build the set at compile time.

## The `coll?` guard cannot fire

`core.lg:2187`, `pipeline.lg:240`, and `pipeline.lg:423` all end the dispatch with `(if (coll? form) (throw ...) form)`. But `coll?` is `(or (list? x) (vector? x) (map? x) (set? x) (seq? x))` (`core.lg:2036`), and every one of those is matched by an earlier arm, since `list?` implies `seq?`. Nothing reaches the throw. I checked the escape hatch too: a record is `coll? false`, so it still passes through as an atom.

So the stated rationale — the next collection type added to the reader fails loudly here instead of silently skipping rewrite — does not hold, because `coll?` would not know that type either. The native `isColl` at `lang.go:2284` tests the `vm.Collection` interface and would catch records and queues, but core.lg's `defn` shadows it. For the loud failure, the fallthrough has to test something other than `coll?`: an allowlist of atom predicates, or the native check.

## Dead `FormSource` call

`reader.go:820` still calls `vm.FormSource.Set(result, ...)`, but `formSourceMap.Set` stores only `*List` and `*Cons` (`source.go:296`), so it is a no-op for `*PersistentSet`. Nothing degrades: error spans inside a set literal still point at the right line, and the frame label improved to "compiling set element". The call and the `startLine`/`startCol` computation above it are now dead.

## Before this can merge

- CI has never run on this PR. `statusCheckRollup` is empty and there are no workflow runs for `96647a94` or the branch, while #734 from the same fork is green across 19 checks.
- `mergeStateStatus` is `DIRTY`. The merge is clean locally because the repo's merge driver regenerates `core_compiled.lgb` and `generated.sums`; GitHub has no driver, so it needs a rebase onto `3d8c9d93`.
- The base is `main`, not `warn-on-core-shadow-fix`, so the PR carries #734's three commits and shows 21 files / +395 where the net change is 10 files / +255. Retarget it, or land #734 first; otherwise the squash message describes the wrong change.

mparrett:
To clarify, since the above reads flatter than I meant it:

- The dedup regression is the blocker. It needs a fix or a decision that let-go diverges from Clojure here on purpose.
- The `coll?` guard I'd either make fire or remove along with the three comments claiming it does. A guard that can't trigger is worse than no guard, because the next person to add a collection type will trust it. Not blocking.
- The `FormSource` call is a delete.

Nothing else in the reshape needs changing from my side.

nnunley:
Second-round findings are fixed at `f01163374123`. Duplicate equal set element forms now fail during reading with `Duplicate key:` rather than collapsing before compilation; regression cases cover scalar, vector, `(gensym)`, and side-effecting `(swap! c inc)` forms. The no-op `FormSource` write/start coordinates are removed, as are the three unreachable `coll?` guards and their misleading loud-fallback claims. Final pre-push suites, IR stress, generated checks, and native/bytecode differential pass.

mparrett (changes requested):
The original findings are fixed, but this new set traversal introduces a scoped native correctness regression.

## `expand-all` collapses distinct source forms after macroexpansion

The new set arm in `pkg/rt/core/ir/passes/pipeline.lg` rebuilds the expanded elements with `(into #{} ...)`. Two distinct source forms are therefore deduplicated if macroexpansion makes them equal, before either form is evaluated.

Reproducer:

```clojure
(defn run []
  (let [c (atom 0)]
    #{(when true (swap! c inc))
      (if true (do (swap! c inc)) nil)}
    @c))
```

`when` expands to the second `if` form. Rebuilding the form collection as a set drops one element, so only one `swap!` remains in native lowering.

Verified with the repository's gogen trampoline:

| revision | bytecode | native | result |
|---|---:|---:|---|
| base `b32acf6` | 2 | 2 | PASS |
| PR head `f011633` | 2 | 1 | FAIL |

This is therefore introduced by this PR, not a pre-existing map/set traversal defect. The simple `#{(when true 1)}` test passes because there is no second form to collide with after expansion.

Please preserve the expanded element sequence until runtime construction—e.g. produce a `(hash-set ...)` call or use another non-deduplicating intermediate representation—instead of rebuilding macroexpanded forms with `into #{}`. A regression test should assert bytecode/native parity and that both side effects occur.

nnunley:
Remediation published at `48188b5158853a2265d701bca3b79c86af11465e`.

`readSet` now routes elements through `appendNonVoid`, so `#?@` splices expand element-by-element, duplicate detection sees expanded forms, and one-shot splicing state is consumed/reset at the set boundary. Regression coverage includes direct splicing, enclosing-vector boundaries, splice-created duplicates, and unmatched-condition state leakage.

Evidence: focused reader tests, compiler suite (61), short suite (1,734), compatibility suite (243), generation/check-generated, gogen-diff, IR stress, and independent re-review passed. The published head now has no failing GitHub checks (14 passed; non-applicable checks skipped).

mparrett:
Rechecked the published head `48188b5158853a2265d701bca3b79c86af11465e`.

That commit correctly addresses reader-conditional splicing inside set literals, but it does not change the blocking `expand-all` set arm in `pipeline.lg`. After rebuilding this exact head, the previously posted collision fixture still reports:

```
bytecode=2 native=1
gogen-trampoline: FAIL
```

The existing changes-requested finding therefore remains open: macroexpanding set elements into `(into #{} ...)` deduplicates distinct source forms that expand identically, dropping an evaluation on the native path.

nnunley:
Restacked onto current `main` (`fc4bec42`) — the branch was conflicting after the base moved. Only `pkg/rt/generated.sums` conflicted, at every commit; resolved by re-running `make generate` per commit bottom-up, never by hand-merging generated artifacts. Zero source conflicts; the four original commits are unchanged in content.

Also added a fifth commit, `test(gold): guard set macroexpansion evaluation` (`053c05f7`), fixing an evaluation-multiplicity bug this branch introduced in `pipeline.lg`'s `expand-all`: rebuilding an expanded set literal with `(into #{} ...)` deduplicates at compile time on *macroexpanded form* equality, silently dropping an element evaluation. Distinct source forms with equal expansions must each still evaluate; only runtime values may dedupe. The fix emits `(apply list 'hash-set ...)` instead.

Verified against JVM Clojure as oracle — for two distinct forms whose macroexpansions are equal, Clojure yields `[2 2 true true]`; before the fix this branch's bytecode gave `[1 1 true false]`. The commit adds a gold parity test (`test/gold/set_macroexpansion_eval.{cljc,out}`), a gogen leg (`test/gogen/set_macroexpansion_eval.lg`, bytecode and native both `[2 2 true true]`), and `test/expand_all_collections_test.lg`.

Validation on the new head: `make generate` twice deterministic, `make check-generated` clean, `go test ./...` all packages pass.

nnunley:
Updated #736 to `ab3c821e` and restacked the full six-commit series onto current `main@dadb8e0b`.

The follow-up replaces equality-bearing host collection forms in the IR expansion pipeline with explicit compiler-only nodes:

- `SetForm.elements` preserves every source element slot through macroexpansion.
- `MapForm.entries` preserves every source key/value slot through macroexpansion.
- IR building evaluates each preserved child exactly once, then constructs the runtime value with qualified `clojure.core/hash-set` / `clojure.core/hash-map`.
- Bytecode intrinsic map literals now also use `hash-map`, consistent with the proposed unordered intrinsic-map contract.

This fixes both reproduced non-injective expansion cases:

- set: bytecode/native `[2 2 true true]`
- map: bytecode/native `[4 2 2 4 true true]`

The constructor-shadowing regression fixture also proves native/bytecode parity with user-defined `hash-map` and `hash-set`: `[1 1 2 true]`.

Validation after the final restack:

- deterministic `make generate` twice, no drift
- `make check-generated`
- focused expansion tests: 4 tests / 18 assertions
- three-fixture bytecode/native trampoline, all native-dispatch
- `make gogen-diff`: 15 full parity, 1 known partial, 0 divergences/fatals
- `go test -race ./pkg/compiler ./pkg/ir -count=1 -timeout 900s`
- `go test ./... -count=1 -timeout 900s`

`make bench-ratchet` continues to report the documented stale Go 1.26.3 baseline failure; no baseline was changed. Relative deterministic deltas remain within the ratchet's 5% budget against the nearby-main control.

mparrett (changes requested):
[P1] Preserve set child slots through field rewriting

At `pkg/rt/core/core.lg:2180`, rebuilding the rewritten elements with `into #{}` deduplicates them at compile time. A bare field `x` and explicit `(.x this)` can be distinct source forms that both perform side effects; after rewriting they become equal and one evaluation disappears. Base `dadb8e0b` returns `2`, while this head `ab3c821e` returns `1` under both bytecode and native lowering.

Please preserve every child slot until runtime construction in both `-dt-rewrite` and the mirrored `dt-rewrite-field` set/map paths. The focused compiler and IR tests otherwise pass. Separately, this PR currently conflicts with the advanced `main` branch and needs rebasing.

nooga (changes requested):
mparrett's Aug 29 P1 still reproduces on HEAD: `-dt-rewrite`'s set arm at `pkg/rt/core/core.lg:2180` and the mirrored `dt-rewrite-field` at `pkg/rt/core/ir/passes/pipeline.lg:236` both still do `(into #{} (map ... (seq form)))`, so two source elements that become equal after field rewriting (bare `x` vs `(.x this)`) collapse and only one is evaluated. It's the same defect class the Aug 15 round fixed in `expand-all` via `SetForm`/`MapForm`; that fix just never reached these two functions.

Please apply the same pattern there (carry a vector until runtime construction, or emit `(hash-set ...)`), add a regression test with two aliasing field elements inside a deftype method body, and rebase — locally the only conflict against main is the generated manifest. Then I'll merge; #675 doesn't conflict with this.

nooga (changes requested):
Reproduced the outstanding dedup bug independently: `-dt-rewrite`/`dt-rewrite-field` in `pkg/rt/core/core.lg` (and the mirrored `pipeline.lg`) still build set elements with `(into #{} (map ... form))`, which dedupes by value. Repro: two side-effecting deftype field forms that alias after rewriting (one bare field symbol, one `(.x this)`) inside a `let [c (atom 0)]` counter — origin/main gives `counter=2`, this branch gives `counter=1` (a `swap!` silently dropped). This is the same defect class flagged in the last review round, recurring in this specific rewrite path.

Also: main has moved since this was written. #770/#854 (merged) added an `r.data`-gated real-set path inside `readSet` for edn/data mode, while this PR makes `readSet` always build a real `PersistentSet` and removes that path outright — two different designs for the same behavior in the same function. Needs a decision on which design wins before a rebase, not just conflict resolution.

Requesting changes: please fix the rewrite-path dedup (same class as the prior round) and reconcile with #770/#854's `r.data` mechanism.

nnunley:
Updated #736 to `e8ba8676`, fixing a lowering-coverage regression this series introduced.

**What was wrong.** The ir-stress lowering-coverage census went from 11 failures to 114 on this
branch, with 95 of the 96 new ones in a single bucket, `:other | unsupported function body shape`.
Bisecting the six commits against the ratchet isolated it exactly:

| Commit | Failures |
|---|---|
| `12921c19` set literals read as sets | 11 (baseline) |
| `e82deb4a` traverse by lattice | 11 |
| `16c301ed` reject duplicate set elements | 11 |
| `16da3acb` splice reader conditionals | 11 |
| `dbd06f2d` guard set macroexpansion | 11 |
| **`5a37f07e` preserve collection cardinality** | **114** |

**Cause.** `quote` is in `build-known-heads`, so `expand-all` rebuilt `(quote datum)` as
`(quote <expand-all datum>)` — descending into quoted *data* and rewriting the datum's own
collections. Once `5a37f07e` moved collection slots into the compiler-only `SetForm`/`MapForm`
records, a quoted map or set came back out as a record. A record has no constant representation,
so the block failed to lower, `lower-body` returned nil, and the whole fn bailed to bytecode.

Every `deftest` in the corpus was affected, because `is` and `testing` embed the quoted source
form in their failure reports. Reduced:

```clojure
(defn f [] '{:a 1})    ; => :fallback "unsupported function body shape"
(defn f [] '#{1 2})    ; => :fallback "unsupported function body shape"
(defn f [] '(a {:b 1})) ; => :fallback "unsupported function body shape"
```

**Fix.** An explicit `quote` arm in `expand-all` that returns the form whole. This matches what
`build.lg` already does for `quote` in both `captures-of` and `free-vars`, and what the
form-heads catalog already declares: *"quote: quoted data is not evaluated"*. The existing guard
on the `expand-fully` result only covers quotes that macros **produce**, never quotes the source
**writes**.

It also fixes an older quiet corruption that predates this series: a quoted `'#{1 2}` expanded to
the list `(hash-set 1 2)` and stopped being a set. That one lowered fine, so nothing caught it.

**Verification.** All three quoted-literal probe shapes now report `:lowered`; the census is back
to 11 failures (2547/2558 lowering natively), all within the known-bucket baseline; `make generate`
and `make check-generated` clean on both the bundle and the `-tags gogen_ir` lowered tree; and both
`go test -short ./test/...` and `go test ./pkg/... ./cmd/...` pass.

mparrett (changes requested):
The previously reported `expand-all` collision and field-rewrite slot loss are fixed, but the current field-rewrite implementation introduces a native-only macro-semantics regression.

## [P1] Keep compiler-only collection nodes out of macro arguments

`dt-rewrite-field` now replaces every map and set literal with `MapForm` / `SetForm` before `build-method-ir-fn` calls `expand-all`. A containing macro is therefore expanded after its collection arguments have stopped being maps or sets, so any macro that inspects the syntax sees compiler records instead of the source literal types.

Reduced fixture:

```clojure
(defmacro literal-kind [form]
  (cond (map? form) :map (set? form) :set :other))

(defprotocol P (probe [this]))
(deftype T [] P
  (probe [this]
    [(literal-kind {:a 1})
     (literal-kind #{1})]))
(defn run [] (probe (->T)))
```

Verified with the repository gogen trampoline and native dispatch:

| revision | bytecode | native | result |
|---|---|---|---|
| parent `5a37f07` | `[:map :set]` | `[:map :set]` | PASS |
| head `e8ba867` | `[:map :set]` | `[:other :other]` | FAIL |

This scopes the regression to the latest PR commit. Please preserve reader-level map/set values until macros have consumed their arguments, and only introduce the slot-preserving compiler representation after that boundary. A native parity fixture for a macro receiving map and set literals inside a deftype method would cover it.

Other validation on `e8ba867` passed: `make check-generated`, `go test -short ./test/... -count=1`, `go test ./pkg/... ./cmd/... -count=1`, and the existing set/map macroexpansion and constructor-shadow trampoline fixtures.

nooga (changes requested):
Every prior blocker (dedup on rewrite, `expand-all` collision, deftype field-rewrite slot loss) was independently reproduced and fixed over the last several commits — good trajectory. But the very latest commit (e8ba867) fixes the field-rewrite dedup by replacing set/map literals with compiler-only `SetForm`/`MapForm` nodes *before* macro expansion, which breaks macros that call `map?`/`set?` on their literal arguments (reproduced with a `literal-kind`-style macro: `dt-rewrite-field` now emits `SetForm`/`MapForm` ahead of `expand-all`).

Requesting changes: keep the literal-type-preserving representation, but delay the `SetForm`/`MapForm` substitution until after `expand-all` has run for that subform — mirror however `expand-all`'s own set arm (which doesn't have this bug) sequences it.

nnunley:
Updated to `9b4254f4`. Both change requests are the same defect, and it's fixed at the source.

`dt-rewrite-field` no longer emits compiler-only nodes at all. It runs over a method body *before* `expand-all` macroexpands it, so anything it substitutes is what a macro receives — that is why `literal-kind` saw records. It now mirrors `-dt-rewrite` exactly: compare the rewritten children against the originals, leave an unchanged literal as the literal it was, and emit a qualified `clojure.core/hash-map` / `clojure.core/hash-set` call only when a field was actually rewritten. `expand-all` still substitutes the slot-preserving forms afterwards, once macros have consumed their arguments — the sequencing @nooga pointed to, and the arm that never had this bug.

`test/gogen/macro_literal_kind.lg` adds the requested native parity fixture. Trampoline over all 11 fixtures, native == bytecode:

```
examples.macro-literal-kind             bytecode=[:map :set]          native=[:map :set]
examples.set-macroexpansion-eval        bytecode=[2 2 true true]      native=[2 2 true true]
examples.map-macroexpansion-eval        bytecode=[4 2 2 4 true true]  native=[4 2 2 4 true true]
examples.collection-constructor-shadow  bytecode=[1 1 2 true]         native=[1 1 2 true]
```

So `[:map :set]` natively, matching the parent-revision row in @mparrett's table — and the cardinality and constructor-shadow fixtures still hold, so removing the records did not reintroduce the dedup bug.

**Correction to my previous comment.** It described introducing `SetForm` / `MapForm` in the field-rewrite path as the fix. That was the wrong design and is exactly what you both flagged; the records now exist only inside `expand-all`, after macroexpansion.

The quote guard described in that comment stands on its own and is unrelated to the field-rewrite path: `expand-all` was descending into `(quote …)` because `quote` is a known head, rewriting the datum's own collections. Bisected across the stack, that cost 95 fixtures in the lowering-coverage census (11 failures -> 114, the `:other | unsupported function body shape` bucket 1 -> 96) starting at `5a37f07e`. Every `deftest` was affected, since `is` / `testing` embed the quoted source form in their reports.

Also verified: `make generate`; `make check-generated` (bundle and lowered tree); `make test` and `go test ./pkg/... ./cmd/...` both exit 0; and the ir-stress census back at 11 failures (2547/2558 lowering natively), all within the known-bucket baseline.

mparrett (changes requested):
Getting closer to no findings, I think!

The fieldless macro fixture is fixed, but the same pre-macro rewrite still changes a literal's type whenever the literal contains a field reference. The new test uses `deftype T []`, so it only exercises the `entries == original` / `elements == original` path.

## [P1] Preserve literal type through macroexpansion when a field is present

Both `-dt-rewrite` and `dt-rewrite-field` replace a changed map literal with a `(clojure.core/hash-map ...)` list before the containing macro is expanded. A macro that inspects its source argument therefore sees a list rather than the map literal it was passed.

Reduced case:

```clojure
(defmacro literal-kind [form]
  (cond (map? form) :map (set? form) :set :else :other))

(defprotocol P (probe [this]))
(deftype T [x] P
  (probe [this] (literal-kind {x 1})))
(defn run [] (probe (->T :field)))
```

Independent results:

| implementation/revision | result |
|---|---|
| JVM Clojure | `:map` |
| `main@9172aa2` bytecode/native | `:map` |
| PR `9b4254f` bytecode/native | `:other` |

This is therefore PR-scoped even though bytecode/native parity now holds. The field rewrite must not consume a source collection before a macro has inspected it; leaving only unchanged literals intact is insufficient. Please extend `macro_literal_kind.lg` with a field-containing map (and preserve the macro boundary before constructorization or field traversal).

The earlier fieldless reproducer and the set/map cardinality fixtures now pass on `9b4254f`.

nnunley:
Rebased onto current main (`369e2a69`) and squashed the seven commits into one, `40b4df88`.

The rebase surfaced one real conflict beyond the generated artifacts: `build-form`'s new `:else` arm threw a bare `(str …)`, which main has since forbidden — #863 converted the core sites to `ex-info` and added `core_throws_ex_info_test.lg` to guard it. That arm now throws `ex-info` with the same `{:fn :type :value}` shape main uses for the pre-existing arm it sits beside. Nothing else changed in the rebase.

Verified on this head: `go test ./pkg/... ./test/` green, `make check-generated` OK, `make ir-stress-gate` at 2572/2573, the known one-failure bucket, not rebaselined.

The commit message now carries the full account — the read-time change, and each traversal and cardinality gap that reading sets as data exposed, including the two that predate this PR (map literals were never rewritten in deftype/defrecord bodies).

---
Delivery verified by local `pr-claim-gate`: exact PR/validated head `40b4df88a3c87eec4f7ac490cfdfeb8399dd7eab`; relevance claims `pkg/rt/core/ir/build.lg` occur in both the delivered delta and the PR feature delta; receipt `/Users/ndn/.local/state/pr-claim-gate/nooga-let-go/pr-736.json`.

Issue #651, opened 2026-07-30
perf(bench-ratchet): seed docs/perf/baseline.json from CI timeline snapshots (retire the local-machine recapture)


nnunley:
## Motivation

`docs/perf/baseline.json` (the ratchet baseline `make bench-ratchet` gates against) is captured **locally on one machine** (`arm64/Apple M3`, 2026-07-10). Refreshing it means a ~25-min quiet-machine `bench-ratchet -full -force update` — a resource burden and the root of #564's "only 52/146 entries have fresh samples." Meanwhile CI **already** captures clean `-full` per-arch snapshots via `perf-timeline.yml` (matrix amd64 ubuntu + arm64 macos-14) onto the `perf-data` branch.

## Key finding — the formats are already identical

A timeline snapshot *is* a `perfdata.Baseline` JSON — `{version:2, machines:{<key>:MachineBaseline}}` with one machine entry; `baseline.json` is the same `Baseline` with N entries. Both use `pkg/perfdata`'s `Baseline`/`MachineBaseline`, and the ratchet only ever compares *within* one machine profile (mixing keys is by design — see the `perfdata.go` doc: *"mixed-hardware in practice … plus amd64 CI"*).

Proven: merging the two `0f3836b7` snapshots (amd64 EPYC + arm64 M1-Virtual) with
```
jq -s '{version:.[0].version, machines:(reduce .[].machines as $m ({}; .+$m))}'
```
produced a valid 2-machine baseline (162 benchmarks each, 3 samples/entry) that `bench-ratchet -baseline <file> show` loaded and gated against with no errors. So no schema work is needed.

## Proposed deliverable

1. **`bench-ratchet seed-baseline`** — new mode in `cmd/bench-ratchet/main.go`:
   - Input: the `perf-data` `timeline/` dir; optional `-sha` (default: newest SHA with snapshots for **all** target machine-keys — a coherent point, not a mix of SHAs).
   - Merge each `<stamp>-<sha>-<key>.json`'s `.machines` into one `Baseline`, write `docs/perf/baseline.json`. Reuse the existing single-machine merge (`update`, main.go:385) and sort keys for deterministic output.
2. **Workflow** — `workflow_dispatch` (or a final job on `perf-timeline.yml`) that clones `perf-data`, runs `seed-baseline`, and opens a PR updating `docs/perf/baseline.json`.
3. Doc note in `docs/perf/ratchet.md` that the baseline is CI-sourced.

## Decisions for @mparrett (perf domain)

[...]

Issue #597, opened 2026-07-21
perf-page: capture the release reference per machine tier so "% vs v1.8.0" is comparable across CI runners


mparrett:
## Problem

The "Are we fast yet?" timeline reports every suite run as a percentage against the v1.8.0 release line, but that release line was captured on a single machine while the timeline points land on a rotating set of CI CPUs. The comparison mixes a real code delta with cross-runner microarchitecture differences, so the headline number is a hardware lottery rather than a measurement.

Two facts drive this:

1. **The v1.8.0 reference is single-machine.** `docs/perf/historical/v1.8.0.json` is schema v1, captured on `arm64/Apple M3`. Every timeline point — all of them amd64 EPYC/Xeon GitHub runners — is compared to that one arm64 capture through `ratio_to_anchor`. That is cross-*architecture*, not just cross-microarchitecture, and `ratio_to_anchor` does not normalize across it (the anchor is a tight ALU loop; the suite is memory/branch/GC-bound, and different CPUs scale the two by different factors, which is the same reason #328 and #329 exist).

2. **The badge is the single latest point.** `cmd/perf-page/main.go` sets the status pill to the latest plotted value vs the reference. It is not a median or a best-of; it is whichever snapshot ran last, on whatever CPU that job happened to draw.

The result on the deployed page: the "End-to-end suite" badge reads "~39% under v1.8.0" not because of a code change but because the most recent snapshot landed on a faster runner tier. The last five suite snapshots, in the chart's own metric (`ratio_to_anchor` vs v1.8.0):

| time (UTC) | CPU | % vs v1.8.0 |
|---|---|---|
| 07-19 01:24 | EPYC 7763 | −27.8% |
| 07-19 01:25 | EPYC 9V74 | −38.9% |
| 07-20 16:42 | EPYC 7763 | −28.4% |
| 07-20 17:27 | EPYC 9V74 | −39.2% |
| 07-20 17:29 | EPYC 9V74 | −41.4% ← current badge |

Consecutive runs of the same code swing 28% ↔ 41% on runner assignment alone. Pooled across all 228 timeline snapshots the two dominant tiers sit at distinct medians: EPYC 7763 at −31% (n=137) and EPYC 9V74 at −43% (n=66). That split is the "bimodal" band structure the chart shows.

## Why the anchor work already merged does not close this

#362 already made the anchor per-machine (schema v2 partitions
[...]

PR #740, opened 2026-08-13
ci(perf): freeze a per-release baseline on every tag, and fix the Pages publish race

nnunley:
## Problem

`docs/perf/historical/` has exactly one file: `v1.8.0.json`. There is no anchor for v1.9.0, v1.10.0, v1.11.x or v1.12.x.

None of the seven perf targets cuts a release anchor:

| target | writes |
|---|---|
| `bench-ratchet`, `bench-ratchet-full` | nothing (check only) |
| `bench-ratchet-update`, `bench-ratchet-full-update` | the rolling `docs/perf/baseline.json` |
| `bench-ratchet-show` | nothing |
| `perf-snapshot` | `docs/perf/timeline/<ts>-<sha>.json` |
| `perf-page` | `docs/perf/index.html` |

So the gate falls back to whatever `baseline.json` happens to hold. Today that is a local-machine recapture at `f154c7fb6bc9` (2026-07-18) — a mid-stream commit between v1.11.1 and v1.12.1, anchored to no release at all. `make bench-ratchet` consequently reports a month of merged work as "regressions" on any branch.

## Change

**`.github/workflows/perf-release-baseline.yml`** — on every `v*` tag, freeze an immutable anchor-normalized baseline at `docs/perf/historical/<tag>.json`, on `ubuntu-latest` with `-full -count 3 -benchtime 2s`. `snapshot` refuses to overwrite, so anchors stay immutable. Lands via auto-opened PR because `main` is protected. `workflow_dispatch` accepts a tag, so past releases can be backfilled.

Capturing in CI rather than on a developer laptop is the point of #651. Two local runs on this machine today drifted the calibration anchor +166% and +86% and disagreed about which benchmarks were regressions, which is what a shared runner avoids.

**`.github/workflows/pages.yml`** — deploy also on `workflow_run` completion of "Perf Timeline". Both currently trigger on the same `main` push, but the timeline job takes ~60 min while the deploy takes ~5, so the published page was permanently at least one snapshot behind. The push trigger is kept for prompt WASM playground updates.

## Backfill feasibility

Verified against each tag's tree:

| tag | `cmd/bench-ratchet` | `make lowered` | backfillable |
|---|---|---|---|
| v1.8.0 | absent | absent | no — the tool postdates the release |
| v1.9.0 | present | absent | yes, with the guard below |
| v1.10.0 | present | present | yes |
| v1.11.1 | present | present | yes |
| v1.12.2 | present | present | yes |

The lowered-runtime step is therefore guarded with `make -n lowered` rather than run unconditionally, so pre-v1.10 tags stay backfillable. At those releases `-tags gogen_ir` matches no files and is already a no-op.

## `v1.8.0.json` stays, with known limits

It cannot be re-frozen under this method, and it differs from what the workflow produces on every axis:

| | `v1.8.0.json` | this workflow |
|---|---|---|
| schema | `version: 1`, flat | `version: 2`, per-machine |
| repetitions | single-shot, no `samples` | `-count 3 -benchtime 2s` |
| machine | Apple M3 laptop, 8 cpu | ubuntu-latest amd64, 4 cpu |
| scope | 75 benchmarks | `-full` |

The single-shot capture is the most likely source of implausible historical deltas, since there is no median and per-benchmark noise is unbounded — the same objection #445 raises about one-shot A/B. Treat `v1.8.0.json` as a progress marker, not a gate reference.

## After merge

Dispatch for **v1.12.2** first: that is the previous release and the one the gate should compare against. Then v1.11.1, v1.10.0, and v1.9.0 to close the gap.

Refs #663, #651, #597, #445.


.github/workflows/pages.yml
```diff
@@ -6,6 +6,16 @@ on:
     branches: [main]
   # Fresh perf-data snapshots are read at deploy time.
   workflow_dispatch:
+  # Perf snapshots are written to the `perf-data` branch by perf-timeline.yml,
+  # which runs on the SAME main push but takes ~60 min — long after this ~5-min
+  # deploy has already fetched perf-data and baked the (stale) timeline into the
+  # static HTML. So a push-triggered deploy is perpetually >=1 snapshot behind.
+  # Re-deploy when "Perf Timeline" finishes, so the freshly-written snapshot is
+  # actually picked up. (Keeps the push trigger for prompt WASM playground updates.)
```
mparrett, line 14: Nice.

.github/workflows/perf-release-baseline.yml
```diff
@@ -0,0 +1,96 @@
...
+          go run ./cmd/bench-ratchet \
+            -full \
+            -count 3 \
+            -benchtime 2s \
+            -timeout 30m \
+            -baseline "$out" \
+            snapshot
+          echo "FROZEN=$out" >> "$GITHUB_ENV"
+
+      - name: Open PR with the frozen baseline
```
mparrett, line 191: Nice.

mparrett (changes requested):
Solid work! Agent review found two issues:

1. **[P1] v1.9.0 backfill always fails.** The workflow invokes `snapshot`, but [v1.9.0’s CLI](https://github.com/nooga/let-go/blob/v1.9.0/cmd/bench-ratchet/main.go#L215) does not support that subcommand. It exits with “unknown mode” before benchmarking. **Recommendation: Use a compatible mode such as `update` with a guaranteed-absent output, or exclude v1.9.0.**

2. **[P2] Reruns do not detect an already-frozen baseline.** Each run checks out the immutable release tag, so a baseline later merged into `main` is never present at the checked-out path. `snapshot` runs again; then the fixed remote branch either rejects the push or a deleted branch permits a duplicate/conflicting PR. **Recommendation: Check `origin/main` for the historical file before running.**

All current PR checks are green. `actionlint` found no workflow syntax errors, only an informational pre-existing `ls` warning in `pages.yml`.

.github/workflows/perf-release-baseline.yml
```diff
@@ -0,0 +1,96 @@
...
+          out="docs/perf/historical/${{ steps.tag.outputs.name }}.json"
+          # snapshot refuses to overwrite an existing baseline, so a re-run for an
+          # already-frozen tag is a no-op failure — intended (baselines are immutable).
+          go run ./cmd/bench-ratchet \
+            -full \
+            -count 3 \
+            -benchtime 2s \
+            -timeout 30m \
+            -baseline "$out" \
+            snapshot
```
mparrett: **[P1] v1.9.0 cannot run `snapshot`.** The workflow advertises v1.9.0 backfill, but that tag's `bench-ratchet` command accepts only `check`, `update`, `show`, `capture`, and `aggregate`; `snapshot` was added later. Dispatching v1.9.0 therefore exits with “unknown mode” before capture. Please use a tag-compatible fallback (for example, `update` into a guaranteed-absent output path) or explicitly exclude v1.9.0.

mparrett (commented):
Adding the two findings inline for precise code context.

.github/workflows/perf-release-baseline.yml
```diff
@@ -0,0 +1,96 @@
...
+            echo "no 'lowered' target at this tag; pre-lowered-Go release, skipping"
+          fi
+
+      - name: Capture immutable per-release baseline
+        run: |
+          set -euo pipefail
+          mkdir -p docs/perf/historical
+          out="docs/perf/historical/${{ steps.tag.outputs.name }}.json"
+          # snapshot refuses to overwrite an existing baseline, so a re-run for an
+          # already-frozen tag is a no-op failure — intended (baselines are immutable).
```
mparrett: **[P2] Reruns cannot see an already-frozen baseline.** Because checkout targets the immutable release tag, this path remains absent even after the generated baseline PR is merged into `main`. A rerun captures again, then either fails while pushing the existing fixed branch or opens a duplicate/conflicting PR if that branch was deleted. Please check `origin/main` for this historical path before capture and exit cleanly when it already exists.

.github/workflows/perf-release-baseline.yml
nnunley: Fixed at `bb64b06f7e00`. Capture now uses v1.9-compatible `update`, guarded by `test ! -e "$OUT"`; the preflight ensures the path is absent, preserving snapshot-style immutability. The v1.9.0 source confirms `update` seeds a missing baseline. A fake-command simulation verified the exact argument vector ends in `update` and refuses an existing output.

.github/workflows/perf-release-baseline.yml
nnunley: Fixed at `bb64b06f7e00`. Preflight explicitly fetches `main` into `refs/remotes/origin/main`, checks `origin/main:<historical-path>`, and also skips when the fixed automation branch already exists. All setup, generation, capture, and publish steps share that skip gate. Temp-remote simulations cover new, already merged, and branch-in-flight cases.

nnunley:
Remediation published at `a788a5f45633d460bd7adfca5589f161d50d9c72`.

- One exact-tag capture now creates the immutable historical snapshot and tighten-only updates the active `docs/perf/baseline.json` with the current-main tool/schema and captured tag SHA.
- Publication re-fetches and branches from `origin/main`, preventing backfills from reverting post-tag history or dropping newer machine profiles.
- Existing open publication skips; orphan/stale branches are rebuilt; closed unmerged PRs are reopened.
- Exact tag refs prevent branch shadowing, and any partial per-package capture warning fails before freezing.

Evidence: actionlint, Bash syntax for every run block, lifecycle state simulations, full test/check-generated gates, schema-v1→v2 merge probe, and independent re-review passed. The published head now has 14 passing GitHub checks and no failures.

mparrett (approved):
Re-reviewed at `a788a5f45633d460bd7adfca5589f161d50d9c72`. Both requested changes are addressed: v1.9-compatible capture/aggregation replaces the unsupported snapshot path, and the preflight now checks current `main` plus in-flight/stale automation state to make backfills idempotent and recoverable. The expanded publication flow branches from current `main`, rejects incomplete captures, and preserves active baseline profiles. All 14 checks pass; no blocking issues remain. Approving.

Issue #97, opened 2026-05-27
Roadmap: Go-package interop (lg-gen, smart wrappers, Mode B AOT lowering)


nnunley:
Filing this as a roadmap for Go-package interop so we can discuss direction before more PRs land. Most of the substrate exists; this is mostly about wiring it together and filling specific gaps.

## Current state

| Capability | Location |
|---|---|
| Callback proxy via `reflect.MakeFunc` (setter shape) | `pkg/vm/func.go:50` `Func.Unbox` |
| Reflection method dispatch (`(.Method obj …)`) | `pkg/vm/boxed.go:45` `Boxed.InvokeMethod` |
| Reflection arg boxing for Go calls (scalars/slices) | `pkg/vm/native_func.go:88` `boxArgForReflect` |
| Generic struct↔Record mapping | `pkg/vm/struct_interop.go` `RegisterStruct[T]` / `ToRecord[T]` / `ToStruct[T]` |
| Go AST construction from Clojure (gogen) | `pkg/rt/gogen.go` |
| `cmd/lginterop` + `scripts/lginterop.lg` (`--smart`/`--skeleton`) | band-next |
| `deps.edn` loader with `:paths` | `pkg/resolver/resolver.go` (#83) |
| WASM build (`.lg` → `.lgb` → embedded Go main → `GOOS=js GOARCH=wasm go build`) | `lg.go:485`, `wasm.go` |
| IR pipeline (typeinfer with lattice, constfold, cse, dce, licm, lower_go) | `pkg/rt/core/ir/` (#94) |

The hand-written `RegisterStruct[…]` lines in `pkg/rt/types.go` are the pain point — six entries today, but the pattern doesn't scale. The point of the roadmap below is to automate that and a few related things, without inventing parallel machinery.

License posture: MIT cleanroom. Glojure (EPL) referenced for architecture only, no source copied.

## Roadmap

**1. Callback coercion in argument position.** Extend `boxArgForReflect` to intercept `target.Kind() == reflect.Func` and emit `reflect.MakeFunc(target, proxy)` directly. The existing setter `Unbox` only works for installing into a function-pointer slot, not for passing a Clojure fn as an argument to a Go function. Single PR, independent of everything else.

[...]

PR #741, opened 2026-08-13
refactor(build): extract the generated-module scaffolding into pkg/gomod

mparrett:
`lg -w` privately owns the logic that scaffolds a throwaway Go module around generated sources before `go build`: pin a released binary's require to its own version (#594), prefer a local checkout in dev builds, fall back to the module proxy when there's neither. The AOT native path (#596) needs exactly those three behaviors, and the module name was baked in — so adding a second caller would have meant a second copy of all of it.

#97 sets the principle for the generator work — avoid two independent generators, factor the internals into reusable pieces first. This applies it before the second copy exists rather than after.

## What changed

The scaffolding moves to `pkg/gomod` with the module name passed in rather than hardcoded. `lg -w` now calls `gomod.Generate(tmpDir, wasmModuleName, version)` and takes the same path it always did — `wasm.go` loses 115 lines and gains 13.

The released-vs-dev decision is split out as `gomod.ReleaseRef`, a pure function, so the branch that #594 fixed can be tested without a network round trip.

Nothing else changes. This is preparation, and it stands alone: one implementation of the module scaffolding is the right shape whether or not the native path lands.

## Why pkg/ and not package main

The helpers could have stayed in `package main` beside their caller. They moved because CI rejects test files at the repo root, so anything in the root package can't carry direct coverage — and this had none. `pkg/gomod` is testable, and the tests below are new.

## Verification

- New `pkg/gomod` tests: the released/dev routing (`ReleaseRef`), the local-source path end to end through `Generate` (module name substituted, `replace` pointing at the checkout, `go.sum` carried across), that the module name is honored for several distinct callers, and that `LETGO_SRC` beats the walk-up.
- `lg -w` end to end on a hello-world `.lg`: builds and emits a 7.7 MB `index.html`, same as before.
- `go vet` clean; builds for `linux/amd64`, `js/wasm`, `plan9/amd64`, `wasip1/wasm`, plus the `bootstrap` tag.
- `generated.sums` untouched; `check-generated` clean.

Groundwork for #596, under #258.

### DIFF
--- a/pkg/gomod/gomod.go
+++ b/pkg/gomod/gomod.go
@@ -1,0 +1,162 @@
+/*
+ * Copyright (c) 2026 let-go contributors; see CONTRIBUTORS.
+ * SPDX-License-Identifier: MIT
+ */
+
+// Package gomod scaffolds the throwaway Go module that `lg` generates around
+// emitted Go sources before handing them to `go build`.
+//
+// It was extracted from the WASM build path, which was its only caller and had
+// the module name baked in. The AOT native path (nooga/let-go#596) needs the
+// same three behaviors — pin a released binary's require to its own version,
+// prefer a local source tree in dev builds, fall back to the module proxy — so
+// this exists to keep one implementation rather than two.
+package gomod
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"strings"
+)
+
+// ModulePath is the let-go module the generated module requires.
+const ModulePath = "github.com/nooga/let-go"
+
+// Files is a generated module's go.mod text and go.sum bytes. Sum is empty when
+// the local-source path supplied no go.sum.
+type Files struct {
+	Mod string
+	Sum []byte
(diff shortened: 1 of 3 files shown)


mparrett:
Self-review: No blocking findings on [[PR #741](https://github.com/nooga/let-go/pull/741)](https://github.com/nooga/let-go/pull/741).

The extraction preserves the prior WASM behavior, including release pinning, local-source discovery, proxy fallback, and Go tool selection. The new module-name parameter is correctly wired as `lg-wasm-app`.

Verification:
- `go test ./pkg/gomod` passes.
- Diff whitespace check passes.
- GitHub’s Go, lint, cross-build, race, generated-artifact, and CodeQL jobs all pass.
- The broader local suite only hit sandbox-related socket permission failures unrelated to this PR.

Minor residual risk: the proxy-backed release/`@latest` branches still lack automated tests because they require network access, but their implementation is effectively unchanged by this refactor.

nnunley (approved):

PR #743, opened 2026-08-14
fix(rt): wake Plan 9 key reads on terminal resize

dharmatech:
## Summary

- add an internal, coalesced wake path to `queuedKeySource` without changing the public `KeySource` interface
- make the Plan 9 terminal source watch the live `/env/WINCH` generation and wake blocked `term/read-key` calls after a resize
- preserve the existing native BEL wake contract, including real-input priority and EOF/error precedence
- document the cross-platform contract and add shared plus Plan 9-specific regression tests

## Root cause

Plan 9 terminals can publish updated `/env/COLS` and `/env/LINES`, but Letgo's queued stdin source previously woke only for keyboard input or EOF. A TUI blocked in `term/read-key` therefore could not observe the new geometry until the user pressed a key.

The Plan 9 adapter now treats a changed `/env/WINCH` generation as the commit signal for updated geometry and queues the same synthetic BEL wake already used by the native `SIGWINCH` path.

## Impact

Letgo TUIs on Plan 9 can redraw while idle immediately after a terminal resize. Resize storms coalesce, a missing `WINCH` environment file remains harmless, and Linux/macOS behavior is unchanged.

## Validation

- `go test -short -count=1 -timeout 300s ./... -skip TestClojureTestSuite`
- `go test -race -short -count=1 -timeout 300s ./pkg/vm/... ./pkg/rt/...`
- `go vet ./pkg/rt/...`
- Plan 9 `amd64` cross-build of both Letgo and the `pkg/rt` test binary
- Plan 9-specific startup-resize test and queued-wake tests, including 100-iteration stress runs
- manual Drawterm/Windows Terminal acceptance with LegMacs: idle resize, maximize/restore, split windows, ordinary input, and clean exit

The complete Plan 9 `pkg/rt` test binary still encounters existing platform failures in loopback-listener and namespace-fixture tests; the resize-specific Plan 9 tests pass.

### DIFF
--- a/docs/design/terminal-resize-wake.md
+++ b/docs/design/terminal-resize-wake.md
@@ -1,0 +1,126 @@
+---
+status: active
+last-verified: 2026-08-14
+authoritative-for:
+  - terminal-resize-wake
+human-verified: 2026-08-14
+---
+
+# Terminal resize wake contract
+
+**Status:** active design for making Plan 9 terminal resizes wake a blocked
+`term/read-key` while preserving the existing native contract.
+
+**Decision:** terminal backends may interrupt a blocked key read with the
+synthetic BEL key (`\x07`). The notification source is platform-specific, but
+the observable `read-key` / `key-pending?` behavior is shared.
+
+## 1. Problem
+
+Interactive applications commonly block in `term/read-key` and check
+`term/size` once per event-loop iteration. Native Unix already wakes that read
+on `SIGWINCH`, but Plan 9's queued stdin source previously woke only for input
+or EOF. Plan 9 terminals could update `/env/COLS` and `/env/LINES` immediately
+while a Letgo application remained parked until the next keypress.
+
+This is a runtime terminal concern, not an application-specific workaround:
+any Letgo TUI needs a chance to observe new geometry without waiting for user
+input.
+
+## 2. Shared contract
+
+The native contract introduced in #165 remains authoritative:
+
+- a terminal resize may make `key-pending?` true;
+- `read-key` consumes the wake as synthetic BEL (`\x07`);
+- real keyboard input takes priority when input and resize are both pending;
+- resize storms may coalesce into one wake; and
+- EOF and reader errors take priority over a synthetic wake.
+
+The application does not interpret BEL as geometry. It handles the event as it
+normally would, then reads `term/size`; a size change triggers its ordinary
+redraw path.
+
+## 3. Platform mechanisms
+
+### Native Unix
+
+The existing backend receives `SIGWINCH`, writes a BEL byte to a self-pipe, and
+polls that pipe alongside stdin. No behavior changes on Linux or macOS.
+
+### Plan 9
+
+The Plan 9 root key source wraps `queuedKeySource` and lazily watches the live
+`/env/WINCH` generation. Terminals publish `COLS` and `LINES` before advancing
+`WINCH`, so a changed generation is the commit signal that new dimensions are
+ready. The watcher then queues an internal wake on the source.
+
+The watcher uses a 100 ms ticker. It is event polling, not a busy loop: the
+goroutine sleeps between reads and reads one small environment file per tick.
+Notifications coalesce in the queued source, so resize drags cannot create an
+unbounded backlog.
+
+If `WINCH` is absent, the watcher quietly waits for it to appear. Input, EOF,
+and `term/size` retain their previous behavior.
+
+## 4. API boundary
+
+`KeySource` remains the public two-method interface: `ReadKey` and
+`KeyPending`. Adding a public `Wake` method would break existing embedders and
+would mix this focused terminal fix with the separate source-lifecycle work in
+#498.
+
+Instead, `queuedKeySource` owns an unexported, coalesced wake flag and method.
+The Plan 9 adapter can use that implementation detail without exposing a new
+host API. WASM's SharedArrayBuffer wake protocol remains separate future work.
+
+## 5. Lifecycle
+
+The Plan 9 source remains the single process-lifetime root over `os.Stdin`, the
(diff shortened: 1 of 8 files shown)


dharmatech:
Video demo:

https://youtu.be/Yy8GPr-0MOM

- Shows legmacs running on plan9 connected via Windows Terminal and drawterm.
- Shows that legmacs properly resizes when Windows Terminal is resized.

mparrett:
@dharmatech Thanks for the PR. Will take a look soon!

pkg/rt/term.go
```diff
@@ -57,7 +57,7 @@ func setupWinch() {
 			for range sigCh {
 				// One BEL per SIGWINCH. Storms collapse — the read side
 				// drains whatever's pending in a single Read regardless.
-				_, _ = w.Write([]byte{0x07})
+				_, _ = w.Write([]byte{terminalWakeKey[0]})
```
mparrett, line 60: Nit (this wording is pre-existing, but the new abstraction makes it more noticeable): now that the wake value is centralized as `terminalWakeKey`, could the implementation comments in this file use that abstraction consistently—perhaps “terminal wake byte (BEL)” where the concrete encoding matters? Several comments still describe the implementation directly as returning or emitting BEL.

docs/design/terminal-resize-wake.md
```diff
@@ -0,0 +1,126 @@
...
+### Native Unix
+
+The existing backend receives `SIGWINCH`, writes a BEL byte to a self-pipe, and
+polls that pipe alongside stdin. No behavior changes on Linux or macOS.
+
+### Plan 9
+
+The Plan 9 root key source wraps `queuedKeySource` and lazily watches the live
+`/env/WINCH` generation. Terminals publish `COLS` and `LINES` before advancing
+`WINCH`, so a changed generation is the commit signal that new dimensions are
```
mparrett: Could we identify which Plan 9 terminal integrations guarantee this `COLS`/`LINES`-before-`WINCH` ordering? The current wording makes `/env/WINCH` sound like a universal Plan 9 terminal contract, while the implementation treats it as optional. Naming the verified producers would make the support boundary clearer.

docs/design/terminal-resize-wake.md
```diff
@@ -0,0 +1,126 @@
...
+- wake bursts coalesce;
+- real input is returned before a simultaneous wake; and
+- the wake remains pending after that real input is consumed.
+
+A Plan 9-only test also moves the `WINCH` generation after root-source
+construction but before the first `ReadKey`, proving that the startup
+correction is not lost.
+
+CI runs the ordinary runtime suite and race detector on supported native
+hosts, while a Plan 9 cross-build covers the build-tagged adapter. Interactive
```
mparrett: The current Go workflow does not appear to include a Plan 9 cross-build job. Should this say that the PR was manually cross-built for Plan 9, or should the cross-build be added to CI?

mparrett (commented):
Code looks solid, thanks! I left a few small comments for consideration while the rest of the review is conducted.

mparrett:
@dharmatech Quick consolidated review/nudge when you get a chance: the code looks solid and CI is green. The one thing I’d like addressed before merging is the new design doc’s statement that CI runs a Plan 9 cross-build—the current workflow doesn’t appear to include one, so that should either say the cross-build was performed manually or be backed by a new CI lane.

Clarifying which Plan 9 integrations provide the `/env/WINCH` ordering would also be helpful. The `term.go` wording comment is just a non-blocking cleanup nit. Thanks!

docs/design/terminal-resize-wake.md
dharmatech: Hey @mparrett 🙋‍♂️

The design doc has been updated to reflect that things were manually cross-compiled.

docs/design/terminal-resize-wake.md
dharmatech: Hey @mparrett 🙋‍♂️

`/env/WINCH` is not a universal Plan 9 terminal facility.
It is currently implemented and verified by the Windows `-G` console path in my drawterm fork,
which publishes `COLS` and `LINES` first and then increments `WINCH`.

The `let-go` side is intentionally producer-agnostic and backward-compatible: when `WINCH` is absent, behavior remains unchanged; terminal integrations that provide the same ordering can wake a blocked read-key so applications can observe the new size. The verified use case is legmacs running through drawterm in Windows Terminal, where this allows it to redraw on an idle resize instead of waiting for another keypress.

I updated the design document to name and link the verified producer and to clarify this support boundary.

dharmatech: @mparrett 

So, this case of "drawterm to plan9 from a Windows (or mac or linux) terminal and run a tui program" is,
as far as I know, *highly unusual* in the plan9 world.
But, awhile back I tried it (with `xsofy`) and to my astonishment, it kinda worked!
So, this work is basically taking this experiment the full distance.
It's saying "let's actually make this work, because it's pretty cool".
What it took to make this happen is basically the patches to `let-go` I share here as well as a modified `drawterm` (my fork).

So, I didn't mean to give the impression that this sort of thing is idiomatic in the plan9 world.
It's definitely very strange but also very cool.

dharmatech: In an ideal world, the upstream drawterm project (from the 9front team) would support this convention.
However, I'm not sure they're interested in this; I haven't discussed this with them.

mparrett: > So, this work is basically taking this experiment the full distance.

Love it, thanks. And appreciate you clarifying the aspects that clearly sit outside my realm of awareness.

mparrett (approved):

mparrett:
All findings have been addressed and checks are green.

PR #744, opened 2026-08-14
glplat: Ebitengine backend behind -tags glplat_ebiten (stacked on #392)

mparrett:
**Stacked on #392.** This branch carries #392's commits rebased onto current main (one `.gitignore` conflict, both sides kept) plus two of ours on top; only the tip two commits are for review here. It will be rebased onto whatever #392 lands as, and closed instead if the commits get folded into #392 directly. Draft until then.

A second `registry.Backend` implementation, on Ebitengine instead of GLFW/OpenGL. Two motivations: prove the backend seam with an independent implementation while #392 is still a draft and the interface is cheap to change, and get the two capabilities the GL backend cannot offer — the browser (ebiten's js/wasm target; the GL backend is cgo+GLFW, native only) and modern native APIs (Metal on macOS, where the GL backend rides deprecated OpenGL 4.1).

[Ebiten](https://github.com/hajimehoshi/ebiten) owns the main thread, which inverts glplat's caller-owned frame loop. The backend absorbs the inversion so the API and game code see no difference: `main` routes through a boot trampoline (`glplat.EbitenBootMain`), the lg program runs on a goroutine, and `EndFrame` hands each frame's batches to ebiten's `Update` over an unbuffered channel, which doubles as frame pacing. MVP matrices are applied on the CPU since ebiten's `DrawTriangles` is 2D screen-space; rendering is at device scale to match the GL backend's framebuffer resolution.

Known limitation: ebiten has no depth buffer, so batches draw in submission order. That is exact for the top-down xsofy frontend (coplanar geometry under an ortho matrix) but mis-sorts first-person-mode geometry; a CPU back-to-front sort for perspective batches is the planned follow-up.

Verified (2026-08-14):

- The unmodified xsofy GL frontend runs on this backend: `XSOFY_GL_DUMP` output byte-identical to the GL backend and deterministic across runs; screenshot pixel diff 0.3%.
- CPU/memory parity over a 10s run; peak RSS flat from 30s to 60s (no leak), constant ~11MB above the GL backend.
- `GOOS=js GOARCH=wasm` builds with the tag and a probe page running an lg draw loop renders correctly in Chromium — the blocking `EndFrame` contract survives wasm because the Go scheduler parks the blocked lg goroutine while ebiten's requestAnimationFrame loop runs.
- Default (untagged) build, `-tags glplat`, and GOOS=linux/plan9/js builds are all unaffected.


nnunley (approved):

nooga:
Delta itself is clean (tag-gated deps confirmed absent from the default build, both build matrices pass, nnunley approved). Parked only because it's stacked on #392, which still has an open P1. Once #392's fixes land, rebase or fold these two commits in and this is a quick approve.

nnunley:
Architecture dependency update (2026-09-07):

#744 remains blocked behind #392, but the intended direction is now explicit: use the merged #572 `surface` host-capability model, put headless font support behind a separate opt-in `glplat_fonts` tag, and define the backend-independent frame/window/input/lifecycle contract before landing either backend.

Please do not treat this as a request for a mechanical rebase yet. #392's unresolved safety/API findings still need fixes after the re-scope, and this Ebitengine backend should be evaluated against the resulting #572-based contract rather than the current `glplat` registry contract.

nooga:
Checked whether the CONFLICTING flag here is just generated-artifact drift (per #747) or a real conflict — ran `git merge-tree --write-tree origin/main` locally against this branch. It's a real conflict: `lg.go` has a genuine content conflict against current main (`.gitignore` auto-merges fine). Needs a manual rebase, not just a regenerate-and-recommit.

nooga:
Not treating this as an independently reviewable PR — it carries #392's own commits verbatim (registry.go, glplat.go, font.go, native backend, interop_glplat.go), so it has no independent existence from #392 right now. #392 is open with its own unresolved P1 and a confirmed conflict, and there's a post-approval architecture-pivot comment (2026-09-07) suggesting this should be re-evaluated against the newer host-capability contract (#572) rather than mechanically rebased once #392 moves.

This PR's own ebiten-only commits are cleanly isolated behind `-tags glplat_ebiten` and build standalone (`go build -tags glplat_ebiten ./...` succeeds on the PR's current head, both natively and cross-compiled) — low risk whenever it does land. Parking this until #392's fate (merge, fold, or re-scope) is settled; will re-review against the current contract at that point rather than doing a mechanical rebase now.

nooga (commented):
Re-reviewing rather than treating the existing approval as still current: real functional commits landed about a month after nnunley's 2026-08-16 approval, including two fixes on 09-02/09-03 — a thread-safety fix for concurrent font glyph rasterization (bdc771d) and a buffer-bounds guard against a C-side texture upload over-read (6dd44bb) — plus further build-tag gating work through 09-13.

Read both fixes: they look correct and are well-tested (dedicated tests, `-race`-covered). CI is fully green across build/race/wasip1/tinygo-wasi/no-http/CodeQL. This is exactly the kind of change (thread-safety, buffer bounds) that warrants a second pair of eyes even with passing tests, so flagging explicitly rather than assuming the original approval still covers it.

No issues found on read — will approve once someone (happy to do it myself) gives those two specific commits a focused look. Given #392 shares this registry-backend design and is currently blocked on its own CI failures, landing this one first seems right.

mparrett:
@nooga @nnunley — rebased onto current main at `c54f73fc` and regenerated. `make check-default-deps`, `make check-generated` and the `-race` glplat tests pass locally; CI is running.

The CONFLICTING flag was the #747 generated-artifact driver gap rather than a content conflict. The only files both sides touch are `pkg/rt/generated.manifest` and `pkg/rt/generated.sums`, which the repo's merge driver resolves and GitHub does not apply; `git merge-tree --write-tree origin/main origin/glplat-ebiten` produced a clean tree locally before the rebase. The PR reads MERGEABLE now.

@nooga, for the focused look you offered: the two commits are `bdc771d` (per-font glyph serialization) and `6dd44bb` (pixel-buffer length guard before the texture upload). Both carry their own tests and run under `-race` in CI.

## What the merge order decides

`pkg/glplat` does not exist on main. Every file here is an add, `internal/registry/registry.go` among them, so merging this PR creates the package and makes the registry the seam that ships.

That is the first of the two questions §5 of the contract doc asks, and both are still open: whether to replace glplat's registry with an `rt.Display` capability bound at `*display*`, and whether C1–C15 stand as written. #874 merged the proposal rather than a decision on it. The 2026-09-07 direction also asked that the contract be agreed before either backend lands, which cuts against landing this one now.

My preference is to land #744 with the registry explicitly provisional and carry the migration in #392's re-scope. The ebiten backend is tag-gated and the registry is `internal/`, so neither is public API yet, and moving both later is a mechanical change rather than a compatibility break. That keeps the green work moving and fixes the `check-default-deps` failure blocking #392, which this branch already carries.

Deciding the seam first and rebasing onto whatever it adopts is the alternative, and I would rather take that than have the merge answer §5 by default.

Either way it needs a sentence from you both: that the registry lands provisionally, or that #744 waits.

mparrett:
Recording the answers, which came in on Discord today.

**nooga:** point 1 yes, land it provisional and document it; points 2 and 4 agreed; point 3 no preference.

**nnunley**, on the larger design: the UI is a scene graph, the surface is the capability a client exposes for presenting a rendered frame, and a renderer sits between them. Whether that renderer is a GL pipeline, a software rasteriser, or a terminal cell mapper depends on what the client supports, so 3D is a renderer choice, not a kind of surface. On the §2 question: the seam holds as long as a display does not imply a window and a GL context, and if C1–C15 encode that assumption anywhere, that is the one amendment the scene-graph direction wants.

C1–C15 and the `Backend` interface, checked against that condition:

- **GL context:** nothing in C1–C15 assumes one. C4 chose painter's order over a depth buffer so a backend without GL qualifies, C15 makes thread ownership a backend property, and the Ebitengine backend in this PR (Metal on macOS, wasm in the browser, no GL API in the contract) is the existence proof.
- **Window:** the rules do not assume one either. C6 and C7 use "window" only as the name for the presentation target's logical size, which a terminal has too. The assumption lives in the interface instead: `Init(width, height, title)` takes a window title, and `PollEventsWindow`, `WindowSize`, and `ShouldClose` are named for one.
- **Renderer:** C1–C5 do assume a triangle-submitting renderer with textures and a matrix. In the three-layer model that is the middle layer with presentation bundled in. A terminal cell mapper would implement it as a software rasteriser whose presentation target is the terminal.

The amendment the scene-graph direction wants is therefore to the seam's shape, not to any rule: the presentation target should be able to be a bound `*surface*` rather than the backend's own window, and `Init`'s title and the `Window*` names should generalise with it. That is a §2 change for #392's re-scope, and landing the registry under `internal/` keeps it open at no cost. Nothing in C1–C15 changes now.

Done on this branch, now at `405cc413`:

- Rebased onto current main (17 commits, the CONFLICTING flag was `.gitignore` against #889's `/perf-page` line plus the generated files the merge drivers resolve). `make check-default-deps`, `check-generated`, the `-race` glplat tests, the linux/wasm/plan9 cross-builds, and both tagged builds pass locally; CI is running.
- One docs commit per nooga's point 1: the package comment marks `glplat` provisional, and `docs/design/glplat-backend-contract.md` gains a §5.1 recording the answers above and the C1–C15 check.

Remaining:

1. @nooga's focused look at the two safety commits (`f40688bc` per-font glyph serialization, `5b2e3749` pixel-buffer guard, same content as before the rebase) and an approval. nnunley's 08-16 approval predates eleven of the sixteen commits here.
2. Squash merge; #392 rebases onto main and becomes the re-scope PR, carrying the seam amendment above, C3, C4, and C10, and the three open P2s.

nnunley:
Yes to all four: land this with the registry explicitly provisional under `internal/`, both backends behind their tags and the default build carrying only the `glplat` namespace; the `rt.Display` / `*display*` migration in #392's re-scope after this merges; C1–C15 standing until amended by number, with C3, C4 and C10 implemented in that re-scope; and `glplat` called experimental with an unstable contract in the release notes.

Quick version of the larger design. The UI is a scene graph; a surface is the capability a client exposes for presenting a rendered frame — the seam #572 shipped, a pixel-buffer sink bound like `*out*`, with a window, fb0, a WASM canvas or a terminal image sink each being one implementation. Rendering sits between them: GL through glplat, a software rasteriser, or a cell mapper for a terminal that advertises the right requirements, so the same scene reaches a terminal client without a separate path. 3D isn't a different surface; it's a renderer, and a GL context is something a client may provide.

For §2: `rt.Display` sits beside the surface, not above the scene — frame, window, input and lifecycle on one side, presentation on the other — so it doesn't change the seam. What remains on #392 is the three P2s (texture Y orientation, `WindowSize`, `Terminate`) and tests, as tracked in nooga's 09-16 review.

nnunley (approved):

Issue #747, opened 2026-08-16
generated.manifest is authoritative but has no merge driver, and the sums driver will hash it mid-conflict


mparrett:
Follow-up from #641, found while rebasing #727 and #692 onto it.

`.gitattributes` gives `pkg/rt/core_compiled.lgb` a `merge=lgb` driver and `pkg/rt/generated.sums` a `merge=sums` driver. #641 made `pkg/rt/generated.manifest` the authority that `generated.sums` is now merely a digest *of* — but the manifest itself is tracked with no driver, so the derived artifact has automated conflict resolution and its source of truth does not.

## The sharp edge

`scripts/git-merge-sums.sh` runs `go run ./cmd/check-generated -o "$2"`, and `genmanifest.Compute` is `hashFile(DepManifestRelPath)` — it never parses what it hashes. So a manifest carrying conflict markers produces a clean-looking digest:

```
$ printf '<<<<<<< ours\n=======\n>>>>>>> theirs\n' >> pkg/rt/generated.manifest
$ go run ./cmd/check-generated -o /tmp/sums.txt
check-generated: wrote manifest digest 3f7cfc362cad
$ echo $?
0
```

`ReadDepManifest` already rejects those lines — they are not 4-field records, so the staleness path fails loudly. Only the digest path is blind, and it is the one the merge driver calls.

That inverts #641's own stated principle. Its body says it "reports failed staleness queries as errors instead of treating them as an empty clean set"; hashing a conflicted manifest is the same class of silent pass.

## The other half, which is a design call

A rebase across #641 now needs `make generate` even when the generated outputs come out byte-identical, because the manifest records a SHA-256 per declared input and any input edit restages it. Both rebases hit this: the `sums` driver fired, the rebase reported clean, and `make check-generated` then failed naming `core_compiled.lgb`, `core_go_lowered/` and `zz_primitives_generated.go` stale. `make generate` fixed both, and in each case only the manifest and digest changed.

CI catches this (`make check-generated-manifest`), so nothing merges broken. But the branches currently show green from a pre-#641 base, and GitHub does not re-run on a base change, so the red only appears on the next push. Cross-referencing the manifest's 276 declared inputs against the 29 open PRs: 18 touch at least one, and 11 carry no digest today — #745, #743, #501, #730, #698, #717, #723, #733, #624, #683, #699.

[...]

PR #748, opened 2026-08-16
fix(genmanifest): validate the dependency manifest before digesting it

mparrett:
`genmanifest.Compute` was `hashFile()` over `pkg/rt/generated.manifest`, so it accepted any bytes as a manifest. `scripts/git-merge-sums.sh` resolves `generated.sums` by calling through `Compute`, which means a rebase or merge that left conflict markers in the manifest still minted a clean-looking digest:

```
$ printf '<<<<<<< ours\n=======\n>>>>>>> theirs\n' >> pkg/rt/generated.manifest
$ go run ./cmd/check-generated -o /tmp/sums.txt
check-generated: wrote manifest digest 3f7cfc362cad
$ echo $?
0
```

`ReadDepManifest` already rejects those lines, since they are not 4-field records, so the staleness path fails loudly and only the digest path was blind. It is the one the merge driver calls. This routes `Compute` through the same parse before hashing.

That restores for the digest what #641 already states for staleness queries: report the failure rather than treat it as a clean result.

## Scope

This is the narrow half of #747. Whether `generated.manifest` should get its own `merge=manifest` driver is the open question there, and it is a design call — the manifest carries output-readiness records, so a driver that regenerates it mid-merge could assert outputs are ready when they have not been rebuilt. A manifest containing `<<<<<<<` is invalid under any answer to that, which is why this piece can land on its own.

## Verification

- `TestComputeRejectsConflictedManifest` fails on `main` (digests the conflicted manifest and returns `6a10250effb2…`) and passes with the change.
- `go test ./pkg/genmanifest/` passes.
- `make check-generated` passes.
- Builds clean for `linux/amd64`, `js/wasm`, `plan9/amd64` and `wasip1/wasm`.

The manifest and digest move in this diff because `pkg/genmanifest/manifest.go` is itself a declared generator input for `core_compiled.lgb` and `core_go_lowered/`, so editing it restages both.


pkg/genmanifest/manifest.go
```diff
@@ -145,7 +145,17 @@ func SourceFiles(repoRoot string) ([]string, error) {
...
 // The manifest-based digest is written deterministically by WriteDepManifest.
+//
+// The manifest is parsed before it is hashed. Hashing bytes alone would accept
+// a manifest that is not a manifest — the case that matters is a merge or
+// rebase leaving conflict markers in it, because scripts/git-merge-sums.sh
+// resolves generated.sums by calling through here. Without the parse, that
+// driver mints a clean-looking digest over a file git has marked as unresolved,
+// and the digest is the artifact later checks trust.
 func Compute(repoRoot string) (string, error) {
+	if _, err := ReadDepManifest(repoRoot); err != nil {
```
nnunley, line 156: Should you cross validate that the manifest not only parses, but that the hashes match what's on disk? That seems like an even stronger check.

If the issue is that there's left over conflict markers, then it's actually unsafe to trust it when there could be orphaned hashes from an allowed merge

mparrett: Heads up before you read this: I'm out this evening, so what follows is Claude's analysis run against the branch rather than mine. The findings are checked against the code, but the scoping call at the end is a proposal, not a decision — I'll DM you tonight to settle it.

The hash cross-check can't go in `Compute`, but your second paragraph found something the PR misses.

`Compute` is what the merge driver calls: `git-merge-sums.sh` runs `check-generated -o`, which lands in `Compute`, and it runs mid-merge — sources have just changed and the manifest hasn't been regenerated yet. Stale hashes are the expected state at that moment, and recomputing the digest afterward is the driver's job. A freshness assertion there would fail the merges it exists to serve.

I'd differentiate well-formedness versus freshness. Parsing is a total invariant: a valid manifest parses at any point in the workflow, so it's safe to assert anywhere, which is why it's in `Compute`. Hash-match-disk is state-dependent and legitimately false mid-merge. We do check it a layer up — the default `check-generated` compares recorded input hashes against disk, and that's the check that caught #743 and turned main red.

Your orphaned-hash case is sharper than the conflict-marker case I wrote this for. Conflict markers are loud: git marks the file unresolved. The quiet failure is a manifest that merged cleanly and is wrong anyway. Two things there:

`generated.manifest` has no merge driver. `.gitattributes` covers `core_compiled.lgb` (`merge=lgb`) and `generated.sums` (`merge=sums`), but not the manifest, so it falls back to git's default text merge. It's a sorted, one-entry-per-line file — the shape that auto-merges without conflict whenever two branches touch different lines. The artifact most able to merge cleanly into a wrong state is the one without a driver.

`ReadDepManifest` has no duplicate detection; it appends every line it parses. Downstream, `staleOutputs` builds `recordedOutputSums` as a map, so if a clean merge lands both sides' line for the same `(output, input, kind)` key, the parse passes and the map keeps whichever comes last. Nothing sees it.

Rejecting duplicate keys is a total invariant like the parse, so it belongs here. The merge driver is the bigger fix and a different change, better filed separately than grown into this PR — but that's the piece I want your read on, since you're often usually running these workflows locally.

mparrett:
@nnunley I'm finally getting to taking a proper look with my meat brain. Apologies for the heavy and less helpful clanker reply above.

So... I think your catch was a good one. Parsing doesn't prevent orphaned hashes surviving a merge. That half is now pushed along with a test. Before this a duplicate quietly resolved to whichever record came last.

Still open from before this PR is the merge driver question expressed in #747. The full hash-versus-disk check was deliberately left out here. Asserting freshness in Compute, which the merge driver calls, would fail merges the driver exists to serve, as far as I can tell. And one layer above, the staleness path already does the comparison.

My borked merges looked like carelessness, and I'll own them. It was easy to merge against green checks, but they were computed against a base that no longer existed. And moving fast + distracted + bots makes that easy to miss, since nothing on either PR could have shown it.

With velocity increasing, should we be able to merge stale but green PRs? Some notes/discussion in #751. We should consider whether the settings-based approach described there is what we want, and if all of this is the right shape.

My personal lean on this, overall, is making green PRs safe to merge. Easier said than done when there's a lot landing concurrently. Curious to hear your thoughts.

pkg/genmanifest/manifest.go
nnunley: That sounds right. I'd argue that a read only verification that all paths in the final product, post parse, exist on disk at the very least. If the hashing method is fast enough, it should add seconds to the check.

A fast follow of adding the gitattributes merge handler seems like a right action.

pkg/genmanifest/depmanifest.go
```diff
@@ -569,7 +570,18 @@ func ReadDepManifest(repoRoot string) ([]HashedEdge, error) {
...
+		edge := Edge{fields[0], fields[1], fields[2]}
+		// This file has no merge driver and is one sorted record per line, so a
+		// clean text merge can land both sides' record for the same edge. Every
+		// downstream lookup is map-keyed and keeps whichever record came last,
+		// so the conflict would resolve silently to one side. Refuse instead: no
+		// digest should be taken over a manifest recording two hashes for one input.
+		if _, dup := seen[edge]; dup {
+			return nil, fmt.Errorf("duplicate manifest entry for %s %s %s",
+				edge.Output, edge.Input, edge.Kind)
+		}
```
nnunley, line 582: This only reports the first duplicate. It would probably be better to have a report of all duplicates, since that could help signal something deeper went wrong with the merge. That said, if this is just triggering a rebuild, i guess the fail fast is fine.

nnunley (approved):
Minor comment left on the duplicates reporting, but I think it's fine as is.

pkg/genmanifest/depmanifest.go
mparrett: Good catch. Happy to do it as a follow-up.

PR #749, opened 2026-08-16
fix(build): refresh the dependency manifest after #733

mparrett:
`main` is red at 1ef26f9d. `check-generated-manifest` reports three outputs stale and both `generated-artifacts` and `build` fail:

```
pkg/ir/op_generated.go
pkg/rt/core/ir/data/generated.lg
pkg/rt/ir_bridge_generated.go
```

The run before it, a9936cd4 (#741), was green.

## Cause

#733 edits `scripts/generate.lg`, which #641's manifest declares as the *generator* for those three outputs, so its recorded hash no longer matches the tree.

#733's own checks were green when it merged because they last ran against a pre-#641 base, and GitHub does not re-run a PR when its base moves. The badge outlived the base it was measured on. This is not a defect in #733 — the same trap is armed for every branch that predates #641 and touches a declared input.

## Fix

`make generate`. Only the manifest and its digest change; no generated output content moves, which is the expected shape when the inputs are restaged but the outputs were already in lockstep.

## Verification

- Reproduced at 1ef26f9d in a clean worktree before the change.
- `make check-generated` passes after it, including `core_compiled.lgb` bundle lockstep and `core_go_lowered/` native dispatch under `-tags gogen_ir`.

## Fallout worth knowing

Branches rebased onto main between #641 and this fix regenerated against the broken manifest and will need `make generate` re-run: #727 and #692 are both in that window. #683 also edits `scripts/generate.lg` and currently shows green from a stale base, so it is the same trap as #733 and should be re-run before it is merged rather than after.

#747 tracks the underlying gap. The gate does its job, but only after the merge: nothing forces a re-run between approval and merge, so a green badge over a moved base is exactly what let this land.


mparrett:
@nnunley when you get a moment — could you take a look? `main` is red right now and this is the fix, so I'd rather not merge it on my own approval.

It is two files: `make generate` restages `generated.manifest` and its digest. No generated output content moves, which is the shape you would expect when the inputs are restaged but the outputs were already in lockstep. All 14 checks are green.

To be clear about the cause, this is not a defect in #733. It edits `scripts/generate.lg`, which #641's manifest declares as the generator for `op_generated.go`, `core/ir/data/generated.lg` and `ir_bridge_generated.go`, so its recorded hash stopped matching the tree. #733's own checks were green because they last ran against a pre-#641 base, and GitHub does not re-run a PR when its base moves — so the badge outlived the base it was measured on. The same trap is armed for every branch that predates #641 and touches a declared input; #733 just got there first. #683 edits the same file and is sitting on green from a stale base right now.

#747 has the broader thread, including the bit I would value your read on most: whether `generated.manifest` should get its own merge driver at all, given it carries output-readiness records and a driver regenerating it mid-merge could assert outputs are ready that were never rebuilt. That felt like your call rather than a patch from me. #748 is only the narrow half — making `Compute` parse before it hashes, so the `sums` driver can no longer mint a clean digest over a manifest still holding conflict markers.

No rush on those two. This one is just to get `main` green again.

nnunley (approved):

PR #750, opened 2026-08-17
fix(build): refresh the dependency manifest after #743

mparrett:
`main` is red at dd1d0246, the same failure shape as #749 fixed for #733:

```
pkg/rt/core_compiled.lgb
pkg/rt/core_go_lowered/
pkg/rt/zz_primitives_generated.go
```

#743 edits `pkg/rt/keysource.go`, `keysource_queued.go`, `term.go` and `term_plan9.go` — four inputs #641's manifest declares — so their recorded hashes stopped matching the tree at merge. Its own checks were green from a pre-#641 base, and GitHub does not re-run a PR when its base moves.

This is the second time in one evening, which is the argument #747 is really about: a green badge measured against a base that no longer exists is not evidence, and nothing currently forces a re-run between approval and merge.

## Fix

`make generate`. Only the manifest and its digest change; no generated output content moves.

## Verification

- Reproduced at dd1d0246 in a clean worktree.
- `make check-generated` passes after the change.

## Still exposed

Open PRs touching a declared input with no digest of their own: #745, #501, #730, #698, #717, #723, #624, #699. Each needs one `make generate` when rebased, and none should be merged on checks that predate #641 landing.


mparrett:
@nnunley sorry to send you a near-duplicate of #749 — same two-file fix, different trigger, and this one was avoidable.

#743 edits four inputs the manifest declares: `pkg/rt/keysource.go`, `keysource_queued.go`, `term.go` and `term_plan9.go`. I had already worked out that it was in that category and merged it anyway on the strength of an APPROVED badge and green checks that predated #641. That is my error, not a problem with #743 — it is a good change and it landed clean.

`make generate` only; no generated output content moves. All 14 checks green.

The rule I should have been applying, and will from here: a PR is only safe to merge if its checks ran after 8ff841dd. Eight open PRs still fail that test — #745, #501, #730, #698, #717, #723, #624 and #699 — so this will keep recurring until something enforces it rather than reporting it after the fact. That is the part of #747 I would most value your read on, whenever you have time. No rush on it or on #748; this one is just to get `main` green again.

nnunley (approved):

PR #754, opened 2026-08-19
feat(test): support thrown? and thrown-with-msg? in is

### LINKED ISSUE #753: clojure.test: is doesn't support thrown? / thrown-with-msg?
`clojure.test`'s `is` special-cases `(thrown? Klass body)` and `(thrown-with-msg? Klass re body)`, but let-go's `is` macro doesn't, so the symbol hits normal resolution:

```clojure
(require 'test)
(test/deftest t (test/is (thrown? Exception (throw (ex-info "b" {})))))
;; => Can't resolve thrown? in this context
```

Expected (Clojure): the assertion passes when the body throws a value satisfying `(instance? Klass e)` — and for `thrown-with-msg?`, whose `ex-message` matches the regex; a normal return counts as a failure, a non-matching thrown value as an `:error`.

This blocks real-world test suites: HoneySQL's `sql_test` has 35 uses and `bigquery_test` won't even load. `instance?`, `Exception`, and `clojure.lang.ExceptionInfo` all already work, so the expansion looks feasible in `pkg/rt/core/test.lg` alone.

abogoyavlensky:
Expand `(is (thrown? Klass body))` to a `try/catch` that passes when the thrown value satisfies `(instance? Klass e)` and fails on a normal return; `thrown-with-msg?` additionally requires `(re-find re (ex-message e))`. A thrown non-matching value rethrows into the runner's catch and counts as :error, matching `clojure.test`. Both arities of is handle the forms. Includes the regenerated `core_compiled.lgb`.

Resolves: #753


mparrett (commented):
Read this closely and found nothing to fix. Recording what I checked, since two of the three looked like problems before I tested them.

The `e#` in `expand-is-thrown` spans two syntax-quote forms — the outer `(try … (catch e# …))` and the inner one built by `~(if with-msg? …)`. Under Clojure's per-syntax-quote gensym map those are two different symbols and the `thrown-with-msg?` branch would reference an unbound one. let-go scopes auto- gensyms per macroexpansion instead, so both resolve to the same symbol: `(probe)` expands to `(let [__G__0 42] (inc __G__0))`. Correct here, and worth knowing it would not port.

`(re-find re (ex-message e#))` throws `re-find expected String` on a nil message, but I could not construct a value that reaches it: `ex-info` rejects a nil message at construction, a non-Exception rethrows before the `re-find`, and runtime errors carry messages. Unreachable rather than guarded, which seems fine.

The probe pattern in the test file is the part I would keep — running the must-fail and must-error cases as plain vars through `run-test-var` rather than as registered deftests, so asserting on failure does not fail the suite.

Verdict: correct. Not approving tonight only because I want to sleep on the batch; no objection in the diff.

mparrett:
Follow-up on the gensym point above, not a change request on this PR.

If the per-macroexpansion scoping is intentional, it is worth one line in `docs/guide/clojure-compatibility.md` under **Behavioral differences**, since it is invisible until a macro ported from Clojure misbehaves, and this PR is now the first place in-tree that depends on it:

> Auto-gensyms (`x#`) are scoped per macroexpansion rather than per
> syntax-quote form, so a nested syntax-quote reuses the outer symbol where JVM
> Clojure generates a fresh one.

Repro:

```clojure
(defmacro probe [] `(let [e# 42] ~(do `(inc e#))))
(macroexpand-1 '(probe))
;; let-go:  (let [__G__0 42] (inc __G__0))   — same symbol
;; Clojure: two distinct gensyms
```

If it is not intentional, then the more forgiving behaviour is the one to question rather than the doc, and `expand-is-thrown` would want an explicit `(gensym)` binding instead. Either way it is a separate change from this one. I am happy to send the doc line; flagging rather than assuming which answer it is.

abogoyavlensky:
@mparrett Good catch, and the answer is that it is not intentional.

`gensymEnv` is allocated fresh in `readSyntaxQuote` with its counter starting at
0, and names are `<prefix>__G__<counter>`, so numbering restarts for every
syntax-quote read rather than being scoped per macroexpansion. The effect is
wider than the nested case in your repro:

```clojure
(defmacro sib1 [] `(let [a# 1] a#))
(defmacro sib2 [] `(let [a# 2] a#))
(macroexpand-1 '(sib1))  ;; (let [__G__0 1] __G__0)
(macroexpand-1 '(sib2))  ;; (let [__G__0 2] __G__0)

**Review by mparrett — APPROVED** [review]
Approving.

I rebuilt at `ae7881aa` and re-checked the parts most likely to break, since the shape has not changed since my August 20 pass.

The explicit gensym holds up. `(gensym "e__")` yields `e__1` — lowercase and unqualified — so `tryCompiler` still parses `(catch e__1 (if …))` as let-go's bare catch rather than a class-dispatch clause. That heuristic keys on a qualified or uppercase-simple first token and a gensym is neither, so it was the plausible failure; macroexpansion confirms the bare form.

The compat suite was the other exposure. `test/compat/clojure/core-test/portability.lg` defines its own zero-class `(defmacro thrown? [& body] …)`, and the new special case in `is` would break any unqualified call to it. Every call site in the jank suite is `p/thrown?`, so nothing collides, and `TestClojureTestSuite` passes on the branch. A bare local `make test` skips that test — the submodule is not initialized by default — so it only ran because CI checks out with `submodules: recursive`. I initialized it locally to confirm.

`make build` and `make test` are both clean; the only `FAIL` lines in the log are the intentional probes. I also ran multi-form bodies, `is` nested inside a `thrown?` body, two assertions in one scope, `are` plus `thrown?`, an empty body, a runtime non-`ex-info` error under `thrown-with-msg?`, and a deliberately shadowed `e__G__0` local. All behave.

Two follow-ups belong to us rather than to this PR, so please do not widen it:

- Eight test files still carry the comment "let-go has no `thrown?` — use try/catch directly" (`gogen_test.lg:91`, `os_unzip_test.lg:48`, `os_paths_test.lg:41`, `async_ns_test.lg:147`, `os_free_port_test.lg:16`, `os_fs_test.lg:68`, `hash_sha_test.lg:39`, `ns_introspection_test.lg:43`). They only become wrong at merge, so the sweep cannot be written before then.
- The `clojure.test` row in `docs/guide/clojure-compatibility.md` is incomplete rather than wrong, and should list `thrown?` and `thrown-with-msg?`.

I will send both once this lands.

Your `gensymEnv` diagnosis is right, and the behaviour is worse than either of us framed it in the thread. The shared name captures:

```clojure
(defmacro outer [x] `(let [v# ~x] ~(do `(let [v# 99] v#))))
(macroexpand-1 '(outer 7))  ;; (let [__G__0 7] (let [__G__0 99] __G__0))
(outer 7)                   ;; => 99, not 7
```

So the reader is the thing to fix, not the compatibility doc. You found it — do you want to file it? Otherwise I will, crediting the diagnosis. Either way it is separate from this PR, and the explicit gensym here is right regardless of how the reader ends up behaving.

The branch is `BEHIND`: `main` picked up `bc425101` and `4d8ce37a` after your last push. Both are perf/CI workflow changes and neither touches `pkg/rt/` or `test/`, so nothing needs re-verifying — it just needs a sync before it can merge.

PR #756, opened 2026-08-19
fix(compiler): var special form resolves private vars across namespaces

### LINKED ISSUE #755: #'other-ns/private-var doesn't resolve (var special form enforces privacy)
The var special form should ignore privacy - in Clojure, `#'other-ns/private-var` resolves anywhere; only *bare* qualified symbol references enforce `defn-`/`^:private`. In let-go the var form goes through the same privacy-enforcing lookup:

```clojure
;; fixture.lg
(ns fixture)
(defn- hidden [] 42)

;; main.lg
(ns main (:require [fixture :as p]))
(@#'p/hidden)
;; => Can't resolve p/hidden in this context
```

Expected: `(@#'p/hidden)` returns 42, while a bare `(p/hidden)` still fails to compile.

Libraries routinely use `#'` to reach private vars in tests — e.g. HoneySQL's `sql_test` (`#'sut/format-on-expr`) fails to load on let-go because of this. The IR builder's `build-var` has the same behavior via `resolve`, so both compile paths need the relaxed lookup.

abogoyavlensky:
Clojure enforces privacy only on bare qualified symbol references - the var special form `(#'ns/sym)` reaches private vars anywhere. Namespace gains `LookupIncludingPrivate`, used by the bytecode `varCompiler`; the IR builder's build-var gets the same semantics via a find-var fallback with alias expansion (strict IR mode used to throw, non-strict silently fell back to bytecode). Bare-symbol resolution stays privacy-enforcing on both paths, locked by tests.

Resolve: #755


nooga (approved):
LGTM. Verified locally: build, vet, `go test ./test/` and `pkg/vm` pass; merges clean into main apart from the generated manifest refresh. Merging.

PR #758, opened 2026-08-19
feat(vm): symbols are invokable for collection lookup, like keywords

### LINKED ISSUE #757: Symbols aren't invokable for collection lookup
Keywords are invokable for collection lookup, but symbols aren't:

```clojure
('a {'a 1})
;; => TypeError: let-go.lang.Symbol is not a function
```

Expected (Clojure): symbols in function position look themselves up in the collection argument exactly like keywords — `('a {'a 1})` → `1`, `('a #{'a})` → `'a`, with the 2-arity default on miss and nil/default for non-collection arguments.

`Keyword` already implements the VM's `Arity()`/`Invoke()` shape in `pkg/vm/keyword.go`; mirroring it on `Symbol` (minus the keyword fast path) looks sufficient. HoneySQL relies on this (`format-do-update-set` breaks `issue-293-sql` and `upsert-test`).

abogoyavlensky:
Resolve: #757


mparrett (commented):
`Symbol.Invoke` mirrors `Keyword.Invoke` line for line apart from the `KeywordLookup` fast path, which the comment already explains is keyword-specific, and `Arity() -1` matches. Arity bounds, the non-`Lookup` fallback, and the 2-arity default all behave the same as the keyword path.

I checked the two predicates that could have shifted when `Symbol` starts satisfying `Fn`. `ifn?` already listed `vm.Symbol` explicitly, so it is unchanged, and `fn?` switches on concrete function types rather than the interface — with a comment saying that is deliberate — so `(fn? 'a)` stays false, matching Clojure.

Verdict: correct. Same as #754: holding the approval stamp until morning rather than for anything in the change.

nooga (approved):
LGTM. Symbol invoke matches Clojure's lookup semantics, tests cover the cases. Verified locally (build/vet/`./test/` green). Merging.

nnunley:
Maintainer update: merged current `main` into the branch at `4b8914c0b`. The only merge conflict was generated `pkg/rt/generated.sums`; it was resolved via canonical `make generate && make check-generated` (exit 0: manifest, primitive registrars, bundle lockstep, lowered-tree native compile). Focused `go test ./pkg/vm` also passes. Required GitHub CI is now running on the updated head; no merge until it is green.

PR #760, opened 2026-08-19
fix(vm): multi-arity dispatch routes exactly-min-arg calls to the variadic variant

### LINKED ISSUE #759: Multi-arity variadic variant rejects calls with exactly its fixed-arg count
A multi-arity fn's variadic variant rejects calls with exactly its fixed-arg count:

```clojure
(defn g ([a] 1) ([a b & c] c))
(g 1 2)
;; => function g doesn't have a 2-arity variant
```

Expected (Clojure): `(g 1 2)` selects the variadic variant with `c` bound to nil; `(g 1 2 3)` binds `(3)`.

The single-fn path already accepts exactly-min args — the bug is in `MultiArityFn.variantFor` (`pkg/vm/func.go`), which requires `count >= rest.Arity()` where the variadic `Func`'s arity counts its rest slot. Related: `(fn [a b & {:as o}] ...)` called with exactly 2 args binds `o` to `{}` instead of nil (the kwargs rest expansion applies `hash-map` to an empty rest unconditionally).

HoneySQL's `assert-statement` hits this via the 3-arity `(sut/format m k v)`.

abogoyavlensky:
Also fix the kwargs rest expansion: an empty rest stays nil through the map destructure `({:as o} binds nil, not {})`, matching Clojure.

Resolves: #759


mparrett (commented):
Verified against lg built at the PR head (`dd6c976a`) and its merge-base (`c39c0a9`), with babashka 1.12.218 as the oracle. No findings.

Both halves land. Every shape I tried matches Clojure, and each one fails on main at exactly-min:

| call | Clojure | main | this PR |
|---|---|---|---|
| `(f 1 2)` for `([a] …) ([a b & r] …)` | `:variadic` | error | `:variadic` |
| `(apply f [1 2])` | `:variadic` | error | `:variadic` |
| multi-arity closure capturing a local, called with 2 | `[:var 9 []]` | error | `[:var 9 []]` |
| `letfn` multi-arity, called with 2 | `:v` | error | `:v` |
| `(g)` for `[& {:as o}]` | `nil` | `{}` | `nil` |

No regressions in the neighbouring shapes. Exact fixed arity still beats the variadic — `([a b] …) ([a b & r] …)` called with two args gives `:two`. Underflow still errors, `:or` defaults still apply through a nil kwargs map, `[& r]` with no args still yields an empty rest, and a three-arm `([a] [a b] [a b c & r])` dispatches 1/2/3/4 arguments to 1/2/3/3.

`restMinArgs` handles the three rest-arm shapes it can encounter and the Go test exercises all three — bare `*Func`, `*Closure`-wrapped, and `NativeFn`. The closure branch is the one that carries the most weight in practice, since `letfn` and any multi-arity `fn` capturing a local route through it.

One pre-existing gap for context, not a request and not caused by this change: Clojure 1.11 trailing-map kwargs are still unsupported. `(g {:a 1})` for `[& {:as o}]` errors on both main and this branch, where Clojure binds `{:a 1}`. That is independent of the nil-rest fix here.

Verdict: correct.

nooga (approved):
LGTM. Dispatch fix and the nil-rest kwargs fix are both right; `make check-generated` and tests pass locally.

Heads up: #787 rewrites the same `fn-expand` line in core.lg (its `seq-to-map-for-destructuring` subsumes this nil fix and adds the 1.11 trailing-map behaviour). I'll land this one first; #787 will need a trivial rebase over it, taking #787's side.

nnunley:
Maintainer update: merged current `main` into this branch at `d3ad0e6f0` to resolve the post-#756 integration blocker and preserve #760-before-#787 ordering. `go test ./pkg/vm` passes.

The local pre-push suite was incomplete in Evener’s restricted sandbox because language tests hardcode `/tmp` and received `operation not permitted`; the linked-worktree attempt also reproduced #679. I pushed the merge commit specifically to run the repository’s authoritative required GitHub checks and will not merge unless those checks pass.

nnunley:
Current-main CI exposed two integration failures (`build`, `generated-artifacts`) after GitHub’s branch update. I am regenerating from exact head `6836c5e699` using the documented `make generate && make check-generated` workflow; I will push only the resulting generated delta after the check exits zero. Other completed required checks are green; several remain in progress.

nnunley:
Confirmed from completed Actions logs: both failures have the same single root cause—manifest recorded `4d2df37e…`, computed `e487300c…`; `build` failed at “Check generated manifest before regeneration” and `generated-artifacts` failed at `make check-generated`. No separate compiler/test failure is indicated. The isolated canonical regeneration is still running.

nnunley:
Artifact repair pushed as `2bf6efbaa`: only `pkg/rt/generated.sums` changed, from `4d2df37e…` to the computed `e487300c…`. The canonical `make generate && make check-generated` completed exit 0, including manifest, primitive registrar, bundle lockstep, and lowered-tree native compile checks. I then updated the branch onto newly merged #707; required CI is rerunning on that actual current head.

PR #762, opened 2026-08-19
feat(rt): reify accepts Object method overrides (toString, hashCode, equals)

### LINKED ISSUE #761: reify can't override Object methods (toString)
`reify` treats every head symbol as a protocol, so `Object` method overrides die:

```clojure
(reify Object (toString [_] "x"))
;; => extend-type* expected Protocol
```

Expected (Clojure): `reify` accepts `Object` method overrides; `toString` drives the value's string representation, so `str`, `pr-str`, and `println` all return `"x"`. `hashCode`/`equals` should at minimum be accepted so forms construct.

This is currently a documented limitation (README + `test/readme_limitations_test.lg`), so supporting it means flipping those. Real-world need: HoneySQL's `unhashable-value-509` test reifies `Object` with `toString` plus a throwing `hashCode`. Since `str`/printing bottom out in the Go `Stringer` on `DTypeInstance`, a `toString` hook on the `DType` looks feasible without touching `deftype` (which could stay unsupported).

abogoyavlensky:
toString is stored on the DType and consulted by DTypeInstance.String(), which str, pr-str, and println all route through. Only string results are accepted (a toString returning the instance itself would recurse fatally); anything else falls back to the default rendering. Unknown Object method names are a compile-time error; hashCode/equals are accepted and ignored — hash/equality stay identity-based. deftype Object methods remain unsupported; limitation tests and docs updated.

Resolves: #761


mparrett (commented):
Reviewed with lg built at the PR head (`61f9c0ca`) and its merge-base (`c39c0a9`), with babashka 1.12.218 as the Clojure oracle. Two findings, neither blocking.

The wiring works everywhere I pushed it. `toString` routes through `str`, `pr-str`, `println`, `format "%s"` and string interpolation, and through nested rendering — an instance inside a vector or as a map value both pick it up. Closure capture works. `hashCode` and `equals` are accepted and ignored, with hash and equality staying identity-based as documented. Unknown Object methods are a compile-time error. `deftype` with Object methods is still rejected, and mixing a protocol with `Object` in one `reify` gives both the protocol method and the override.

## P2 — a `toString` that stringifies `this` takes the process down

The `String`-result check in `DTypeInstance.String()` catches an override that *returns* the instance, directly or inside a collection — I confirmed both fall back to the default rendering. It does not catch an override that *calls* `str` on the instance, which is the more natural way to write it wrong:

```clojure
(reify Object (toString [this] (str "x" this)))
```

`str` on that is `fatal error: stack overflow`, exit code 2, 388 lines of Go runtime dump, and `(try … (catch Throwable e …))` does not intercept it — it is a Go fatal error, so `recover` cannot reach it either. In a REPL or an embedded host that takes the whole process with it.

Calibrating this: printing a self-referential structure already does the same thing on `main` — `(def a (atom nil)) (reset! a a) (str a)` is the same fatal overflow — so this is a new door onto a failure mode the runtime already tolerates rather than a new failure mode, which is why it is a P2 and not a blocker. It is newly reachable, though: before this change `reify Object` was rejected outright. Since the comment above the check is already reasoning about fatal recursion, the gap seems worth closing — a re-entrancy flag on the `DType` consulted in `String()` would do it, with a test alongside the existing non-string fallback case.

## P3 — duplicate `toString` takes the first, Clojure takes the last

`(some (fn [[_ mname fn-form]] (when (= 'toString mname) fn-form)) object-entries)` returns the first match. Given `(reify Object (toString [this] "a") (toString [this] "b"))`, let-go prints `a` and Clojure prints `b`. Pathological input, and cheap to align while the file is open.

## Verdict: correct

Nothing existing breaks — `reify Object` was rejected before this change, so no existing code depends on it, and the identity-based hash and equality paths are untouched. Both findings need new user code to reach.

README.md
```diff
@@ -135,7 +135,9 @@ so shared `.cljc` stays JVM-loadable. See
 
 Not a drop-in JVM Clojure. The main gaps: no coordinated STM or async agents
 (`ref`/`agent` are atom-backed aliases), no `clojure.spec`, unchunked lazy seqs,
-no custom `*data-readers*`, no JVM host interop on `deftype`/`reify`, and no
+no custom `*data-readers*`, no JVM host interop on `deftype` (`reify` accepts
```
nnunley: Nit: I'd suggest that this documentation moves out of README.md to docs/guide/compatibility.md with a clear link from this README to the doc for further reading.

That said, there's already a lot of other compatibility notes here, so this depends on your own level of energy. It would be good as a fast follow, at any rate.

abogoyavlensky: Thanks for pointing that out, I will take a look.

abogoyavlensky:
@mparrett Thanks for the review. Both findings are addressed.

**P2, self-stringifying `toString`.** `DTypeInstance` now carries a per-instance `atomic.Bool`, set only while that instance's own override is running. A re-entrant `String()` renders the default form, which terminates and preserves the override's own output:

```clojure
(str (reify Object (toString [this] (str "x" this))))
;; => "x#<reify__1>"
```

The flag is per instance rather than per type, so unrelated instances are unaffected. The trade, noted in the comment, is that two goroutines calling `String()` on the same instance may see the default rendering. That seemed clearly better than a fatal stack overflow no catch or recover can intercept.

**P3, duplicate `toString`.** Now takes the last match rather than the first, so `(reify Object (toString [_] "a") (toString [_] "b"))` prints `b` on both runtimes.

Tests added in `test/reify_object_test.lg`: the self-stringifying case, that the guard is not sticky across repeated `str` calls, an instance nested in a collection, and the duplicate-method case. They sit alongside the existing non-string fallback test.

The branch is also updated from main. `make test` and `make check-generated` pass locally.

README.md
abogoyavlensky: Good point. The guide entry in `docs/guide/clojure-compatibility.md` already
carries the full `reify` `Object` semantics, and this README section is a
summary that links there, so spelling it out inline was duplication. Trimmed the
claim back to "limited JVM host interop on `deftype`/`reify`" and left the
detail in the guide.

Moving the rest of the compatibility notes out of the README is a bigger change
than this PR introduced, so I would rather do that as the fast follow you
suggested than widen this diff.

nooga (approved):
LGTM. mparrett's two findings (re-entrant `toString` stack overflow, first-vs-last duplicate method) are fixed in the diff — checked the atomic re-entrancy guard and the `(last (keep ...))` change. `make check-generated` and `./test/` pass locally. The README reorg thread from nnunley can stay a fast-follow. Merging.

nnunley (changes requested):
Blocking correctness issue in `DTypeInstance.WithMeta`: `cp := *d` now copies the active atomic `rendering` guard. If a `toString` override calls `(with-meta this ...)` and retains the result, that copy is created while `rendering == 1` and permanently skips its override on future `String()` calls. Concurrent `WithMeta` and `String` also mix a plain struct-field read with atomic writes.

Please construct the copy explicitly from `dtype`, `fields`, and `meta` so each new instance starts with `rendering == 0`, and add a regression test where `toString` stores `(with-meta this ...)` and the stored value still uses the override afterward.

The rest of the implementation and existing regression coverage look sound. The branch will also need conflict resolution against current `main` after #756.

abogoyavlensky:
@nnunley Fixed in `5c2bfd7`. `WithMeta` now constructs the copy explicitly from `dtype`, `fields`, and `meta`, so every new instance starts with the guard down and no plain read of `rendering` races the atomic writes; field storage stays shared as before. Regression coverage from both sides: the e2e you described in `test/reify_object_test.lg` - `toString` stores `(with-meta this {:tag :copy})` and the stored copy must still render through the override (it rendered the default form before the fix) - plus a white-box `pkg/vm` test that raises the guard and asserts the copy starts at 0. Branch is also merged with current main (post-#756), so the conflict is gone.

nnunley (approved):
Re-reviewed exact head `cc696682d808`. The prior `WithMeta` blocker is fixed at the root: clones receive a fresh rendering guard rather than copying active atomic state, and the focused white-box regression plus `reify_object_test.lg` cover the behavior. Independent focused verification passed (`go test -race ./pkg/vm -run TestDTypeInstance_WithMeta_FreshRenderingGuard`; reify fixture: 9 tests / 21 assertions). Approving. The branch still needs an author-owned update after current queue integrations because GitHub reports a base conflict.

Issue #742, opened 2026-08-13
nth accepts maps and sets (Clojure throws); indexed?/counted? invert for seqs; indexed? untested


nnunley:
## Summary

`nth` succeeds on maps and sets, which Clojure rejects. Two predicates in the same lattice also disagree with Clojure. Verified against both babashka 1.12.218 and JVM Clojure 1.12.5 (identical results), on `lg dev (cf43894)`.

## `nth` ignores `indexed?`

`indexed?` is the capability that governs positional access. It correctly reports `false` for maps and sets, but `nth` accepts them anyway and returns an element in hash order.

| expression | Clojure | let-go |
|---|---|---|
| `(nth #{1 2 3} 0)` | `UnsupportedOperationException: nth not supported on this type: PersistentHashSet` | `2` |
| `(nth (sorted-set 3 1 2) 0)` | `UnsupportedOperationException: … PersistentTreeSet` | `1` |
| `(nth {:a 1 :b 2} 0)` | `UnsupportedOperationException: … PersistentArrayMap` | `[:b 2]` |
| `(nth (sorted-map :b 1 :a 2) 0)` | `UnsupportedOperationException: … PersistentTreeMap` | `[:a 2]` |
| `(nth [1 2] 0)` | `1` | `1` |
| `(nth (list 1 2) 1)` | `2` | `2` |
| `(nth (seq [1 2]) 1)` | `2` | `2` |

Sorted collections throw in Clojure too, so a total order is not sufficient for `nth` — `Indexed` is required. Out-of-range indices do throw in let-go, so only in-range positional access is affected.

The answers are stable across processes because our hash iteration is content-determined, so a test asserting `(nth #{…} 0)` passes consistently and would only break when a hash function or capacity rule changes. That makes this quiet rather than harmless.

**Fix:** gate `nth` on `indexed?` (vectors, and the linear-scan path for lists/seqs) and throw for associative collections.

## Two predicate divergences

| expression | Clojure | let-go |
|---|---|---|
| `(indexed? (seq [1 2]))` | `false` | `true` |
| `(counted? (seq [1 2]))` | `true` | `false` |

`(seq [1 2])` is a chunked seq: `Counted` but not `Indexed`. We have it backwards in both directions.

## Why the suite did not catch any of this

- `test/clojure-test-suite/.../indexed_qmark.cljc` **does not exist** — `indexed?` has no coverage at all.
- `nth.cljc` has 4 `thrown?` assertions but **zero** map or set cases.
[...]

PR #764, opened 2026-08-19
feat(vm): insertion-ordered small maps matching Clojure array-map semantics

### LINKED ISSUE #763: Small maps don't preserve insertion order (Clojure array-map semantics)
Maps never preserve insertion order, even small ones:

```clojure
(keys {:_id 1 :name "cat"})        ;; => (:name :_id) — varies per run
(seq (array-map :a 1 :b 2 :c 3))   ;; => ([:b 2] [:a 1] [:c 3])
```

Expected (Clojure's array-map contract): map literals and `array-map` preserve insertion order up to 8 entries (`PersistentArrayMap`); assoc past 8 promotes to hash order; `hash-map` never guarantees order. Builders on top — `assoc` chains, `into`, `zipmap`, `merge`, `select-keys` — preserve order at small sizes too, because they build through array maps (including transient ones).

This is the single biggest source of Clojure-compat test failures in practice: 9 of HoneySQL's 16 failing tests on let-go reduce to it (`test-value`, `inline-record-body`, `records-statement`, `patch-statement`, `object-record-expr`, `navigation-dot-index`), and any code that formats/prints small maps for humans depends on it. Equality and `hash` must stay order-insensitive. An insertion-ordered small-map representation behind the existing public `Map` type (paired key/value slice, linear scan — cheap at ≤8 entries) looks feasible.

abogoyavlensky:
`PersistentMap` gains an internal array-map mode behind the same public `MapType`: maps with at most 8 entries keep insertion order in a paired key/value slice with linear-scan lookup (root == nil means ordered mode); the 9th assoc promotes to the HAMT and dissoc never demotes a non-empty map, matching Clojure's `PersistentArrayMap`/`PersistentHashMap` split. Draining a promoted map to empty deliberately lands back in the empty ordered mode. `TransientMap` gets the same mode so the into/transduce path preserves order for small maps. PersistentSet internals switch from direct root access to mode-agnostic accessors. Equality and hash stay order- and representation-insensitive.

Map literals, `array-map`, `assoc` chains, `zipmap`, `merge`, `select-keys`, and into all preserve insertion order up to 8 entries; hash-map order stays unspecified (now incidentally insertion-ordered when small).

Resolves: #763


mparrett (commented):
Reviewed with lg built at the PR head (`a5dddaf7`) and at its merge-base (`c39c0a9`), with babashka 1.12.218 as the Clojure oracle.

The map work holds up. Cross-mode invariants are intact — an ordered map and a HAMT map with the same entries compare `=` in both directions, hash identically, dedupe in a set, work as map keys, and carry meta through assoc, dissoc and promotion. The assoc-chain threshold matches Clojure exactly now: ordered at n=8, hash order at n=9. Two findings, neither blocking.

## P2 — `PersistentSet` picked up ordered mode, which exposes a reversed set-literal path

`NewTransientSet` now routes through `NewTransientMap` and `TransientSet.Persistent` through `TransientMap.Persistent`, so a set of at most 8 elements comes out insertion-ordered. Clojure has no array-set — `PersistentHashSet` is hash-ordered at every size — so this is a new divergence, and the PR body's semantics list covers maps only.

It also surfaces a second-order problem. `readSet` (`pkg/compiler/reader.go:788`) accumulates into `vm.EmptyList` with `ret.Conj(form)`, which prepends, then emits `(hash-set …)` with the arguments in reverse source order. `readMap` builds a `[]vm.Value` forward and calls `NewArrayMap` directly. That asymmetry was invisible while set order was hash-determined:

| form | Clojure | main | this PR |
|---|---|---|---|
| `#{1 2 3 4 5 6 7 8}` | `(7 1 4 6 3 2 5 8)` | `(8 2 1 3 6 4 5 7)` | `(8 7 6 5 4 3 2 1)` |
| `(set [1 2 3 4 5 6 7 8])` | `(7 1 4 6 3 2 5 8)` | `(8 2 1 3 6 4 5 7)` | `(1 2 3 4 5 6 7 8)` |

On main and in Clojure the literal and the constructor agree. Here they disagree, and the literal is exactly reversed. Nine elements or more is unaffected.

Worth deciding whether sets should be ordered at all. If not, forcing HAMT mode in `NewTransientSet` resolves both rows and the reader asymmetry goes back to being harmless.

## P3 — direct `array-map` past 8 entries still diverges

Clojure's `(array-map …)` stays a `PersistentArrayMap` at every size and preserves order; this PR promotes past 8:

| n | Clojure | this PR |
|---|---|---|
| 4, 8 | in order | in order |
| 9, 16, 32, 64 | in order | hash order |

nnunley measured the cost of the Clojure behavior in #742 — last-key lookup at n=512 is 166.8 ns against a flat 16.2 ns hash control — and argued for capping at 8 rather than offering ordered maps at arbitrary size. The threshold looks right; it wants a line in `docs/guide/clojure-compatibility.md` rather than staying implicit. `(type …)` also still returns one `let-go.lang.Map` for both constructors.

## Coverage

#742 Item 4 says this divergence is pinned by a `KNOWN-DIVERGENCES` ledger that fails once the divergence is fixed. That ledger is not on `main` or any remote ref, which is why CI is green here. I have asked @nnunley to push it.

## Downstream

xsofy is unaffected. Its determinism is an explicit salt-key hash chain rather than map iteration, and its entity maps run 19 to 34 entries, so they stay in HAMT mode. Six seeds through its headless replay driver under both builds give bit-identical final seed, entity set, player position and turn count, and its suite passes identically at 318 tests and 2825 assertions.

## Verdict: correct

Map and set iteration order was never specified, so nothing existing breaks, and the cross-mode equality and hash invariants that would otherwise break map keys and set membership all hold. Both findings are fidelity and documentation rather than correctness.

nnunley:
So, I'd prefer a bit more discussion on the issue. I purposefully made the change to make the ordering non-deterministic back in  https://github.com/nooga/let-go/pull/397

mparrett:
> So, I'd prefer a bit more discussion on the issue. I purposefully made the change to make the ordering non-deterministic back in #397

I remember this, too. I mentioned that to the agent that did the initial review. Surprised it didn't make it into the comment. I'll check the transcripts and see how they got dropped.

nooga (dismissed):
Two things need settling before this can move:

1. nnunley's point that map/set iteration order was deliberately randomized in #397 hasn't been answered. That's a design question, please reply on his thread so we can decide it rather than re-review code around it.
2. `PersistentSet` picked up ordered mode as a side effect of sharing the transient-map internals, and it now disagrees with the reader: `(seq #{1 2 3 4 5 6 7 8})` gives `(8 7 6 5 4 3 2 1)` while `(seq (set [1 2 3 4 5 6 7 8]))` gives `(1 2 3 4 5 6 7 8)`. Either sets are ordered too (and the reader is fixed) or `NewTransientSet` forces HAMT mode.

Also needs a rebase — it's the only PR of yours that didn't get the 08-28 round. The map mechanics themselves (equality/hash order-insensitivity, the 8→9 promotion, transients) check out locally.

nnunley (changes requested):
Architecture decision: adopt Clojure-style small-map behavior—map literals and builders are insertion-ordered through the small-map threshold, then promote.

One blocking representation side effect remains: `PersistentSet` currently inherits ordered transient-map mode, so small `#{...}` literals sequence in reverse source order while `(set [...])` sequences forward. Clojure has no array-set contract. Please keep sets HAMT-backed/unordered (for example, force `NewTransientSet` into HAMT mode), add coverage that prevents accidental map-order leakage into sets, and rebase/resolve the current conflicts with `main`. The ordered-map mechanics otherwise look coherent.

abogoyavlensky:
@nnunley Fixed in `58bf5ad`, and the audit turned up more entry points than the transient: the ordered mode could reach a set through `EmptyPersistentSet`/`NewPersistentSet` seeding from the (now-ordered) empty map, through `Disj` draining to the empty ordered map, and through an empty HAMT transient persisting with a nil root — not just through `NewTransientSet`. All paths now land on a shared `emptySetImpl` (empty HAMT-rooted map), and `NewTransientSet` forces its backing transient out of the ordered mode, so set iteration order is a function of element hashes alone.

Coverage from both sides: `TestSetsStayHAMTBacked` white-box walks every construction path (constructors, conj chain, transient round-trips, drain-and-regrow) asserting the impl never reads as ordered, and `test/set_order_test.lg` pins the observable property — literal, `(set ...)`, `into`, `conj`, and drain-regrow builds of equal sets iterate identically, including 32-element sets against reversed and shuffled inserts. Branch is merged with current main, so the conflicts are resolved; `make test`, `check-generated`, and lint are green locally.

nnunley:
Re-reviewed exact head `2a7c7241d95e` against the selected contract. The substantive request is satisfied: small maps preserve the Clojure-style behavior, sets remain independently HAMT-backed, equality/hash remain order-insensitive, and focused tests pass. The only remaining item from my review is integration: GitHub reports this head conflicting with current `main`, so the generated artifacts must be refreshed during an author-owned update. To avoid repeated churn, this can wait until the PRs ahead of it in the merge queue land; I can clear the review immediately after verifying that refreshed head.

nnunley:
The substantive map/set contract was verified at `2a7c7241d95e`, and the current head `3f46a8184403b128ef03da9d70477e945d1ce123` was fully green against `f26eb497` before #800 landed. GitHub now reports a conflict with current `main` (`3aa9f4ea86d41dd55ced66fb3ca800c4bc66c355`).

@abogoyavlensky, please update/rebase the contributor-owned branch and refresh the generated artifacts. I will not push to it. I’ll clear the requested-change review after verifying the refreshed exact head still keeps sets HAMT-backed and CI is fully green.

nnunley (approved):
Re-reviewed the ordered-map/set work at requested SHA `3f46a8184403b128ef03da9d70477e945d1ce123` against current `main` `3aa9f4ea86d41dd55ced66fb3ca800c4bc66c355`. During review the branch advanced to `d46738fee317dc0648250236b26b06d63f54f30b`; that delta only integrates #800 and refreshes generated metadata, with no changes to the reviewed map/set implementation or tests.

The prior blocker is fixed. Small maps preserve insertion order through 8 unique entries and promote on the ninth; replacement does not move a key, and non-empty HAMTs do not demote. Equality and hash remain order- and representation-insensitive. Sets now stay independently HAMT-backed through constructors, persistent and transient updates, empty round-trips, drain, and regrow paths, with white-box and language-level regression coverage.

Focused VM tests pass, as do `test/map_order_test.lg` (25 assertions) and `test/set_order_test.lg` (7 assertions). GitHub reports the updated head mergeable with current main. All completed CI checks are green; remaining in-progress checks continue to gate merge. Approving.

nnunley:
@nooga The set-representation blocker from your review is fixed and independently re-verified at current head `d46738fee317dc0648250236b26b06d63f54f30b`; the branch is on current `main`, conflict-free, and every required check is green. I approved the refreshed head, but your earlier CHANGES_REQUESTED review still keeps the branch policy blocked.

Please re-review/clear that review when convenient. I am deliberately not enabling auto-merge; after reviewer clearance I will recheck the exact head/base and checks immediately before merging.

nooga:
@nnunley I dismissed my stale review, lgtm!

nooga (approved):

Issue #751, opened 2026-08-18
Stale or red PR merges unblocked; no required status checks on main


mparrett:
Main went red twice within three hours on the evening of 2026-08-16, both times from merges I made, both times with the same failure:

```
check-generated: dependency manifest stale for 3 output(s):
  pkg/rt/core_compiled.lgb, pkg/rt/core_go_lowered/, pkg/rt/zz_primitives_generated.go
```

| time (UTC) | event |
|---|---|
| 2026-08-16 22:32 | #641 merges — provenance dependency manifest |
| 2026-08-16 23:02 | #733 merges → main red |
| 2026-08-17 00:21 | #749 merges — hand-refreshed manifest |
| 2026-08-17 00:42 | #743 merges → main red again |
| 2026-08-17 19:35 | #750 merges — hand-refreshed manifest |

#749 and #750 are the same commit written twice by hand: a regenerated `generated.manifest` plus its digest, nothing else.

## Why CI did not stop it

#747 assumed CI covers this — "CI catches this (`make check-generated-manifest`), so nothing merges broken". I wrote that, and it turns out to be wrong.

The `main` ruleset (active, as of 2026-08-17) contains:

deletion, non_fast_forward, pull_request, required_linear_history

There is no `required_status_checks` rule, so CI is advisory at merge time. Nothing blocks a merge on a red PR, and nothing invalidates a green one that has gone stale.

The job itself is built correctly. `generated-artifacts` uses a bare `actions/checkout@v4`, so on a `pull_request` event it tests `refs/pull/N/merge`: the merge result, not the branch tip. It would have caught both breaks.

**The green was stale, not wrong.** #743's `generated-artifacts` ran at 22:13:19Z. #641 landed at 22:32:24Z, 19 minutes later. Nobody pushed to #743 afterward, so GitHub never recomputed the merge ref, and the check stayed green against a base that no longer existed. #747 already noted this shape: "GitHub does not re-run on a base change, so the red only appears on the next push."

## Two quiet failure modes

**A conflicting PR runs no checks at all.** When GitHub reports `DIRTY` it cannot build a merge ref, so nothing runs — `gh pr checks` returns zero rows. Not a red X, an absence. Seen on #748 today. With no required checks, such a PR reads as merely unreviewed while never having been tested.

**GitHub does not run custom merge drivers.** `.gitattributes` registers `merge=sums`, which recomputes the digest rather than taking a side, so local merges resolve cleanly. Server-side there is no driver:

```
[...]

PR #765, opened 2026-08-19
build: gate pushes on generated-artifact manifest freshness

mparrett:
## Problem

#641 made `pkg/rt/generated.manifest` record a SHA-256 per declared generator input. Editing any of those inputs restages `core_compiled.lgb`, `core_go_lowered/` and `zz_primitives_generated.go`, whether or not the generated content moves. `pkg/rt/os.go` is an ordinary source file, and nothing about editing it signals that a regeneration is now owed.

Nothing local catches the omission. No hook ran `check-generated` at any stage: the config has twelve hooks — nine at pre-commit, three at pre-push, with `forbid-compiled-binaries` at both — and none of them is this one. `make install-hooks` registers only the two merge drivers.

And no server-side operation can supply it, because none of them has a generator. A rebase moves a base. A merge driver resolves a file. The stack UI does both and force-pushes the result. None recomputes a hash.

The cost so far:

- `main` went red twice in three hours on 2026-08-16. #749 and #750 are the same commit written by hand twice: a regenerated manifest and its digest.
- The #698/#699/#717 stack stayed red through two full rounds of restacking. Each member needed its own `make generate`, and each one restaled the members above it, so the chore scaled with stack depth rather than with the size of the change.

There is also no backstop above it: per #751 the `main` ruleset has no `required_status_checks`, so a red branch reads as merely unreviewed.

## Change

Add `check-generated` as a pre-push hook.

It is a hash comparison of the manifest's declared inputs against the tree, with no regeneration and no build, so it costs about 1.3 seconds. That makes it the cheapest hook in the file — the three already at pre-push are `go test`, the vendored Clojure compat suite, and the ir-stress ratchet.

`always_run` is set rather than a `types` filter, because a declared input can be a `.go` or a `.lg` file and a filter on either would miss half the sources.

## Verification

With `prek`, the hook passes on a clean tree, and a committed edit to `pkg/rt/os.go` fails it:

```
generated-artifact manifest freshness (#641).............................Failed
  check-generated: dependency manifest stale for 3 output(s): pkg/rt/core_compiled.lgb, pkg/rt/core_go_lowered/, pkg/rt/zz_primitives_generated.go
  Run `make generate` to regenerate the bundle + lowered Go tree and refresh the manifest.
```

The failure only reproduces when the edit is committed. `prek` stashes unstaged changes before running hooks, so probing with a dirty working tree passes vacuously.

## What this does not cover

This is the local layer only, and it is worth being clear about the two it leaves open.

It would not have prevented either `main` outage. Those were stale-green *merges* — #743's checks ran 19 minutes before #641 landed, and GitHub does not recompute a merge ref when the base moves. That is #751's missing `required_status_checks`, and it needs a ruleset change rather than a hook.

It also does not settle the `merge=manifest` question in #747. My read is that this check weakens the case for a driver rather than complementing it: the sharp edge you identified there is a manifest that merges *cleanly* into a wrong state, since the file is sorted and one record per line, and a hash comparison against the tree catches exactly that. #747 already notes the risk that a regenerating driver "could assert outputs are ready when they have not been rebuilt". I would rather land this, see whether the driver still has a job, and decide that separately. You run these workflows locally more than I do, so if there is a conflict pattern this check misses, that changes the answer.

Refs #641, #747, #751.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -149,6 +149,24 @@ repos:
         pass_filenames: false
         always_run: true
         stages: [pre-push]
+      # Generated-artifact provenance gate (#641). ~1.3s: a pure hash
+      # comparison of the manifest's declared inputs against the tree, with no
+      # regeneration and no build. Editing any declared input restages the
+      # outputs, and no server-side operation can refresh the manifest for you
+      # — not a rebase, not a merge driver, not the stack UI — so this has to
+      # run on a developer machine. always_run: a declared input can be a .go
+      # or a .lg file, so it must run on every push rather than on a types
+      # filter.
+      # NOTE: jj does not run git hooks — jj users gate with `make
+      # check-generated-manifest` (the same check) before pushing.
+      - id: check-generated-manifest
+        name: generated-artifact manifest freshness (#641)
+        description: Fail if a declared generator input changed without `make generate`
+        entry: go run ./cmd/check-generated
+        language: system
+        pass_filenames: false
+        always_run: true
+        stages: [pre-push]
   - repo: https://github.com/checkmake/checkmake.git
     rev: v0.2.2
     hooks:
--- a/Makefile
+++ b/Makefile
@@ -421,7 +421,7 @@ install-hooks:
 # pkg/rt/generated.sums. This catches the exact drift that a prior go generate
 # or lgbgen invocation would otherwise mask in CI.
 check-generated-manifest: $(GO)
-	@go run ./cmd/check-generated -stale || { \
+	@go run ./cmd/check-generated || { \
 		echo "ERROR: dependency manifest stale or check errored — run 'make generate'."; \
 		exit 1; \
 	}


nnunley (approved):

.pre-commit-config.yaml
```diff
@@ -149,6 +149,24 @@ repos:
...
+      # — not a rebase, not a merge driver, not the stack UI — so this has to
+      # run on a developer machine. always_run: a declared input can be a .go
+      # or a .lg file, so it must run on every push rather than on a types
+      # filter.
+      # NOTE: jj does not run git hooks — jj users gate with `make
+      # check-generated-manifest` (the same check) before pushing.
+      - id: check-generated-manifest
+        name: generated-artifact manifest freshness (#641)
+        description: Fail if a declared generator input changed without `make generate`
+        entry: go run ./cmd/check-generated
```
nooga, line 165: **Pre-push hook doesn't match its own "same check" claim for jj users.**

This hook runs `go run ./cmd/check-generated` (no flags) — the stricter path that checks both per-input staleness *and* the `generated.sums` vs `generated.manifest` digest.

But the NOTE above (lines 160-161) says jj users get "the same check" via `make check-generated-manifest`, which passes `-stale` and therefore runs *only* the per-input staleness check — skipping the digest check entirely.

Reproduced directly: with `pkg/rt/generated.sums` corrupted (digest mismatch) but `generated.manifest` untouched, `go run ./cmd/check-generated` correctly fails ("STALE"), while `go run ./cmd/check-generated -stale` (what the Makefile target runs) passes silently.

That's exactly the #747 scenario this gate is meant to catch — a bad manifest/sums merge. As written, a git contributor pushing through prek would be blocked, but a jj contributor running the 'equivalent' Makefile target would not — the opposite of the parity this hook is supposed to guarantee.

mparrett: Confirmed, and fixed in 6faeed6f by dropping the flag rather than rewording the note.

Reproduced your case directly: with the digest line in `generated.sums` replaced by zeroes and `generated.manifest` untouched, `-stale` exits 0 silently while the unflagged command exits 1 with `STALE — .lg/lgbgen sources changed without regeneration`. So `make check-generated-manifest` was skipping exactly the #747 case it gets relied on for.

The unflagged mode is a strict superset, which is why this seemed the safer direction than documenting around it: a stale input still exits 1 with the same output, and a digest mismatch now exits 1 as well. Both callers — `make check-generated` and the `generated-artifacts` CI job — get the wider check, and clean `main` passes it unchanged. That also makes the jj note true as written instead of needing a caveat.

On how the wrong claim got in: I read the target from a checkout sixteen commits behind `origin/main`, which predates the `-stale` flag. Everything else in the body I checked against the remote; this line I did not.

nooga (approved):

Issue #266, opened 2026-06-19
gogen_ir: residual timeouts + unresolved-dep symbols in lowering


nnunley:
**Buckets: 3 timeouts + 5 unresolved-dep symbols (8/749).** Sub-issue of #262. (local `lg-mnbv`)

`:stress/timeout` (3): `core.lg::for-emit`, `ir/lattice.lg::mask->type`, `ir/lower_go.lg::inst-rhs`.

`ir/build: unresolved symbol` (5): `vm-import` @ `ir/passes/pipeline.lg::lower-ns-to-go`, `recv-type` @ `ir/lower_go.lg::lower-method`, `tname` @ `ir/lower_go.lg::oos-sequence-copies`, `func` @ `core.lg::with-redefs-fn`, `refs` @ `ir/passes/constfold.lg::try-primitive`. Most are deftype/interop-shaped locals build can't resolve.

Done = both bucket groups → 0 in the #262 repro.

nnunley:
Post-#249-merge refresh (merged main `23bfd87f`): the unresolved-dep symbols are gone; the remaining failures in this bucket are **4 `:stress/timeout`** defns: `core.lg::for-emit`, `ir/lower_go.lg::inst-rhs`, `ir/lower_go.lg::try-assign-stmts`, `ir/lattice.lg::mask->type`. This is now the sole open bucket under #262.

nnunley:
**Done — both bucket groups → 0 in the corpus repro.**

Latest full `ir-stress.lg corpus` run (per-defn TSV), all 8 named defns now `:ok`:

**Timeouts (3) — cleared by #280** (`closure-info*` per-nid memo):
| defn | status | lower ms |
|---|---|---|
| `core.lg :: for-emit` | `:ok` | ~9922 |
| `ir/lattice.lg :: mask->type` | `:ok` | ~1747 |
| `ir/lower_go.lg :: inst-rhs` | `:ok` | ~4952 |

**Unresolved-dep symbols (5) — cleared by #287 (pre-declare generated names) + #288 (eval top-level requires):**
`pipeline.lg :: lower-ns-to-go`, `lower_go.lg :: lower-method`, `lower_go.lg :: oos-sequence-copies`, `core.lg :: with-redefs-fn`, `constfold.lg :: try-primitive` — all `:ok`.

[...]

PR #767, opened 2026-08-20
fix(lower-go): assign each branch its own closure at a disagreeing join

### LINKED ISSUE #766: gogen: a capturing closure inside a conditional is emitted as the function's return value, discarding the branch
## Summary

When gogen lowers a function whose conditional branch evaluates to a **closure that captures a local** (so it can't be lifted to a top-level fn and becomes a `rt.BoxNativeFn`), the closure is emitted as the enclosing function's unconditional `return`, and the conditional collapses to an empty `if {} else {}`. The other branch is silently dropped.

Non-capturing closures are fine — those get lifted to `Fn__liftedN` and the branches assign correctly. It's specifically the `BoxNativeFn` path.

## Repro

`repro.lg`:

```clojure
(ns repro)

(defn one-branch [flag x]
  (if flag (fn [] x) nil))

(defn both-branches [flag x y]
  (if flag (fn [] x) (fn [] y)))

(defn via-when [flag x]
  (when flag (fn [] x)))

(defn nested-let [entry]
  (if (:a entry)
    (let [h (:a entry)] (fn [] (h)))
    nil))
```

```sh
LG_SOURCE_PATHS=".:pkg/rt/gogen:/tmp/repro" \
  ./lg scripts/lg-compile /tmp/out repro/build /tmp/repro/repro.lg
```

Generated (all four, abridged):

```go
func OneBranch(ec *vm.ExecContext, arg0 vm.Value, arg1 vm.Value) vm.Value {
	if vm.IsTruthy(arg0) {
	} else {
	}
	return rt.BoxNativeFn(func() vm.Value {
		return arg1
	})
}

func BothBranches(ec *vm.ExecContext, arg0, arg1, arg2 vm.Value) vm.Value {
	if vm.IsTruthy(arg0) {
	} else {
	}
	return rt.BoxNativeFn(func() vm.Value {
		return arg1        // arg2's closure is gone entirely
	})
}

func NestedLet(ec *vm.ExecContext, arg0 vm.Value) (vm.Value, error) {
	var v2 vm.Value
	var h vm.Value
	v2, _ = rt.InvokeValueEC(ec, vm.Keyword("a"), []vm.Value{arg0})
	if vm.IsTruthy(v2) {
		h, _ = rt.InvokeValueEC(ec, vm.Keyword("a"), []vm.Value{arg0})
	} else {
	}
	return rt.BoxNativeFn(func() (vm.Value, error) {   // returned even when h was never set
		...
	}), nil
}
```

`(one-branch false 1)` returns a closure instead of `nil`; `(both-branches false 1 2)` returns the `then` closure; `(nested-let {})` returns a closure over an unbound `h`.

For contrast, this one lowers correctly, because neither closure captures anything:

```clojure
(defn pick [flag]
  (if flag (fn [] :from-then) (fn [] :from-else)))
```

## Expected

Each branch assigns its own value to the result temp, exactly as the non-capturing case already does — the `BoxNativeFn` expression should be emitted *inside* the branch, not hoisted to the return.

## Why this one is nasty

It's silent. Go only rejects it when the lowering happens to leave a dead temporary behind, which is how we found it — `go build` in [legmacs](https://github.com/nooga/legmacs)' native build:

```
# legmacs/build/legmacs/modes
legmacs/modes/modes.go:56:6: declared and not used: v12
legmacs/modes/modes.go:59:6: declared and not used: v33
legmacs/modes/modes.go:416:6: declared and not used: v12
```

In the same build, another function with this shape (`legmacs.modes.prog/make-carry-advance`) compiled *without* an error and just silently had the wrong semantics in the native binary. So "it builds" isn't evidence the lowering was right.

Two real functions that hit it, both ordinary Clojure:

```clojure
(defn carry-advance [mode-kw]
  (let [entry (get @registry mode-kw)]
    (cond
      (:highlight-carry entry) (:highlight-carry entry)
      (:highlight-line entry)  (let [h (:highlight-line entry)]
                                 (fn [carry line] (:carry (h carry line))))
      :else (fn [_carry _line] nil))))
```

The workaround we used is to build each closure in its own tiny named function, so the branches contain calls rather than `fn` literals:

```clojure
(defn- spans-dropping-advance [h] (fn [carry line] (:carry (h carry line))))
(defn- no-carry [_carry _line] nil)

(defn carry-advance [mode-kw]
  (let [entry (get @registry mode-kw)]
    (or (:highlight-carry entry)
        (when-let [h (:highlight-line entry)] (spans-dropping-advance h))
        no-carry)))
```

That lowers correctly, but it's a shape you have to know about — worth fixing in gogen so a returned-conditional closure just works.

## Version

Reproduced on `main` @ c39c0a93. legmacs pins c879db76 in `.let-go-version` and hits it there too.

mparrett:
Fixes #766.

## Why

A closure that captures a local cannot be lifted to a top-level `fn`, so it lowers to an inline `rt.BoxNativeFn`. When such a closure was a *branch* of a conditional, gogen emitted it as the enclosing function's unconditional `return` and collapsed the conditional to an empty `if {} else {}`, dropping the other branch:

```clojure
(defn both-branches [flag x y]
  (if flag (fn [] x) (fn [] y)))
```

`(both-branches false 1 2)` returned the `then` closure. `(if flag (fn [] x) nil)` returned a closure instead of `nil`, and a closure over a `let` binding that only the taken branch established escaped with that binding unset.

The failure is silent. Go rejects the result only when the lowering happens to strand a dead temporary; otherwise the binary builds, exits 0, and returns the wrong branch's value. In the reported case one function failed with `declared and not used`; a sibling of the same shape compiled clean and had wrong semantics in the native binary, so "it builds" was not evidence the lowering was right.

## Root cause

Closures are compile-time entities in this pass: they carry no runtime Go local, and `closure-expr` re-materializes them at each use site. `closure-info*` resolved the `:block-arg` case with `some`, taking the first incoming edge that resolved to a closure. At a join whose edges disagree — two different `fn` literals, or a closure against `nil` — the parameter was still classified as a closure value, so it got no local and both edges skipped assignment. The single use site then materialized whichever closure came first.

Non-capturing closures were unaffected: they lift to top-level `fn`s, so the edges carry `:load-var` nodes and the join gets an ordinary local.

## What changed

- Resolve every incoming edge, and treat the parameter as a closure value only when they unanimously agree. A disagreeing join falls back to a real local.
- `emit-assignments-for-target` no longer screens the arg side, so each edge materializes its own `rt.BoxNativeFn` into that local. A template threaded into a DCE-killed param is still skipped by the existing `live?` and const-param tests on the param.
- The visited guard yields a `:cycle` sentinel rather than `nil`, so unanimity can tell a loop's back edge apart from an edge that carries no closure. Without that distinction a loop-carried closure would lose the local-free treatment the lineage walk exists to provide (#266).
- A `:cycle` source is set aside rather than counted as disagreement, which makes the surviving answer true only for a walk with that loop header on the stack. Each result therefore carries the set of still-on-the-stack nids it leaned on, and only results that leaned on none are memoized. A parameter's own back edge is discharged where it is found instead of recorded: `p = agree(inits…, p)` has `agree(inits…)` as its fixed point, which is a property of the block graph rather than of the walk, so the ordinary loop-carried closure stays fully cacheable.

  A plain boolean "derived through a cycle" flag is not enough in either direction. Excluding only a literally-`:cycle` result caches a path-dependent concrete answer, and a disagreeing join nested in a loop then loses a branch the same way #766 did — `(choose-loop false 1 2)` in the fixture below returned `2` instead of `1`. Excluding every cycle-derived result instead locks the ordinary loop-carried closure out of the memo, and the per-param-per-edge re-walk of shared block-arg lineage is what made lowering time quadratic to begin with (#266).

`pkg/rt/generated.manifest` and `pkg/rt/generated.sums` are refreshed because `lower_go.lg` is a declared generator input (#765).

## Verification

Two new native-entry fixtures. `test/native-entry/conditional_closure.lg` covers the acyclic disagreeing joins: closure against a different closure, closure against `nil`, and a closure over a `let` binding only one branch establishes. `test/native-entry/loop_carried_closure.lg` covers a loop-carried closure that a conditional either replaces or retains, which is the case the memo rule gets wrong. Each lowers through the production path, matches a committed AST shape pinning the per-branch assignment, builds, and prints byte-exact output that the pre-fix lowering got wrong. Both semantic mutants die, so the assertions are not vacuous.

`make native-entry-gate`, `make test`, and `make gogen-diff` pass; the engine-parity gate reports 0 output divergence.

`make parity-full` reports the `lower-go` leg as diverged, but that is pre-existing: `main` shows the same single `:stress/timeout` failure and the same bucket hash `d88d9278`. This branch passes two more assertions than `main` on that leg, and single-run wall time was slightly lower on both engines.


mparrett:
_Posted by Claude on @mparrett's behalf._

@nnunley — this came out of a local adversarial review of the PR head. The P1 it raised is real, and reproduced exactly as written: `(choose-loop false 1 2)` returned `2`, from a binary that built and exited 0.

Fixed in dd050cc7. Instead of a boolean "derived through a cycle" flag, each resolution now carries the set of still-on-the-stack nids it leaned on, and only results that leaned on none are memoized. A parameter's own back edge is discharged where it is found rather than recorded — `p = agree(inits…, p)` has `agree(inits…)` as its fixed point, which is a property of the block graph rather than of the walk — so the ordinary loop-carried closure stays fully cacheable.

That distinction turned out to matter. A plain boolean taint was the first attempt, and it cost a second `:stress/timeout` on the untagged ir-stress `lower-go` leg: excluding every cycle-derived answer also locks the ordinary loop-carried closure out of the memo, which is the #266 re-walk.

`test/native-entry/loop_carried_closure.lg` covers conditional replacement and retention of a loop-carried closure, as you asked.

## Status

Green on the new head: `make native-entry-gate` (both new fixtures, semantic mutants dying), `make test`, `make gogen-diff`, and all GitHub checks.

Still open: re-running `make parity-check` to confirm the deps-set version leaves the ir-stress `lower-go` timing where it was — the boolean version did not, so this is the number I most want to see before calling it done. I will post those counts here within the hour, and follow up with a fix rather than leave it if they moved.

For reference, the baseline that leg diverges against is pre-existing: `main` shows one `:stress/timeout` and bucket `d88d9278` on the untagged side.

mparrett:
_Posted by Claude on @mparrett's behalf._

Parity follow-up, as promised — with the honest version of the answer.

The `gogen_ir` leg is clean and stable on this head: `pass=824 fail=0`, bucket `d41d8cd9`, in every run I did. `main` measures `822/0`.

I could not get a trustworthy number out of the untagged leg. It replays the lowering passes through the interpreted engine, its only failure bucket is `:stress/timeout`, and the count therefore tracks machine load rather than the diff. On this same commit I measured `fail=0` at 67s wall and `fail=5` at 268s wall, depending on what else was running; `main` measured `fail=1` at 90s. An earlier read of mine looked like this change removed a pre-existing timeout — I don't believe that is established, and I'd rather retract it here than have it quoted back later.

So the timing question the boolean-taint attempt raised is still open on my side. What does not depend on timing, and is green: `make native-entry-gate` (both fixtures, semantic mutants dying), `make test`, `make gogen-diff` at 0 output divergence, and the GitHub checks on this head.

If the untagged `lower-go` leg is load-bearing for the merge decision, it wants a quiet machine — CI rather than my laptop, which is heavily contended right now.

pkg/rt/core/ir/lower_go.lg
```diff
@@ -667,56 +667,129 @@
...
+   the first resolvable source instead silently published one branch's closure
+   as the whole join's value (#766).
+
+   A :cycle source is set aside rather than counted as disagreement, and the
+   nid it cycled back to is recorded in deps. Cycling back to THIS parameter is
+   discharged on the spot: `p = agree(inits…, p)` has `agree(inits…)` as its
+   fixed point, which is a property of the block graph and not of the walk, so
+   the answer stays cacheable. That is the loop-carried closure the lineage walk
+   exists to serve (#266). A cycle back to some OTHER parameter still on the
+   stack is not discharged — it stays in deps and travels upward."
```
nnunley: A few things
1. This is a documentation style comment. Is this something that needs to be exposed at runtime to repl users? If so, it should document what the function does, not the history of why it exists.
2. It should document what it returns if it is intended to be consumed by repl/ide/llm users.
3. Including issues/pr ids is not great for long term maintenance. Especially when it's unclear whether it's a PR or not.

I'd cut down the comments verbosity and try to stick to narrow set of what, why, and the return signature (optional).

pkg/rt/core/ir/lower_go.lg
```diff
@@ -667,56 +667,129 @@
...
+                           (filter (fn [r] (not= (nth r 0) :cycle)) results)))
+        all-deps (reduce (fn [acc r] (into acc (nth r 1))) #{} results)
+        deps     (disj all-deps nid)]
+    (cond
+      (empty? real)
+      [:cycle (reduce (fn [acc r] (into acc (nth r 1))) #{} cyc)]
+
+      (some (fn [x] (nil? x)) real)             [nil deps]
+      (every? (fn [x] (= x (first real))) real) [(first real) deps]
+      :else                                     [nil deps])))
```
nnunley: You have a number of extra reduce/map passes on the results list (bad name for the incoming list of dependencies)

```
(let [step (fn [acc [tag deps]]
             (let [acc (update acc :all-deps into deps)]
               (if (= tag :cycle)
                 (update acc :cyc-deps into deps)
                 (update acc :real conj tag))))
      {:keys [real cyc-deps all-deps]}
      (reduce step {:real [] :cyc-deps #{} :all-deps #{}} results)

      deps (disj all-deps nid)]
  (cond
    (empty? real)     [:cycle cyc-deps]
    (some nil? real)  [nil deps]
    (apply = real)    [(first real) deps]
    :else             [nil deps]))
```

pkg/rt/core/ir/lower_go.lg
```diff
@@ -667,56 +667,129 @@
     (get srcs nid)
     (get (compute-param-sources f) nid)))
 
+(defn- unanimous-closure
```
nnunley: Nit: odd function name, and doesn't actually match its intent.

pkg/rt/core/ir/lower_go.lg
```diff
@@ -667,56 +667,129 @@
...
+  ;;
+  ;; Tracking only whether the final info is :cycle is not enough. A disagreeing
+  ;; join nested inside a loop resolves its back edge to :cycle, sets it aside,
+  ;; and returns a perfectly concrete closure that nonetheless holds only for
+  ;; this walk. Caching that republished one branch of the inner join as the
+  ;; join's whole value — the same silent miscompile as #766, reintroduced
+  ;; through the cache. Tracking deps as a plain boolean is not enough either:
+  ;; it would also lock the ordinary loop-carried closure out of the memo, and
+  ;; the per-param-per-edge re-walk of shared block-arg lineage is what made
+  ;; lowering time quadratic in the first place (#266).
```
nnunley: Another bit of comment slop. Interesting that it switched from a doc comment to a pure comment. I'd suggest picking a comment style and sticking with it. Might be a broader discusson.

pkg/rt/core/ir/lower_go.lg
```diff
@@ -667,56 +667,129 @@
...
 (defn closure-info
   "Resolve the closure a value refers to: {:template <aux> :captures [nid …]}
    or nil. Walks make-closure/push-closed and follows block-parameter lineage
    so closures threaded through a loop header still resolve (captures on
-   push-closed nodes are preserved across the walk). Memoized per function on f."
+   push-closed nodes are preserved across the walk). Memoized per function on f.
+
+   This is the boundary that normalizes the internal :cycle sentinel to nil:
+   callers ask whether a value denotes one closure, and a value that only ever
+   resolves to itself does not."
```
nnunley: These lines are editorial, and can be deleted. It's a contract that isn't enforced anywhere.

mparrett:
_Posted by Claude on @mparrett's behalf._

Correcting myself on the last paragraph above: I suggested CI as the place to settle the untagged `lower-go` leg. That was wrong twice over.

It does not run in CI at all — `.github/workflows/go.yml` defers it explicitly ("Bootstrap-parity gate (deferred — runs as `make parity-full` locally)"), which matches the note in the contribution policy. And if it were wired up it would be a worse instrument, not a better one: `:stress/timeout` is a wall-clock budget (`LG_STRESS_TIMEOUT_MS`, default 5000ms per defn), so a free shared runner would be more contended than a laptop, not less. The `perf-pr` job is opt-in behind a `perf` label and benches the `pkg/vm` fleet, so it does not cover lowering either.

The useful consequence is that I had conflated two questions that want different instruments:

- **Parity** — do the two engines agree? Raise `LG_STRESS_TIMEOUT_MS` until nothing times out, then compare counts and bucket hashes. Contention cannot corrupt that, so it is answerable here and it is the half that matches the policy checkpoint.
- **Lowering speed** — a separate question, and a pass/fail count was never the right way to measure it. That one does want a quiet machine.

Running the raised-timeout parity now and will post the counts. The speed half I will flag as unmeasured rather than guess at it.

mparrett:
_Posted by Claude on @mparrett's behalf._

Parity numbers, measured the way that actually works: `LG_STRESS_TIMEOUT_MS=60000`, so the wall-clock bucket cannot fire and the counts stop tracking machine load. Both runs below were taken on a contended machine, which is the point — the method is what makes them comparable, not the conditions.

**ir-stress `lower-go`**

| | untagged | gogen_ir | verdict |
| --- | --- | --- | --- |
| `main` @ a6657612 | `pass=822 fail=0` `d41d8cd9` | `pass=822 fail=0` `d41d8cd9` | PARITY |
| this branch @ dd050cc7 | `pass=824 fail=0` `d41d8cd9` | `pass=824 fail=0` `d41d8cd9` | PARITY |

So the `lower-go` divergence we were both looking at — the one I reported as pre-existing on `main` — was a timeout artifact on **both** sides. Give the corpus enough time and the two engines agree, before and after. No parity divergence, and no bucket delta, which is the policy checkpoint.

The count moves `822 → 824`: two ir-stress cases that did not lower cleanly before now do. That is the direction the fix predicts.

Two things I am explicitly **not** claiming:

- **Lowering speed is still unmeasured.** The wall times in those runs (`377s` for `main` untagged against `53s` for the branch) were taken at load ~20 and ~8 respectively and are not comparable. I am not reading a speedup into that, and it is a separate question from parity.
- `main`'s **jank** leg reported DIVERGED in its run, with `compile=1 runtime=2` skips on the `gogen_ir` side and 239 of 242 files reached. That has the shape of load noise rather than a finding; this branch's jank leg was clean at `pass=6311 fail=0` on both engines. Flagging it rather than burying it, but I would not treat it as real without a quiet re-run.

mparrett:
_Posted by Claude on @mparrett's behalf._

Better evidence for the leg I left unmeasured above, from a machine neither of us controls. I ran `make parity-check` on a free 4-core `ubuntu-latest` runner, on this branch and on `main`, at the repo-default per-defn timeout and at a raised one:

| ref | `LG_STRESS_TIMEOUT_MS` | jank | lower-go untagged | lower-go gogen_ir | verdict |
| --- | --- | --- | --- | --- | --- |
| `main` @ a6657612 | 5000 (default) | `6311/0` both | `822/0` `d41d8cd9` | `822/0` `d41d8cd9` | PARITY |
| `main` @ a6657612 | 60000 | `6311/0` both | `822/0` | `822/0` | PARITY |
| this branch @ dd050cc7 | 5000 (default) | `6311/0` both | `824/0` `d41d8cd9` | `824/0` `d41d8cd9` | PARITY |
| this branch @ dd050cc7 | 60000 | `6311/0` both | `824/0` | `824/0` | PARITY |

`822 → 824`, no failures on either side, identical bucket hash, `OK: all suites parity-identical` in all four. That is the same delta I measured locally, now reproduced somewhere contention cannot reach.

It also retires the caveat I raised earlier: the raised timeout turned out to be a workaround for **my** machine, not something the measurement needs. On a dedicated runner the default 5000ms budget never fires — 48.71s against 49.36s on the untagged leg — so the counts were never at risk there. The `:stress/timeout` noise I hit was purely local contention.

Whole suite runs in roughly 4.5 minutes. I have written that up separately as an infrastructure suggestion rather than piling it onto this PR.

nooga:
Nice, tight fix — and the `:cycle` sentinel is the right call given `closure-info*` is memoized per-nid for perf (#266).

One observation for future reference rather than a request to change this PR: `closure-info*`'s shape (resolve a block-param's value by walking incoming edges, needing to survive loop back-edges) is the same problem `const-param-map` (line ~2941) already solves — "does this block-param provably hold one value on every path?" — but the two solve it with different algorithms:

- `const-param-map` is an **optimistic SCCP-style fixed point**: every param starts at `:top`, lowers to the meet of incoming args, iterated to convergence. Its own comment explains why: a pessimistic "only promote once every input is known" walk deadlocks on a loop-carried cycle, and starting optimistic sidesteps that by construction — no separate cycle marker needed.
- `closure-info*` is a **pessimistic DFS + visited-set**, which hits exactly that deadlock on loop-carried closures, and this PR fixes it by bolting on a `:cycle` sentinel to distinguish "still resolving a cycle above me" from "genuinely not a closure."

Both are correct now, but there are two bespoke "phi consistency across a possibly-cyclic CFG" algorithms living ~2000 lines apart in this file. Modeling closure identity as a lattice value (`:top` / `{:template … :captures …}` / `:bottom`, meet = "equal → same value, else `:bottom`") and computing it with the same worklist machinery `const-param-map` (or `ir/lattice.lg`'s) already uses would drop the `:cycle` sentinel entirely and give the five `closure-value?` call sites (`collect-local-ids` x2, `local-decls`, both sides of `emit-assignments-for-target`) one precomputed map to consult instead of a lazily-triggered recursive resolver.

Not proposing it here — this fix is small and well-tested, and the unification wouldn't be an efficiency win (SCCP-style fixpoints do potentially several sweeps vs. one memoized pass here). But if another "compile-time value, no local" category shows up later and needs the same treatment, that's the signal to fold `closure-info*` into `const-param-map`'s machinery rather than adding a third bespoke walker.

nnunley:
Reviewed with a dual-engine comparison (bytecode vs `scripts/lg-compile --entry-frame` native binary) on the exact head `dd050cc7` against base `a6657612`.

**The fix works as claimed.** The conditional-join case:

```clojure
(loop [i 0 acc []]
  (if (< i 4)
    (recur (inc i) (conj acc (if (even? i) (fn [] i) (fn [] (- i)))))
    (mapv (fn [f] (f)) acc)))
```

On base, native lowering loses branch identity entirely: `[4 4 4 4]`. On this head each branch materializes its own closure: `[4 -4 4 -4]`. Bytecode reference: `[0 -1 2 -3]`.

**Related pre-existing defect, not introduced or worsened here, and not blocking this PR:** the per-iteration capture value is still wrong in loops on both base and this head. Minimal case:

```clojure
(defn capture-each [n]
  (loop [i 0 acc []]
    (if (< i n)
      (recur (inc i) (conj acc (fn [] i)))
      (mapv (fn [f] (f)) acc))))
```

Bytecode: `[0 1 2]`. Native (base and this head): `[3 3 3]`. Same shape with a loop-body `let` binding (`[0 10 20]` vs `[20 20 20]`).

Root cause is adjacent to the code this PR touches but distinct from it: `closure-expr`/`box-as-value` splice capture expressions (e.g. `vm.Int(i)`) inline into the inner Go func literal at `:load-closed` sites. A Go func literal captures the outer local by reference, and the lowered loop mutates that local in place, so all iterations observe the final value. The fix shape is to materialize each capture into a fresh local at the closure creation site (`cap_n := vm.Int(i)`) before `rt.BoxNativeFn`, which composes cleanly with the per-branch assignment this PR introduces. Happy to file it as its own issue with the repro if useful.

While fixing #766 independently before finding this PR, I also wrote structural unit tests in the `lisp_lower_go_test.go` render-assertion style: they pin that `rt.BoxNativeFn` count per branch matches, that no closure escapes past the function's top-level `return`, and that a dropped `nil` branch fails loudly. This PR's implementation is the better one (the path-aware memoization correctly handles the disagreeing-join-inside-loop case mine got wrong), but the tests are complementary to the end-to-end fixtures; patch available if you want them folded in.

nnunley (approved):
Reviewed at `981e7655106f5f29698c23aef5f6fa03c7ccbc8c`. No blockers found.

The active-stack-before-memo ordering, dependency-carrying cycle sentinel, and dependency-free cache gate preserve the loop fixed-point invariant without leaking path-dependent answers. Disagreeing joins retain a real local, and the destination-only edge screen lets each incoming branch materialize its own closure.

Regression coverage exercises distinct closures, closure-versus-nil, branch-local capture, and both retain/replace paths for a loop-carried closure. The generated Go pins the per-branch assignments and the semantic mutants die.

Local validation:
- `make lowered`
- `make native-entry-gate`

All passed. GitHub checks are also green on the reviewed head.

Residual non-blocking risk: memo query order and mutually recursive multi-parameter SCCs are not directly unit-tested. Malformed anchorless CFG cycles and cycles through incomplete `push-closed` construction remain outside source-producible builder invariants; no valid counterexample was found.

PR #771, opened 2026-08-23
docs: catalog known Clojure divergences

nnunley:
Adds `docs/KNOWN_DIVERGENCES.md`, a decision-record ledger for known behavioral differences between let-go and Clojure JVM, and links it from the root README and the docs topical map.

The ledger separates intentional language decisions from temporary compatibility mismatches:

- **Intrinsic map traversal order** — intentional. Map literals and `hash-map` have unspecified traversal order at every size, following Go's map philosophy. Two independent reasons are recorded: generic maps should not acquire an ordering contract from a small representation detail, and the order side table removed in #397 had unfavorable access patterns during AOT/IR optimization. The Clojure `PersistentArrayMap` behavior it diverges from is documented from measurement (8-entry promotion threshold, direct large `array-map` staying ordered).
- **Temporary `array-map` mismatch** — `array-map` currently routes through the unordered map. Intended contract: direct construction preserves insertion order at any size; assoc past eight entries promotes, matching Clojure. Notes that #763's original acceptance criteria need updating to match.
- **Character model** — intentional. let-go characters are Unicode scalar values (Go runes), so supplementary characters like `(char 65895)` work as single characters instead of throwing as on the JVM. Two current validation defects are recorded: surrogate code points U+D800–U+DFFF are accepted, and huge BigInts truncate through `Int64()` before range checking.
- **Possible compatibility controls** — a marked-speculative section on mixed build/runtime features (opt-in strict randomized traversal for unordered collections; opt-in heavyweight bridges like gojava).

Each entry links the shared-suite `:lg` overrides that consume it (jank-lang/clojure-test-suite#927, #939) and the ledger documents its own maintenance rules. All five current upstream `:lg` occurrences are accounted for.

Refs #763.

### DIFF
--- a/docs/KNOWN_DIVERGENCES.md
+++ b/docs/KNOWN_DIVERGENCES.md
@@ -1,0 +1,201 @@
+---
+status: active
+last-verified: 2026-08-23
+authoritative-for:
+  - known-clojure-divergences
+---
+
+# Known Clojure divergences
+
+This ledger records known behavioral differences between let-go and Clojure
+JVM. It includes both deliberate language decisions and temporary compatibility
+mismatches. Intentional entries explain why let-go does not follow Clojure;
+temporary entries state the intended contract and resolution condition.
+
+This is separate from let-go's engine-differential ledgers:
+`test/parity-divergence.txt` tracks bytecode-versus-gogen output, while
+`test/gogen_aot_xfail.txt` tracks known gogen fixture divergences.
+
+## Intrinsic map traversal order
+
+**Classification:** Intentional divergence, with related compatibility work.
+
+### Clojure behavior
+
+Clojure JVM's representation is observable for small maps:
+
+- A map literal with up to eight entries is a `PersistentArrayMap` and traverses
+  in insertion order.
+- A map literal with nine entries is a `PersistentHashMap`.
+- Adding a ninth distinct entry to a `PersistentArrayMap` with `assoc` promotes
+  it to `PersistentHashMap`, which no longer preserves or guarantees insertion
+  order.
+- A directly constructed `(array-map ...)` remains a `PersistentArrayMap` and
+  preserves insertion order even when constructed with more than eight entries.
+
+The `PersistentArrayMap` threshold is 16 alternating key/value array slots,
+which means eight map entries. Clojure's common collection chunk size of 32 is
+not this promotion threshold.
+
+### let-go decision
+
+Traversal order for intrinsic map literals and `hash-map` is unspecified at
+every size. let-go does not actively randomize traversal by default; an order
+that appears stable for particular values or a particular release remains an
+implementation detail.
+
+Ordering should come from an explicitly ordered collection type. `sorted-map`
+provides comparator order today; `array-map` is intended to provide insertion
+order after the temporary mismatch below is resolved.
+
+### Why not Clojure?
+
+There are two independent reasons:
+
+1. A generic map should not acquire an ordering contract merely because its
+   current representation is small. This follows the host language Go's map
+   philosophy and avoids code accidentally depending on an implementation
+   detail.
+2. Preserving insertion order required extra state and unfavorable access
+   patterns. The order side table was deliberately removed while optimizing the
+   AOT/IR pipeline in
+   [PR #397](https://github.com/nooga/let-go/pull/397).
+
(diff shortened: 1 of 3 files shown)


docs/KNOWN_DIVERGENCES.md
mparrett, line 1: Nit: Can we name it `known-divergences.md` for consistency with the rest of `docs/*.md`? README.md follows the uppercase convention as it should.

docs/KNOWN_DIVERGENCES.md
```diff
@@ -0,0 +1,201 @@
...
+authoritative-for:
+  - known-clojure-divergences
+---
+
+# Known Clojure divergences
+
+This ledger records known behavioral differences between let-go and Clojure
+JVM. It includes both deliberate language decisions and temporary compatibility
+mismatches. Intentional entries explain why let-go does not follow Clojure;
+temporary entries state the intended contract and resolution condition.
```
mparrett, line 13: Nit: Consider tightening the specification for the known-divergences ledger:

```suggestion
JVM. It includes entries for two specific types of divergence:

- **Intentional divergence / language decisions**: explains _why_ let-go deliberately does not follow Clojure
- **Temporary compatibility mismatches**: states the intended contract and resolution condition(s).
```

docs/KNOWN_DIVERGENCES.md
```diff
@@ -0,0 +1,201 @@
...
+
+Portable code must not depend on intrinsic-map traversal. Today, sort the
+entries or use `sorted-map` when order is part of the program's contract. After
+the temporary mismatch is resolved, explicit `array-map` construction will also
+request insertion order.
+
+### Temporary `array-map` mismatch
+
+let-go currently routes `array-map` through its unordered persistent-map
+implementation. That is not the intended long-term contract.
```
mparrett, line 72: We should consider linking the implementation back to the divergence record in this document, with the idea being drift mitigation. Alternatively, make agents aware at a higher level via CLAUDE.md/AGENTS.md.

docs/KNOWN_DIVERGENCES.md
```diff
@@ -0,0 +1,201 @@
...
+  regardless of its initial size.
+- Adding a ninth distinct entry to a smaller array map promotes it to the
+  intrinsic unordered map.
+
+[Issue #763](https://github.com/nooga/let-go/issues/763) tracks the surrounding
+compatibility problem, but its original acceptance criteria propose ordered
+small literals and limit `array-map` ordering to eight entries. Those criteria
+do not match the contract established above. The issue must be updated or
+replaced by a dedicated implementation tracker before this work is considered
+fully scoped.
```
mparrett, line 86: Similar to [my last comment](https://github.com/nooga/let-go/pull/771/changes#r3891452127), I recommend we link #763 back to the recorded divergence, so it can be updated appropriately when updated or replaced.

docs/KNOWN_DIVERGENCES.md
```diff
@@ -0,0 +1,201 @@
...
+- `Int` and `BigInt` inputs in U+D800–U+DFFF pass the outer range check even
+  though the intended scalar-value contract excludes surrogate code points.
+- A `BigInt` is converted with `Int64()` before validation. Values outside the
+  64-bit range can therefore truncate into the accepted range; for example,
+  `(int (char 18446744073709551681N))` currently produces `65`.
+
+Resolution requires checking arbitrary-precision inputs before conversion and
+rejecting the surrogate range while retaining direct support for valid
+supplementary values. No dedicated issue tracks these defects yet; this entry is
+the local record until an issue is created or the validation is fixed.
```
mparrett, line 150: Nit: This paragraph implies the entry shouldn't exist if an issue is created. If the intent is to make known-divergences.md the source of truth for divergences, the entry should remain as long as there is a divergence, issue or not.

mparrett (commented):
Thanks for this PR— overall a solid addition that should clear up confusion. I left a few small suggestions to consider.

nooga (approved):
Approving — spot-checked five of the documented behaviours against the binary and all match. Two things I'd fold in before merge: (1) mparrett's filename nit is right, `docs/known-divergences.md` matches everything else in docs/, so `git mv` plus the README/docs/README links; (2) #764 is about to change the array-map ordering section — plan the follow-up edit when it lands. The other copy nits are take-or-leave.

PR #773, opened 2026-08-23
feat(interop): self-contained interop packages and an importable pkg/cli

### LINKED ISSUE #772: lginterop cannot generate bindings for use outside the let-go tree
## Problem

`cmd/lginterop` only runs inside a let-go checkout, and only emits files that compile *as part of* `pkg/rt`. A third-party module can't use it to generate bindings for an arbitrary Go package.

- It walks up to `go.mod` to find a checkout, builds a fresh `lg`, and shells out with `-source-paths <root>/scripts` — none of which exists for someone running `go run .../cmd/lginterop@<version>` from their own module.
- The codegen script hardcodes `package rt` and unqualified `RegisterInstaller` / `RegisterNS`.

## `RegisterInstaller` can't work out-of-tree

Worth calling out because it fails *silently*. `pkg/rt` drains its installer queue during its own package init (`pkg/rt/zz_run_installers.go`), and Go runs an imported package's `init` before the importer's. Anything an out-of-tree file enqueues lands after the drain and never runs — no error, just a missing namespace.

Calling the install function directly from `init()` is safe: `rt` is fully initialized by then, `RegisterNS` is mutex-guarded, and ordering relative to `LoadCore` matches an in-tree installer.

## Possible shape

- Embed the codegen script and run it in-process — `gogen` comes with `pkg/rt`, so any binary importing the runtime already has the emitter.
- An opt-in flag for self-contained output: `package <name>`, `pkg/rt` import, qualified `rt.RegisterNS`, direct install from `init()`. Omitting it stays byte-identical to today.

Also worth documenting: `go/types`' source importer resolves against the current directory's module, so a package is only scannable from a module that requires it.

### LINKED ISSUE #774: The lg CLI is not importable, so custom binaries can't be full-featured lg
## Problem

The whole `lg` command line lives in `package main` at the module root. A third-party module can import `pkg/rt` and build an `lg-runtime`-style bytecode host, but nothing that behaves like `lg`: no REPL, no resolver, no `-c`, no `-b`.

That bites as soon as you want a binary wrapped around your own generated interop, because it has to fill two roles at once:

1. **Running scripts** — the generated native namespace must resolve.
2. **Being the compiling side of `lg -b`** — top-level forms execute at AOT time, so a veneer's `(:require)` must resolve the native namespace *at build time* too. A bytecode-only host can't, and bundling against a stock `lg` yields a binary with the natives missing.

## Possible shape

An importable `pkg/cli` behind e.g. `func Main(version, commit string) int`, so a custom binary is roughly:

```go
func main() { os.Exit(cli.Main("dev", "none")) }
```

with the generated interop package blank-imported alongside.

Two constraints worth flagging:

- **`version`/`commit` have to stay in `package main`** — goreleaser and the Makefile both set `-X main.version` / `-X main.commit`, and ldflags only reach the declaring package. Hence passing them in.
- **Flags register in `init()` today.** In an importable package that mutates the global flag set at import, so a custom main can't register its own first. Registering from `Main` avoids it.

Mechanically this looks close to a pure move; happy to discuss the entry-point signature first.

abogoyavlensky:
Resolves #772
Resolves #774 

Two commits, reviewable in order.

**1. `feat(lginterop)`** - `-out-pkg <name>` emits a self-contained interop package that imports `pkg/rt` and installs from `init()`, so generated bindings work outside the let-go tree. Ambiguous or invalid names
(`-out-pkg rt`, Go keywords, non-identifiers) are rejected up front rather than failing later.

**2. `feat(cli)`** - moves the CLI out of `package main` into `pkg/cli` behind `Main(version, commit) int`. A third-party module can now build a binary that *is* lg - repl, resolver and `-b` included - not just a bytecode host. Flag registration moves from `init()` to a `sync.Once`, so importing the package no longer mutates the global flag set. `version`/`commit` stay at the root because ldflags only reach the declaring package.

Together they're what it takes to build a custom `lg` around generated interop: the first produces the package, the second produces something that can host it. The e2e exercises exactly that - generate for a stdlib package, blank-import it beside `pkg/cli`, run a script through the resulting binary, then bundle with that binary as its own bundle base.

Also fixes an emitter bug the e2e surfaced: `ex-struct?` tested the export vector's *length*, but all three `:type` branches serialize to 6 elements, so every named type registered as a struct. `crc32.Table` (`type Table [256]uint32`) panicked at init with "expected struct, got array". This is latent on `main` today - the in-tree xxh3 golden misses it because xxh3 uses `-opaque-structs`. Golden output is byte-identical.

New: `docs/guide/custom-lg.md`.


abogoyavlensky:
The idea behind this PR is what `lgx` needs to support Go dependencies. For `lgx.edn` to name a Go library, lgx needs both halves of this PR: bindings that work outside the let-go tree, and a way to build a binary that is a real lg and can host them. Then `lgx run` and `lgx build` work on projects backed by Go libraries.

nnunley (changes requested):
Reviewed at `ba4f3be64a694b3d4eec8be96e8e3e68419755f6`. Requesting changes for public API and generator contract failures that the current tests do not cover.

## Must fix

1. **Custom-host metadata overwrites let-go runtime identity.** `pkg/cli/cli.go:407-409` assigns the host strings passed to `cli.Main` to `rt.Version`/`rt.Commit`, while `pkg/rt/system.go` exposes those values as `let-go.version` and `let-go.commit`. The new guide explicitly says these strings describe the host. A custom binary calling `cli.Main("host-v9", ...)` therefore returns `"host-v9"` for `(System/getProperty "let-go.version")`, which can break runtime-version feature checks. Keep host CLI metadata distinct from the let-go dependency/runtime metadata and add a test where they differ.

2. **The documented local-replace custom-host `-w` path selects nonexistent `v0.0.0`.** `pkg/cli/wasm.go:54-62` reads `dep.Version` but ignores `dep.Replace`. The guide and e2e build custom hosts with `require github.com/nooga/let-go v0.0.0` plus a local replace, so real BuildInfo returns `v0.0.0` with a non-nil replacement. `gomod.Generate` then treats that as a release before considering `LETGO_SRC`. Reproduced with `GOPROXY=off`: stock-namespace `-w` fails on `github.com/nooga/let-go@v0.0.0` even with `LETGO_SRC` set. Model the real `Module{Version:"v0.0.0", Replace:...}` shape in tests and honor the replacement.

3. **Accepted self-contained generator inputs can produce invalid or overwritten output.** The new `-out-pkg` contract says the result is usable as an imported package, but:
   - aliases `vm`, or `fmt` with `-smart`, collide with generator-owned imports in `cmd/lginterop/lginterop.lg:341-348`; an out-of-tree `vm` alias reproduced `vm redeclared` and undefined target symbols;
   - `-out-pkg main` passes `validateOutPkg`, generates successfully, then cannot be blank-imported (`is a program, not an importable package`);
   - aliases `foo-bar` and `foo_bar` pass the raw `seenAlias` check (`cmd/lginterop/main.go:110-123`) but both normalize to `interop_foo_bar.go`; the second silently overwrites the first and the tool reports `2/2`.

   Some collision mechanics predate this PR, but the new self-contained-package contract knowingly exposes them as accepted external inputs. Either reject these names before writing or choose collision-free generated aliases/paths. Add compile/output-count tests for each case.

4. **Generation success and reproducibility claims are false for valid workflows.** `generatePackage` returns nil when a package has zero eligible exports (`cmd/lginterop/main.go:343-345`), so the new failure accounting reports `generated 1/1`, exits 0, and writes no file. I reproduced this with a local package containing only unexported declarations. Also, the new docs promise that the generated header round-trips, but `cmd/lginterop/lginterop.lg:367-383` omits a deps.edn alias, output path, and generator version. A custom alias therefore regenerates under the default namespace/filename; an unversioned `lginterop` command cannot pin the documented `go run ...@<version>` workflow. Define skip accounting explicitly, and either make the header reproducible or narrow the documentation claim. Add byte-for-byte regeneration coverage for a non-default alias.

## Should fix

`docs/guide/custom-lg.md:43-45` says the host controls shutdown because `Main` returns an exit code, but `flag.CommandLine` uses `ExitOnError`; help and malformed flags still call `os.Exit` inside `Main`, as `pkg/cli/cli.go:598-602` admits. Narrow the guide claim or give callers ownership of the FlagSet/args.

## Validation

The existing focused suites pass:
- `go test ./cmd/lginterop ./pkg/cli`
- `go test ./test/e2e -run 'TestCustomMain|TestLginterop' -short=false -count=1`

All GitHub checks are green on the reviewed head. These findings are uncovered contract cases, not existing-test failures.

abogoyavlensky:
@nnunley Thanks for the review. All five items are addressed, and the branch is updated from main.

1. **Runtime identity.** `cli.Main`'s `ver`/`com` now describe the host binary only and drive `-v`. `rt.Version`/`rt.Commit` resolve independently from let-go's own `BuildInfo` entry, so a custom host cannot leak into `let-go.version`. Covered by `TestRuntimeMetadataFrom` plus a `TestCustomMain` subtest where the two differ.
2. **Local-replace `-w`.** `letgoVersionFrom` follows `dep.Replace`: a directory replace resolves to `dev` (local-source path), a versioned replace pins to the replacement. Tests use the real `Module{Version: "v0.0.0", Replace: ...}` shape.
3. **Generator inputs.** New `validateEntries` runs before any write: it rejects aliases colliding with the emitted file's own imports (`vm`, and `fmt` under `-smart`), rejects `-out-pkg main`, and detects collisions on the normalized filename so `foo-bar` and `foo_bar` cannot overwrite each other.
4. **Accounting and header.** Zero-export packages are now reported as skipped rather than generated. `-packages` gained a `path=alias` form that the header records, so aliased files reproduce from their own command, plus a version line under `go run ...@<version>`. `-out` is deliberately omitted, since the file's location supplies it and an absolute path would differ per checkout. The docs state this.
5. **Docs.** The shutdown claim now notes that `flag.CommandLine` is `ExitOnError`, so `-h` and malformed flags exit inside `Main`.

`make test` and `make check-generated` pass locally.

pkg/cli/wasm.go
```diff
@@ -26,6 +28,62 @@ import (
...
+	for _, dep := range info.Deps {
+		if dep.Path != gomod.ModulePath {
+			continue
+		}
+		if dep.Replace != nil {
+			dep = dep.Replace
+		}
+		v := strings.TrimPrefix(dep.Version, "v")
+		if v == "" || v == "(devel)" {
+			return "dev", "none"
```
mparrett, line 121: **[P1] Preserve the replacement source for custom `-w` builds.** Following `dep.Replace` but returning only its version discards the replacement path. With the documented `require v0.0.0` plus `replace => /local/let-go`, this returns `dev`; `gomod.Generate` cannot discover that path from the custom module or executable and falls back to `@latest`. I reproduced `customlg -w` failing with `GOPROXY=off`; online it can silently build against a different release. Please carry the replacement path/mapping into the generated module.

cmd/lginterop/main.go
```diff
@@ -87,87 +117,121 @@ func main() {
...
-			continue
+		if ent.smart && normalized == "fmt" {
+			return fmt.Errorf("alias %q for %s collides with the fmt import smart mode adds to the emitted file — set a distinct alias", alias, ent.pkg)
+		}
+		if prev, dup := seenFile[normalized]; dup {
+			return fmt.Errorf("aliases %q (%s) and %q (%s) both normalize to interop_%s.go — outputs would collide; set a distinct alias",
+				prev.alias, prev.pkg, alias, ent.pkg, normalized)
 		}
-		okCount++
+		seenFile[normalized] = owner{pkg: ent.pkg, alias: alias}
```
mparrett, line 196: **[P2] Reject aliases that are invalid Go identifiers.** This validates import and filename collisions, but not whether `normalized` can be used as a Go import name. `-packages hash/crc32=for` exits 0 and emits `for "hash/crc32"` plus `for.Checksum`; `_` similarly produces invalid selectors. Validate the normalized alias against Go keywords and the blank identifier before generation.

docs/guide/custom-lg.md
```diff
@@ -0,0 +1,173 @@
...
+
+```go
+var version, commit = "dev", "none"  // -X main.version=... -X main.commit=...
+
+func main() { os.Exit(cli.Main(version, commit)) }
+```
+
+These describe *your* module. let-go's own version is read separately from build
+info where it matters (`-w` resolves it from your `go.mod`'s let-go
+requirement), so passing your version here is correct and won't mispin anything.
```
mparrett: **[P3] Correct the runtime metadata description.** This section says the host strings feed `System/getProperty`, contradicting both the earlier guide text and the new implementation, which deliberately exposes the linked let-go dependency as `let-go.version`/`let-go.commit`. The arguments drive host `-v` output, not those runtime properties.

mparrett (changes requested):
Reviewed the updated head. The focused CLI/lginterop suites and custom-main e2e tests pass, and GitHub checks are green. I found two remaining code-contract issues plus one documentation inconsistency; details are inline.

pkg/cli/wasm.go
abogoyavlensky: You were right that carrying
only the version discards the path. `pkg/gomod` gained `GenerateFrom`, which
takes an explicit let-go source directory and emits `replace <let-go> => <dir>`
through the existing `localFiles` path, validating the directory first and
falling back to ordinary resolution when it is not let-go's module. `pkg/cli`
extracts the path from `dep.Replace` in build info and passes it through.

Worth recording: my first cut still failed your repro. Go writes
`Replace.Version` as `"(devel)"` for a directory replace, not `""`, so the
guard I wrote rejected the real shape and the build still fell through to
`@latest`. `TestCustomMain` now has a subtest that runs `customlg -w` with
`GOPROXY=off`, which is what caught it: it fails with exactly your
`module lookup disabled by GOPROXY=off` without the fix.

cmd/lginterop/main.go
abogoyavlensky: `validateEntries` now
checks the normalized alias against the Go keyword set and the identifier
rules, sharing `isGoIdent` with `validateOutPkg`. `-packages hash/crc32=for`
and `=_` exit non-zero before anything is written; `=my-pkg` still normalizes
to `my_pkg` and is accepted. Covered by table cases in `TestValidateEntries`
and an e2e rejection test alongside the existing `-out-pkg` one.

docs/guide/custom-lg.md
abogoyavlensky: Corrected. That section now says the
arguments feed `-v` only, and states that the runtime resolves
`let-go.version`/`let-go.commit` separately from build info, honoring a local
`replace`. It no longer contradicts the earlier section or the implementation.

pkg/cli/wasm.go
```diff
@@ -26,6 +28,107 @@ import (
...
+		// A directory replace records the replacement PATH and no real
+		// version: Go writes "(devel)" there, or leaves it empty. A module
+		// replace records another module path WITH a version, and that one
+		// resolves through the proxy like any release, so it is not a source
+		// directory.
+		if dep.Replace == nil || dep.Replace.Path == "" {
+			return ""
+		}
+		if v := dep.Replace.Version; v != "" && v != "(devel)" {
+			return ""
```
mparrett: **[P1] Preserve the module path for versioned replacements.** A versioned replacement is reduced to `dep.Replace.Version`, while this branch discards `dep.Replace.Path`. `GenerateFrom` then receives no source directory and `Generate` constructs `github.com/nooga/let-go@v<replacement-version>`. For a normal fork replacement such as `replace github.com/nooga/let-go => example.com/fork v1.2.3`, that fetches the wrong upstream module/version (or fails) rather than reproducing the linked dependency. Please carry the full replacement directive into the generated module.

pkg/cli/wasm.go
```diff
@@ -26,6 +28,107 @@ import (
...
+		p := dep.Replace.Path
+		if !filepath.IsAbs(p) {
+			// go.mod may spell it relative to the main module. Resolving
+			// against the working directory is the best guess available;
+			// GenerateFrom validates the result and falls back if it is wrong.
+			if abs, err := filepath.Abs(p); err == nil {
+				p = abs
+			}
+		}
+		return p
```
mparrett: **[P2] Do not resolve relative replacements against runtime cwd.** Go build info preserves a replacement such as `../let-go` verbatim; `filepath.Abs` therefore anchors it to wherever the custom binary happens to be run, not the main module directory against which the replace was defined. Against this exact head I built a custom host with `replace => ../let-go-pr773-review`, ran it from another directory with `GOPROXY=off`, and `-w` fell through to `@latest` and failed. Either establish a recoverable module-root contract for relative paths or reject/document this unsupported case rather than claiming the local replace is honored.

mparrett (changes requested):
Re-reviewed the updated head at ef0e155b. The prior three findings are addressed and the focused suites pass. Two replacement-semantics gaps remain; both are based on the current head, not workspace drift. Details are inline.

pkg/cli/wasm.go
abogoyavlensky: You were right, and the fix is broader than the finding: the directive is now carried as a `(path, version)` pair and reproduced verbatim in the generated module, whether it points at a directory, a fork, or a pin of the stock module at another version. `gomod.Replacement` and `gomod.GenerateWithReplace` replace round two's `GenerateFrom`. For a module replacement the generated `go.mod` carries the `replace` line as-is and `go get github.com/nooga/let-go` on the placeholder require resolves the replacement and writes `go.sum` for it — no `go mod tidy`, same read-back path as the pinned and `@latest` cases.

Covered end to end: `TestCustomMain` now builds a host against `replace github.com/nooga/let-go => example.com/fork v1.2.3`, served from a `file://` module proxy that holds only the fork (this worktree, zipped), with `GOSUMDB=off`. `-w` can only succeed against that proxy by requiring the fork, and `let-go.version` through that host reports `1.2.3`. `letgoDepMeta` keeps reporting the replacement's version for runtime identity, which for a fork is the right answer.

pkg/cli/wasm.go
abogoyavlensky: Agreed. Build info records `../let-go` verbatim and carries no module root to anchor it, so nothing a built binary can do recovers that path honestly. I took the "reject and document" branch rather than a heuristic: a relative directory replacement now fails `-w` with

    let-go replace path "../let-go" is relative and cannot be resolved from a built binary; set LETGO_SRC to the checkout or build the host with an absolute replace

and never reaches the proxy. `LETGO_SRC` is checked *before* build info is consulted (`wasmLetgoSource`), so it can rescue exactly that case; round two had the precedence backwards, which is why setting it did not help against your repro. An absolute directory that is not a let-go checkout is an error too, for the same reason — silently building against `@latest` was the real bug in both shapes.

E2e: a host built with a `../..`-style replace, run from an unrelated cwd with `GOPROXY=off`, fails with that message (and the test asserts the `module lookup disabled` text is absent, so it did not fall through), then passes with `LETGO_SRC` pointing at the checkout. Unit tables pin the precedence and each build-info shape. The guide's `-w` section now states the contract.

abogoyavlensky:
@mparrett thanks — both addressed at `ed81d2b`, branch merged with `main`.

- **P1:** the whole replace directive (path *and* version) is reproduced in the generated module; a fork replacement builds against the fork. E2e uses an offline `file://` proxy that serves only `example.com/fork v1.2.3`.
- **P2:** relative directory replacements are rejected with an error naming `LETGO_SRC`, and `LETGO_SRC` now wins before build info is consulted. E2e runs the host from a foreign cwd with `GOPROXY=off`.
- While adding those I noticed `TestCustomMain` was never in the CI expensive-e2e lane (it is `-short`-gated and the lane runs a named list), so "checks green" did not cover this PR's headline e2e. It is in the lane now.

`make generate`, `make test`, `check-generated`, `make lint`, and the full `TestCustomMain` pass locally.

mparrett:
@nnunley appreciate a final review, if you have bandwidth

nooga (approved):
Approving. Every item from nnunley's and mparrett's rounds is in the diff (version/commit split in `pkg/cli/cli.go`, replace-path carrying and `LETGO_SRC` precedence in `pkg/cli/wasm.go`, alias/collision validation in `cmd/lginterop`). Build/vet/tests/check-generated pass locally.

One nit, non-blocking: the new `relative replace is rejected` subtest in `test/e2e/custom_main_test.go` fails on stock macOS because `filepath.Rel` at ~L132 doesn't see through the `/var` → `/private/var` symlink that Go's build tooling resolves — CI is green only because it's ubuntu. An `EvalSymlinks` on both sides before computing `rel` fixes it. Happy to merge with or without that.

nooga (approved):

abogoyavlensky:
@nooga Good catch — fixed in `b689c1d`: `EvalSymlinks` on both sides before the `Rel`, with a comment naming the /var → /private/var boundary. Verified the subtest still passes on linux; test-only change. Branch is already up to date with main.

nnunley (approved):
Re-reviewed current head `b689c1d`. The previous runtime-identity, replacement, generator-validation, accounting/reproducibility, and documentation findings are resolved in the current diff. The later full-replacement and relative-replacement findings are also covered, and the final macOS symlink-path test nit is fixed. Focused unit suites pass independently and all GitHub checks are green. Approving.

Note: the e2e test independently reached `TestCustomMain`; its fork-replace case could not complete only because the read-only source tarball lacked `.git` for `git ls-files`. The same e2e path is covered by green repository CI.

nnunley:
@mparrett I re-reviewed current head `b689c1d` and confirmed the replacement-semantics findings from your two requested-change rounds are addressed; focused unit suites and repository CI are green. I approved above. Could you re-review/dismiss the stale requested-changes state so this can merge after updating onto current `main`?

mparrett (approved):
Re-reviewed at `2aa872ed`. Both replacement-semantics findings are fixed, and fixed at the level they were about rather than at the call site. Clearing the requested-changes state.

**P1, the fork replacement.** `letgoReplacementFrom` now carries both halves of the directive, and `GenerateWithReplace` emits them: `replace github.com/nooga/let-go => example.com/fork v1.2.3` reproduces the fork, where before the path was dropped and the version alone produced a require against the stock module at a version that is not its own. `TestReplacedGoMod` pins the fork case and the same-path pin, and `TestLetgoReplacementFrom` covers both at the build-info boundary.

**P2, the relative directory replace.** Taken as the second of the two options I offered: refuse it rather than resolve it. `../let-go` now returns an error naming `LETGO_SRC`, so the case that previously fell through to `@latest` fails loudly instead. I checked the escape hatch works in the order the comment claims — `wasmLetgoSource` short-circuits on a non-empty `LETGO_SRC` before build info is consulted, so a host whose replace is unrecoverable still has a way out. `docs/guide/custom-lg.md` documents the limitation and the two workarounds.

The docs state the failure mode more usefully than my finding did: "so a fork is built against the fork, never against `github.com/nooga/let-go` at the fork's version number."

`go test ./pkg/gomod/ ./pkg/cli/` passes. The diff since the head I last reviewed is confined to these two fixes plus their tests and docs; the rest is `main` merging in.

Still needs updating onto current `main` before it can go in, as nnunley noted.

mparrett:
Heads-up on generated-artifact churn, since it affects your open PRs more than this one.

I merged a four-PR bytecode stack today — #501, #745, #502, and #624 — and all four touch `pkg/rt/generated.sums` and `pkg/rt/generated.manifest`. You rebased your open PRs a couple of hours before the first of those went in, so they are now behind main and will each want one more `make generate`.

The source changes are disjoint from yours; the overlap is the generated pair and nothing else. That makes it a regeneration rather than a conflict to work through, including on the branches where GitHub currently reports a conflict on those files.

Nothing else of ours is queued against them, so this should be the last of it.

PR #776, opened 2026-08-23
feat(vm): reflect proxy boxes multi-return funcs as vectors

### LINKED ISSUE #775: Reflect proxy silently drops Go return values past the second
## Problem

`pkg/vm/native_func_reflect.go` assumes a boxed Go function returns at most two values, and silently drops the rest:

```go
// assume lr == 2 && res[1] is error
```

`func() (int, string)` yields just the int; `func() (a, b, error)` yields `a` with the error ignored.

The drop happens on the Go side *before* boxing, which is what makes it awkward: a `.lg` veneer can never recover the lost values, and a wrapper library has no Go of its own to shim with. Nothing downstream can work around it.

`(a, b, ok)` and `(a, b, err)` are ordinary modern Go — `strings.Cut` returns three — so this rules out wrapping a fair number of APIs.

## Possible shape

Peel a trailing `error` as a throw, then box what remains as nil / a single value / a vector.

One case deserves care. **`func() error` currently returns the error as a value and never throws**, and every reflect-boxed `Close`/`Write`/`Flush` has that shape — peeling there would quietly turn `(if (.Close f) ...)` into a throw across a lot of working interop. Restricting the peel to "only when a non-error result remains", i.e. from `(T, error)` upward where it already applies, leaves existing behavior untouched.

Testing the *declared* result type also matters, so a function returning a concrete `*MyError` reads as a value rather than a failure.

Separately: `BoxValue` doesn't handle Go arrays (`[N]T`) — its `case reflect.Slice, reflect.Array` calls `IsNil`, which panics for arrays. Independent of this, but a multi-return change makes it reachable through m

abogoyavlensky:
Resolves: #775 

Same goal as #773: letting lgx projects depend on Go libraries. That only pays off if calling those libraries works, and the reflect proxy assumed at most two results, silently dropping the rest. Any `(a, b, ok)` or `(a, b, err)` function came back wrong, and `strings.Cut` returns three. The values were dropped on the Go side before boxing, so no `.lg` wrapper could recover them.

Independent of #773 and can land in any order.

### DIFF
--- a/docs/guide/go-interop.md
+++ b/docs/guide/go-interop.md
@@ -115,6 +115,31 @@ So the core of the wrapper is nearly free:
 `database/sql` is standard library, so the wrapper adds **no new go.mod
 dependency**. Only drivers are external, and they stay host-side (see below).
 
+### How Go return values arrive
+
+A boxed Go function's results map to let-go like this:
+
+| Go signature | let-go result |
+|---|---|
+| `func(…)` | `nil` |
+| `func(…) T` | the boxed `T` |
+| `func(…) error` | the boxed error **as a value** — does not throw |
+| `func(…) (T, error)` | boxed `T`, throwing when the error is non-nil |
+| `func(…) (A, B)` | vector `[a b]` |
+| `func(…) (A, B, …, error)` | vector `[a b …]`, throwing when the error is non-nil |
+
+A trailing `error` becomes a throw only when some other result remains to
+return. `func() error` — `Close`, `Flush`, and friends — hands the error back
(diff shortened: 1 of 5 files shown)


nooga (approved):
LGTM. Small, correct, tested (`strings.Cut`-style 3-return). Build/vet/tests/check-generated pass locally. Merging.

nnunley:
Queue update: #758 has merged and advanced `main`. GitHub now reports this approved PR as conflicting at `fd23603cd713`. Please update it onto current `main`; once the refreshed head is green, it is next in the integration order.

PR #778, opened 2026-08-23
fix(vm): make collections cross the Go/let-go boundary as native values

### LINKED ISSUE #777: []any does not cross the Go/let-go boundary as native values
Two mirror-image gaps in `pkg/vm`. Neither is exercised by anything currently in the let-go tree, which is why they survived. Both surface immediately when building a wrapper over `database/sql`.

**Boxing (Go to let-go).** `BoxValue`'s slice branch switches on the slice's *static* element type. For a `[]any` that kind is `Interface`, so every element misses the string/int/float paths and becomes an opaque `vm.Boxed`:

```clojure
(= "Ada" (first (sql/ScanRow rows)))   ;; => false
```

No error, just a wrong answer. The boxes print as `<go.string Ada>` and compare equal to nothing.

**Unboxing (let-go to Go).** Converting a let-go vector into a Go `[]any` parameter gives up when an element has no Go counterpart and hands the whole slice over as `[]vm.Value`, which the reflect call rejects:

```
reflect: Call using []vm.Value as type []interface {}
```

A let-go `nil` is the common trigger, so every SQL NULL parameter fails.

`[]any` is the natural Go type for "a row of unknown column types", and `nil` is the natural let-go value for SQL NULL, so a wrapper hits both at once. The worked `database/sql` example in `docs/guide/go-interop.md` does not work as written, which is the first thing a wrapper author copies

abogoyavlensky:
Resolves: #777

Slices and maps now cross the Go/let-go boundary as native values, in both directions, including when nested. Both gaps were found by building real wrappers: `[]any` from a `database/sql` wrapper (#777), maps from a [Wails v3](https://v3.wails.io) wrapper that had to declare `vm.Value` on every options parameter and convert by hand.

Maps land as commits on top of this PR rather than a second one, because the nesting fix edits code this PR introduces.

## What changes

**Boxing (Go to let-go).** `BoxValue` walked slices and maps by their static element type. For `[]any` and `map[string]any` that kind is `Interface`, so every element missed the scalar fast paths and became an opaque `vm.Boxed` that compares equal to nothing. Elements are now boxed by their dynamic type.

**Unboxing (let-go to Go).** Conversion into `any` is now total, and a new `unboxMapInto` converts a let-go map into any Go map. It is wired into both `unboxInto` and `boxArgForReflect`. Previously:

```
reflect: Call using []vm.Value as type []interface {}
reflect: Call using *vm.PersistentMap as type map[string]interface {}
```

**Nesting.** A collection reaching an `any` target converts to its natural Go shape. Without it the fix only works one level deep.

## Please review: unwrapping is an allowlist

Unwrapping routes elements into `BoxValue` paths the wrapped form never reached, and three of those panic rather than error. `ChanType.Box` spawns a goroutine calling `Recv`, which on a send-only channel panics in another goroutine and kills the process, with no way to recover at the call site. The paths recurse, so only a positive list is safe: predeclared scalars, `[]byte`/`[]int64`/`[]float64`, `[]any`, and exactly `map[string]any`. Everything else boxes as before.

## Key coercion

Keyword keys become string keys and do not come back as keywords, so `{:a 1}` round-trips as `{"a" 1}`.

Going to Go this follows existing behaviour: `unboxInto`'s `reflect.String` case already accepts a `Keyword`. Coming back, a Go string key boxes to a let-go string. Keywordising automatically would be wrong, since Go map keys can contain spaces and dots. This matches `clojure.data.json` with no `:key-fn`.

One consequence: a numeric key into a string-keyed Go map is an error, not a conversion. Go reports `int64` as `ConvertibleTo` `string` and converts it to a rune, so the fallback would turn the key `65` into `"A"`.

## Implementation notes

- Maps are identified by an explicit type switch on `vm.Map`, `*vm.PersistentMap` and `*vm.SortedMap`. Every map type returns `EmptyList` from `Seq()` when empty, so an empty map and an empty vector cannot be told apart by inspecting the first element.
- Maps are checked before sequences, since every let-go map is also `Sequable`.
- The nesting rule keys on what `Unbox()` returns (a `[]Value`), not on `Sequable`. `String` and `NIL` both implement `Sequable`, and `NIL.Seq()` returns itself, which recurses without end. Keying on `Unbox` also excludes a `Seq`, which may be infinite: a `LazySeq` over an infinite range stays unrealized rather than being iterated.
- An unhashable map key is a conversion error, not a panic. `reflect.Type.Comparable` cannot decide this, since a `struct{ X any }` holding a slice reports comparable and still panics when hashed, so the insert runs behind a local recover. It matters because `unboxMapInto` is reachable from `RecordToStruct`, which has no recover of its own.


abogoyavlensky:
Same goal as #773 and #776: letting lgx projects depend on Go libraries. A wrapper's first real job is returning a row of unknown column types, which in Go is `[]any`. Today every element of that slice arrives as an opaque box, so `(= "Ada" (first row))` is false with no error, and passing `nil` back fails the call outright. Wrappers can hand-roll around both, but then every wrapper author reimplements boxing.

nooga (approved):
LGTM. The `dynamicBoxSafe` allowlist is the right shape — it deliberately keeps interface elements away from paths like `ChanType.Box`. Verified locally, all green. Merging.

nnunley (approved):
The implementation and tests look sound, and focused `go test ./pkg/vm` passes independently. Approving the intentionally closed dynamic-unwrapping allowlist and the map/nested-collection support in this PR as the bounded fix for #777.

I filed #790 to design a broader conversion policy before extending automatic conversion to additional Go types. The branch will need conflict resolution against current `main` after #756.

PR #779, opened 2026-08-23
fix(ir): preserve capture-free closure identity

nnunley:
## Why

Under `*ir-compile*`, a capture-free `fn*` literal is a `:const` node carrying a `:fn-template` aux. The lowering treated it like any other cheap replayable constant: multiple local references re-ran `template-value` at each use site, manufacturing a distinct `*vm.Func` per reference.

Observable damage: closure identity breaks. `(let* [f (fn* [x] ...) g f] (identical? f g))` is false under IR compilation and true under the direct compiler; `=` on the rebound closure fails the same way.

## What changed

`pkg/rt/core/ir/lower.lg`:

- `fn-template-const?` recognizes `:fn-template`/`:multi-fn-template` const nodes.
- The cheap-load rematerialization path excludes them, so a function literal is materialized once per lowering site and every local reuse goes through `DUP_NTH` of that one value.
- The use-count sink rule treats a fn-template const as always worth a slot: it is an allocation with observable identity, not a replayable scalar.

Distinct textual `fn*` literals still produce distinct closures (`(identical? (fn* [x] x) (fn* [x] x))` stays false).

`pkg/rt/generated.manifest` / `generated.sums` are refreshed because `lower.lg` is a declared generator input.

## Verification

New `TestIRBytecodePreservesCaptureFreeClosureIdentity` in `pkg/ir/lisp_closure_template_test.go` pins rebinding identity, multi-reference identity, and the distinct-sites negative case under `*ir-compile*`.

`make generate` twice is deterministic, `make check-generated` passes, and `go test ./...` including the e2e suite passes on this head.

Complementary to #767: that fixes capturing closures at disagreeing joins in `lower_go.lg`; this fixes capture-free closure identity in bytecode `lower.lg`. No file overlap.

### DIFF
--- a/pkg/ir/lisp_closure_template_test.go
+++ b/pkg/ir/lisp_closure_template_test.go
@@ -57,3 +57,33 @@ func TestNestedMultiFnTemplateStillExecutesViaBytecodeLowering(t *testing.T) {
 		t.Fatalf("expected captured one-arity branch to produce 3, got %T %v", result, result)
 	}
 }
+
+func TestIRBytecodePreservesCaptureFreeClosureIdentity(t *testing.T) {
+	ensureLoader()
+
+	runLispExpr(t, `(binding [*ir-compile* true]
+		(eval (quote
+			(defn ir-closure-identity-regression [eq]
+				(let* [f (fn* [x] (+ x 1))
+				       f-again f]
+				  (eq f f-again))))))`)
+	runLispExpr(t, `(binding [*ir-compile* true]
+		(eval (quote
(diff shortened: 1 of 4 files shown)


nooga (approved):
LGTM. Identity across rebinding, multi-reference reuse via `DUP_NTH`, and the distinct-literal negative case are all covered by `TestIRBytecodePreservesCaptureFreeClosureIdentity`; read `lower.lg`'s diff independently and found nothing. Only the manifest conflicts. Merging after a rebase.

PR #780, opened 2026-08-23
build: bench-ratchet infrastructure — deterministic rebaseline, Go 1.26.5 baseline, pre-push gate

nnunley:
Three bench-ratchet infrastructure changes, split out of a local stack (the companion perf change follows separately with A/B evidence).

1. **fix(bench): make forced rebaseline reset deterministic bars** — `cmd/bench-ratchet`: a forced rebaseline now also resets the deterministic (alloc/bytes) bars instead of carrying stale ones forward; unit tests cover the reset path. `docs/perf/ratchet.md` updated to match.
2. **chore(perf): regenerate Go 1.26.5 baseline** — refreshes `docs/perf/baseline.json` after the toolchain moved from Go 1.26.3. Numbers are from the same Apple M3 machine as the existing baseline; single-machine baselines are what the ratchet fingerprint expects, flagging for reviewer awareness.
3. **build: enforce benchmark ratchet before push** — pre-push hook + Makefile target so a push fails locally when the ratchet regresses, instead of finding out in CI.

Each change is independent of the others in content but ordered (baseline regen assumes the deterministic-bar fix; the push gate assumes a current baseline).

Validation: `make generate` deterministic across two runs, `make check-generated` clean, `go test ./...` including e2e passes on this head.


nnunley:
Restacked onto `main@dadb8e0b` and corrected the M3 baseline provenance at `ea34ee4c`.

The original split accidentally retained `BenchmarkInitFromLGB` values captured from #781’s lazy-metadata candidate (`a998eead`): about 13,834 allocs and 1,004,581 B. Because #780 is independent, that made current main fail its new deterministic bars.

The baseline was recaptured under the exclusive benchmark baton from exact reachable `main@dadb8e0b` plus benchmark-only infrastructure:

- Apple M3, Go 1.26.5, `n=10`
- 26,294 allocs/op in every sample
- 1,599,971 B/op aggregate
- `captured_at_sha` / measured `best_since_sha`: `dadb8e0b`
- no diff from main under `pkg/compiler`, `pkg/ir`, `pkg/vm`, or `test`

Final validation:

- exclusive `make bench-ratchet`: 0 regressions / 0 missing
- actual `prek run bench-ratchet --hook-stage pre-push`: pass
- `make check-generated`
- bench-ratchet unit tests
- `make test`
- full `go test ./...`

No #781 implementation code is included in this PR.

mparrett (approved):
Reviewed the forced-rebaseline semantics, regenerated baseline, and pre-push integration.

Validated locally:
- `go test -count=1 ./cmd/bench-ratchet`
- `prek validate-config .pre-commit-config.yaml`
- semantic baseline comparison: 3 measured benchmarks changed across all 6 profiles
- prospective merge with current `main`: deterministic benchmark ratchet passes

No blocking findings. One non-blocking process caveat: the comments reference a repository `jj push` alias, but that alias is not tracked here, so JJ enforcement depends on external configuration.

mparrett:
@nnunley What do you think about sharing/documenting any jj-specific workflow pieces, like the `jj push` mentioned in the latest review round? This would be helpful for folks like myself on a cautious/exploratory adoption path.

nooga:
Reviewed cmd/bench-ratchet/main.go, main_test.go, and the ratchet.md delta (skipped the baseline.json regen itself beyond confirming it's a wholesale regen, not hand-edited).

forceRebaseline and filterUnstableBenchmarks are both well tested — the new tests actually exercise the cross-profile propagation and the six-variant filter, not just happy-path. Two things worth your own judgment call rather than a rubber stamp from me, since they're repo-policy rather than correctness:

1. The go1.26.5 baseline.json is captured on a single Apple M3 — you flagged this yourself. Worth noting the design already limits the blast radius: 'check' gates ns/op and ratio_to_anchor only against the *same machine profile* (main.go:454-459), and only allocs/bytes — which are supposed to be machine-independent — are compared globally. So a contributor on different hardware isn't held to this M3's wall-clock, only to its allocation counts, which is the fairer of the two gates to make global.

2. The pre-push hook makes bench-ratchet blocking on every push (always_run: true), gated on a baseline that currently lives on one contributor's laptop. That's a real repo-ergonomics tradeoff (fails-closed for everyone vs. only in CI) that's yours to decide, not something I'd wave through.

Nothing blocking from a correctness read.

mparrett (approved):

mparrett:
@nnunley — nooga's read covers correctness, and the two items he left open are policy calls on your own infra, so they belong with you rather than with me.

On the pre-push hook, one measurement that isn't on the thread yet: `make bench-ratchet` is about a minute, and the Makefile comment above the target says outright that it is what CI runs. The hook it is modelled on, `ir-stress-gate`, is deliberately not in CI, which is what makes its `always_run: true` pre-push stage the only gate there is. `bench-ratchet` already has a server-side gate, so this hook duplicates it and takes the per-push cost from roughly one minute to two. That duplication is what separates it from the sibling hook.

On the M3 baseline, nooga's own reading limits the blast radius: wall-clock is gated only within the same machine profile, and only allocations and bytes compare globally, which are machine-independent. That reads fine to me. The one addition I'd want is a line in `ratchet.md` recording who owns a profile, so a stale one can be retired rather than quietly gating.

Two mechanical things:

- The branch is 5 behind, and `main` now requires branches be up to date before merge. It lives on your fork, so the rebase is yours to push — I don't force-push contributor forks.
- This PR and #684 overlap on `cmd/bench-ratchet/main.go`, `cmd/bench-ratchet/main_test.go`, and `docs/perf/ratchet.md`, and both want to own `docs/perf/baseline.json`: this one regenerates it wholesale, and I said on #684 that the re-seed rides along there. My default is to land #684's re-seed first and rebase this on top of it, but I'll drop the re-seed and take your regen here if you'd rather.

mparrett:
Correcting myself on the pre-push hook, because I had the central fact backwards.

I said `bench-ratchet` already has a server-side gate and that the hook duplicates it. It does not gate. `perf-pr.yml` runs `bench-ratchet check` under `|| true`, and the comment right above it says so: "Inform-only: never fail the job on a regression … the exit code is ignored". `bench-ratchet` is also not among `main`'s required status checks. What misled me was the Makefile comment above the target, "Default gate (~1 min) … This is what CI runs": CI runs that profile, but only to produce the report.

That inverts my argument rather than weakening it. With no server-side enforcement, this hook would be the only thing standing between a perf regression and `main`, which is the same position `ir-stress-gate` is in and the reason its `always_run: true` pre-push stage is justified. The parallel you drew holds.

What survives is the cost, not the duplication: about a minute on top of `ir-stress-gate`'s, on every push. And nooga's point about the baseline sharpens rather than softens — you report on #781 that `make bench-ratchet` currently fails on `main` itself, the committed baseline predating go1.26.5. An always-run blocking hook plus a stale baseline fails every contributor's first push, so the refreshed baseline has to land in the same change as the hook. It does, both being here, which is an argument for this PR staying one change rather than being split.

The rebase and the `baseline.json` ordering question from my earlier comment still stand.

nooga (approved):
LGTM — mparrett's two approvals plus a local build/vet/test pass. The 3.6k-line delta is the regenerated `docs/perf/baseline.json`, not code.

Needs a rebase (6 behind main). Note this has a real conflict with #684 in `cmd/bench-ratchet/main.go`; since #684 isn't finished yet, this lands first and #684 rebases over it.

PR #787, opened 2026-08-27
feat(rt): JVM string/keyword interop surface and a StringBuilder shim

abogoyavlensky:
Adds the JVM host-interop surface Clojure libraries reach on their `:clj` branches. Motivated by running seancorfield/honeysql under let-go, but each item is a general facility.

- `java.lang.String` methods on native strings: `toString`, `length`, `isEmpty`, `charAt`, `indexOf` (string, char, and codepoint forms, optional fromIndex), `concat`, `substring`, `startsWith`, `endsWith`, `contains`, `toUpperCase`/`toLowerCase` (optional locale argument accepted and ignored, since Go's case mapping is locale-independent), `trim`. Indices are rune indices, which match Java's UTF-16 units for BMP text. Previously `(.concat "a" "b")` resolved to core `concat` through the name fallback and returned `(a b)`.
- `clojure.lang.Keyword` accessors: `.sym`, `.getName`, `.getNamespace`, `.toString`, `.hashCode`.
- `MetaValue` delegates `InvokeMethod` to the wrapped value. A `^Tag` hint in expression position compiles to a runtime `with-meta` wrap, so invokable values wrapped into `MetaFn` silently lost their `Receiver` surface: `(.sym k)` worked while `(.sym ^clojure.lang.Keyword k)` failed.
- `java.lang.StringBuilder` shim following the `host_arraydeque.go` pattern: ctor forms with an optional capacity or initial string, `.append` returning the builder for chaining, `.toString`, `.length`.
- `java.util.Locale` constants plus `getDefault`/`setDefault`/`forLanguageTag` as no-ops, since a default-locale change cannot affect let-go's case mapping.
- `java.net.URLEncoder/encode` over `url.QueryEscape`, which implements the same form-urlencoded rules.
- `unchecked-long/int/short/byte` accept characters as their codepoint, matching Java's char to int widening.
- Universal `.toString` fallback for any non-nil value. `nil` still fails loudly.
- `clojure-version` and `*clojure-version*` report the tracked language level.
- `seq-to-map-for-destructuring` implements Clojure 1.11 kwargs: pairs assoc'd in order, a single trailing map assoc'd in place, empty kwargs binds `nil`, odd count without a trailing map still throws. This supersedes the one-line `core.lg` change in #760.
- `readFormComment` loops past VOID, so `#_` skips comments before the form it discards instead of discarding the comment.

Tests: `test/host_string_interop_test.go`, `test/host_jvm_statics_test.go`, `test/clojure_version_test.lg`, `test/kwargs_trailing_map_test.lg`, `test/reader_discard_comment_test.lg`. `make test` and `make check-generated` pass.


nooga (approved):
LGTM. Rune-indexed string semantics and the StringBuilder shim look right, coverage is thorough, `make check-generated` passes locally.

One ordering note: #760 (yours as well) edits the same `fn-expand` line in core.lg. I'm landing #760 first; please rebase this over it and keep this PR's `seq-to-map-for-destructuring` version — it subsumes #760's nil-rest fix. Then I'll merge.

Issue #788, opened 2026-09-02
pkg/rt links pkg/glplat unconditionally, so x/image and x/text land in every default binary


mparrett:
On the glplat branch a plain, untagged build of `lg` links `golang.org/x/image` and `golang.org/x/text`. It costs **718,208 bytes, +3.60%** — 19,958,434 to 20,676,642 on darwin/arm64, measured 2026-09-02 — and `./cmd/lg-runtime` is affected too, which is the binary #658 shrank by 30%.

mparrett:
Fixed by #789 (merged 2026-09-02), which referenced this issue but did not close it.

Verified on `origin/main` @ `031ddc6b`: there is no `glplat` reference outside `pkg/glplat/`, and neither `golang.org/x/image` nor `golang.org/x/text` appears in `go.mod`. The premise here — those packages landing in every default binary — no longer holds.

One note for anyone arriving from #744: that branch still turns default-deps red after a rebase past #789. That is branch-local, not a `main` problem.

PR #789, opened 2026-09-02
build: gate untagged builds against the optional heavy subsystems

mparrett:
The GL and Ebitengine backends and the font stack they use are meant to be reachable only behind a build tag. One untagged file anywhere in their import closure defeats that, and nothing in CI notices, because the build still succeeds and the tests still pass while every shipped binary grows.

`scripts/check-default-deps.sh` runs `go list -deps` over both shipping targets, `.` and `./cmd/lg-runtime`, and fails if either links a tag-gated subsystem. It is wired up as `make check-default-deps` and as a `default-deps` CI job.

Motivating case and the measurement: #788.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -159,6 +159,28 @@ jobs:
   # Keeps the lg_no_http lane from rotting. Nothing in the default build
   # exercises pkg/rt/http_stub.go, so without this a change to http.go's
   # signatures breaks the tagged build silently.
+  # Default-build dependency gate. The optional heavy subsystems (glplat's GL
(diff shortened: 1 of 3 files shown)


nnunley (approved):
Reviewed the full three-file change. The gate covers both shipping targets, anchors external-module patterns so stdlib `vendor/golang.org/x/text/...` does not false-positive, and propagates `go list` failure separately from a deny-list hit. I independently ran the exact script on current main (pass) and with a fake dependency graph containing `github.com/ebitengine/oto/v3` (exit 1 with both targets reported). The script is executable and all GitHub checks, including the new `default-deps` job, are green. Approving.

Issue #791, opened 2026-09-04
bench-ratchet: IRCompile and InitFromLGB regressed since the #780 baseline (#727 allocs, #723/#730 timing)


nnunley:
The pre-push ratchet (`make bench-ratchet`) fails on current `main` (bdd8268, #730) against the baseline captured by #780 on 2026-09-02, independent of what a push contains. Bisected on an Apple M3, Go 1.26.5, plugged in (Low Power Mode off), in-repo `.workspaces/` checkouts, median of the lowest three of six samples per point.

| Commit | IRCompile [bytecode] ns/op | InitFromLGB ns/op | IRCompile allocs/op | bytes/op |
|---|---|---|---|---|
| 87579506 #780 (baseline) | 15.05M | 1.378M | 114,255 | 4,963,167 |
| 700a6b54 (before #727) | 15.01M | 1.357M | 114,255 | 4,963,144 |
| 95c17ec5 #727 PreparedCall in reduce | 15.07M | 1.495M | 116,149 | 5,080,395 |
| 3fd93d85 #723 cold error arms out of dispatch | 16.26M | 1.572M | 116,149 | — |
| bdd8268c #730 remove runLoop hop (tip) | 17.01M | 1.676M | 116,149 | 5,080,290 |

Net vs baseline: IRCompile [bytecode] +13%, IRCompile [gogen_ir] +6.5%, InitFromLGB +22%, IRCompile bytes/op +2.4% (over the 2% deterministic bar).

**#727 mechanism.** `ExecContext.PrepareCall` allocates an args slice, a `Frame`, and the `PreparedCall` struct on every `reduce` invocation (`pkg/vm/prepared_call.go`). The IR compiler performs many short reductions per compile, so it now pays ~1,900 extra allocations per `BenchmarkIRCompile` iteration. A length threshold before preparing, or reusing a prepared call across invocations of the same fn, would recover it; happy to send a PR.

**#723 / #730** each add several percent to both compiler benchmarks on this machine while targeting the Clojure test-suite benchmark. I have not looked at why.

**Two false positives worth a note in `docs/perf/ratchet.md`.** (1) With macOS Low Power Mode on (battery), the anchor read 1.85 ns/op vs the 1.061 baseline, +74%, which made every normalized number untrustworthy; `pmset -g | grep lowpowermode` is the first thing to check when the anchor moves on the same machine class. (2) The deterministic bytes/op bar is not arch-dependent: an amd64 build under Rosetta allocated the same as arm64.

[...]

PR #792, opened 2026-09-04
fix(browser-inspector): drop the tracked copy of coi-serviceworker.js; ignore dist/

nnunley:
`lg -w` already writes `coi-serviceworker.js` (the COOP/COEP shim the page registers) beside every `index.html` it builds, from the constant in `wasm.go`. The example carried an identical hand-placed copy under `dist/`, tracked in git, so `make clean`'s `rm -rf dist` deleted a tracked file, and the ~8 MB generated `index.html` next to it was never ignored, so every jj/git snapshot refused it.

- Remove the tracked copy; `dist/` is now entirely build output and ignored as a directory.
- README: the IR, optimized-bytecode, and lowered-Go panes have worked since #379; it no longer calls them placeholders and says what `build` writes.

Verified with `make -C examples/browser-inspector clean build`: `dist/` comes back with both files, the page loads and "Inspect Active Cell" fills all four panes, and `go test ./pkg/rt/wasm` passes.

Pushed past the pre-push ratchet: it is red on current `main` independent of this change, see #791.


mparrett (approved):
Reviewed 063ebb4. No findings. Confirmed the deleted worker is byte-for-byte identical to the generator constant; clean removes dist, a fresh browser-inspector build recreates index.html and coi-serviceworker.js, both outputs are ignored, and go test ./pkg/rt/wasm passes. The README description matches the current analysis-toggle and inspect-all wiring.

Issue #739, opened 2026-08-13
feature: fully implement the clojure.test.* namespace.


nnunley:
Implement the `clojure.test` contract so unmodified Clojure test tooling runs against let-go.

`clojure.test` is the surface the Clojure testing ecosystem depends on: tests are vars with `:test` metadata, results flow through the `report` multimethod, output goes to `*test-out*`, fixtures live on namespace metadata, and runners return a summary map. kaocha, the cognitect test-runner (#738), `clojure.test.tap`, `clojure.test.junit`, and editor integrations all consume that surface and none of them has a let-go mode. Our `clojure.test` has the same names but not the contract: it prints results directly, finds tests through a private registry, keeps fixtures globally, and signals success with a flag. That serves our own harness and nothing else, and it leaves a hole in compatibility at the first place a Clojure developer touches a new runtime, their existing suite and the tools they run it with.

Closing the gap pays three ways. Existing harnesses and reporters work unmodified, which is how a Clojure developer decides whether let-go is a Clojure. Running real suites through real runners turns "compatible" from a claim into a measurement, so future gaps surface as ordinary test failures rather than harness quirks. And the registry, the direct prints, the global fixtures, and the string-compiling Go harness are code only we maintain; the standard contract makes each of them unnecessary.

"Close enough" is not a target here. A harness that reads `(:test (meta var))` gets nothing from a registry, and a reporter bound over `report` sees nothing from a `println`. The bar is the JVM's observable behavior, with each deviation listed rather than implied.

Design: #793. Refs #738.

Issue #738, opened 2026-08-13
Feature Request: Get the cognitech test runner working against standard let-go


nnunley:
It looks like this might be another strong contender for providing good signal on clojure compatibility:

https://github.com/cognitect-labs/test-runner

From its readme:

> Clojure's 1.9 release included standalone tools for dependency resolution, classpath construction, and launching processes. Clojure also ships with a straightforward testing library, clojure.test.
> 
> Using these tools, however, there was no standard way to discover and run unit tests. Including a heavyweight project tool such as Leiningen or Boot just for the purpose of testing is overkill. Projects can build their own ad-hoc test runners, but these tend to lack features that will eventually be desired, and tend towards the "quick and dirty," besides being nonstandard from project to project.
> 
> This library aims to fill in the gap and provide a standardized, easy-to-use entry point for discovering and running unit and property-based tests while remaining a lightweight entry in Clojure's suite of decomplected project management tools.

Getting this to run unmodified will ensure that we conform to the clojure.test protocol that folks expect externally.

PR #793, opened 2026-09-04
docs(specs): specify clojure.test conformance, clojure.test.tap, and stack-trace primitives

nnunley:
Adds `docs/specs/clojure-test-conformance.md`, an NLSpec-style specification for porting `clojure.test` to Clojure 1.12.5 semantics so unmodified harnesses (kaocha, cognitect test-runner, #738) can drive let-go: metadata-based test discovery, the `report` multimethod and `do-report`, `*test-out*` / `with-test-out`, per-namespace fixtures, `test-var` / `test-vars` / `test-ns`, runners that return the summary map, and a port of `clojure.test.tap` with the plan printed last as the JVM oracle does.

It also specifies the runtime pieces the port needs. A stack trace is the error chain the runtime already builds as an error unwinds: each call site wraps the error with its callee and source position, the catch keeps that chain on the caught `ExInfo`, and `ex-trace` / `Throwable->map` and a `clojure.stacktrace` namespace read it, across the Go boundary and on both backends. Nothing is captured at throw time and there is no `current-stack-trace`. The rest: namespace metadata, exposing the reader's `FormSource` positions through `meta` / `&form` / `def`, `*file*`, `*out*` accepting `io/buffer`, and a Go harness that runs every namespace in which a loaded file defined a test, first through the summary `run-tests` returns and then through a `report` bridge that turns each deftest into a Go subtest.

It names #673, #671, #754 as the work whose behavior survives and whose mechanism (the registry) it replaces, relies on the typed-catch dispatch from #472/#476 as-is, and lists every deviation from the JVM oracle in an appendix with the oracle transcripts.

`testing-and-conformance.md` yields `clojure-test-api-design` to the new spec via `superseded-by`, and the docs README gains a topical-map row. Docs only; no code.

Pushed past the pre-push ratchet: it is red on current `main` independent of this change, see #791.

Refs #739 (fully implement the `clojure.test.*` namespaces) and #738 (run the cognitect test-runner unmodified). This PR is the specification for both; it does not close them. The implementation lands in the three delivery slices the spec defines (traces; runtime extensions plus the `clojure.test` / `clojure.test.tap` port; the Go harness bridge), and the slice that makes the harnesses run will carry the closing reference.


mparrett (changes requested):
I reviewed docs-only head ba70c599040ba16631f0cc974525f6373c113380 against the current runtime and the Clojure 1.12.5 oracle. CI is green and the document is impressively comprehensive, but I found five blocking correctness/design issues:

1. **The trace contract has no viable ownership model for arbitrary thrown values.** Sections 5.2-5.3 require `ex-trace` to work for every value, distinguish a never-thrown value, preserve a trace on rethrow, and assign a new trace to a newly thrown value. Today `ThrownError` owns only the raw `Value`, and `handleError` discards that wrapper when it pushes `errorToValue(err)` into a catch. A side table keyed by `Value` cannot satisfy the contract: some values (for example `ArrayVector`) are not Go-comparable, and equal scalars such as two independent throws of `"s"` have no identity with which to select the right trace. Please specify the trace-bearing representation/provenance that survives catch and rethrow while preserving the caught value's observable type/equality, and add an acceptance case for two equal scalar values thrown at different sites (including a nested catch).

2. **The specified `clojure.stacktrace` output is not the 1.12.5 output and the differences are absent from Appendix A.** The oracle prints ex-data on a new line, uses `" at "` only for the first frame and four spaces for later frames, and prints `" at [empty stack trace]"` for an empty trace. Sections 5.4 and 16.4 instead put ex-data on the header line, prefix every frame with `" at "`, omit the empty-stack line, and require two `at` lines at depth 2. Please either match the tagged implementation or explicitly document/test each intentional deviation. Oracle: https://github.com/clojure/clojure/blob/clojure-1.12.5/src/clj/clojure/stacktrace.clj#L40-L70

3. **The `*test-out*` initialization order cannot produce the promised WASM behavior as written.** Section 10.2 says bootstrap captures the root after an embedder installs its writer and cites `pkg/rt/wasm/rendermain.go`, but compiler/runtime package initialization (`pkg/compiler/init.go`) finishes before generated `main`, while that template installs `HostWriter` only inside `main`. No hook or generated-main update is specified to refresh `*test-out*` after `*out*.SetRoot`. The proposed capture therefore freezes the old writer and contradicts the Section 16.9 WASM acceptance case. Please define an explicit post-host-install operation (and require every root-writer replacement path to use it), or have the generated host setup update both roots atomically.

4. **The report bridge protocol is not self-contained and can misattribute or deadlock valid runs.** `ReportEvent` carries no snapshot/rendered form of `*testing-vars*`, `*testing-contexts*`, or trace output, yet `fail_text`/`error_text` read those dynamic vars and invoke let-go formatting on the harness goroutine, contradicting the stated rule that only the runner goroutine touches let-go state. In addition, composed deftests are explicitly supported, but `consume_var` treats nested `BEGIN_TEST_VAR` as an `ELSE` and returns on the nested `END_TEST_VAR`, so the rest of the outer test escapes its Go subtest. Finally, the goroutine pseudocode has no deferred close/recover/result send; an unexpected panic leaves `consume_events` blocked forever despite the behavior paragraph promising propagation. Please make the envelope self-contained on the runner goroutine, define nesting/depth handling, and define a deferred terminal result that closes the event stream and transports panics/errors to the harness goroutine.

5. **`print-tap-diagnostic` preserves a trailing empty field that Java's oracle drops.** Section 11.1 explicitly uses `keep_trailing_empty = true`; Clojure 1.12.5 calls Java `String.split("\\n")`, which discards trailing empty strings. For example, the oracle renders `"a\\n"` as only `"# a\\n"`, whereas the specified algorithm adds a second `"# \\n"` line. This also affects multiline error text captured with `with-out-str`. Please use the oracle split semantics and add trailing-newline coverage. Oracle: https://github.com/clojure/clojure/blob/clojure-1.12.5/src/clj/clojure/test/tap.clj#L51-L57

Once these contracts are resolved, the rest of the document gives a strong implementation and acceptance-test roadmap.

nnunley:
Updated and restacked in `ed0edff4502e6679b1261c3c9583602e1edf8edc` onto current main `a6763e774fd05bc90ce2d1663fcfde7354046c1e`.

All five review blockers are addressed:

1. **Throw ownership:** the spec now defines occurrence-owned `ThrowOccurrence`/`EvalDatum` transport, a context-preserving `DatumFn` invocation ABI, typed `LoweredDatum<T>` adapters, one-time `ErrorToDatum` normalization, datum-preserving collections/refs/natives/closures, and cause-datum storage in `ExInfo`. Acceptance covers non-comparable vectors, independent equal strings/integers, nested catches, rethrow, cause wrapping, and interpreter/lowered-Go specialized-path parity.
2. **`clojure.stacktrace`:** ex-data is on its own line; only the first frame has ` at `; later frames use four spaces; empty traces and `n <= 0` match the 1.12.5 oracle.
3. **`*test-out*`:** `install-host-output-roots` plus `register-test-out` handles either initialization order. WASM uses the paired root operation, while `api.WithStdout` binds both vars per execution context and temporary captures still bind only `*out*`.
4. **Go bridge:** runner-side freezing produces immutable Go-only envelopes; scope IDs recursively preserve composed deftests; `t.Error`/`t.Log` avoid format-string interpretation; and buffered terminal, panic recovery, cancellation, sequence validation, and drain-on-protocol-fault rules prevent stranded producers/deadlocks.
5. **TAP splitting:** the algorithm now specifies Java `String.split("\\n")` limit-zero behavior. Direct Java 26 and Clojure 1.12.5 probes confirmed that trailing empty fields are dropped, an interior empty field is retained, `"\n"` emits nothing, and `""` emits `"# \n"`. `String.strip` is unrelated and is not used.

The source/oracle audit also corrected canonical `:test/*-fixtures` keys, thrown-string trace rendering, `Throwable->map` root/omission/phase rules, `NO_SOURCE_PATH`, the explicit `read-string` metadata deviation, one-pass namespace discovery, the exact `are` root-cause class/message, and the Appendix B TAP exception transcript. Each now has a falsifiable Definition-of-Done case.

Verification on the exact pushed head:

- `git diff --check`: pass
- docs frontmatter for all three changed docs: pass
- internal spec contract/heading/link validator: pass
- `make check-generated`: pass (using a disposable in-workspace module cache because the sandbox-private external cache cannot be source-loaded by Go 1.26)
- independent adversarial source/oracle review performed; every reported finding was incorporated

@mparrett, please re-review when convenient.

mparrett (changes requested):
I reviewed the current restacked head (`ed0edff4502e6679b1261c3c9583602e1edf8edc`). The earlier five blockers appear substantially addressed, but I found three remaining correctness/design blockers:

1. **[P1] Scalar `is` forms can fail during macroexpansion.** Section 6.2 sends non-sequential forms to `:default`, but the normative expansion then evaluates `function?(first(form))` without first checking `sequential?`. As written, `(is true)`, `(is false)`, `(is :ok)`, and `(is x)` can error instead of going through `assert-any`. Clojure 1.12.5 uses `(and (sequential? form) (function? (first form)))`. Please restore that guard and add scalar acceptance cases.

2. **[P1] The trace design still lacks an execution-wide frame carrier.** `current-stack-trace` and `capture_live_frames(ec, err)` require the caller's live frames, but `DatumFn` receives only `ExecContext`, whose current contract contains bindings and scope rather than a logical frame chain. Interpreter parent links exist only inside one VM run, native-to-Lisp callbacks start another run, and lowered Go has no VM `Frame` objects. At the same time, `ThrowOccurrence` is immutable and an existing occurrence is never normalized or captured again, so outer native/caller frames cannot be appended after a callback returns. That makes the cross-Go-boundary requirement in Section 5.2(4), and the parity cases in 16.4, unrealizable as specified. Please define an execution-wide logical-frame stack, including interpreter, native, callback, and generated-Go push/pop/finalization rules.

3. **[P1] The mandatory `no test namespace` failure breaks the existing corpus.** The current `TestRunner` walk includes 18 `.lg` files with no `ns` form: all 16 `test/gold-aot/*.lg` fixtures, `test/top_level_do_test.lg`, and `test/in_ns_auto_refer_test.lg`. The last intentionally uses only `in-ns`, which Section 16.11 explicitly says must leave `saw_ns_form` false. Implementing Sections 12.2 and 16.11 literally therefore makes `go test ./test/...` fail, contradicting the Definition of Done. Please specify whether these are excluded/load-only or enumerate safe namespace migrations for them.

CI is green and `git diff --check` passes, but I do not think the specification is implementation-safe until these contracts are resolved.

mparrett:
Agent-drafted, posted by @mparrett. This continues the 2026-09-04 review on this PR (the five blockers). The 2026-09-08 review came from the same account through a different tool; this comment builds on it rather than restating it.

## Round-1 blockers: resolved

I re-read the restacked head `ed0edff4502e6679b1261c3c9583602e1edf8edc` against the runtime and the Clojure 1.12.5 sources. All five are closed:

1. **Trace ownership.** Section 3.5 now names an owner, a `ThrowOccurrence` allocated per throw and carried by `EvalDatum`, instead of a table keyed by `Value`. Section 16.4 covers the cases the old design could not: non-comparable vectors, two equal scalars thrown at distinct sites, nested catch, rethrow, and cause retention.
2. **`clojure.stacktrace`.** Checked against the tagged source. Ex-data on its own line, ` at ` on the first frame only, four spaces on the rest, `[empty stack trace]`, and `take (dec n)` all match.
3. **`*test-out*` initialization.** `install-host-output-roots` plus `register-test-out` is order-independent, and treating `api.WithStdout` as a per-execution binding matches what it already does.
4. **Report bridge.** Freezing on the runner goroutine, scope IDs for composed deftests, sequence validation, drain-on-fault, and a buffered terminal close the misattribution and stranded-producer paths. I traced the abandonment cases, and every early return drains first.
5. **TAP splitting.** Matches `String.split` at limit zero, and Section 11.2's `print-diagnostics` matches `tap.clj` exactly, including the `:pass` versus non-`:pass` branch.

## Three further findings

All three are about Section 3.5, and they build on the frame-carrier finding in the 2026-09-08 review rather than competing with it. If the capture side gains a logical-frame stack, the transport side still needs these answered.

**[P1] The datum carrier widens every typed lowered value, with no stated budget.** Section 3.5 makes `LoweredDatum<T>` the representation of typed parameters, captures, locals, and results in lowered Go, and `EvalDatum` the representation of operand-stack slots and locals in the interpreter. This is not a reboxing claim: the arithmetic stays a native `int64` or `float64` operation on `.value`, with no interface header and no allocation. Two costs survive that. A typed local goes from one register to two, and a slot that was pointer-free becomes pointer-containing, so locals and typed slices the collector skips today join the scan set. `Frame.stack` is a separate widening, from 16 to 24 bytes on an already-pointerful slot. `make bench-ratchet` is a blocking pre-push hook at 5% anchor-normalized, so Slice 1 has to clear it, and Section 14 already prices a cheaper case when it declines construction-time capture to avoid a stack walk per `ex-info`. I have measured none of this, which is why the ask is a budget rather than a number: please state the accepted budget and add a Definition of Done case that runs the ratchet, or narrow the carrier to values reachable from a catch binding.

**[P2] Datum-carrying collections change the exported `pkg/vm` surface, and no migration row covers Go embedders.** Section 3.5 lists collection key, value, and element storage as datum storage, and Section 16.4 requires a caught vector to keep its occurrence through `[caught]` and `first`. Since `EvalDatum` never implements `Value`, `vm.ArrayVector` (declared `[]Value` in `pkg/vm/vector.go`) has to become a slice of datums, along with the `Value`-typed slots in `PersistentMap` and `vm.Var`'s root and binding stack, which `pkg/api` writes directly. `ArrayVector` alone appears in 80 Go files, and `docs/guide/go-interop.md` documents this surface. Section 13's migration table covers only Lisp-visible changes. This is a completeness gap rather than a defect, so it does not bear on the verdict: please say what the exported surface looks like after Slice 1 and add the embedder row.

**[P3] The datum-preserving set is illustrative rather than closed.** Section 3.5 introduces the `DatumFn` operations with "including" and routes everything else through the legacy `[]Value` path, which drops provenance. Nothing states the rule that decides membership, so `(first [caught])` rethrows with the original trace while `(first (vec [caught]))` allocates a fresh one, because `vec` is an ordinary native. Both are user-visible through `ex-trace`. The same boundary carries an unpriced cost: handing a datum-carrying collection to a legacy native means projecting it down to `[]Value` per call. Please make the set normative and closed, or state the rule that determines membership.

The existing verdict stands. The corpus finding in the 2026-09-08 review is a real blocker, and none of the three above changes that.

nooga:
Thanks for the restack. The round-1 fixes are good: the trace-ownership rewrite, the clojure.stacktrace formatting, and the TAP split semantics (with the Java 26 / Clojure 1.12.5 probes). Keep those as is.

The remaining findings need two different responses, not one more revision pass.

Fix these directly in the doc:
1. The `is` macroexpansion guard (round 2, finding 1).
2. The `ns`-less corpus issue. Ruling: gold-aot fixtures are load-only and excluded from the discovery walk. `in_ns_auto_refer_test.lg` gets an explicit carve-out because it only uses `in-ns` on purpose.

Section 3.5 (the stack-frame/datum carrier) is different. Two review rounds have said it's unrealizable as specified against the current VM, and nobody has measured what it costs against `make bench-ratchet`. More spec text won't answer either question. What would: a small throwaway branch that threads a frame chain across one native-to-Lisp callback and reports the actual bench delta. Also, `make bench-ratchet` is already red on `main` for unrelated reasons (#791), so no cost number in this doc is trustworthy until that's fixed.

Before doing that work, answer this: does the kaocha / cognitect-test-runner path actually need `ex-trace` on non-exception values (a thrown vector, a thrown string), or does everything real go through `ExInfo`? If it's the latter, restrict the trace guarantee to Throwable-like values and document the rest as a deviation. That removes the datum-carrier redesign and the changes it forces on `ArrayVector`, `PersistentMap`, and `vm.Var` across about 80 files.

Requested changes:
1. Mark section 3.5 provisional/deferred, not blocking this PR. The spike is separate follow-up work.
2. Add a one-line ruling on the trace-fidelity bar (Throwable-like only, deviation documented) unless you have a concrete case that needs more.
3. Fold in the `is` guard and corpus fixes.
4. Rebase onto current `main` (13 commits behind).

Sections 6-11 (assertions, fixtures, running tests, reporting, TAP) are unaffected and have held up across two rounds. They're ready once the above lands.

nnunley:
Pushed `90f717aa` (child of `ed0edff4`) to the PR branch. It predates your comment and already lands most of what you asked for, so this is a status report against your list rather than a new revision.

**Your question first.** No, nothing real needs `ex-trace` on a non-exception value. Clojure cannot throw a non-Throwable, so kaocha, cognitect test-runner, and every ported library only ever see `ExceptionInfo` and host exceptions. The only scalar throws in this repo are 19 test fixtures, and Section 13 already migrates the 49 core `(throw (str ...))` sites to `ex-info`. So the bar is exactly what you propose: the trace guarantee covers exception values only, and thrown scalars are a documented deviation (Appendix A, "Trace of a thrown scalar").

**Section 3.5 is not deferred, it is replaced.** The datum carrier is gone. The trace is now the `ExecutionError` chain the VM already builds: every `OP_INVOKE` wraps the outgoing error with `calling <fn>` plus `SourceInfo` on the way out, and the only reason `ex-trace` could not see it was that the catch seam called `errorToValue` and dropped the chain. The catch now stores the chain on the caught `ExInfo`, set once by CAS, as a computed field excluded from equality and hashing. No live frame walk, no `ThrowOccurrence`, no `EvalDatum`, no change to `ArrayVector`/`PersistentMap`/`vm.Var`, and nothing for the ratchet to measure on the happy path. That also answers the round-2 frame-carrier finding without a frame stack, and it is why the spike is unnecessary: the design no longer threads anything across a callback. The one real gap it exposed is now Section 4.10: lowered Go propagates errors with a bare `return nil, err` (2,578 sites in the generated core, zero wraps), so the emitter gains a call-site wrap from the IR's source spans, error path only; and 76 natives flatten errors with `fmt.Errorf("%v")`, which become `%w`. `current-stack-trace` and `file-position` are dropped; `fail_position` uses the `is` form's source position from 4.2.

**Corpus ruling: matches yours.** 12.2 now skips `gold-aot` in the walk like `gogen`, and a file without an `ns` form is load-only. The failure is a `deftest` defined in such a file, since it would silently never run. `in_ns_auto_refer_test.lg` and `top_level_do_test.lg` pass unmodified (16.11).

**`is` guard: done.** The `sequential?` guard is in the 6.2 `:default` row with scalar cases in 16.5.

**Two things you did not ask for but affect review:**
- 4.9 pins `clojure.string/split` to `Pattern.split` limit-zero (vector, trailing empties dropped, limit arity). The runtime's split wraps Go `strings.Split` and diverges; the fix is #843.
- Section 17 makes the spec's assertions executable in-document: typed fences (`clj-repl` forms with inline `;=>`/`;; out:` expectations, `clj-table` rows, `clj-test` deftests) tangled and run against any Clojure engine, with `--accept` promotion of mismatches back into the doc. The first three fences ran green under Clojure 1.12.5, and that run caught a wrong assumption in the smoke test (a nested `run-tests` leaves the outer var on `*testing-vars*`), which is the argument for the mechanism.

**Still to do on this PR:** rebase onto current `main`, and the reply on the round-3 datum findings, which are moot by construction now.

Sections 6 through 11 are unchanged from what you reviewed. Implementation of them is under way on a separate branch off `main`, task by task against this spec, and is at the runners now.

nooga (changes requested):
Following up on the standing changes-requested review: the section 3.5 ruling asked for the stack-frame/datum carrier design to be marked provisional/deferred pending a throwaway benchmark spike. The latest push instead replaces 3.5 with a new, unreviewed normative design (ExecutionError-chain trace ownership, CAS-set trace on caught ExInfo, a call-site wrap across ~2,578 lowered-Go error-return sites, `%v->%w` on 76 natives) — that's a materially different ask than what was requested, and it's already being implemented against on a separate branch. There's also a new, unrequested Section 17 (executable doc fences with `--accept` promotion) that hasn't had a design pass.

The branch is also 13 commits behind main per its own last comment, with the rebase still outstanding.

Requesting changes: please either follow the original deferral ruling for 3.5, or make the case for the replacement design explicitly so it can be reviewed on its own; drop or separately propose Section 17; and rebase.

docs/specs/clojure-test-conformance.md
```diff
@@ -0,0 +1,1915 @@
...
+
+| Kind | Meaning |
+|---|---|
+| `LG` | Rendered as `at <fn> (<file>:<line>:<column>)`. Counted toward `*stack-trace-depth*`. |
+| `NATIVE` | Rendered as `at <fn> (<file>:<line>:<column>)` using the let-go call site, matching the runtime's behavior for `native fn`. Counted. |
+| `GO` | Rendered as `at <fn> (<file>:<line>) [go]`. Counted. Omitted entirely when a reporter requests `lg_only`. |
+
+A trace is a vector of `Frame`, innermost first.
+
+**The trace is the unwound error chain.** No capture happens at the throw site and no live frame stack is walked. A `throw` returns `ThrownError{value}` as a Go `error`; every call site the error passes through on the way out wraps it in an `ExecutionError` carrying that site's function name and `SourceInfo`, which is what the bytecode VM does today at `OP_INVOKE`. The chain that arrives at a `catch` therefore already lists every frame between the throw and the catch, innermost first. The design adds three things: the lowered-Go backend wraps the same way (Section 4.10), the catch keeps the chain instead of discarding it, and `ex-trace` renders it.
```
mparrett: **[P1] Record a source-bearing link at the throw instruction.** This chain contains only call-site `ExecutionError` wrappers. In the current VM, `(try (throw (ex-info "x" {})) (catch ...))` dispatches a bare `ThrownError` directly to the same frame's handler, before any call site can wrap it, so `frames_of` yields no frame at all. A throw inside a helper likewise makes the first recorded position the caller's invoke site, not the throwing subform. That contradicts 16.4's non-empty/throw-site requirement and 10.1's ERROR positioning. Please wrap `OP_THROW` and the lowered throw operation with their source, or specify another mechanism that records the throw instruction before local catch dispatch.

docs/specs/clojure-test-conformance.md
```diff
@@ -0,0 +1,1915 @@
...
+4. A trace crossing the Go boundary more than once (Lisp calls a native, which invokes Lisp, which throws) is one vector in call order, because the native returns the callback's error unchanged and its own call site wraps it. Each crossing contributes a `NATIVE` frame named for the native (`apply`, `map`, `reduce`, multimethod and protocol dispatch). These are the frames `internal?` filters in Section 10.1.
+5. A tail call (`OP_TAIL_CALL`) reuses the frame, so the tail-called function does not appear between its caller and callee. A self-recursive function in tail position shows one frame, not its recursion depth. This matches `recur` on the JVM.
+6. A catch stores the chain on the `ExInfo` only if it has none. Rethrowing a caught exception, `(throw e)`, starts a new `ThrownError` around the same object; the outer frames wrap it, and `ex-trace` returns the stored chain followed by the new one. Throwing a new `ex-info` from a catch gives it a new trace, and when the caught value is its cause, `ex-cause` reaches the old trace.
```
mparrett: **[P1] Reconcile first-catch CAS with rethrow semantics.** `catch_bind` refuses to replace a non-nil `chain`, and `ex_trace` reads only that stored chain. After `(throw e)`, the newly wrapped outer error chain reaches the second catch but is discarded because `e.chain` is already set; nothing can produce "stored chain followed by the new one" or satisfy 16.4's "gains the outer frames" case. This also conflicts with line 328's JVM claim: a Clojure 1.12.5 probe confirmed that rethrowing the same exception leaves its stack unchanged. Either keep the first trace unchanged and update this rule/DoD, or define storage that actually represents additional throw occurrences.

docs/specs/clojure-test-conformance.md
```diff
@@ -0,0 +1,1915 @@
...
+    chain   : Error | None              -- set once at first catch; never overwritten (CAS)
+
+FUNCTION catch_bind(err : Error) -> Value:
+    -- The single seam both backends use (vm.ErrorToValue today).
+    value = thrown_value(err)                       -- ThrownError.Value, or a boxed exception for a Go error
+    IF value IS *ExInfo AND value.chain IS None:
+        compare_and_swap(value.chain, None, err)    -- first catch wins; a reused exception keeps its first trace
+    RETURN value
+
+FUNCTION box_go_error(err : Error) -> *ExInfo:
```
mparrett: **[P1] Make the boxed Go cause chain advance and terminate.** The only boxing rule shown creates `ExInfo(cause = err)`. When `ex-cause` exposes that raw Go cause by boxing it, the result is another `ExInfo` whose cause is the same `err`; repeating `ex-cause` never reaches nil, so `Throwable_to_map`'s `cause_chain` can loop forever. Please distinguish the initial catch box from cause boxes and advance through `GetCause`/`Unwrap` (or make an opaque host-error box terminal) rather than retaining the exact error as its own repeatable cause.

docs/specs/clojure-test-conformance.md
```diff
@@ -0,0 +1,1915 @@
...
+            send(done, terminal)                   -- capacity one: never waits
+            close(events)                          -- runner is the sole closer
+
+        TRY:
+            WITH rt.WithBinding(ec, report_var, bridge_report(ec, events, cancel)):
+                loaded, load_error = load_file_with_result(path)
+                IF load_error IS NOT None:
+                    terminal = Terminal(INVOKE_ERROR, error_text = copy_text(load_error))
+                    RETURN
+                IF NOT loaded.saw_ns_form:
```
mparrett: **[P1] Apply the load-only rule in bridge mode too.** Section 12.2 now correctly accepts a no-`ns` file when its final namespace contains no `:test` vars, but the eventual default bridge rejects every such file here. Once `bridge` becomes the default this still fails the existing `in_ns_auto_refer_test.lg` and `top_level_do_test.lg`, contradicting 16.11 and the claimed corpus fix. Please mirror the 12.2 no-`ns`/deftest check in this path.

docs/specs/clojure-test-conformance.md
```diff
@@ -0,0 +1,1915 @@
...
+FUNCTION box_go_error(err : Error) -> *ExInfo:
+    -- A Go error that reaches let-go is boxed with the Go error retained as cause.
+    RETURN ExInfo(innermost_message(err), {}, cause = err)
+
+FUNCTION ex_trace(v) -> Vector<Frame> | None:
+    IF v IS *ExInfo AND v.chain IS NOT None: RETURN frames_of(v.chain)
+    RETURN None
+
+FUNCTION frames_of(err) -> Vector<Frame>:
+    -- Walk ExecutionError links outward. A "calling <fn>" link with SourceInfo is an LG or NATIVE frame.
```
mparrett: **[P2] Preserve the frame kind at the wrap site.** Current `ExecutionError` links carry only message, source, and cause, and the lowered wrapper in 4.10 has the same shape. Although `wrapCallErr` knows whether its callee is native or let-go code, that fact is discarded; function names cannot reliably recover it, especially across native-to-Lisp callbacks. `frames_of` therefore has no defined way to emit the required `:lg` versus `:native` kind or satisfy the `nth`/crossing cases in 16.4. Please add a kind discriminator to the link and define how VM and lowered call sites set it.

mparrett (changes requested):
Re-reviewed the latest docs-only head (`c1547720`) against the current VM control flow and the Clojure 1.12.5 oracle. The earlier `is` guard is fixed, the datum carrier is gone, and the branch is rebased; CI, `git diff --check`, and the docs frontmatter check are green. The replacement trace design still has four blocking internal contradictions plus one frame-model gap, detailed inline. I also ran a local Clojure oracle probe for rethrow behavior: rethrowing the same exception preserved exactly the original stack (`before == after`, 16 frames both times).

nnunley:
You asked for one of two things on §3.5: follow the deferral, or make the case for the
replacement so it can be reviewed on its own. This push makes the case, in §15 ("Why is the
trace the unwound error chain instead of a captured frame stack?"), and moves §17 out.

**Why not the spike.** The spike was meant to price a frame carrier: live frames captured at the
throw and threaded across native, callback, and lowered-Go boundaries. The replacement has no
frame carrier, so there is nothing for a spike to measure. What it relies on already exists and
is checked against `main`:

- The VM wraps every error leaving a call site as `ExecutionError("calling <fn>")` with the
  site's `SourceInfo` (`wrapCallSite` / `wrapCallErr` in `pkg/vm/vm.go`, both `//go:noinline`,
  error path only). Callback errors return through the same path.
- The chain is lost at one seam: `errorToValue` (`pkg/vm/errors.go`) returns a thrown value and
  drops the chain around it. Non-thrown runtime errors already get that chain as `:trace`
  strings in `ex-data`.
- `ExInfo` is an existing pointer struct and `WithMeta` copies it whole, so one added field
  (`chain`) carries through. No change to `ArrayVector`, `PersistentMap`, `vm.Var`, or the
  exported `pkg/vm` surface.
- #863 does not block on it: `test.lg` calls `ex-trace` only through `resolve` and falls back to
  the assertion position.

**Corrections to my 09-11 comment.** Two numbers there were wrong:

- The lowered core has about **19,000** error-return blocks, not 2,578 (`pkg/rt/core_go_lowered`,
  40 files, about 4.7 MB; 18,446 `return nil, callErr` plus try/catch paths), and none wrap.
  The emitter change touches each one, inside the existing `if err != nil` branch.
- **63** natives rewrap an error with `%v`, not 76.

**Costs, stated in the doc.** Generated-code size and `-tags gogen_ir` compile time grow; the
bytecode binary does not. The happy path does not change. Slice 1 records a before/after
`make bench-ratchet` on the same base (main is red for unrelated reasons, #791) and the lowered
core's size and `go build` time before and after.

**Section 17** now points to `docs/specs/executable-evidence.md` (#861); nothing in this spec
depends on it.

**Rebase:** done; the branch is on current `main`.

mparrett (changes requested):
Re-reviewed head `4fdbd415`. This push resolves the process/scope concerns from the previous round: the branch is rebased, Section 17 now points to the separate #861 proposal, Section 15 makes the case for the replacement error-chain design and states its measurement plan, and the `ExInfo` record now includes the existing `class` field. CI and `git diff --check` are green.

It does not change the five correctness contracts called out in [the preceding inline review](https://github.com/nooga/let-go/pull/793#pullrequestreview-5214966339), so all five remain:

1. **[P1] No throw-site frame:** 3.5/5.2 still say nothing is captured at `OP_THROW`; a throw caught in the same frame therefore reaches the handler as a bare `ThrownError`, and a helper throw starts at the caller's invoke site rather than the throwing subform. This still contradicts 16.4's non-empty trace at the throw site's file/line.
2. **[P1] Rethrow accumulation is unrepresentable:** `catch_bind` still sets `chain` only when nil and `ex_trace` still reads only that stored chain, while 5.2(6)/16.4 require the stored trace followed by new outer frames. The new push also leaves the conflict with 3.5's JVM claim; the Clojure 1.12.5 oracle keeps the original stack unchanged on rethrow.
3. **[P1] Boxed Go causes do not advance:** `box_go_error(err)` still stores that exact `err` as cause, and no rule says an `ex-cause` box advances through `GetCause`/`Unwrap`. Repeated `ex-cause`, and therefore `Throwable->map`'s `cause_chain`, can keep boxing the same error indefinitely.
4. **[P1] Bridge mode still rejects the load-only corpus:** 12.3 still returns `"no test namespace"` for every file without an `ns` form instead of applying 12.2's no-deftest/load-only rule. That contradicts 16.11 for `in_ns_auto_refer_test.lg` and `top_level_do_test.lg`.
5. **[P2] Frame kind is still missing from the chain:** `ExecutionError`/the 4.10 lowered wrapper still carry name, source, and cause but no native-vs-LG discriminator. `frames_of` therefore has no reliable input for the required `:kind`, particularly across native-to-Lisp callbacks.

The new rationale explains why the design was selected, but it does not make these acceptance cases implementable as currently written. The existing changes-requested verdict therefore remains.

mparrett:
A small implementation-oriented follow-up: I think the shortest path through the remaining trace findings is to use seams that already exist in this tree, without introducing a live frame carrier or datum-wide representation change.

1. **Create the missing throw-site link at `OP_THROW`.** `pkg/vm/vm.go` already has the current instruction pointer and `CodeChunk.LookupSource`. Before `handleError`, wrap the `ThrownError` in a source-bearing `ExecutionError` (error path only). Mirror that operation in the lowered-Go throw emission. This gives a direct/same-frame throw a real source frame; the existing invoke wrappers can continue adding caller frames during unwind. Store only immutable frame data, not a `Frame` pointer, since frames are pooled.

2. **Make rethrow explicitly first-trace-wins.** At the `errorToValue`/catch seam, attach the complete incoming chain to a caught `ExInfo` only when its stored chain is nil. If that `ExInfo` is thrown again, preserve the stored chain unchanged. That matches the Clojure/JVM probe and makes the CAS representation sufficient; Sections 5.2/16.4 should no longer require appending rethrow frames.

3. **Put frame kind in `ExecutionError`.** Add a small explicit LG/native discriminator and set it at construction sites where the boundary is known (`OP_THROW`, VM call wrapping, native invocation/adapters, and lowered-Go wrappers). Have `frames_of` consume that field; do not derive `:kind` from `"calling ..."` or a function-name convention.

4. **Advance before boxing a cause.** A single `nextCause(error) error` helper can use the project's `GetCause` convention and then Go `Unwrap` as the fallback. `ex-cause` should advance once and box that successor. `box_go_error(err)` may represent `err`, but it must not make that same `err` the next cause again. Nil terminates the chain; a small cycle guard would make malformed host errors harmless.

5. **Share the no-`ns` classification.** Let both the discovery walk and the explicit bridge call one classifier with three outcomes: namespace-backed test file; no-`ns`/no-`deftest` load-only file; no-`ns` file containing `deftest` (error). That removes the current 12.2/12.3 split and naturally covers `in_ns_auto_refer_test.lg` and `top_level_do_test.lg`.

The minimum closing tests I would want are: direct same-frame throw reports the throw line; helper throw reports the helper's throw line; rethrowing the same `ExInfo` leaves its trace unchanged; repeated `ex-cause` over a mixed `GetCause`/`%w` chain terminates; native-to-Lisp callback frames have explicit kinds; and both discovery and bridge accept the two existing no-`ns` load-only files. Run the relevant trace cases through bytecode and `gogen_ir` so the two wrappers prove the same contract.

Those changes seem sufficient to resolve the five current findings while keeping the replacement design's main advantage: all extra work stays on error paths and the existing public value/container representations remain untouched.

nooga (changes requested):
The 09-15 push resolved the process complaint from the last review (rebase done, Section 17 spun out to #861, design rationale for the replacement trace model added) — thank you for that. But on substance, the five P1 correctness contradictions from the prior inline review are still open: no throw-site frame capture, rethrow accumulation unrepresentable, boxed Go causes don't advance (infinite-loop risk in `ex-cause`/`Throwable->map`), bridge mode still rejects the load-only `.lg` corpus, and no native/LG frame-kind discriminator.

This is a docs-only spec, so nothing runs today, but merging it would encode a contract that (per the above) doesn't look implementable as written. Given this has gone through several async rounds without converging on the trace-ownership model, requesting changes here mainly as a marker — I think this one is better resolved with a direct sync between the two of you on that model rather than another round of review comments talking past each other.

nnunley:
@mparrett @nooga — new head pushed: one commit on top of `4fdbd415`, single file, +297/−98. This is still a docs-only spec; nothing here changes the runtime. It does name the runtime additions Slice 1 would make, listed at the end so they can be accepted or rejected on their own.

Drafted with Claude and checked by me; every runtime claim below was confirmed by running `lg`, not by reading code alone. That distinction turned out to matter.

## One correction to the review's premise

Finding 1 (no throw-site frame) assumed `OP_THROW` runs. Nothing emits it: it has been declared and dispatched since `1a1012748`, but `throw` is the native `CoreThrowf`, called through `OP_INVOKE`, and the VM wraps a native's error with the call site's source *before* the local handler runs. So `(try (throw …) (catch …))` already arrives with one link, `calling throw`, at the throw form:

```
(defn helper []
  (throw
    (ex-info "boom" {})))
(defn outer [] (helper) 1)
(outer)

stack trace:
  at throw (f.lg:3:5)
  at helper (f.lg:4:16)
  at outer (f.lg:5:1)
```

I made the same assumption in my first pass at this; a probe caught it. The spec now describes the mechanism that runs.

## The five findings

1. **Throw-site frame.** The throw site is the `throw` native's ordinary call-site frame: `:native`, `:fn` `throw`, at the throw form (3.5, 4.10, 5.2 rule 1). No throw-time capture, no special link kind.
2. **Rethrow.** Your option one: the first trace is kept unchanged. 5.2 rule 6 and the 16.4 item are rewritten; "gains the outer frames" is gone. Re-ran the oracle here (1.12.6): same object, same 16 frames before and after. Recorded in B.3.
3. **Boxed Go causes.** `box_caught` for the catch, and a cause box built per `ex-cause` read whose own cause is the *next* link (3.5). One chain step — `GetCause`, else `errors.Unwrap` — is shared by catch dispatch, `frames_of`, and the cause walk, because `unwrapThrown` steps only through `ExecutionError` today and 4.10's `%w` rule was unimplementable without it. `ex-info` keeps only `*ExInfo` causes, so the "thrown scalar" cause is removed.
4. **Bridge and load-only files.** Fixed, then replaced — see below; 12.2's own rule failed one of the files 16.11 names.
5. **Frame kind.** `ExecutionError` gains `kind` and `fn`. `kind` comes from whether the callee carries `SourceInfo`; lowered functions are boxed `NativeFn` with nothing on them today, so `NativeFn` gains an optional `source` and the emitter populates the existing `name` (4.10).

## What probing turned up beyond the review

- **12.2 contradicted 16.11.** `test/in_ns_auto_refer_test.lg` has no `ns` form and defines two deftests in a namespace it enters with `in-ns`, so "deftest outside a namespace" failed a file required to pass. clojure.test never asks how a namespace was entered, so `saw_ns_form` is gone. Both harness modes now share one classifier (12.2): snapshot every `:test` var and `test-ns-hook` before the load, then run every namespace that gained one, in one `run-tests` call. A deftest left in the starting namespace still fails the file, as harness policy. No loader change.
- **A call's position is its last argument's today.** The probe above reports `3:5`, the `ex-info` form, not `2:3`. The IR builder attaches the call form's own position to a `:call`, so the lowered backend would disagree with the VM once it wraps. 4.2 now requires the compiler to re-record the call form at the invoke instruction.
- **`map` is `.lg` and lazy**, so no `map` frame ever exists; the throw surfaces at the realizing call (`seq` / `dorun` / `doall`). 5.2 rule 4's example is now an observed `reduce` run; rule 9 covers lazy sequences; `map` leaves the 16.4 "holds through" item.
- **`nth` reports `native fn`.** It is registered by both `ns.Def` and a `//lg:native` marker, and the generated registrar's `SetRoot` path never calls `SetName`. The spec requires a var-bound native to report its var's name; `native fn` stays the honest name of a Go closure bound to no var.
- **`ERROR` position.** `internal?` is deleted. clojure.test's `do-report` takes the first stack element unfiltered — `(is (= 1 (/ 1 0)))` prints `ERROR in (t) (Numbers.java:190)` — so 10.1 does the same.
- Smaller: `frames_of` reverses the links but keeps a panic's Go frames in captured order; `Throwable->map` takes `:trace` from the innermost cause that has one and omits empty `:data`; the depth-100 item states observed counts.

## Runtime additions this spec now names

`ExecutionError.kind` / `.fn`; `NativeFn.source` and name population for lowered functions; the shared chain step in `unwrapThrown`; one source-map entry per call at the invoke instruction; `SetName` where a generated primitive replaces an existing var's root; `test/gold-aot` added to the walk's skip list. All error-path or compile-time; Section 15's measurement plan is unchanged.

Not in this PR, to be filed separately: the dead `OP_THROW` (delete or wire up), the IR treating `throw` as a call that returns, and the double native registrations.

A delta review against `4fdbd415` should be enough; the items most worth a second opinion are 1 and the harness classifier, since both change the shape of what was asked for.

mparrett:
The new head closes four of the five prior blockers cleanly and replaces the cause-loop bug with a finite traversal. The shared namespace classifier also looks sound. I would still request changes for three issues:

1. **[P1] Current-main CI is red.** The branch is 19 commits behind; main now contains #861, so its evidence blocks execute automatically. The [build check fails](https://github.com/nooga/let-go/actions/runs/35402158581/job/105784186960) because `clojure.test.tap` and the new trace behavior have not landed. Rebase and add `evidence: skip` to this planning spec until its implementation lands. [Section 17](https://github.com/nooga/let-go/pull/793/files) is now stale about #861 being merely proposed.

2. **[P1] Specialized VM operations still produce frameless caught errors.** The claim that every error starts at a call-site link overlooks `OP_ADD`, `OP_SUB`, `OP_MUL`, `OP_INC`, bit operations, comparisons, etc. Their failures go directly to `handleError(err)` without `wrapCallErr`, so `frames_of` sees only a kind-less host error and returns no frame. Re-recording source before `OP_INVOKE` does not affect these opcodes. The [capture rules](https://github.com/nooga/let-go/pull/793/files) need either an error-path wrapper for specialized operations on both backends or an explicit reduced guarantee and acceptance case.

3. **[P1] `box_caught` duplicates the head host error.** It uses the host error’s message to represent the caught exception, then stores that same host error as its cause. A terminal Go error therefore produces the caught box plus a second box with the same message, making `Throwable->map :via` one entry longer than the Go host-error chain—contradicting the DoD. Additionally, [`class = go_type_name(c)`](https://github.com/nooga/let-go/pull/793/files) cannot populate the existing `*ExceptionClass` field without a new canonical class facility, while Section 14 says finer host classes are out of scope. The simpler consistent model is for the normalized box to represent `host_error(err)` and store `next_link(host_error)` as its cause, using `ClassException`.

Two smaller corrections:

- [`NativeFn.source` as the LG/native discriminator](https://github.com/nooga/let-go/pull/793/files) conflates provenance with optional debug information. Stripped or sourceless functions can lose source data without becoming native. An explicit origin bit/enum is safer; source remains optional location data.
- [Fresh cause boxes are claimed to be `=`](https://github.com/nooga/let-go/pull/793/files), but current let-go and JVM exceptions use identity equality. Fresh boxes should be neither equal nor identical.
- The old “provenance-bearing datum path” sentence at [line 409](https://github.com/nooga/let-go/pull/793/files) should be removed.

nnunley:
Head `09b921b8`, rebased onto current main. All three, plus the small ones:

1. **CI.** The masthead now carries `evidence: skip` (the opt-out `scripts/spec-evidence.lg` honours) with the reason: `clojure.test`, `clojure.test.tap` and the trace primitives this spec describes are not implemented yet, so a spec-evidence run would exercise capabilities that do not exist. Section 17 describes #861 as landed and states the opt-out, and lifts when the implementation does.
2. **Specialized opcodes.** You are right that `OP_ADD`, `OP_SUB`, `OP_MUL`, `OP_INC`, the bit ops and the comparisons go straight to `handleError` without `wrapCallErr`. 4.10 now requires them to wrap their error the way the call to the native they specialize would — `CALL_NATIVE`, `fn` the operator's name, position the form's own (through the 4.2 source-map rule) — on both backends; 5.2 rules 1 and 2 say so; 16.4 pins `(+ 1 nil)` yielding a first `:native` frame named `+` at the form; Section 15 counts the error-path wrap.
3. **`box_caught`.** It is now the box of the head host error `h` itself: message `h.Error()`, `cause = next_link(h)`, so the head is not duplicated as its own cause, and `:via` has exactly one entry per link of the Go host-error chain (5.3 states it; a 16.4 item pins `(count (:via …))`). `go_type_name` is gone: boxes are class `Exception`, with finer host classes out of scope per Section 14. The `nth` item now reads: first frame `:native` `nth` at the call site, `(ex-message e)` the Go error text, `(ex-cause e)` nil when the Go error wraps nothing.

- Origin is an explicit `Origin` enum (`LG` / `NATIVE`) on `NativeFn`; `source` stays optional location data. Agreed that presence-of-source conflated provenance with debug information.
- Fresh cause boxes are neither `=` nor identical — identity equality, as on the JVM.
- The datum-path sentence is removed.

Docs frontmatter check clean; docs only. Drafted with Claude and checked by me.

nnunley:
New head `1768cf52`, one commit on top of `09b921b8`, single file.

Closes the last open tension between Section 1.7 (the reference TAP example, whose `math-test` fails on purpose) and Section 12.2 (a file with a failing deftest fails the harness). The example stays failing as written. It gains `^:expected-failure` on the deftest form, which is a var-metadata key (3.4) that `clojure.test` never reads: counts, printed output, and TAP lines are exactly those of any failing test. Only the Go harness reads it — the summary path compares `fail + error` against the count of marked vars, and the report bridge inverts the per-var subtest verdict (`t.Log("expected failure: ...")` when it fails, `t.Error("expected failure passed")` when it does not). Three DoD items added in 16.11; the 16.13 prose case now reads `--- PASS` for that subtest.

The spec-side change #863 needs is the same marker on `test/tap/tap-example.lg`; noted there.

mparrett:
Agent-drafted, posted by @mparrett. Reviewed head `1768cf5244febc58c678739b459880c198a64aeb`, diffed against merge base `ebd758a6`. CI is 13/13 green at this head.

## Closed since the last round

1. **CI.** Green at this head, `evidence: skip` is in the masthead, and `evidence-skip?` (`scripts/spec-evidence.lg:569`) reads it off the masthead lines as the opt-out. Section 17 no longer describes #861 as merely proposed.
2. **`box_caught`.** It now represents `h = host_error(err)` with `cause = next_link(h)`, so the head is not its own cause, and the `:via` length matches the Go chain. `go_type_name` is gone; boxes are class `Exception`.
3. **Specialized opcodes.** Section 4.10 now requires the wrap. The mechanism is right; the list is short, which is finding 3 below.

The `Origin` enum, the identity rule for fresh cause boxes, and the removal of the datum-path sentence all landed as described.

I checked the premise correction in your 22:34 comment by running it, and you are right: `(try (throw (ex-info "x" {})) ...)` already arrives carrying a `calling throw` link at the throw form, because `throw` is `CoreThrowf` reached through `OP_INVOKE` and `wrapCallErr` runs before the enclosing frame's handler. My earlier finding 1 assumed `OP_THROW` was live. It is not, and the mechanism you describe is the one that runs.

## Findings

**[P1] The expected-failure verdict compares an assertion count to a var count.** Both `run_file_tests` (12.2) and `run_file_tests_bridged` (12.3) compute `successful = (summary.fail + summary.error) == expected_failures(namespaces)`, where `expected_failures` counts *vars* carrying `:expected-failure` but `fail + error` counts *assertions*. A marked `deftest` with two failing `is` forms gives `fail = 2` against `expected = 1`: the file fails while `consume_scope` logs both failures through `t.Log` and issues no `sub.Error`, so the Go test fails with nothing attributed, which is the failure mode the bridge exists to remove. It breaks the other way too: two marked vars, one failing twice and one passing, gives `2 == 2` and a `successful` Terminal, while `consume_scope` calls `t.Error("expected failure passed")` on the second. 12.3's behavior note states the opposite — "`terminal.successful` and the per-var Error calls agree by construction: both compare `fail + error` against the same `expected_failures` count" — but `consume_scope` compares nothing of the sort; it keeps a per-scope `saw_failure` boolean, which is the correct shape. The identity holds only because Section 1.7's `math-test` fails exactly one assertion. The bridge already has the exact per-var answer and can use it for the Terminal; summary mode, which cannot attribute, needs either a per-var count or an explicitly weaker contract. No 16.11 item pins a marked var with more than one failing assertion.

**[P2] Section 4.9 specifies work that has already landed.** `clojure.string/split` was brought to `Pattern.split`'s limit-zero contract by #843 (`91e14248`, 2026-09-11), which is an ancestor of this head. I ran all nine rows of 4.9's own evidence table against a build of `1768cf52`: every one passes unchanged, `(vector? (str/split "a" #"\n"))` included. The sentence about the raw primitive in `pkg/rt/lang.go` is still accurate as written, since `core/split` is unchanged and still returns a list with trailing empties. But the conforming wrapper #843 added at `pkg/rt/core/string.lg:98` is already the `string_split` this section specifies, so 4.9 is not a remaining runtime extension, and Section 4's preamble ("the runtime lacks (September 2026), verified against the working tree") no longer covers it. The masthead's `last-verified: 2026-09-10` is one day before #843. A sentence recording that `split` conforms as of #843, with the evidence block kept as a regression guard, says the same thing truthfully.

**[P2] The specialized-opcode list omits three families.** 4.10 enumerates `OP_ADD`, `OP_SUB`, `OP_MUL`, `OP_INC`, seven bit operations and four comparisons. `tryFastOpcode` (`pkg/compiler/compiler.go:1023`) also emits `OP_DEC`, `OP_UNSIGNED_BIT_SHIFT_RIGHT`, and `OP_UNCHECKED_ADD`/`SUB`/`MUL`, and each of those calls `f.handleError` directly with no `wrapCallErr`, which appears only on the invoke arms at `pkg/vm/vm.go:1068` through `1185`. Verified by running it: `(dec -9223372036854775808)` inside a helper prints only the caller's frame, the same as `(+ x nil)` does today. Implemented as enumerated, `inc` overflow gains a frame and `dec` overflow stays frameless, which contradicts rule 1 of 5.2. The operator list a few words later ends in "...", so the intent may already be the general rule; the opcode sentence is the one an implementer will work from.

## Smaller

- Section 17 says the spec "carries no `@R-` tags of its own outside Section 4.9". There are three more: `@R-tap-diagnostic-split` at line 1144 (11.1) and `@R-smoke-run-and-tap` / `@R-smoke-error-text` at 1871 and 1932 (16.13). 16.10 requires the first of those to pass.
- 4.4 lists `:file` and `:line` among the `def` metadata keys to add. Both are already attached; `(meta (var f))` returns them today. `:ns`, `:name` and `:column` are genuinely missing.

## Open question

#863 solves the deliberately-failing reference example a different way: its `test/tap_test.lg` does not intern `math-test` with `:test` at load time and attaches the body only for the duration of each driving test, and `test/language_test.go` takes its verdict from `test/successful?` with no expected-failure accounting. If that approach stays, the whole `^:expected-failure` path here (the envelope field, `consume_scope`'s inversion, `expected_failures`, and the three 16.11 items) has no consumer. You flagged that #863 needs the marker. Which of the two is the mechanism? Worth settling before either is built against.

## Verdict

Docs-only; nothing here can break existing code or tests, so none of the above blocks the merge. The P1 is an internal contradiction worth settling before Slice 3 is implemented rather than after.

What I ran: built `lg` at the pinned head and probed the throw chain, the `dec` / `unsigned-bit-shift-right` / `unchecked-add` / `/` error paths, the 4.9 split table, `alter-meta!` on a namespace, `(meta (read-string ...))`, and var metadata. I did not run the Go suite; nothing in the diff reaches it.

mparrett (changes requested):
Independent re-review of `1768cf5244febc58c678739b459880c198a64aeb` (docs-only; 13/13 checks green). **No-go for merging this as the authoritative implementation spec yet; requesting changes.** The earlier CI, throw-site, cause-chain, and origin concerns appear resolved. My blocker is in the new expected-failure contract, not in running code.

**[P1] The file verdict counts assertions, while the marker and bridge verdict count vars.** In 12.2, `expected_failures` counts marked *vars* (lines 1292-95), but `run_file_tests` compares that with `summary.fail + summary.error`, a count of *events* (1309-10). 12.3 repeats the comparison for `Terminal.successful` (1564-68), while `consume_scope` correctly treats a marked var as expected-failing when it sees *one or more* fail/error events (1453-63). Thus a single marked var with two failing `is` forms passes its Go subtest but produces a false Terminal (and the summary mode fails it); two marked vars where one passes and the other fails twice produce a true Terminal but an `expected failure passed` subtest error. The statement that these verdicts “agree by construction” (1601-04) is false. This is observable as an unattributed failure or a false-positive summary result, and the spec's reference example hides it by having exactly one failing assertion. Please make the bridge Terminal derive its verdict from per-var outcomes and give summary mode an explicit, sound policy (or remove expected-failure support there); add both counterexamples to 16.11. Aggregate assertion counts cannot establish which marked vars failed.

Non-blocking accuracy fixes from this pass:

- Section 4.10's specialized-opcode list (line 517) omits `OP_DEC`, `OP_UNSIGNED_BIT_SHIFT_RIGHT`, and `OP_UNCHECKED_ADD`/`SUB`/`MUL`; the compiler emits them and their VM error paths call `handleError` directly. State the rule for *every error-producing specialized opcode*, with the list as examples, so `dec` and `inc` cannot diverge in Slice 1.
- Section 4.9 describes `clojure.string/split` as work to add, but `pkg/rt/core/string.lg` already supplies the vector/limit-zero wrapper from #843. Keep the evidence as a regression guard and label this an existing dependency, not a missing runtime extension. The raw `core/split` primitive can remain different.
- Section 4.4's `:file`/`:line` var metadata are already attached; `:ns`, `:name`, and `:column` are the missing pieces. Section 17's “no `@R-` tags ... outside Section 4.9” is also stale (11.1 and 16.13 have tags).

I would reconcile #863's temporary test-body approach with the marker before implementing the harness, but that cross-PR choice is **not** the basis of this changes request. `git diff --check` and the docs frontmatter check pass locally; I did not rerun the Go suite because this PR changes only docs and its CI is green.

nnunley:
@mparrett — head `f876d6d8`, one new commit on `1768cf52`, single file. Your P1 is right: the two verdicts did not agree, and the reference example hid it by failing exactly once.

**Bridge (12.3).** The consumer decides the file's result, from per-var outcomes. `ConsumerState` gains `all_vars_ok`: `consume_scope` clears it when a scope's outcome does not match its marker (unmarked and failed, or marked and clean), and `consume_events` clears it for a `:fail`/`:error` with no open var scope. `Terminal.successful` is gone; the summary rides on the Terminal for display only, and `run_file_tests_bridged` returns `state.all_vars_ok AND NOT protocol_failed`. A nested deftest's failures belong to the inner scope only, stated in place. The "agree by construction" bullet is replaced with one that walks both of your counterexamples.

**Summary mode (12.2).** Support removed rather than approximated: `(successful? summary)` is restored for unmarked files, and a file whose touched namespaces hold any `:expected-failure` var is skipped via `t.Skip("expected-failure requires the report bridge")` — loaded once, none of its tests run. I considered handing such a file to the bridge from there and rejected it: the marker is only discoverable after the load, so the file would load twice and repeat its load-time effects. `expected_failures` (a count) became `any_expected_failure` (a boolean). 1.8 and 13 now schedule `test/tap/tap-example.lg` with the bridge slice.

**16.11.** Both counterexamples added as items — one marked var with two failing `is` forms passes; two marked vars where one is clean and one fails twice fails, with `expected failure passed` on the first — and the tap-example item split into its bridge and summary-mode outcomes.

**Accuracy notes, all taken:** 4.10 states the rule for every error-producing specialized opcode, with the list as examples and `OP_DEC`, `OP_UNSIGNED_BIT_SHIFT_RIGHT`, `OP_UNCHECKED_ADD/SUB/MUL` added (checked against `pkg/vm/vm.go`). 4.9 is relabelled existing, kept as a regression guard; it also claimed the raw `core/split` has no limit arity, which is false, so that sentence now says what differs (list result, Go's `SplitN` count). 4.4: probed on current main, which since #863 merged attaches all of `:ns`, `:name`, `:file`, `:line`, `:column`, so it is marked existing too. Section 17 names 4.9, 11.1 and 16.13 as the tagged sections.

mparrett:
Re-review disposition for head `f876d6d817b7f982ae91038a047b93b250b7b7b3`: **NO-GO as-is, but the prior expected-failure blocker is closed.**

The new 12.3 consumer verdict is per-var (`all_vars_ok`), not an assertion-count/marked-var-count comparison. Summary mode now explicitly skips marked files instead of guessing, and 16.11 includes both counterexamples from my review. The opcode, `split`, metadata, and evidence-tag accuracy notes were addressed. I found no new algorithmic blocker in this commit; all 13 checks are green, and my local diff and frontmatter checks pass.

The remaining issue is the spec's baseline after #863 merged into main. Section 1.2 still calls the missing `report`, `run-all-tests`, `clojure.test.tap`, metadata discovery, and summary returns the *current* status quo; Section 1.8 still describes the test port as a future Slice 2 that depends on Slice 1 traces. Section 17 says `clojure.test` and `clojure.test.tap` are unimplemented as part of its `evidence: skip` rationale. Those statements are no longer true on current main, even though the trace/bridge work remains. Section 4.4, meanwhile, correctly describes metadata from #863 as existing, so the document now mixes pre- and post-#863 baselines.

Please rebase onto current main and make the overview, delivery status/order, and evidence opt-out distinguish what #863 already delivered from the remaining trace and bridge work. Keeping `evidence: skip` for the unimplemented trace/bridge evidence is reasonable; the explanation should name the actual gap. This is a documentation/merge-readiness no-go, not a renewed objection to the expected-failure fix. I have not rerun the Go suite because the PR diff is docs-only.

nnunley:
@mparrett — head `cf12ab41`: the stack is rebased onto current main (no conflicts) with one new commit that re-baselines the document. I audited each runtime piece the spec names against main — by reading the source and by probe — rather than editing the prose by memory.

**What the audit found.** Delivered by the port: the whole `clojure.test` / `clojure.test.tap` surface of Sections 6–11, `:test`-metadata discovery, per-namespace fixtures, the `report` seam, summary-map returns, `(meta form)` / `&form`, `def` metadata, namespace `alter-meta!`, `*out*` into an `io/buffer`, the `throw (str …)` → `ex-info` conversions (none left under `pkg/rt/core`), and the harness's summary mode of 12.2 including the touched-namespace classifier. Not delivered: the trace mechanism (3.5, 4.8, 4.10, 5 — `ExecutionError` has no `kind`/`fn`, and `ex-trace`, `Throwable->map`, `clojure.stacktrace` do not resolve), the call-position rule of 4.2, the report bridge (12.3), `LG_TEST_REPORTER`, harness support for `^:expected-failure`, and `test/tap/tap-example.lg`.

**Edits.**
- **1.2** — status quo, pain, and solution now describe main: the remaining gap is attribution and trace fidelity, not harness compatibility.
- **1.8** — the slice table gains a status column. Slice 2 is delivered with two named exceptions; the false "Slice 2 depends on Slice 1" sentence is replaced with what actually happened: the port landed first, `error-position` resolves `ex-trace` defensively and falls back to the assertion's own line until Slice 1 lands.
- **17** — `evidence: skip` stays, and the rationale names the real gap: `@R-string-split-limit-zero` and `@R-tap-diagnostic-split` already pass on main; `@R-smoke-run-and-tap` does not.
- Same treatment where the sweep found stale baseline language: Section 4's intro, 4.1, 4.2's reader list, 4.5 (which also wrongly claimed `io/writer` is resolved — it is not), 2.2's alias table, Section 12's intro, and Section 13's conversion row. Reviewed design in 3.5, 4.10, 5 and 12.3 is untouched, and no DoD box is ticked.

**One real finding from the audit, now Section 4.11.** `(deftest ^:expected-failure math-test …)` — the spec's own reference example — does not compile on main. The reader yields `(with-meta sym m)` for `^:k sym`; `def` unwraps it, but `varCompiler` asserts a `Symbol`, and `deftest` splices its name into `(var ~name)`, so any reader metadata on a `deftest` name dies with `interface conversion: vm.Value is *vm.List, not vm.Symbol`. `(def ^:k x 1)`, `(defn ^:k f [] …)` and a macro splicing into `(def ~n …)` are all fine; only `var` is affected. 4.11 specifies the fix (`var` accepts a `with-meta` operand; any other non-symbol is a compile error, not a panic), 1.8 lists it as the bridge slice's prerequisite, 16.11 gains a DoD item, and I'll file the compiler bug separately since it stands on its own.

mparrett (approved):
Approved from my review at head `cf12ab4182918e209ba3d104d20b15829f657d67`.

The expected-failure verdict contradiction I requested changes for is resolved: the bridge derives the file result from per-var outcomes, while summary mode skips marked files it cannot judge. The two counterexamples are in the DoD. This rebased head also corrects the current-main baseline after #863, distinguishing the delivered `clojure.test`/TAP and summary harness from the remaining trace and bridge slices. Section 4.11 identifies a real prerequisite for the bridge: I reproduced the `varCompiler` type-assertion panic from `(deftest ^:expected-failure ...)` on this head, and the spec now requires a non-panicking, metadata-aware `var` operand on both backends.

This is a docs-only PR. All 13 checks are green; `git diff --check` and the docs frontmatter check pass locally. I found no remaining blocking issue in the latest change.

Two non-blocking wording nits remain: the masthead comment above `evidence: skip` still says `clojure.test` and TAP await implementation, and Section 1.5 still calls the old registry-based `test.lg` design “current.” Section 17 and the revised overview correctly state the new baseline. This approval resolves **my** prior changes request; it does not speak for the owner's separate standing review.

nnunley:
@mparrett — thanks. Both wording nits taken in `c6fae4be` (one commit, two lines): the masthead comment above `evidence: skip` now names the actual gap (trace primitives and the report bridge) and says `clojure.test` / TAP are already in; Section 1.5 puts the registry-based `test.lg` in the past tense. No other change, so your approval's substance is untouched.

@nooga — your standing changes-requested on this PR predates the last several rounds; it is the remaining gate. The compiler bug the audit turned up is filed as #903.

---
Delivery verified by local `pr-claim-gate`: exact PR/validated head `c6fae4bed3d372745ec77103a413da6dbea0e098`; relevance claims `docs/specs/clojure-test-conformance.md` occur in both the delivered delta and the PR feature delta; receipt `/Users/ndn/.local/state/pr-claim-gate/nooga-let-go/pr-793.json`.

nnunley:
Head `5dc9bc55`, folding in your `run-all-tests` refinement from #929. Section 13's migration row for `(run-tests)` now states the failure mode rather than just the remedy — an aggregate runner holds no test vars of its own, so it runs nothing, reports no failures and exits 0 — and gives the regex form, since `all-ns` sees only loaded namespaces and the `:require` list has to stay the source of truth.

Same commit also carries the two wording nits from your approval: the masthead comment above `evidence: skip` and Section 1.5's tense. Nothing else changed, so your approval's substance stands.

@nooga — #929 is a concrete cost of this one sitting unmerged: #863's commit message cites this spec as the contract it implements, so someone migrating goes looking for a file that isn't in the tree. Your changes-requested predates several rewrites; a re-review or a dismissal is the only thing gating it.

---
Delivery verified by local `pr-claim-gate`: exact PR/validated head `5dc9bc551c0cf6d0eb15e2ff0cbecbd84eb68e19`; relevance claims `docs/specs/clojure-test-conformance.md` occur in both the delivered delta and the PR feature delta; receipt `/Users/ndn/.local/state/pr-claim-gate/nooga-let-go/pr-793.json`.

PR #794, opened 2026-09-04
perf(vm): make PrepareCall allocation-free for native fold loops

nnunley:
Fixes the allocation regression from #727 and restores a green `make bench-ratchet` on `main`. Bisect and numbers in #791.

#727 routed `reduce` and `some` through `PrepareCall`, which heap-allocated a `PreparedCall` and an argument slice per invocation and resolved the target through `resolveBytecodeCall`, which packs a rest list for variadic fns before the preparation could reject them. The IR compiler performs thousands of short reductions per compile, so `BenchmarkIRCompile` gained ~1,900 allocs/op, over the ratchet's 2% deterministic bar.

- `PrepareCallInto(p, fn, arity)` fills a caller-owned `PreparedCall`, so the hot loops keep it on their stack.
- Argument slots live in a new `Frame.prepArgs` that survives pool reuse and does not alias `argbuf`, which `installBytecodeCall` overwrites on a same-frame tail call.
- `unwrapBytecodeFn`, split out of `resolveBytecodeCall`, inspects the target without allocating, so variadic and wrong-arity fns are rejected before any packing. `PrepareCall` remains as the allocating form.
- `TestPrepareCallIntoDoesNotAllocate` pins the zero-allocation contract; the existing PreparedCall tests (tail-call rebind, error reuse, arity mismatch) still pass.

On an Apple M3, plugged in, `BenchmarkIRCompile [bytecode]`:

| | allocs/op | bytes/op | ns/op |
|---|---|---|---|
| #780 baseline | 114,255 | 4,963,229 | 15.05M |
| main (bdd8268) | 116,149 | 5,080,290 | 17.01M |
| this PR | 113,722 | 4,935,179 | 15.48M |

`make bench-ratchet` on this tree: deterministic OK, anchor +0.7%, IRCompile [bytecode] +2.0% ok, InitFromLGB -2.4%, IRCompile [gogen_ir] -10.2%, 0 regressions. `go test ./... -skip TestClojureTestSuite` and `go test ./test/e2e` pass; `make check-generated` clean (manifest and sums refreshed).

### DIFF
--- a/pkg/rt/generated.manifest
+++ b/pkg/rt/generated.manifest
@@ -143,7 +143,7 @@ pkg/rt/core_compiled.lgb pkg/rt/lg_bind_marker.go generator e9732750420945f868e1
 pkg/rt/core_compiled.lgb pkg/rt/math.go generator d5860fe3242df6e2afd6a6b69b0e5224e60ce94041a47d823d98225f878d5ed2
 pkg/rt/core_compiled.lgb pkg/rt/native_direct.go generator db57f47b71f19c1be0bf941797a8a0b065ba9575de22201a3abeab1dff0fc83a
 pkg/rt/core_compiled.lgb pkg/rt/native_direct_install.go generator 674bede5ceee03bb0d2af8b88f3290cfe15d1826768abcfeeb0aaf63e081cabe
-pkg/rt/core_compiled.lgb pkg/rt/native_prims.go generator df4e87fb76a3378156ff19a37b39c4317fb8f6399efedee180d34f430917ccec
+pkg/rt/core_compiled.lgb pkg/rt/native_prims.go generator f5c6734ede69cb6a3e01f9501c34a79faa685a7d83b3cb5b150d7e5232db242b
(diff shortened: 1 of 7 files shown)


mparrett (approved):
Reviewed head 1b4d8f01e3ef8bf7b160e1fc13e1b8d2323d7b1f. No findings.

I checked the PreparedCall lifecycle and frame-pool ownership paths, including same-frame tail-call rebinding and rejection of unsupported targets. `go test ./pkg/vm ./pkg/rt -count=1`, the corresponding race run, 100 repetitions of the PreparedCall tests, `make check-generated`, and `git diff --check` all pass.

Independent base/head allocation checks reproduced the intended improvement. On the ordinary bytecode path, steady `BenchmarkIRCompile` samples dropped by 2,427 allocs/op and about 145 KB/op. With generated IR enabled, steady samples also dropped by about 1,486 allocs/op and 55 KB/op. I did not use wall-clock results as review evidence because this host was under heavy unrelated load.

The current merge conflict appears limited to `generated.manifest` / `generated.sums`; rebasing and regenerating those files should be the remaining integration step.

PR #795, opened 2026-09-04
tooling: add bench-baton, a machine-quiescence lease for benchmarks

nnunley:
Benchmarks and the pre-push ratchet need a quiet machine; a test suite or a second benchmark running alongside produces numbers the ratchet cannot tell from a regression (part of what happened in #791). `cmd/bench-baton` arbitrates one pool per machine with two lanes:

- `bench` holds the pool exclusively (benchmarks, `make bench-ratchet`).
- `build` runs concurrently up to `--max-shared` slots (builds, suites, `make generate`).
- A waiting `bench` closes the gate, so a stream of builds cannot starve it.

Leases are `flock(2)`s, so a dead holder releases by itself; `state.json`, `ledger.jsonl`, and per-run logs live under `$BENCH_BATON_HOME` (default the user cache dir) per pool, in the same shape as the Python tool this replaces so an existing pool stays readable. A `!unix` stub keeps `go build ./...` clean elsewhere.

```
go run ./cmd/bench-baton bench --owner pr641 --cwd WT -- make bench-ratchet
go run ./cmd/bench-baton build --owner pr641 --cwd WT -- go test ./...
go run ./cmd/bench-baton status | ledger -n 10 | reap
```

The skill card at `.agents/skills/bench-baton/SKILL.md` (linked from `.claude/skills`, like `docs-status`) carries the rules: timing gates go through `bench`, never `build`; stable `--owner` per worker; cheap commands take no lease.

Tests pin the lane semantics: shared leases overlap up to the cap, an exclusive waits for shared holders, a waiting exclusive defers new shared work, `reap` drops dead holders, and `run` logs, ledgers, and returns the wrapped exit code.

Pushed past the pre-push ratchet, which is red on current `main` (#791) until the fix lands.


mparrett:
Nice one! Review incoming.

mparrett (changes requested):
Two blocking findings on 94c182d:

1. cmd/bench-baton/main.go:469 kills only the immediate child on timeout, then releases the lease. I reproduced this with a wrapped shell that started sleep in the background: run returned rc=127, status reported no holder, and the descendant PID was still alive. A timed-out make or test process can therefore keep consuming CPU while the next benchmark runs. Please terminate and wait for the full process group before releasing; the same lifecycle needs consideration when the baton itself is terminated so a workload cannot outlive its lease.

2. cmd/bench-baton/main.go:315-345 implements max-shared as a per-invocation slot search rather than a pool capacity. With one live holder started using --max-shared 1, a second default invocation immediately acquired slot1 and ran concurrently. Different workers can therefore silently defeat the requested cap, while status always reports the hard-coded value 4. Please make the capacity a pool invariant and reject mismatches, or otherwise enforce one consistent value and report the actual capacity.

The package build, vet, unit tests, and 10 repeated race-enabled test runs otherwise pass.

nooga (changes requested):
Confirming both bugs from the earlier review are still present at the current head (no commits since that review):
1. `main.go:~469` `cmd.Process.Kill()` kills only the immediate child, not the process group — a timed-out wrapped process can outlive its released lease.
2. `acquireSlot(pool, maxShared, deadline)` treats `max-shared` as a per-invocation parameter rather than a pool-stored invariant — a second worker with a different `--max-shared` silently defeats the cap, and `status` always reports a hard-coded 4 regardless of actual capacity.

Package tests and 10x race runs are clean, but neither of these is exercised by the current test suite. Requesting changes per the standing review — please fix both before merge.

mparrett (changes requested):
Requesting changes on current head `4a2eda6`. The rebase leaves the two previously reported blockers unchanged, and I found one additional portability blocker:

1. **Timeouts release the lease while descendants can still run** (`cmd/bench-baton/main.go:469`). `cmd.Process.Kill()` kills only the immediate child. On this head, wrapping `sh -c 'trap "" HUP; sleep 300 & echo CHILD_PID=$!; wait'` with `--timeout 0.2` returned 127 and cleared the holder while `pgrep` still showed the descendant `sleep 300`. Please run the workload in its own process group, terminate/wait for that group before releasing, and handle baton termination so a workload cannot outlive its lease.

2. **`--max-shared` is still per invocation, not a pool invariant** (`cmd/bench-baton/main.go:315-345`). I held slot 0 with `--max-shared 1`; a concurrent default invocation immediately took slot 1 and ran, while `status` reported the hard-coded capacity 4. Please persist/enforce one capacity per pool (or use one fixed capacity) and reject mismatches; status should report the actual value.

3. **The `unix` build tag includes targets where this code does not compile** (`cmd/bench-baton/main.go:1`). `GOOS=aix GOARCH=ppc64 go build ./cmd/bench-baton` and `GOOS=solaris GOARCH=amd64 go build ./cmd/bench-baton` both fail because `syscall.Flock` (and, on Solaris, the `LOCK_*` constants) are undefined. Since `main_other.go` is tagged `!unix`, those targets get no fallback. Please narrow the supported build constraint or use a portable locking implementation plus target-specific stubs.

Verification otherwise passes: `go test ./cmd/bench-baton`, `go test -race -count=10 ./cmd/bench-baton`, and `go vet ./cmd/bench-baton`; cross-builds for Linux, FreeBSD, OpenBSD, NetBSD, DragonFly BSD, and illumos also pass.

nnunley:
All three blockers are fixed in the new commit.

1. **Workloads cannot outlive their lease.** The wrapped command runs in its own process group
   (`Setpgid`). A timeout, a signal to the baton itself (SIGINT, SIGTERM, SIGHUP), or the command
   exiting while processes it started keep running all send SIGTERM to the group, wait up to 2s,
   then SIGKILL, and the lease is released only once the group is empty. A signal exits
   128+signal. Tests reproduce your case (a `sleep 30 &` child) for a timeout, a normal exit, and
   a stop signal, and require the child to be gone.
2. **`--max-shared` is a pool invariant.** The capacity is recorded in the pool's `state.json`.
   An idle pool records the requested value (default 4); a lease without the flag uses the
   recorded value; a different explicit value is refused while the pool has holders or waiters
   (the error names the recorded capacity and suggests `reap` if its holders are gone). `status`
   reports the recorded capacity instead of a constant. Your repro — a `--max-shared 1` holder
   plus a default invocation — now blocks the second, and a mismatched explicit value is refused.
3. **Build constraint.** `unix` is replaced by `linux || darwin || freebsd || openbsd || netbsd ||
   dragonfly || illumos`; `main_other.go` takes the exact negation, so aix and solaris build the
   stub instead of failing on `syscall.Flock`.

Verified: `go vet` (also under `GOOS=aix`), `go test -race -count=3 ./cmd/bench-baton`, and
cross-builds for linux (amd64, arm64), freebsd, openbsd, netbsd, dragonfly, illumos, darwin, plus
the aix, solaris, and windows stubs.

mparrett (changes requested):
One blocking signal-handling regression remains on current head `4026574`.

`signal.Notify` is installed before `run`, but `run` first blocks in `acquire` (`cmd/bench-baton/main.go:537`) and acquisition never observes the stop channel. Because notification suppresses the signals' default behavior, SIGINT/SIGTERM/SIGHUP cannot cancel a worker while it is waiting for the pool. I reproduced this by holding an exclusive lease, queuing a second `bench`, and sending SIGTERM to the queued process: it remained alive and present in `status` until the holder released; it then acquired the exclusive lease, started its command, and only afterward consumed the buffered signal and exited 143. A shared worker waiting for a slot is worse: it already holds the shared lease, so an attempted cancellation can continue delaying an exclusive benchmark.

Please make lease/gate/slot acquisition observe the stop signal and unwind its waiter record and open locks before returning (or avoid intercepting these signals until acquisition has completed), and add a queued-worker cancellation test.

The three previously requested fixes otherwise look correct. `go test ./cmd/bench-baton`, `go test -race -count=10 ./cmd/bench-baton`, and `go vet ./cmd/bench-baton` pass; the process-group and capacity tests pass, and the AIX/Solaris fallback builds now succeed.

nooga (changes requested):
All three prior blockers (timeout only killing the immediate child not the process group; `--max-shared` treated as per-invocation instead of pool-invariant; `unix` build tag not compiling on aix/solaris) are fixed in 4026574 and verified (`go test -race -count=10`, cross-builds for AIX/Solaris pass).

New gap found on this round: `signal.Notify` is installed before `run`, but a worker blocked in `acquire()` never observes the stop channel, so SIGINT/SIGTERM/SIGHUP can't cancel a queued (not-yet-running) worker — reproduced concretely, a queued process outlives SIGTERM and only exits after it eventually acquires and runs.

Requesting changes: route lease/gate/slot acquisition through a select on the stop channel, plus a test covering cancellation of a queued (pre-acquire) worker.

nnunley:
Pushed `1409796` (rebased onto `main` @ `ebd758a`), addressing the signal-handling blocker.

**What acquisition now observes.** Every wait inside `acquire` — the gate, the lease, and the shared slot — goes through the stop channel. `flockUntil` takes an extra `stop <-chan os.Signal` and polls with a `select` instead of blocking in the kernel whenever there is a deadline *or* a stop channel (it still blocks in `syscall.Flock` only when there is neither, so the uninterruptible path is unchanged). `acquireSlot`'s sleep becomes the same `waitPoll`. `acquire` also checks for an already-delivered signal before it joins the queue, so a signal that arrives just before the call cannot let the command start on an idle pool.

**How the waiter record and locks unwind.** A signal turns into a `*canceled` error carrying the signal, which flows out through the existing `fail` closure: it calls `unlockAll()` (gate, lease, and any slot already taken) and removes the token from `state.waiters` before returning. That is what makes the shared case safe — a shared worker takes the lease *before* it looks for a slot, so cancelling it while it waits for a slot now really does give the lease back rather than continuing to delay a queued exclusive. `mainWithArgs` maps `*canceled` to 128+signal, so a cancelled-while-queued worker exits with the same code as one cancelled while its command was running. Process-group termination of a running workload is untouched.

**Tests** (all written first and watched fail against the previous head — both reproduced your symptom exactly, "a queued worker sent SIGTERM did not return"):
- `TestCancelWhileQueuedForExclusive` — an exclusive holder is held, a second worker queues behind it (the test waits for its waiter record to appear, so it signals a worker that has really reached the queue), SIGTERM is sent. Asserts it returns within 5s, the error is a `*canceled` with exit code 143, the command's marker file was never created, no waiter record remains, and after the holder releases a fresh lease is grantable — proving no gate or lease lock was left held.
- `TestCancelWhileWaitingForSharedSlot` — `max_shared=1` with the slot occupied; a second shared worker gets SIGINT while waiting for the slot. After the occupant releases, an **exclusive** lease must be granted within 2s; it could not be if the cancelled waiter still held the shared lease.
- `TestCancelBeforeQueueing` — a signal already buffered when the run starts cancels it on an idle pool, leaving holders and waiters empty.

**Verification.** `go test ./cmd/bench-baton -count=1` (13 pass), `go test -race -count=10 ./cmd/bench-baton` (130 pass), `go vet ./cmd/bench-baton`, `gofmt -l cmd/bench-baton` (clean), `bin/lg scripts/lint.lg cmd/bench-baton` (no development-note findings), and cross-builds for linux, freebsd, openbsd, netbsd, dragonfly, illumos, aix/ppc64, solaris/amd64 and windows — all nine pass.

Note on the push: `SKIP=bench-ratchet` on the pre-push hooks, per the standing exception in #791/#899 (the ratchet needs a quiesced machine — which is what this tool exists to arrange). Every other hook ran and passed, including the full `go test ./...`, the Clojure compatibility suite, the ir-stress ratchet, and the generated-artifact freshness check.

mparrett (approved):
Agent-drafted, posted under `mparrett`. It supersedes my own `CHANGES_REQUESTED` reviews of 2026-09-15 and 2026-09-16 on this PR.

All four blockers from the earlier rounds are fixed at `1409796`. Each repro below was re-run against the built CLI:

- **Queued worker cancellation.** An exclusive holder plus a second `bench` queued behind it: `kill -TERM` on the queued process returned **143** in 0.03s, its command never ran (marker file absent), and the waiter record was gone. A fresh lease was grantable afterwards, so no gate or lease lock was left held. `SIGINT` to a shared worker waiting for a slot behaves the same way, and an exclusive lease was granted 0.04s after the occupant released.
- **Process groups.** `--timeout 0.5` wrapping `sh -c 'sleep 300 & ... wait'` left no descendant, and the pool showed no holder.
- **`--max-shared` as a pool invariant.** `status` reports the recorded `1`; a default second lease is refused a slot; an explicit `4` is refused with the recorded value named.
- **Build constraint.** Twelve cross-builds pass, including `aix/ppc64`, `solaris/amd64`, `windows/amd64`, and `js/wasm` on the stub.

One check the earlier rounds had not run: a cross-process stress of the lane invariants. Thirteen concurrent workers (ten `build`, three `bench`, `--max-shared 3`), each wrapped command timestamping its own start and end, over four repetitions. Zero violations every time: no two exclusives overlapped, no exclusive overlapped a build, and the shared lane peaked at the cap. The unit tests cover this with goroutines in one process; this is the same invariant across real processes.

## Exit code for a signal-killed command

Not a merge condition, and the only thing I found this round.

`rc = exitErr.ExitCode()` at `cmd/bench-baton/main.go:659` is `-1` when the wrapped command dies by signal rather than exiting, so the baton exits **255** and the ledger row carries `rc: -1`. Verified with `kill -9` on the wrapped process. The file already spells the shell's convention twice, at `main.go:655` for a signal to the baton and `:156` for a cancelled acquisition, and the package comment promises the wrapped command's exit code, so this is an inconsistency with the tool's own contract rather than an outside standard. A `make bench-ratchet` killed by the OOM killer hits it. One branch fixes it, here or as a follow-up; anything checking `rc != 0` behaves the same either way.

## Test gaps

Nothing exercises `mainWithArgs`: flag parsing, the `--` split, the exit-code mapping, and the `signal.Notify` wiring. Every test drives `run`/`acquire` with a hand-fed `stop` channel. The `rc = -1` path above is untested, and all tests run as goroutines in a single process, so the multi-process behaviour the tool exists for is asserted nowhere.

## Verification

`go test ./cmd/bench-baton -count=1`, `go test -race -count=10 ./cmd/bench-baton` (35s), `go vet`, `gofmt -l` clean, twelve cross-builds, and the five repros above, all at `1409796` in a worktree pinned to that commit. Fifteen CI checks are green and six are skipped.

The PR reads `BEHIND`. The maintainer's 2026-09-16 request for changes predates this commit and still stands on the PR.

nooga (approved):

Issue #796, opened 2026-09-04
entry-frame binaries run -main twice: main-chunk replay, then the frame


mparrett:
An `lg-compile --entry-frame` binary runs the program's entry twice when the
source ends with the documented top-level guard — once interpreted, then once
natively.

## Repro

Any program shaped like the idiom in `docs/guide/usage.md:78`:

```clojure
(ns demo)
(defn -main [] (println "hello"))
(when-not *compiling-aot* (-main))
```

Built the way `examples/aot/native-entry/build.sh` builds its example:

```sh
lg scripts/lg-compile --entry-frame out demo demo.lg
lg -c out/program.lgb demo.lg
cat > out/go.mod <<EOF
module demo
go 1.23
require github.com/nooga/let-go v0.0.0
replace github.com/nooga/let-go => /path/to/checkout
EOF
(cd out && go mod tidy && go build -o demo.native . && ./demo.native)
# prints "hello" twice
```

Verified on `bdd8268c` with the mandelbrot demo from mparrett/let-go-lab: the
20-frame benchmark ran 40 frames, the first 20 at VM speed and the second 20 on
the lowered path. That build needed three `math/*` sites hand-unboxed to compile
at all, per #562. The duplicate run is independent of that patch, and the
`hello` repro above reproduces it with no patching.

## Cause

The generated frame does both:

```go
if err := rt.RunProgramMainChunk(unit); err != nil { … }   // runs top-level forms
if _, err := prog.Main(ec); err != nil { … }               // enters natively
```

`RunProgramMainChunk` executes the program's top-level forms, and
`*compiling-aot*` is `false` at runtime by design (`pkg/rt/lang.go:3651`; set
`TRUE` only during `-c`/`-b`/`-w` at `lg.go:476`, documented at
`docs/guide/usage.md:71`). The guard passes, the top-level `(-main)` fires on the
VM, and then the frame calls the lowered `Main`.

This is an interaction with #729 rather than an oversight: the replay is there
on purpose, because `lg -c` emits bundles whose NS table is empty, so
`LoadProgramNamespaces` alone defines nothing and a VM-backed callee resolves to
a Nil var. Removing the replay would reintroduce what #729 fixed.

`examples/aot/native-entry/fib.lg` has no top-level `(-main)` call, which is why
the reference example doesn't show this.

## Why fix rather than document

The guard is the idiom the guide recommends for this situation, so the programs
most likely to be AOT-compiled are the ones that hit it. And the duplicate run
[...]

Issue #562, opened 2026-07-18
AOT lowering: native math/* intrinsics (sqrt, cos, sin, …)


mparrett:
**TL;DR: a float kernel that calls `math/sqrt` doesn't just stay boxed — it lowers to Go that doesn't compile. `math/*` calls emit `vm.Value` trampolines, but inference has already typed the surrounding sites `float64`, so the emitted assignment is `float64 = vm.Value`.**

## Context

AOT lowering (`lg-compile`) turns scalar arithmetic into native Go: the mandelbrot `escape` kernel lowers to a pure `float64` loop, ~72× over interpreted. But `escape` only ever uses `+ - *` and comparisons. Any numeric kernel that reaches for a transcendental — `sqrt`, `cos`, `sin`, `pow`, … — can't lower, and worse, the output doesn't build.

## What breaks

A path tracer is a good stress case: its inner loop is `sqrt`-heavy (normalize, intersect, cosine-weighted bounce). I extracted a scalar, allocation-free `radiance` + `trace-pixel` kernel (scalar `float64` ray params, scene in `double-array`, per-depth scratch) and ran `lg-compile` against current `main`.

`lg-compile` emits Go without complaint, but `go build` fails — 11 errors, all the same shape:

```
cannot use ec.Invoke(rt.CachedVarFn(&__v_math_sqrt, "math", "sqrt"),
  []vm.Value{…}) (value of interface type vm.Value) as float64 value in
  assignment
```

13 `math/{sqrt,cos,sin}` sites, each a boxed trampoline dropped into a `float64` slot.

## Root cause

`math/*` fns have no native lowering, so a call emits `ec.Invoke(rt.CachedVarFn(&__v_math_sqrt, "math", "sqrt"), …)`, which returns `vm.Value`. But the backward type pass has already typed the result site `float64` (the value flows straight into float arithmetic: `(let [sd (math/sqrt disc) r (- b sd)] …)`), so codegen emits `sd float64 = <vm.Value>` — a type error, not a fallback.

This is a notch worse than the collections gap (#358): there, everything stays `vm.Value` and at least *compiles* (boxed, slow). Here inference and codegen disagree about the type at the `math` boundary, and the result is non-compiling Go.

## Two things this is

[...]

Issue #551, opened 2026-07-17
AOT lowering: infer a float param type from float-contaminated arithmetic (so `^double` isn't required)


mparrett:
## Motivation

#357 gave float params a native lowering via an explicit `^double`/`^long` hint, and #534 fixed the mixed int/float widening at the op. Together a hinted float kernel lowers to native Go and hits the AOT speedup (see #425). But the hint is still required: the backward arg-type pass (`infer_arg_types.lg`) types any param used in arithmetic as `:int`, so an unhinted float param lowers to an `int` Go param and truncates the argument at the boundary.

The mandelbrot `escape` that surfaced #357 — `(defn escape [cx cy mi] ...)`, no hints — still lowers to `Escape(ec, arg0 int, arg1 int, arg2 int)` on current `main` and computes the wrong set. You only get the correct float lowering by adding `^double`. #357 was explicit that inference stays open; this tracks closing it.

## The gap

`infer-arg-types` maps `int-constraint-ops` (`+ - * < <= > >= inc dec`) → `:int`, with nothing for `:float`, so a param that flows into float arithmetic defaults to `:int`. The sound-join objection from #357 is real: `(+ x 1)` alone leaves `x` int-or-float, whose lattice join is boxed (`vm.Value`), not `float64`.

But there's a narrower, sound signal the pass ignores: float contagion. When a param meets a float-typed operand (a float literal, or a value already typed `:float`), the result is float and the param must be float to lower natively. In `escape`, `(+ (- zx2 zy2) cx)` adds `cx` to a `:float` (`zx2` is `(* zx zx)` with `zx` seeded `0.0`), so `cx` is provably float-contaminated — inferable without the hint.

## Direction (not a plan)

Propagate float contagion backward through arithmetic: if a param feeds an op whose other operand is `:float` (literal or inferred), constrain the param to `:float` rather than `:int`. Call-site coercion already handles an `int` argument reaching a `float64` param (`coerce-arg-to-type`), so an occasional int caller stays sound. Scope is the same `infer_arg_types.lg` backward pass, and the guard added in #530 (`assert-hint-agrees-with-inference!`) is already the tripwire for the contradiction this would introduce.

This is the "improve the type-inferencing logic to widen types" from the #536 review. Design belongs to whoever owns typeinfer; this is the capture plus repro.

## Repro

```clojure
(defn escape [cx cy mi]            ; unhinted
  (loop [zx 0.0 zy 0.0 i 0]
[...]

PR #797, opened 2026-09-04
fix(examples): ys-lower-report calls the public compile-form

mparrett:
`examples/aot/yamlstar/ys-lower-report.lg` has been unable to run since
`compile-form*` became private. Every invocation dies at load:

```
error: compiling let body
caused by: Can't resolve ir.passes.pipeline/compile-form* in this context
  --> examples/aot/yamlstar/ys-lower-report.lg:45:32
```

It never reports on a single form, so the per-fn lower/fallback report described
in `examples/aot/README.md` is currently unavailable.

## Why the public wrapper, beyond it resolving

`compile-form` normalises a `:multi-fn-template` result to `:status :lowered`.
Its own comment gives the reason: public callers key off `:status`, and without
the normalisation "a form that lowered via lambda-lift reads as unlowered".
`ys-lower-report` is one of those callers — it branches on `:status` and prints
anything other than `:lowered` as a fallback. So the wrapper is the correct
entry point here, not just the reachable one.

## Verification

Ran the tool over five in-repo namespaces, which exercises the whole of its
contract:

| File | Result |
|---|---|
| `examples/aot/native-entry/fib.lg` | `ok=2 fallback=0 threw=0` |
| `examples/aot/cross-package/src/lib.lg` | `ok=2 fallback=0 threw=0` |
| `test/native-entry/closure_capture.lg` | `ok=2 fallback=0 threw=0` |
| `test/native-entry/loop_carried_closure.lg` | `ok=3 fallback=0 threw=0` |
| `test/native-entry/conditional_closure.lg` | `ok=4 fallback=0 threw=0` |

`fib.lg`'s `ok=2` agrees with the `2 fns lowered` that `lg-compile` reports for
the same file.

The three closure fixtures are the lambda-lift shapes, and they only show that
the fixed tool reports them lowered. There is no before-and-after to compare,
because nothing ran before.

## Scope

Two lines: the call site and the header comment that describes it. Other
references to `compile-form*` in the tree describe the private function itself
and are left alone.

Found while using the tool to survey a demo's lowerability against tip, which
also produced #796 and the comments on #551, #562, #660 and #358.

### DIFF
--- a/examples/aot/yamlstar/ys-lower-report.lg
+++ b/examples/aot/yamlstar/ys-lower-report.lg
@@ -1,6 +1,6 @@
 ;; ys-lower-report — actually ATTEMPT to lower each single-arity defn in a file
 ;; and report :lowered vs fallback (with the error). Loads the file form-by-form
-;; first (so refers resolve), then runs compile-form* per lowerable form under
+;; first (so refers resolve), then runs compile-form per lowerable form under
 ;; *target* :go, catching failures.
 ;;
 ;;   ./lg scripts/ys-lower-report.lg <file.lg>   (one ns per invocation)
@@ -42,7 +42,7 @@
     (let [r (reduce
               (fn [acc f]
                 (let [nm  (form-name f)
-                      res (try (ir.passes.pipeline/compile-form* f)
+                      res (try (ir.passes.pipeline/compile-form f)
                                (catch e {:status :threw :err (str e)}))
                       st  (:status res)]
                   (cond


nooga:
## Review summary

Verified end to end:
- Reverted the fix locally and reran the tool — reproduces the exact reported error (`Can't resolve ir.passes.pipeline/compile-form* in this context` at the same line).
- Restored the fix and ran it against all five files listed in the PR body (`fib.lg`, `lib.lg`, and the three closure fixtures) — output matches the PR's table exactly, including `ok=2`/`ok=3`/`ok=4` and `fallback=0 threw=0` across the board.
- Confirmed the `compile-form` vs `compile-form*` distinction actually matters here, not just resolves: read `compile-form`'s normalization in `pkg/rt/core/ir/passes/pipeline.lg` (a `:multi-fn-template` result gets `:status :lowered` attached, since the private path keys off `:kind` instead) and confirmed the three closure fixtures are exactly the lambda-lift shape that would read as unlowered without it.
- Checked scope: grepped for remaining `compile-form*` references in the tree — all are the definition itself, doc/comment mentions, or internal test files that intentionally exercise the private function directly, not other broken external callers.
- `go build ./...`, `go vet ./...` clean.

Small, precise, well-verified fix. LGTM — approving.

nooga (approved):
Verified: reproduced the original error, confirmed the fix resolves it and matches the PR's reported output across all five test files, and confirmed compile-form's :multi-fn-template normalization is the correct (not just reachable) entry point here.

nnunley:
Approved and green at `cb90b439a05e2d3ee04ac1ae3709babe8d31747f`, but GitHub currently reports this branch as behind `main` (now `3aa9f4ea86d41dd55ced66fb3ca800c4bc66c355` after #800).

@mparrett, please update/rebase the contributor-owned branch and let CI rerun. I will not push to it. Once it is current-base and fully green, it is next eligible to land.

mparrett:
Rebased onto `3aa9f4ea` and force-pushed; head is now `1710dacc`. Clean rebase, diff unchanged — the same two lines in `examples/aot/yamlstar/ys-lower-report.lg`.

Re-verified on the new base. The pre-fix script still fails at the same line, and the fixed version reproduces the PR body's table: `fib.lg` `ok=2`, `lib.lg` `ok=2`, `closure_capture.lg` `ok=2`, `loop_carried_closure.lg` `ok=3`, `conditional_closure.lg` `ok=4`, `fallback=0 threw=0` throughout.

CI is green on the new head: all 15 checks pass.

nnunley:
Thanks for updating this to `1710daccaea872d062c8f722538bba2e8d479ca2`; its fresh checks are all green. Main advanced once more when the previously blocked #764 cleared and merged as `638b4a6aa1ab0aaadcfa42d29458e02bcbcd7bdd`, so the PR is again one base step behind our exact-current-base merge rule.

@mparrett, please update the contributor-owned branch once more and let CI rerun. I will not push it. I’ll merge after confirming the new head/base, approval, mergeability, and every required check.

mparrett:
Updated again: head is now `9ee86d1f`, a single commit on `638b4a6a`.

The branch had picked up a merge commit from the Update-branch button (`acfe213b`, "Merge branch 'main' into fix/ys-lower-report-public-api"). I replaced it with a rebase, so the branch is one commit on current main rather than a merge. `main` requires linear history here, and it keeps the head/base check unambiguous.

Re-verified on the new base, since #764 changes small-map ordering and the IR pipeline reads those maps: the five files still report `ok=2`, `ok=2`, `ok=2`, `ok=3`, `ok=4`, with `fallback=0 threw=0`.

All 15 checks green.

PR #798, opened 2026-09-05
fix(test): restore namespace after explicit run-tests

jasalt:
## Summary

Fix `test/run-tests` failing after otherwise successful tests when called with explicit namespaces.

The explicit-namespace branch saves `*ns*` as a namespace object, but attempted to restore it with:

```clojure
(in-ns (symbol (name old-ns)))
```

`name` only accepts symbols, keywords, or strings in let-go—not namespace objects—so restoration throws:

```text
error: name expected Symbol, Keyword or String
```

The no-argument branch already handles namespace objects correctly with `ns-name`. This change makes the explicit branch use the same conversion:

```clojure
(in-ns (ns-name old-ns))
```

## Reproduction

```clojure
(ns example-test
  (:require [test :refer [deftest is]]))

(deftest passes
  (is true))

(test/run-tests 'example-test)
```

Before this fix (lg 1.12.2), the assertion passes, but `run-tests` throws while restoring the caller namespace, causing the command to fail:

```clojure
example-test=> (test/run-tests 'example-test)
Running tests...
PASS true
PASS true
error: name expected Named
  --> <embedded:test>:80:29

stack trace:
  at name (<embedded:test>:80:29)
  at anonymous fn (<embedded:test>:59:33)
  at anonymous fn (<embedded:test>:72:6)
  at run-tests (REPL:1:1)
```

After this fix, the test completes successfully and the caller’s namespace is restored:

```clojure
...
example-test=> (test/run-tests 'example-test)
Running tests...
PASS true
Finished running tests. Tests: 1 Pass: 1 Fail: 0 Error: 0
#'test/*test-result*
```

### Clojure reference

Babashka v1.13.219:
```clojure
(ns example-test
  (:require [clojure.test :refer [deftest is]]))

(deftest passes
  (is true))

(clojure.test/run-tests 'example-test)

;; Testing example-test
;; Ran 1 tests containing 1 assertions.
;; 0 failures, 0 errors.
;; => {:test 1, :pass 1, :fail 0, :error 0, :type :summary}
```

## Changes

- Restore the saved namespace with `ns-name` when explicit namespaces are passed.
- Add a regression test verifying:
  - tests execute in the explicitly requested namespace;
  - `run-tests` returns without a restoration error;
  - the caller namespace is restored afterward.
- Regenerate the embedded runtime bundle and generated manifests.

## Testing

- `make check-generated`
- `make test`
_____
LLM assisted PR (gpt-5.6-sol)


nooga:
## Review summary

Verified the fix directly: reverted just the one-line change (`(in-ns (symbol (name old-ns)))` → `(in-ns (ns-name old-ns))`), regenerated `core_compiled.lgb`, and reran `test/run_tests_test.lg` — it reproduces the exact reported error (`name expected Symbol, Keyword or String`). Restoring the fix makes it pass again, so the new regression test genuinely exercises the bug rather than passing vacuously.

Also checked:
- `go generate ./pkg/rt/` reproduces `core_compiled.lgb`/`generated.manifest`/`generated.sums` byte-for-byte — consistent with what's checked in.
- `go build ./...`, `go vet ./...` clean.
- `go test ./...` — green apart from environment-only failures in this bare worktree (missing `test/clojure-test-suite` git submodule, no `.lgbgen-tree.sum` for the lowered-tree parity tests) and one pre-existing flaky `os_paths_test.lg` case that also fails standalone on `main` and passes in isolation — none attributable to this change.

Small, correct, well-tested fix. LGTM — approving.

nooga (approved):
Verified the fix reproduces and resolves the reported error; regression test is genuine (confirmed it fails without the fix). See summary comment.

nnunley:
This remains approved and its head checks are green at `7b3d2c1fd9f1f2605cdba345e5e9d35fee89d895`, but GitHub now reports conflicts with `main` (`3aa9f4ea86d41dd55ced66fb3ca800c4bc66c355`).

@jasalt, please update/rebase the contributor-owned branch and resolve the generated-artifact overlap, then let CI rerun. I will not push to it. Once it is current-base, conflict-free, and fully green, it is eligible to land.

jasalt:
@nnunley thank's for heads up, merged now and while CI here waits for maintainer approval, the "Go" workflow shows green at https://github.com/jasalt/let-go/actions/runs/34447244194
CC @nooga

nooga:
Already approved and independently verified — the regression test genuinely fails pre-fix and passes post-fix. No new review needed.

Current `mergeStateStatus` conflict is confined to `pkg/rt/generated.manifest` (confirmed via `git merge-tree` — pure generated-artifact drift, the #747 class; `core_compiled.lgb` and `generated.sums` auto-merge cleanly). A second author rebase attempt already re-conflicted for the same reason since main keeps touching `core.lg` — regenerating artifacts at merge time will likely be faster than another round-trip.

Separate note for whoever merges #798/#863 in either order: #863's rewritten `run-tests` already uses the equivalent-correct namespace-restoration pattern this PR fixes, so landing order only affects which PR's diff on that file becomes redundant, not correctness.

jasalt:
Also here "Expected — Waiting for status to be reported" needing some help @nooga @mparrett

mparrett:
@jasalt Checks are now green — fixed some issues with actions yesterday. Please rebase and we can land this.

nooga (approved):
Verified locally against the PR head (this is a fork PR with 0 CI runs, so I compiled and ran the suite myself rather than trusting GitHub's mergeable flag): `go build ./...` clean, full `.lg` integration suite (`go test ./test/ -run TestRunner`) green. The namespace-restore fix is unchanged since the original commits — later commits are just upstream merge-syncs.

Approving.

mparrett:
@jasalt #803 landed, can you please merge/rebase the PR? Don't want to clobber anything you have locally. I'll approve CI again when needed.

Issue #631, opened 2026-07-24
Strip debug sections from the embedded core and -w bundles


mparrett:
Sub-issue of #503 (footprint). Sibling lever to #502. Where #502 compresses the embedded `core_compiled.lgb`, this removes data from it: the source maps and local-variable tables that exist only for error reporting. The two compose — strip drops the debug bytes, compression shrinks what remains, so strip-then-compress beats either alone.

## Background

#624 added `bytecode.StripDebug` and taught `lg -strip` to omit source maps and local-var tables from `-c`/`-b` bytecode. It deliberately parked two follow-ups: the always-shipped embedded core, and the `-w` wasm program bundle. This tracks finishing both.

## What it covers

- `lg -w -strip` strips the per-app `program.lgb` in the emitted web app, mirroring the `-c`/`-b` paths.
- `lgbgen -strip` regenerates `core_compiled.lgb` debug-free, so a release build can ship a stripped core. Default off — the committed core stays full-debug so dev builds keep source locations in core stack traces.

## Measured

On the current core:

- Core bundle: 258,956 → 201,624 bytes (−22.1%, −57 KB).
- Native `lg` binary (`-s -w` ldflags): −66 KB.
- The stripped core boots and passes the `rt`, `vm`, and `compiler` suites.

## The tradeoff

Stripping is not free: a runtime error originating in `clojure.core` reports without a source location, and local-var names are gone from core frames. So this stays opt-in and release-only, the same posture #502 takes for compression — the committed artifact is untouched and the size win is a release build's choice, not a silent default.

A workable shape for that choice is a `build-strip` make target: regenerate the core with `-strip`, build, then restore the full-debug core (so the working tree is never left with a stripped core checked in). Whether the release lever is a make target, a build tag, or folded into the compression path is open — it should probably be decided together with #501/#502 rather than piecemeal.

## Scope note

This strips the `-w` *program* bundle. The core embedded *inside* the wasm binary is `pkg/rt`'s fixed `//go:embed core_compiled.lgb`, which `buildWasm` can't swap without an embed-override mechanism — so stripping the wasm core is a separate follow-up, not part of this.

## Status

There's a working prototype: the two flag wirings, the `build-strip` recipe, and a test that asserts the regenerated core decodes debug-free and is smaller than the committed one. It's held on a branch rather than opened as a PR, pending the compression direction in #501/#502, so the footprint changes land as one coherent story instead of colliding on the same core generator. I'll open the PR once that settles.

PR #800, opened 2026-09-06
feat(lg): extend -strip to -w bundles

mparrett:
## What changes

`-strip` split debug info out of `-c` and `-b` artifacts in #624, but rejected `-w`. A wasm bundle therefore had no way to ship without its source maps and local-variable tables. This makes the flag compose with `-w`, closing half of #631.

## Where the companion goes

Beside the bundle directory, not inside it. Every file in the output directory is served; a debug sidecar is build output, not something to hand to each visitor. `-debug-output` overrides the path.

Three details that are easy to get wrong:

- The output directory is canonicalized before the sibling is derived, so `dist/`, `dist/.`, and `.` all name the directory itself. Without that, a trailing separator gives a companion named `.debug` inside the bundle, and `dist/.` gives `dist/..debug`.
- An override is refused when it names the bundle directory or anything under it. A same-path check alone lets `-debug-output dist/index.html` through, and the build then replaces a generated asset with binary debug data while still reporting success. The same holds for `coi-serviceworker.js` and, in external mode, `main.wasm`, so the check is "inside the served directory" rather than a list of filenames.
- The companion is written last, after the bundle succeeds. These artifacts are digest-bound, so a sidecar left behind by a failed build would claim to describe something that was never produced, which is worse than having none.

## Sizes, not oversold

On a real program bundle, xsofy's `main.lg` compiled to `.lgb`:

| | raw | gzip -9 |
|---|---:|---:|
| plain | 709,129 | 232,854 |
| stripped | 583,963 | 164,882 |
| saving | 125,166 (17.7%) | **67,972 (29.2%)** |

The compressed saving is the larger percentage because source maps and local-variable tables are structured numeric data that compresses worse than bytecode and strings, so removing them takes a disproportionate share of the shipped bytes.

Against a whole wasm bundle, though, the program is a small term. A three-line app shrank by 48 bytes on an 8.1 MB bundle, and xsofy's saving is roughly 1.5% of its bundle. This is worth having and nearly free, but the footprint argument in #631 rests on the embedded core and on `.lgb` artifacts shipped directly, not on the wasm bundle. The embedded-core half is a separate change, the way #501 and #502 were split.

## Tests

The strip step lives in `pkg/rt/wasm` so its tests run without a full wasm build, which would have put minutes and a module fetch into the suite. They cover: the companion landing outside the served directory; every spelling of the directory deriving the same sibling; `-w .` resolving to the current directory's sibling under its real name; overrides inside the bundle, including the generated asset names, being rejected while siblings and unrelated paths are allowed; a filesystem root being refused; and a stripped bundle plus its companion still decoding together.

Each rejection was checked against the reintroduced bug: with canonicalization disabled the `-w .` case fails, and with only the same-path check the inside-bundle case fails.

`gofmt`, `go vet`, the `pkg/rt/wasm`, `pkg/cli`, and `pkg/bytecode` suites, `make check-generated`, and cross-builds for linux/amd64, js/wasm, plan9/amd64, and wasip1/wasm are clean.

### DIFF
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -403,7 +403,7 @@ func registerFlags() {
 	flag.StringVar(&bundleOutput, "b", "", "bundle .lg file into a standalone executable (specify output path)")
 	flag.BoolVar(&compressBundle, "z", false, "with -c/-b: DEFLATE-compress the bundle body (smaller .lgb / standalone binary; transparently inflated at load)")
 	flag.StringVar(&bundleBase, "bundle-base", "", "path to target-platform lg binary for cross-OS bundling (defaults to current executable)")
-	flag.BoolVar(&stripDebug, "strip", false, "split source maps and local-variable tables into a digest-bound .debug companion (with -c/-b)")
+	flag.BoolVar(&stripDebug, "strip", false, "split source maps and local-variable tables into a digest-bound .debug companion (with -c/-b/-w)")
 	flag.StringVar(&debugOutput, "debug-output", "", "path for the -strip debug companion (defaults to <output>.debug)")
 	flag.StringVar(&wasmOutput, "w", "", "build .lg file into a WASM web app (specify output directory)")
 	flag.StringVar(&wasmShell, "w-shell", "xterm", "shell for -w: 'xterm' (default), 'none' (emit core only; client supplies its own shell via window.LetGoHost), or a path to a custom HTML template containing __LG_HOST_JS_BODY_PLACEHOLDER__")
@@ -561,8 +561,8 @@ func runMain() int {
 		fmt.Fprintln(os.Stderr, "error: -z requires -c or -b")
 		return 2
 	}
-	if stripDebug && compileOutput == "" && bundleOutput == "" {
-		fmt.Fprintln(os.Stderr, "error: -strip requires -c or -b")
+	if stripDebug && compileOutput == "" && bundleOutput == "" && wasmOutput == "" {
+		fmt.Fprintln(os.Stderr, "error: -strip requires -c, -b or -w")
 		return 2
 	}
 	if debugOutput != "" && !stripDebug {
--- a/pkg/cli/wasm.go
+++ b/pkg/cli/wasm.go
@@ -198,6 +198,23 @@ func buildWasm(ctx *compiler.Context, nsRes *resolver.NSResolver, src string, ou
 		}
 	}
 
+	// -strip splits source maps and local-variable tables out of the embedded
+	// program before it is baked into the wasm binary. The companion is written
+	// beside the bundle directory rather than inside it: everything in outDir is
+	// served, and a debug sidecar is build output, not something to ship to every
+	// visitor.
+	var debugData []byte
+	var debugPath string
+	if stripDebug {
+		stripped, companion, path, err := wasmassets.SplitProgramDebug(lgbBuf.Bytes(), outDir, debugOutput)
+		if err != nil {
+			return err
+		}
+		lgbBuf.Reset()
+		lgbBuf.Write(stripped)
+		debugData, debugPath = companion, path
+	}
+
 	// 2. Create temp build directory
 	tmpDir, err := os.MkdirTemp("", "lg-wasm-*")
 	if err != nil {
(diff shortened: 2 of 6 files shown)


mparrett:
## Self-review

I found two path-handling issues that should be fixed before merging:

1. **[P2] Normalize `outDir` before deriving the companion path.**

   `lg -strip -w dist/. app.lg` derives `dist/..debug`, and `-w .` derives `..debug`. Both put the companion inside the served bundle directory, exposing the debug data this change intends to keep private. `outDir` should be canonicalized, with `.` and `..` handled explicitly, before the sibling path is derived.

   https://github.com/nooga/let-go/pull/800/files#diff-a385b4de6fe3b94273d51d0bd440ee2f310e31aad7ba1ba7495a30b534d7209cR33

2. **[P2] Prevent `-debug-output` from overwriting generated bundle assets.**

   The validation only compares the companion path with the directory itself. For example, `-debug-output dist/index.html -w dist` passes validation, then replaces `index.html` with binary debug data while the command still succeeds. The same applies to `coi-serviceworker.js` and, in external mode, `main.wasm`. Canonicalize and compare the companion path against every generated output before writing the bundle.

   https://github.com/nooga/let-go/pull/800/files#diff-a385b4de6fe3b94273d51d0bd440ee2f310e31aad7ba1ba7495a30b534d7209cR41

Focused validation passed for the root package, `pkg/rt/wasm`, and `pkg/bytecode`; the PR's CI checks are green.

mparrett:
Both addressed at `994e03b8`, which also rebases the branch onto the CLI's new home in `pkg/cli` after #773.

1. The output directory is canonicalized before the sibling is derived. `.` and `..` spellings resolve to an absolute path so the companion takes the directory's real name; a filesystem root is refused. `dist/.` now derives `dist.debug`, and `-w .` from inside `dist` does the same.
2. An override is refused when it is the bundle directory or anything under it, compared on canonical absolute paths. That covers `index.html`, `coi-serviceworker.js`, and external-mode `main.wasm` without enumerating them. `-debug-output dist/index.html -w dist` now fails before the build starts and leaves the existing `index.html` untouched.

Verified end to end with the rebuilt `lg` on a three-line app: `-w dist/.`, `-debug-output dist/index.html`, `-w-wasm external -debug-output dist/main.wasm`, and `-strip` without a target all behave as described.

Observed and left alone: an override whose parent directory does not exist (`-debug-output symbols/app.debug`) fails at write time, after the wasm build has run. `-c` and `-b` fail the same way today, so it is the existing convention rather than something introduced here.

nooga:
## Review summary

Ran a multi-angle review (correctness, reuse/simplification/efficiency, altitude, conventions) plus hands-on repro of the wasm build path.

**Found two real path-handling bugs** in the original `SplitProgramDebug`:
1. `-debug-output` pointing at a file *inside* the served output directory (e.g. `-w dist -debug-output dist/index.html`) wasn't rejected — the companion write landed last and silently clobbered the generated asset with binary debug data, exiting 0.
2. The collision guard compared raw, unnormalized strings, so `-debug-output ./dist` vs `-w dist` escaped it and only failed late with an opaque `is a directory` error.

Both were independently caught in your self-review and fixed at 994e03b8 (`canonicalBundleDir` + `filepath.Abs`/`Rel`-based `rejectInsideBundle`). I rebuilt from that commit and re-ran my original repros — both now fail fast, before the wasm build starts, with clear errors. The new tests (`TestSplitProgramDebugRejectsOverrideInsideBundle`, `TestSplitProgramDebugCanonicalizesOutDir`, `TestSplitProgramDebugDotResolvesToCwdSibling`, `TestSplitProgramDebugRejectsRoot`) cover exactly the cases that were previously untested. `go test ./pkg/rt/wasm/... ./pkg/bytecode/...` is green.

**Left open, non-blocking:**
- No test for the widened `-strip requires -c, -b or -w` flag-guard error text (`pkg/cli/cli.go`) — same pattern as the existing `TestCompressFlagRequiresCompileOrBundle` for `-z` would cover it cheaply.
- The wasm-embedded runtime never reads a `.debug` companion back at runtime (the generated `main.go` only calls the plain decoder) — unlike `-c`/`-b`, where a sibling `.debug` is transparently reattached. Likely an accepted limitation given the browser sandbox has no filesystem access to a sibling file, but worth a doc note so users don't expect `-c`/`-b`'s auto-reattach behavior here.

The missing-parent-directory case you flagged yourself as "observed and left alone" checks out — same pre-existing convention as `-c`/`-b`, not a regression.

LGTM — approving.

nooga (approved):
Verified the path-handling fix at 994e03b8 fixes both bugs found in review (repro'd before/after). See summary comment for details.

PR #803, opened 2026-09-06
feat(api): expose CompilerContext accessor for host nREPL integration

jasalt:
### Summary

`api.LetGo` compiles and evaluates guest code against an internal `*compiler.Context`, but that context is a private field (`l.c`). Anything that needs to share the running application's live compiler state — most notably `nrepl.NewNreplServer(ctx *compiler.Context)`, which already accepts one as its public constructor argument — has no way to obtain it from a `*LetGo` instance. A second, independently constructed context is not the running application: it lacks the same namespaces, vars, and live redefinitions.

This adds a single accessor:

```go
// CompilerContext returns the shared compiler context for host integrations
// such as nREPL servers that must evaluate in the embedded runtime's namespace.
func (l *LetGo) CompilerContext() *compiler.Context {
	return l.c
}
```

### Motivation

Downstream, embedding `api.LetGo` in a GTK development host alongside `nrepl.NewNreplServer` requires evaluating both from the same compiler context, so the nREPL client can inspect and mutate live application state (redefine a `view`, `reset!` a reactive cell, and reconcile the running window via `reload!`). Reproduced without this patch: `LetGo.Run` works standalone, but there is no supported way to hand `nrepl.NewNreplServer` the context backing that same running instance.

The accessor is generic — it makes no GTK or windowing assumptions — and mirrors the existing public surface: `NewNSResolver(ctx *compiler.Context, ...)` and `NewNreplServer(ctx *compiler.Context)` already treat `*compiler.Context` as a plain public type; this just closes the gap so a `*LetGo` can hand its own context to either.

### Downstream usage

[jasalt/let-go-time-tracker](https://github.com/jasalt/let-go-time-tracker) is the consumer that motivated this patch — a GTK time-tracker app that embeds `api.LetGo` and hot-reloads its UI over nREPL. Without a public accessor, it currently has to reach for a local interface type assertion to get at the patched method (`cmd/gtt-letgo-gtk/main.go`):

```go
var server *nrepl.NreplServer
if configuredPort := os.Getenv("LGTT_NREPL_PORT"); configuredPort != "" {
	port, err := strconv.Atoi(configuredPort)
	if err != nil {
		return fmt.Errorf("invalid LGTT_NREPL_PORT: %w", err)
	}
	contextProvider := any(lg).(interface{ CompilerContext() *compiler.Context })
	server = nrepl.NewNreplServer(contextProvider.CompilerContext())
	if err := server.Start(port); err != nil {
		return fmt.Errorf("start nREPL: %w", err)
	}
	defer server.Stop()
}
```

With `CompilerContext()` upstream, that call site collapses to `nrepl.NewNreplServer(lg.CompilerContext())` — no local interface shim needed.

### Changes

- `pkg/api/api.go`: add `(*LetGo).CompilerContext() *compiler.Context`.

### Compatibility

Purely additive — no existing signatures change. `go build ./...` and `go test ./pkg/api/...` pass locally.

### Test plan

- [x] `go build ./...`
- [x] `go test ./pkg/api/...`
- [ ] Maintainer review of whether a narrower interface (rather than the concrete `*compiler.Context`) is preferred for this seam

_____
LLM assisted PR (implementation gpt-5.6-sol / claude review)


mparrett:
Hi @jasalt - thanks for the PR. Initial agentic review found no issues. One design consideration, mentioned below.

The accessor returns the context initialized by `NewLetGo`, and both `LetGo.Run` and `NreplServer` correctly create transient evaluation children from it. Validation passed. The remaining consideration is API design: callers receive a mutable, low-level compiler context with its existing concurrency constraints. Given that `NewNreplServer` already requires this concrete type, that seems intentional and non-blocking.

When you can, please rebase and we'll take another look + architectural checks against the design.

nooga:
Confirming @mparrett's review — checked independently and reached the same conclusion.

- The accessor just returns `l.c`, the same context `NewLetGo` constructs and `Run` already evaluates against via `ChildForEval()` — purely additive, no behavior change to existing paths (`go build ./...`, `go test ./pkg/api/...` clean).
- Traced the one sanctioned consumer (`nrepl.NewNreplServer`): it already calls `n.ctx.ChildForEval().CompileMultiple(...)` per eval (pkg/nrepl/server.go:274), not `ctx.Compile` directly — so the "hand out a mutable, low-level compiler context" concern doesn't introduce a new race beyond what already exists between concurrent `LetGo.Run` calls on one instance, which the package doc already documents as the user's responsibility to serialize (the `WithStdout` concurrency caveat: "vm.Var's binding stack is process-global... serialize Run calls").
- No test was added for the accessor itself, but it's a zero-branch getter — a test would only assert `lg.CompilerContext() == lg.c`, not much value there; not blocking.

Agree this is safe and non-blocking as-is.

jasalt:
Sounds good, can check later this week. Feel free also to finalize meanwhile.

nooga (approved):
@jasalt Clean, minimal addition — `*compiler.Context` is already public API surface via `NewNreplServer`/`NewNSResolver`, so this accessor just closes the last gap for embedders. I checked the concurrency question raised in the thread directly: both `LetGo.Run` and the nREPL eval path go through `ChildForEval()`, so handing out the shared context here doesn't introduce a new race. Build and pkg/api tests are green. Green light — you're 3 commits behind main but none touch this file, so a rebase should be conflict-free.

jasalt:
@nooga Thank's for heads up. Merged but maintainer approval needed for test run here. "Go" workflow is green at https://github.com/jasalt/let-go/actions/runs/34447004113

nooga (approved):
Trivial additive getter, no behavior change. `go build ./...` and `go test ./pkg/api/...` (28 tests) clean. The concurrency question raised earlier is a non-issue — both `LetGo.Run` and nREPL eval already go through `ChildForEval()`.

Note: this PR's sole hunk is byte-identical to a hunk already present in #798's diff (same fork history). Whichever of #798/#803 merges first will make the other a no-op on that file — not a blocker, just a heads-up so the second one isn't chased as a failed merge.

Approving.

jasalt:
GH Workflow stuck still at "Expected — Waiting for status to be reported"?
@mparrett @nooga

nooga:
@jasalt looks like I ran out of github action limits, I'm looking into how I can fix this.

mparrett:
> @jasalt looks like I ran out of github action limits, I'm looking into how I can fix this.

@nooga @jasalt — made some fixes to help actions yesterday. Checks are green now and will run faster. Rebase this branch and we'll be good to land the PR.

nooga (approved):
Trivial, correct accessor addition — verified `go build ./...` clean at the PR head (fork PR, no CI ran; exercised locally instead, including via #798's test run which carries the identical `pkg/api/api.go` hunk).

Approving.

Issue #801, opened 2026-09-06
Reader compatibility: metadata yields a with-meta list and map-interior #_ discard fails


nnunley:
## Summary

While implementing [Strange Lettractor](https://github.com/nnunley/strange-lettractor), we found two differences between let-go's `read-string` and JVM Clojure that prevent using Clojure reader syntax for data/configuration. We want the Clojure superset of EDN, not a restricted hand-written EDN parser, and do not want to evaluate configuration to compensate for reader behavior.

## Environment

- Local let-go binary: `lg dev (bdd8268)`
- Local source HEAD: `bdd8268c9cb3acf369f0854bada47af79d1d673d`
- Reproduced by running the expressions below with `lg -e` and `clojure -M -e` on macOS.
- These observations concern that local build; I have not verified them against the latest upstream HEAD. The local checkout has unrelated documentation/test-suite changes.

## 1. Metadata is returned as an executable form rather than attached to data

```clojure
(let [v (read-string "^{:doc \"mapping\"} {\"a\" \"b\"}")]
  (prn v)
  (prn (map? v)))
```

Expected / JVM Clojure:

```clojure
{"a" "b"}
true
```

Additionally, `(meta v)` is `{:doc "mapping"}` in Clojure.

Actual local let-go:

```clojure
(with-meta {"a" "b"} {:doc "mapping"})
false
```

In the inspected source, `pkg/compiler/reader.go`'s `readMeta` constructs a `with-meta` list. That differs from reading a metadata-bearing value; evaluating the returned form is not an appropriate fix for a data reader.

## 2. Reader discard between a map key and value causes an odd-form error

```clojure
(read-string "{\"a\" #_[:ignored] \"b\"}")
```

Expected / JVM Clojure: `{"a" "b"}`.

Actual local let-go: `map literal must contain even number of forms`.

A leading discard, `(read-string "#_[:ignored] {\"a\" \"b\"}")`, does work in the same binary.

## Expected contract and regression coverage

The [Clojure reader reference](https://clojure.org/reference/reader) specifies that metadata attaches to the following form and `#_` skips the following form entirely. Both expected results above were also verified directly in JVM Clojure.

Please preserve those semantics for `read-string`, with regressions for metadata-bearing maps and discards before/between/after collection entries. Fixing the reader should not require application code to use `eval` or implement its own restricted grammar.

Separately, `read-string` consuming only the first form is normal Clojure behavior and is **not** being reported as a bug here. Our application must enforce its own single-value input contract.

[...]

PR #819, opened 2026-09-07
fix(reader): attach metadata to read-string collections

nnunley:
## Why

`read-string`, `read-all-string` and `compiler.ReadString` read with the code reader, so `(read-string "^{:a 1} [1 2]")` returned the form `(with-meta [1 2] {:a 1})` rather than a vector carrying metadata. Clojure's `read-string` returns data.

## What

- The three entry points use the data reader (the existing `data` field), so metadata attaches to the value where the runtime can carry it: lists, vectors, maps and sets.
- A bare-symbol tag is attached as the symbol itself, Clojure's metadata value; the code reader keeps quoting it because it emits an evaluated `(with-meta …)` form.
- A target that cannot carry metadata keeps the `(with-meta …)` wrapper instead of reading as itself — `vm.Symbol` is a plain string with no `IMeta`, and the bare form would drop the metadata silently.
- Data read back in must still compile, as in Clojure: a vector that gains metadata is a persistent vector, so the fn compiler accepts either vector representation for an arity (`^long [^long n]` used to be rejected as "unexpected fn form"), and the `deps.edn` consumers (`pkg/resolver`, `cmd/lginterop`) accept either too.
- The two `return_hint_metadata` fixtures normalize a hinted parameter's wrapper form alongside the shapes they already normalized.

Addresses the metadata section of #801 only; map-interior `#_` discard handling remains open there.

**Stacked on #736**, which supplies the set half (set literals read as sets in both modes). Merge after it.

## Verified

`make generate`; `make check-generated`; `go test ./pkg/compiler ./pkg/resolver ./cmd/lginterop ./pkg/api ./test ./test/e2e`; `scripts/gogen-trampoline.lg` on all fixtures, native == bytecode; regression in `test/read_string_eval_test.lg`.


nooga:
Still a draft, so holding off on a formal review action — a couple of things to flag before this is ready:

- Needs a rebase: the branch carries an intermediate #770 commit rather than the final merged version, so `git merge-tree` shows real conflicts in `eval.go`/`reader.go` plus add/add conflicts in test files, not just generated-file drift.
- Overlaps #736's edit to the same `readSet` function — the two need a joint decision (unconditional real sets vs. the `r.data`-gated mechanism #770/#854 already introduced) before either merges, or whichever lands second will re-conflict with the other's design, not just its diff.
- The PR description's framing against #801 is now stale — #854 (merged) already covers metadata via the edn path; this PR's real remaining scope is extending that same behavior to the plain (non-edn) `read-string`/`read-all-string` entry points, which is still genuinely unaddressed on main. Worth updating the description to reflect that narrower, still-valid scope.

Once rebased and the #736 design question is settled, please mark ready for review.

nnunley:
Rebased the 8-commit stack onto main@b3fd3bbf (upstream); only `pkg/rt/generated.sums`/`generated.manifest` (and `pkg/rt/core_compiled.lgb`) conflicted in every commit, resolved by regenerating with `make generate` at each conflicted commit in turn (bottom-up), no empty commits produced. The PR's own files (reader.go, core.lg/build.lg/forms.lg/lower_go.lg/pipeline.lg, and all the gold/gogen/set-literal tests) are unchanged in substance from the reviewed head f05ea1f0. `make check-generated`, `make build`, `go test ./pkg/...`, and `go test ./test` all pass. Pushed with SKIP=bench-ratchet (the local timing gate is red independent of this change — #791, #899). Note: mergeStateStatus now shows BEHIND again since main moved further mid-review — another rebase may be needed before merge.

nnunley:
Rebased directly onto main@ebd758a6; this PR no longer carries #736's commits (one commit now). `ReadString` reads `#{…}` as a set, so `TestDataReaderFollowsClojureDataSemantics`'s last assertion now expects `SetType` — the same change #736 makes to that line, so the two cannot conflict. `make check-generated`, `make build`, `go test ./pkg/compiler ./pkg/rt` and the full `./test` suite pass. Pushed with SKIP=bench-ratchet (#791, #899).

nnunley:
Correction to my previous comment: that push did not land — the PR is unchanged at `d821c5c0`. The pre-push `ir-stress-gate` fails on the standalone commit: the lowering-coverage ratchet reports 143 failures against a baseline of 11, with 133 new `:build/unrecognized-form` entries, i.e. corpus forms that stop lowering to native Go once `read-string` returns data-reader values (sets as `PersistentSet`, collections carrying metadata). So this change needs a position on the IR/gogen path before it can go up; the standalone rebase itself is ready locally. Posted early by mistake; the correction is mine.

nnunley:
Rebased standalone onto `main` (ebd758a6) — this PR is no longer stacked on #736 and is a single commit.

**`read-string` keeps data semantics.** It reads data everywhere, so `#{…}` comes back as a `PersistentSet` and metadata attaches to the collection. The follow-on is that the IR now has to know about sets, because `scripts/ir-stress.lg` parses the corpus with `read-string` and a set VALUE reached two places that had only ever seen the code reader's `(hash-set …)` form:

- **`pkg/rt/core/ir/build.lg`** gains a `set?` arm. `build-set` mirrors `build-map` — a `Const` when every element is a literal (which keeps the value and its metadata intact), otherwise a `(hash-set …)` call, exactly the form the code reader produced. `captures-of` and `free-vars` gain matching arms for the same reason their `map?` arms exist: a local referenced only inside a set literal was invisible to block-arg threading and to closure capture (this is what `captures-of`'s own `#{(:binding catch-clause)}` tripped over).
- **`pkg/rt/core/ir/lower_go.lg`** gains `boxed-set-expr`, so a set *constant* — `(quote #{1 2})`, which is how `clojure.test`'s `is` captures a set-bearing form — lowers to `vm.NewSet` / `vm.EmptyPersistentSet` instead of failing the whole function with "unsupported function body shape".

**Position on the three compiler paths.** `pkg/compiler` (default, bytecode) is unchanged: a `PersistentSet` is self-evaluating there, which is why this PR's own tests already passed; the new test pins IR/bytecode agreement on set literals so they cannot drift. `*ir-compile*` gets the `build.lg` change — `ir/lattice` already classified a set `Const` as `:set` and `ir/lower` carries a `Const` generically, so nothing else was needed. `-tags gogen_ir` gets the `lower_go.lg` change; `:set` was already in the Go type map.

**ir-stress gate is back at baseline.** On the un-fixed tree the lowering-coverage ratchet went to **143** failures (133 of them `:build/unrecognized-form`); it is now **11/2557**, the same count and the same buckets as `main`. No rebaseline.

`test/ir_set_literal_test.lg` is new and covers both directions: IR-compiled set literals (constant, empty, non-constant elements, nested, metadata, loop-crossing and closure-captured locals) under `*ir-compile-strict*`, and the emitted Go for quoted set constants via `pipeline/lower-ns-to-go`.

**Reader test.** `TestDataReaderFollowsClojureDataSemantics`'s last assertion now expects `code.Type() == vm.SetType` instead of `(*vm.List).First() == 'hash-set` — the same hunk #736 changes, so the two cannot conflict when #736 lands.

**Verified:** `make generate`; `make check-generated`; `make build`; `go test ./pkg/... -count=1`; `go test ./test/ -count=1`; `make ir-stress-gate` — all green. Pre-push hooks ran with `SKIP=bench-ratchet` (#791/#899 — the ratchet is unrelated to this change and is known-flaky on this machine); every other hook ran and passed.

nnunley:
Rebased onto current main as `d89e8a73` (still one commit). The only conflict was `pkg/rt/generated.manifest` / `generated.sums` after today's merges; both were regenerated with `make generate`, no source change. `make check-generated`, `go test ./pkg/... ./test/` and `make ir-stress-gate` (2568/2569, the known one-failure baseline) pass on this head.

---
Delivery verified by local `pr-claim-gate`: exact PR/validated head `d89e8a735ad472f199fe4e5de704523e429f3978`; relevance claims `pkg/rt/generated.manifest` occur in both the delivered delta and the PR feature delta; receipt `/Users/ndn/.local/state/pr-claim-gate/nooga-let-go/pr-819.json`.

mparrett:
I looked at the current `d89e8a73` head. The standalone rebase and the IR set handling address the earlier integration concerns; getting the stress gate back to baseline is reassuring. Two cases seem worth closing before marking this ready:

1. **A metadata-bearing vector still fails the ordinary read/eval round trip.** With this PR, `(eval (read-string "^:flag [1 2]"))` passes a `vm.PersistentVector` to `compileForm`. The new `paramVector` handles fn parameter vectors, but `compileForm` still only has the `vm.ArrayVectorType` vector arm. I reproduced the equivalent on the unchanged compiler using `(eval (read-data-string "^:flag [1 2]"))`: it fails with `ExecutionError: runtime error: index out of range [-1]`; `(eval (read-data-string "[1 2]"))` succeeds. An ordinary persistent-vector form needs a compiler case and a regression test.

2. **The new AOT set-constant path appears to drop reader metadata.** A data-read `^:flag #{1 2}` reaches `build-set` as a `Const`, but `boxed-set-expr` reconstructs it with `vm.NewSet(...)` and never applies the source set's metadata. The Go emitter test currently checks for constructor text, so it would miss this. A behavior test for `(meta ...)` on a data-read set in generated Go would pin down the intended semantics. The bytecode `SetType` arm also rebuilds through `hash-set`, so it may be useful to state the eval contract across the compiler paths.

I would treat the vector failure as blocking the stated read/eval contract. The approach otherwise looks sound to me.

nnunley:
@mparrett — head `d7619899` (still one commit). Both points taken, and the first turned out to be a symptom of something wider.

**1. The vector round trip.** The root cause is that `compileForm`'s `switch o.Type()` had no `default` arm: a type no arm names compiled to *nothing* and returned `nil`, and the caller's missing stack slot surfaced later as `index out of range [-1]`. `PersistentVector` was one such type; `Atom`, `Range`, `PersistentQueue`, `SortedMap`, `SortedSet`, `TransientVector` and `*Namespace` reached as a form all failed the same way (a macro splicing one in is enough). Two changes:
- The vector arm takes both vector tags, matching `vector?`, through one helper, `orderedLiteralElems`, which `paramVector` now shares. Elements come from `Seq()`, not index + count. `paramVector` still refuses a bare `MapEntry`; the literal arm accepts one, as before.
- A `default` arm returns a `CompileError` naming the type and value. Nothing in the suite or `make generate` relied on an unlisted type being embeddable, so no tag was added to the constants arm.

`pkg/vm` has no ordered/sequential marker to dispatch on instead — `Indexed` and `Sequable` also cover strings, ranges and maps — so this remains a tag match, confined to that one helper. Forms reach `compileForm` from four producers (the code reader, macro output, `eval` of data, and tagged/data readers, whose return values are unconstrained), so the set of types is open; letting a value declare what kind of form it is looks like the right follow-up, separate from this PR.

**2. Metadata on evaluated collections.** The contract is now stated and tested on each path: evaluating a collection form that carries metadata yields a collection carrying it, and the metadata map is itself evaluated — `^{:a (+ 1 1)} [1]` carries `{:a 2}`, as the code reader's `(with-meta …)` form always has. Bytecode: `compileAggregateWithMeta` wraps the vector, map and set arms. AOT Go: `boxed-with-meta` in `lower_go.lg` wraps all three constant emitters, so a set `Const` no longer loses it. The new tests assert `(meta …)` of the evaluated value, not constructor text, under plain eval and under `*ir-compile*` strict.

**Left open, deliberately:**
- In the AOT path a metadata-bearing **vector or map** `Const` loses its metadata somewhere in the optimize pipeline (typeinfer/CSE, not pinned) before it reaches `boxed-value-expr`; sets are unaffected. It predates this PR and I have not fixed it here. I'll file it.
- `SortedMap` / `SortedSet` as forms were never compilable; they now fail closed instead of corrupting the stack. Clojure compiles them as ordinary map/set literals — a small follow-up.
- A behaviour fixture through `scripts/gogen-trampoline.lg` was abandoned: its parity check reads the fixture with the code reader on one side and the data reader on the other, which a metadata-bearing quoted literal cannot satisfy. The `*ir-compile*` + `read-string` pattern in `test/ir_set_literal_test.lg` is used instead.

Verified on this head: `go vet`, `go test ./pkg/... ./test/`, `make check-generated`, `make ir-stress-gate` at 2569/2570 (the known one-failure bucket, no rebaseline).

---
Delivery verified by local `pr-claim-gate`: exact PR/validated head `d76198991dfa151109d56dcc03ebe376476feb71`; relevance claims `pkg/compiler/compiler.go` occur in both the delivered delta and the PR feature delta; receipt `/Users/ndn/.local/state/pr-claim-gate/nooga-let-go/pr-819.json`.

mparrett:
Rechecked `d7619899`: the new vector arm/default error and the set metadata wrapper address my two original findings. Thanks for tightening those paths.

One follow-up in `compileAggregateWithMeta`: `readMeta` deliberately stores a bare type hint in data-reader metadata as a **symbol** (`^long [1]` gets `{:tag 'long}` as a value), but the new helper passes that metadata map to `compileForm`. Compiling the map evaluates its `:tag` value, so `long` resolves to the core native function instead of staying a symbol. The code-reader path quotes the tag specifically to avoid this. I verified the distinction in the existing runtime: `(meta ^long [1])` has a `vm.Symbol` tag, while `(meta (with-meta [1] {:tag long}))` has a `*vm.NativeFn` tag. The new data-read/eval path appears to take the latter route for `(meta (eval (read-string "^long [1]")))`; an unresolved hint could fail compilation outright.

Could you add a regression for a bare symbol hint through `read-string` + `eval` and preserve/quote that tag when compiling metadata? The `^:flag` and expression-valued metadata tests don't exercise this case.

nnunley:
@mparrett — head `7edba67a` (one commit). Correct, and I introduced it: compiling the metadata map as a form is what made `^{:a (+ 1 1)} [1]` carry `{:a 2}`, and it also sent the data reader's bare-symbol `:tag` through symbol resolution.

`compileAggregateWithMeta` now passes the map through `quoteTagHint` first: when `:tag` holds a `vm.Symbol` it is replaced with `(quote sym)`, which is the same thing `readMeta` does on the code path (`m` vs `dataM`). Everything else in the map is still evaluated.

Regressions in `test/read_string_eval_test.lg`, confirmed red without the helper with your exact symptom (`actual: (not (= long <native-fn long …>))`):
- `(:tag (meta (eval (read-string "^long [1]"))))` is the symbol `long`, and equals `(:tag (meta ^long [1]))`
- an unresolvable hint, `^no.such.Hint [1]`, stays a symbol instead of failing compilation
- the same on a set and a map
- a hint combined with expression metadata: `^long ^{:a (+ 1 1)} [1]` carries `{:tag long :a 2}`

One thing worth stating rather than leaving implicit: Clojure itself evaluates the hint here (`(meta ^long [1])` there carries the `long` function object, and an unresolvable tag is a compile error). This runtime's code reader deliberately keeps the symbol, so the data path now matches this runtime rather than Clojure. If that choice should change, it should change for both paths together.

`go vet`, `go test ./pkg/... ./test/`, `make check-generated` and `make ir-stress-gate` (2569/2570, known bucket) pass on this head.

---
Delivery verified by local `pr-claim-gate`: exact PR/validated head `7edba67ac63890981a93db74a764c2c4e6c0d2e0`; relevance claims `pkg/compiler/compiler.go` occur in both the delivered delta and the PR feature delta; receipt `/Users/ndn/.local/state/pr-claim-gate/nooga-let-go/pr-819.json`.

PR #820, opened 2026-09-07
fix(json): preserve string map keys

### LINKED ISSUE #817: json/write-json adds literal quotes to string map keys
## Reproducer

Using the local `lg dev (bdd8268)` binary, source revision
`bdd8268c9cb3acf369f0854bada47af79d1d673d`:

```clojure
(require '[json :as json])
(println (json/write-json {"key" "value"}))
;; actual JSON: {"\"key\"":"value"}
;; expected:    {"key":"value"}

(let [value {"key" "value"}]
  (= value (json/read-json (json/write-json value))))
;; actual false; expected true
```

The encoded object key contains literal quote characters. This corrupts native
provider event fixtures and any read-modify-write JSON object with string keys.
Keyword-key maps work for the simple fixture keys; that is our temporary fixture
workaround, not a general lossless conversion for arbitrary user keys.

## Likely cause

`pkg/rt/json.go`, `fromMapValue`, derives every key with `k.String()` and only
special-cases keywords. `vm.String.String()` produces its printed/quoted
representation. The value conversion in `fromValue` already uses the underlying
Go string for `vm.String`; map keys need the analogous treatment.

## Suggested regression coverage

- String-key map roundtrip, including nested maps.
- Empty keys, quote/backslash/newline characters, Unicode, and slash-containing keys.
- Preserve existing keyword-key behavior.

Found while implementing StrongDM Attractor in strange-lettractor. No runtime
checkout changes were made.

nnunley:
## Summary

Fix `json/write-json` corrupting VM string map keys by serializing their underlying text instead of their readable EDN representation.

- preserve raw string keys at every nesting level
- retain existing keyword-key spelling and fallback behavior for other key types
- cover plain, nested, empty, quote-escaped, Unicode, and slash-containing keys

Closes #817.

## Reproduction

Before:

```clojure
(json/write-json {"key" "value"})
;; => {"\"key\"":"value"}
```

After:

```clojure
(json/write-json {"key" "value"})
;; => {"key":"value"}
```

## Validation

- `go test ./test -run 'TestRunner/json_test\.lg$' -count=1`
- `go test -race ./test -run 'TestRunner/json_test\.lg$' -count=1`
- `go test ./pkg/rt -count=1`
- `go vet ./pkg/rt ./test`
- `make generate`
- `make check-generated`
- `git diff --check`
- independent code review: PASS, no findings

A broader `go test ./pkg/rt ./test -count=1` pass before the final mechanical restack had `pkg/rt` green; the `test` package was blocked only by this sandbox denying the suite's existing literal `/tmp` writes. The focused JSON suite is green before and after the restack; CI remains the unrestricted full-suite gate.

### DIFF
--- a/pkg/rt/json.go
+++ b/pkg/rt/json.go
@@ -72,9 +72,14 @@ func fromMapValue(v vm.Value) (any, error) {
 			if e != nil {
 				return vm.NIL, vm.NewExecutionError("invalid VM value")
 			}
-			nk := k.String()
-			if k.Type() == vm.KeywordType {
-				nk = nk[1:]
+			var nk string
+			switch k := k.(type) {
+			case vm.String:
+				nk = string(k)
+			case vm.Keyword:
+				nk = string(k)
+			default:
+				nk = k.String()
 			}
 			r[nk] = vv
 		}
--- a/pkg/rt/generated.manifest
+++ b/pkg/rt/generated.manifest
@@ -134,7 +134,7 @@ pkg/rt/core_compiled.lgb pkg/rt/ions.go generator bc46c36ed70a734ed6323f95157883
 pkg/rt/core_compiled.lgb pkg/rt/iort.go generator d56aae517e4680a3aed6eaaf1fd12541494999d73918a5528448f037d4095a87
 pkg/rt/core_compiled.lgb pkg/rt/ir_bridge_generated.go generator e9147537efd6da82ca9023b8fe7ef249cecaae89bae5893a51db4113d95aa2d9
 pkg/rt/core_compiled.lgb pkg/rt/js.go generator 0009511f66420745103c8c17881ad52a5f31227172557dc539b2a4b171dd4738
-pkg/rt/core_compiled.lgb pkg/rt/json.go generator 1a47708464ebcd2958f804ba4723f59f9394fee26b8e725ece03088851902397
+pkg/rt/core_compiled.lgb pkg/rt/json.go generator cde0a6755b094b1e187a2615e82cc578d871245fa80757ee82bf8be8c5212d6c
 pkg/rt/core_compiled.lgb pkg/rt/key_tokenizer.go generator 197eba584a6f0a33fc02b429e4dd45c8f3a63e0b86d0c4d3ba7c3700dcf81d9e
 pkg/rt/core_compiled.lgb pkg/rt/keysource.go generator f7331de29ceb2055b3be9b1a9f4aee6d07b16b01a336306014dff76d86f5f2cd
 pkg/rt/core_compiled.lgb pkg/rt/keysource_js_wasm.go generator d790cddc0c77a3993b668d205cf38264db312d3cff314f48b92a81d82e08073c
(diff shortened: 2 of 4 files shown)


pkg/rt/json.go
```diff
@@ -72,9 +72,14 @@ func fromMapValue(v vm.Value) (any, error) {
...
+			var nk string
+			switch k := k.(type) {
+			case vm.String:
+				nk = string(k)
+			case vm.Keyword:
+				nk = string(k)
+			default:
+				nk = k.String()
 			}
 			r[nk] = vv
```
mparrett, line 84: [P2] Reject collisions after coercing map keys

Now that both `"x"` and `:x` become `"x"`, a valid VM map containing both silently drops one entry here because `r` is a Go `map[string]any`. This is particularly easy to hit in the advertised read-modify-write flow: `(assoc (read-json "{\"x\":1}") :x 2)` retains two VM keys, but `write-json` emits only `{"x":2}`. Please detect a previously emitted `nk` (and return an error, or otherwise define a deterministic policy) rather than silently losing data.

mparrett (commented):
Found one P2 correctness issue, commented inline. Focused JSON and race tests pass; vet and generated-artifact validation also pass.

mparrett (changes requested):
Requesting changes for the unresolved P2 inline finding: key coercion can silently collapse distinct VM map entries (for example, string "x" and keyword :x) into one JSON object key.

nooga (commented):
The string-key fix is correct and minimal. Verified locally: the new tests fail on main and pass here; `go test ./pkg/rt`, `go vet`, and `make check-generated` are all clean; and the `generated.manifest` hash matches the new `json.go` sha256. The pods JSON encoder (`JSONEncodeArgs` in transit.go) and `js/emit` share `fromMapValue` and pick up the fix with no dependency on the old quoted form.

+1 to @mparrett's P2. The silent drop on a `"x"` / `:x` collision is a Go-map artifact rather than JSON semantics (data.json and cheshire would emit duplicate keys instead), and it is a few lines to detect in this loop. Note it was already reachable for other type pairs (`{1 1 "1" 2}`, `{:x 1 'x 2}`), so this PR widens the path rather than creating the class; it could equally land as a follow-up if you prefer to keep this PR to the issue's scope.

One small pre-existing thing in the same switch: the `default` branch still emits reader syntax for the remaining key types, so `(write-json {\a 1})` gives `{"\\a":1}` and a nil key gives `"nil"`. Routing `vm.Char` through `string(rune(k))` here would be a one-liner if you want to sweep it in.

nooga (approved):
Verified: `git merge-tree` against current main is clean (no conflicts, only the registered `generated.sums` merge driver firing). `go build ./...` clean, `go test ./pkg/rt/...` and the json test suite (34/34) pass. Minimal, correct fix.

Noting the known follow-up (distinct keys that stringify identically, e.g. string `"x"` vs keyword `:x`, still collapse) — agreed this is pre-existing Go-map behavior and fine as a separate PR, not a blocker here.

Approving.

pkg/rt/json.go
mparrett: Re-checked this against both builds, and the finding holds as a bug but not as a blocker on this PR. Moving it to #875.

| input | `main` @ `5c9e1e4c` | this PR @ `122b2598` |
|---|---|---|
| `{:x 1 (symbol "x") 2}` | `{"x":2}` | `{"x":2}` |
| `{"y" 1 :y 2}` | `{"\"y\"":1,"y":2}` | `{"y":2}` |

The collision is already there on `main` and this PR leaves it byte-identical: a keyword and a symbol of the same name have always coerced to one key. What the PR adds is one more pair to the class, since string keys stop carrying their quotes and can now collide too. The underlying defect is older than the change in front of it, and the read-modify-write example I gave reproduces on `main` today.

So my changes-requested was mis-scoped, and I've cleared it. Thanks for the patient argument, @nooga — the `{1 1 "1" 2}` framing is what sent me back to measure it properly.

mparrett (approved):
Approving. The string-key fix is correct and the quoted-key output it replaces was plainly wrong.

Clearing my Sep 7 changes-requested: the collision I raised reproduces unchanged on `main` at `5c9e1e4c`, so it is a pre-existing defect rather than something this PR introduces, and it does not belong in the way of the fix. Filed as #875 with the measurement, against `main` where it lives.

Verified at `122b2598`: `{"q-id" 7}` emits `{"q-id":7}`, and the keyword/symbol case is identical before and after.

Posted by Claude under my account; the re-measurement and the issue text are its work.

nooga (approved):
Verified the fix: `fromMapValue` now type-switches on the map key (string → raw, keyword → bare name via `string(k)`, else `.String()`), which is behavior-preserving for the keyword case and fixes the quoted/corrupted string-key bug in `json/write-json`. CI is green and this already carries two approvals.

Note on the `CONFLICTING` flag: it's spurious — `git merge-tree --write-tree origin/main origin/pr/820` resolves with zero conflicts. The only files besides `json.go`/the test are `generated.manifest`/`generated.sums`, which have a registered `merge=sums` driver that GitHub's mergeability check doesn't invoke. A local rebase (or `make generate` after merging) clears it — no real conflict here (same #747-class drift).

Approving.

mparrett:
@nnunley This looks good- feel free to rebase and merge when you're ready.

nnunley:
Rebased onto `main@3bbde90a` and pushed as `2c300910`; the only conflict was `pkg/rt/generated.sums`, resolved by `make generate`. `json.go` and `json_test.lg` are byte-identical to the previously approved head `122b2598`. `make check-generated`, `make build`, `go test ./pkg/rt/... ./pkg/genmanifest/...` and the `json_test.lg` runner pass. Pushed with `SKIP=bench-ratchet`: the pre-push timing gate on this machine is red independent of this change (#791, #899); the deterministic bar was a false positive against a stale baseline row, fixed in #900. Merging on the two standing approvals once CI is green.

nnunley:
Rebased again onto main@ec18f334 after #896 landed; no conflicts / regenerated pkg/rt/generated.sums, pkg/rt/generated.manifest; verification as above.

PR #821, opened 2026-09-07
fix(interop): preserve typed-array backing slices

### LINKED ISSUE #813: Go interop copies mutable byte-array arguments, discarding io.Reader.Read output
# Mutable byte-array copying during Go interop

Observed 2026-09-07 with the local let-go binary; checkout HEAD was
`bdd8268c9cb3acf369f0854bada47af79d1d673d` (user-owned checkout may contain edits).

## Reproducer

```clojure
(require '[io :as io])
(let [r (io/string-reader "ABC")
      b (byte-array 3)]
  (println :n (.Read r b) :actual (vec b) :expected [65 66 67]))
```

Actual: `:n 3 :actual [0 0 0] :expected [65 66 67]`.
The reader consumes the input, but mutations to the Go byte slice are not
visible in the original let-go array. The same result occurs with `.Read` on a
child process stdout pipe; this is not a subprocess-specific problem.

## Cause

`pkg/vm/native_func.go` function `boxArgForReflect` handles slice/array targets
by treating every `Sequable` as a sequence to copy with `unboxSliceInto`.
`TypedArray` implements `Sequable`, so its existing mutable Go slice returned
by `Unbox()` is not passed through. A correctly typed mutable byte-array should
retain its backing storage for calls whose contract mutates that buffer.

Potential upstream repair: preserve compatible typed-array backing slices
before generic sequence conversion. Keep vector/list conversion behavior and
add regression coverage for byte-array identity/mutation and other array types.
No runtime source was modified in this investigation.

## Attractor workaround and return point

Buffered single-byte reads do not require caller-owned mutable buffers:

```clojure
(let [r (.Buffered (io/string-reader "ABC"))]
  [(.ReadByte r) (.ReadByte r) (.ReadByte r)])
;; verified result: [65 66 67]
```

Use this boundary when implementing bounded Codex frame accumulation; do not
replace bounded framing with an unbounded line reader. Revisit chunked reads
after an upstream fix, with a regression proving the caller's array changes
and a split-UTF8 protocol test. EOF/error propagation and AOT remain separate
transport verification obligations.

nnunley:
## Summary
- preserve compatible `TypedArray` backing slices at the Go reflection boundary
- keep generic vector/list and incompatible-element conversion behavior unchanged
- cover backing identity and host mutation for byte, int, float, object, and named slice types
- reproduce `io.Reader.Read` at the language level and verify caller-visible bytes

## Verification
- focused VM/API/language regression tests
- focused race tests
- `go vet ./pkg/vm ./pkg/api ./test`
- full `go test ./pkg/vm ./pkg/api`
- all Go-level `./test` tests outside the fixture runner
- `make generate`
- `make check-generated`
- independent review: PASS

The full local fixture runner was also attempted: 810 tests passed and 10 fixture cases failed only because this sandbox denies writes to literal `/tmp` paths. The exact #813 regression passes.

Fixes #813

### DIFF
--- a/pkg/rt/generated.manifest
+++ b/pkg/rt/generated.manifest
@@ -230,7 +230,7 @@ pkg/rt/core_compiled.lgb pkg/vm/map.go generator f8d314a14ca7c5c91055b7ec963d986
 pkg/rt/core_compiled.lgb pkg/vm/meta_value.go generator 457b232d744935a2be249ab6d0a7b1029a6c7315086c0065fa6cb08e74bcead2
 pkg/rt/core_compiled.lgb pkg/vm/multifn.go generator faabb5b0a048d17c744e7e8bed33087f7d671063f8b5e2142252674bf7576efe
 pkg/rt/core_compiled.lgb pkg/vm/namespace.go generator b847271519c2b7eeba5bee173e31b7d4dd4c49361476ded23baf4a6d2cd5942b
-pkg/rt/core_compiled.lgb pkg/vm/native_func.go generator 63b68016e0875f5b2dc6d3615297d7923dc74bece0676cca728cb84a929afdc9
+pkg/rt/core_compiled.lgb pkg/vm/native_func.go generator 32527c009e38ed2f19724f9e4fe0035c56e4156e6c0e7310394b025689ee341a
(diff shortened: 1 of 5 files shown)


mparrett (approved):
No actionable findings. The compatible-slice fast path preserves TypedArray backing identity while retaining the existing conversion path for fixed arrays, incompatible element types, and persistent collections. Focused VM/API/language tests, race tests, vet, and generated-artifact validation pass; CI is green.

PR #825, opened 2026-09-08
docs: rename KNOWN_DIVERGENCES.md to known-divergences.md

mparrett:
Every page under `docs/` is lowercase except the two `README.md` files; `KNOWN_DIVERGENCES.md` from #771 is the one exception, and nothing depends on the name. This renames it and updates the two links that point at it (the root README and the `docs/README.md` topical map). Content and frontmatter are unchanged, so `docs_frontmatter_hook.py --check` passes as before.


nooga (approved):
@mparrett This looks merge-ready — pure rename plus the two inbound link updates in README.md/docs/README.md, no content changes. CI is green and there's no open review thread. Green light from me.

One heads-up: this collides with #826 (edits `docs/KNOWN_DIVERGENCES.md` at the old path). Merge this one first, then rebase #826 onto it so the content edit lands on the renamed path.

PR #826, opened 2026-09-08
docs(divergences): narrow the array-map entry to construction past eight

### LINKED ISSUE #763: Small maps don't preserve insertion order (Clojure array-map semantics)

mparrett:
## Why

`docs/KNOWN_DIVERGENCES.md` still says `array-map` is routed through the unordered persistent map and points at #763 for scoping. #764 resolved #763 on 2026-09-07 without touching the ledger, and the ledger's own maintenance rule asks for temporary entries to track their resolution. The entry is not gone, though: #764 orders up to eight entries and promotes on the ninth `assoc`, but `(array-map ...)` with more than eight pairs is built through the same transient and promotes during construction, so it traverses in hash order where Clojure keeps a directly constructed array map ordered at any size.

## What changed

- The "Temporary `array-map` mismatch" entry now records what #764 fixed, the residual with a transcript from `main` at `a6763e77`, the intended contract, and #827, which tracks the fix.
- The decision paragraph and the portable-code guidance say `array-map` gives insertion order for up to eight entries, instead of "after the mismatch is resolved".
- `last-verified` bumped. No other entry changed; both `char` validation defects still reproduce on the same binary.

## Verification

```clojure
(keys (array-map :i 1 :h 2 :g 3 :f 4 :e 5 :d 6 :c 7 :b 8 :a 9))
;; a6763e77: (:b :g :e :f :a :c :h :d :i)
(keys (array-map :h 1 :g 2 :f 3 :e 4 :d 5 :c 6 :b 7 :a 8))
;; a6763e77: (:h :g :f :e :d :c :b :a)
(int (char 55296))                  ;; 55296, still accepted
(int (char 18446744073709551681N))  ;; 65, still truncated
```

`vm.NewArrayMap` in `pkg/vm/persistent_map.go` builds on a transient, which is where the ninth-entry promotion lives. `scripts/docs_frontmatter_hook.py --check` passes.

### DIFF
--- a/docs/known-divergences.md
+++ b/docs/known-divergences.md
@@ -1,6 +1,6 @@
 ---
 status: active
-last-verified: 2026-08-23
+last-verified: 2026-09-08
 authoritative-for:
   - known-clojure-divergences
 ---
@@ -45,8 +45,11 @@ that appears stable for particular values or a particular release remains an
 implementation detail.
 
 Ordering should come from an explicitly ordered collection type. `sorted-map`
-provides comparator order today; `array-map` is intended to provide insertion
-order after the temporary mismatch below is resolved.
+provides comparator order. Since [PR #764](https://github.com/nooga/let-go/pull/764)
+(2026-09-07), `array-map` provides insertion order for up to eight entries, and
+`assoc` of a new key onto an array map holding eight promotes it to the
+intrinsic unordered map, as Clojure's `PersistentArrayMap` does. Construction
+with more than eight pairs is the temporary mismatch below.
 
 ### Why not Clojure?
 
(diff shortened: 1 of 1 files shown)


nooga (approved):
@mparrett Content checks out — I verified the array-map construction-past-eight framing against #764 (merged) and the open #827 tracking issue, and the code in `NewArrayMap`/`NewTransientMap` matches what this entry now claims. CI green, no open threads. Green light.

Just sequence this after #825 — both currently touch `docs/KNOWN_DIVERGENCES.md`/its renamed successor, so rebase onto #825 once that's merged rather than merging in parallel.

Issue #835, opened 2026-09-08
lint: flag breadcrumb and devlog comments


mparrett:
@nnunley asked for this in review on #735:

> We need a linter that checks for breadcrumbs and dev logs in comments. This needs to be cleaned up

Tracking it here rather than in that PR.

## The rule

Flag comments that narrate the change instead of describing the code:

- change narration — "moved from the script", "an earlier version of this test", "used to be", "no longer"
- review dialogue — "as you asked", "per your review", "thanks for"
- roadmap breadcrumbs — "TRANSITIONAL", "will move to"

They go stale as soon as the change lands, and what they carry belongs in the PR body, a design doc, or the issue tracking the follow-up.

## Notes

Comments are dropped before the reader builds a form tree, so this needs a raw-source pass, over `.lg` and `.go` both.

Open: the phrase set and keeping false positives low; report-only or a CI gate; where it lives.

mparrett:
Draft PR: #836 — `scripts/lint.lg`, written in let-go and report-only. The phrase list and the report-only-versus-gate question are both still open there.

mparrett:
Two phrase-list measurements from the #846 sweep. Both are scanner findings, not textual greps, taken by adding the phrase to `devlog-phrases` and running `lg scripts/lint.lg pkg scripts cmd` against `main` at `5dea980a` on 2026-09-11. Baseline there is 13 findings across 566 files.

**"moved to" is a clean add: +2, both true positives.** The list carries "moved from" and "moved into" but not "moved to". Adding it gives 15.

- `pkg/rt/core/ir/passes/typeinfer.lg:169` — a forwarding address for four helpers that live in `ir.ops` now.
- `pkg/vm/var.go:221` — "spawn sites that have not yet moved to a child ExecContext", a roadmap breadcrumb of the kind the rule names.

The two substring shapes I expected to be a problem are not: `moved toward` in `cmd/bench-ratchet/main.go` fails the padded whole-word test, and `total-removed` in `trace.lg` is code rather than a whole-line comment. The scanner already excludes both.

**"no longer" is +22, and that is the argument against it.** It's in the rule text on this issue but not in the implementation. Adding it takes the report from 13 to 35, so it would nearly triple a list that is currently short enough to read end to end.

[...]

PR #836, opened 2026-09-09
feat(scripts): lint.lg — flag breadcrumb and devlog comments

mparrett:
Requested in review on #735 — "We need a linter that checks for breadcrumbs and dev logs in comments" — and tracked as #835.

A devlog comment narrates the change instead of the code: the shape the code had earlier, what a reviewer asked for, what a later change will do to it. It reads as current, goes stale as soon as the change lands, and what it carries belongs in the PR body, a design doc, or the issue tracking the follow-up. There are some on `main` today; they are listed below.

## What this adds

`scripts/lint.lg`, run the way the rest of the let-go tooling in `scripts/` is run:

```
lg scripts/lint.lg [paths...]        # default: pkg scripts
```

It is written in let-go, alongside `generate.lg`, `lgbdump.lg`, and `ir-stress.lg`. Comments never reach the reader's form tree, so this is a raw-source pass over whole-line comments in `.lg` and `.go`, which also gives each finding an exact `file:line` instead of an enclosing-form anchor.

Report-only: it never edits source, and it always exits 0.

## The phrase list

Deliberately small, and limited to wordings that can be true of the change but not of the code. "previously", "originally", and "for now" are left out because they read the same either way.

Matching is on whole words over normalized prose. Without that, `this pr` also matched "this process" and "this primitive": 49 findings over `pkg/` collapsed to 13 once word boundaries were in and the over-firing `no longer` was dropped.

Comments carrying a license header, a tool directive, or `lint:ignore` are skipped.

## The skip-marker bug

`"//go:"` could never match, for the reason @nooga gave: `comment-body` strips the leading `//` from a Go comment before `skip?` runs, so the marker was tested against text that can no longer contain it.

Blocks now carry the raw comment lines alongside the stripped bodies, and `skip?` matches against the raw text. Every marker written with its comment introducer now works, this one included.

`test/lint_lg_test.go` pins it. The fixture holds one line per skip marker plus the matching rules the phrase list leans on — word boundaries, and a phrase that wraps across two lines — and asserts the exact line numbers reported. Reverting the fix turns it red on the `//go:` line.

The tree numbers below are unchanged by the fix. No `//go:` comment on `main` carries a devlog phrase, which is the low frequency @nooga read it as.

## Verification

Against the review that prompted it, over the five source files of #735 — the state just before `ab699061`, the commit that did that PR's comment cleanup, and the merged result:

| | findings |
|---|---|
| before the cleanup | 3, including the header @nnunley's review comment anchored on |
| after the cleanup | 0 |

On the current tree, the default scan (`pkg scripts`) reports 12 findings across 545 files; adding `cmd` brings it to 13 across 566. Measured at `4ef37ddb`, rebased on `main` 2026-09-10. Three are unambiguous: shipped code pointing at a pull request that merged long ago.

```
pkg/nrepl/server.go:295          same hazard this PR fixes for the CLI REPL
pkg/resolver/resolver.go:286     losing the source snippet and stack this PR restores
pkg/ir/lisp_lower_go_test.go     the miscompile this PR fixes
```

`pkg/rt/core/ir/ops.lg:33` carries `(moved from typeinfer.lg's infer-one arms)`, and `pkg/rt/lang.go:412`, `pkg/cli/wasm.go:130`, and `pkg/vm/var_deref_bench_test.go:15` are the same shape.

I read 11 of the 13 closely: nine I would call real, two I would not. `pkg/ir/spike_rpnvm_test.go:128` describes a transformation the pass performs at runtime rather than a change to the code, and `cmd/lginterop/main.go:82` is about deps.edn's history rather than this file's.

## Open questions

@nooga confirmed these are calls for us to make rather than blockers. Here is where I land; adjust any of them and I will follow.

- **The phrase list stays yours to set.** `devlog-phrases` is a literal vector at the top of the file, so adding or dropping a wording is a one-line edit. Shipping it report-only is what keeps that cheap: a wrong entry costs a line of output rather than a red build.
- **Report-only, with nothing wired into the Makefile or CI.** Gating reads better as a follow-up, once the list has been through a few reviews. A gate today would also need the findings above cleaned first, which couples two changes that review better apart.
- **The findings go in a follow-up.** They touch thirteen files across eight packages and are comment-only; that diff is easier to read on its own than bolted onto the tool that found it.


nooga:
Nice heuristic redesign — the word-boundary matching genuinely fixes the false-positive case you mention (`"this pr"` no longer matches `"this process"`), and spot-checking the 12 findings against current `pkg`/`scripts`, ~10 are real breadcrumbs.

Found one concrete bug though: the Go-comment skip marker `"//go:"` can never match. `comment-body` strips the leading `//` from a Go comment line *before* `skip?` checks for the `"//go:"` prefix, so that marker string never appears in the text being tested. Reproduced directly — a synthetic `//go:generate this pr adds a flag` comment gets flagged as a devlog hit instead of being skipped, while `// Code generated by mockgen...` is correctly skipped via the still-working `"code generated"` marker. Low real-world frequency (directive comments rarely narrate PR history) but it's dead code contradicting the stated intent — a one-line fix (check for the marker before stripping `//`, or match on the un-stripped raw line).

Also confirming this genuinely isn't wired into the Makefile or CI yet, so the phrase-list ownership and report-vs-gate questions in your PR description are real open calls for us to make, not blockers on your end.

mparrett:
Fixed in `4ef37ddb`, rebased on current `main`.

`skip?` now matches against the raw comment lines, which each block keeps alongside the stripped bodies. Every marker written with its comment introducer works now, `//go:` included. `test/lint_lg_test.go` pins it: the fixture carries one line per marker plus the word-boundary and wrapped-block cases, and reverting the fix turns it red on the `//go:` line.

Tree numbers are unchanged, so your read on the frequency holds — no `//go:` comment on `main` carries a devlog phrase.

I have answered the three open questions in the body: report-only with nothing wired into the Makefile or CI, the phrase list yours to set, and the findings in a follow-up PR. Out of draft on that basis.

mparrett:
Follow-up cleanup is #846: the 13 findings on `main`, plus three clauses the linter does not report.

Two of those three are notes for the phrase list here. `typeinfer.lg:169` uses "moved to", where the list carries "moved from" and "moved into". `eval.go`'s "now lives" is on the list but was masked — a block reports its first matching phrase and stops, so fixing one phrase in a block can expose a second. Worth knowing for any sweep: re-run rather than trust one pass.

I also read two of the 13 as false positives when I wrote the body above, `spike_rpnvm_test.go:128` and `cmd/lginterop/main.go:82`. Reading them closely for the cleanup, both are devlog after all: the first cites `compactDeadParams`, a symbol no longer in the tree, and the second narrates deps.edn's history rather than what the code does. The linter was right and I was not.

nnunley (approved):

PR #837, opened 2026-09-09
fix(wasm): resolve wasm_exec from active Go toolchain

mparrett:
## Summary

- resolve the Go root by asking the same `go` executable selected for the WASM build
- preserve the existing preference for let-go's build-time toolchain while it is still installed
- fall back cleanly to the PATH toolchain when a cached let-go binary outlives its original Go installation
- cover active-toolchain resolution plus both current and legacy `wasm_exec.js` layouts

## Why

[xsofy#191](https://github.com/nooga/xsofy/pull/191) exposed this after GitHub Actions restored a cached let-go binary built with Go 1.26.7 onto a runner using Go 1.26.8. `runtime.GOROOT()` remained `/opt/hostedtoolcache/go/1.26.7/x64`, which no longer existed, so `lg -w` failed with:

```
error: wasm_exec.js not found in GOROOT (/opt/hostedtoolcache/go/1.26.7/x64)
```

`gomod.GoToolPath()` already handles this lifecycle correctly: it uses the build-time Go executable if that path still exists and otherwise falls back to `go` on PATH. Asking that selected executable for `go env GOROOT` makes the runtime asset lookup follow the toolchain that will actually perform the WASM build.

This preserves the toolchain-consistency intent introduced in #379 rather than always preferring an unrelated PATH toolchain.

## Related

- Cache behavior that created the stale-binary scenario: [nooga/xsofy#163](https://github.com/nooga/xsofy/pull/163)
- Failure that exposed the stale build-time GOROOT: [nooga/xsofy#191](https://github.com/nooga/xsofy/pull/191)
- Companion cache-identity fix: [nooga/xsofy#193](https://github.com/nooga/xsofy/pull/193)
- Post-merge documentation refresh: [nnunley/let-go-wiki#20](https://github.com/nnunley/let-go-wiki/issues/20)

## Verification

- `go test ./pkg/cli ./pkg/gomod`
- `go test -short -timeout 60s ./... -skip TestClojureTestSuite`

### DIFF
--- a/docs/contribution-policy.md
+++ b/docs/contribution-policy.md
@@ -212,6 +212,10 @@ explicit rules:
 - If the default binary grows to carry deps/generator machinery,
   verify the wasm build does not pay for authoring-only code unless
   the user opted into it.
+- `lg -w` resolves Go support assets such as `wasm_exec.js` from the
+  same selected Go toolchain used for the build (while honoring an
+  explicit `GOROOT`). A cached or relocated `lg` binary must not retain
+  its build-time GOROOT as a permanent runtime dependency.
 
 ## 8. Distribution UX target
 
--- a/pkg/cli/wasm.go
+++ b/pkg/cli/wasm.go
@@ -15,7 +15,6 @@ import (
 	"os/exec"
 	"path/filepath"
 	"regexp"
-	"runtime"
 	runtimeDebug "runtime/debug"
 	"strings"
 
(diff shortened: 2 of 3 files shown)


nooga (approved):
Traced `resolveGoRoot`/`readGoWasmExecJS`: it now always queries `gomod.GoToolPath()`'s `go env GOROOT` instead of falling back to `runtime.GOROOT()`, and `GoToolPath()` correctly prefers the build-time toolchain while it's still on disk, falling back to `go` on PATH once it's gone — matches the described lifecycle. I ran the actual fix end-to-end rather than trusting the unit tests alone: this sandbox happens to reproduce the exact GOTOOLCHAIN-switching scenario the bug describes (base `go` 1.25.4, `GOTOOLCHAIN=auto` switching to 1.26.5), and `lg -w <dir> hello.lg` built a working wasm app end to end on the PR's branch. Targeted `go test ./pkg/cli/... ./pkg/gomod/...` and the full suite pass; both the current `lib/wasm` and legacy `misc/wasm` `wasm_exec.js` layouts are covered by the added tests.

Approving — this one's a single docs commit behind current main, please rebase and merge.

PR #838, opened 2026-09-09
build(lgbgen): bootstrap ir.data after the bundle is written

mparrett:
## Why

`ir.data` carries `;; lgbgen:skip`, so it has no namespace chunk in `core_compiled.lgb` and loads from source when `ir.build` requires it. But `lgbgen` compiled it into the shared const pool before any bundled namespace (phase 0), and `EncodeBundleOrdered` emits every const in the pool, each func const dragging its chunk in. So its compiled functions were in the file anyway, under a source name the bundle never executes. `cmd/lgbstat` on `a9c183f9` attributes 111 of 937 chunks, 4,773 of 49,821 code int32s, and 753 sourcemap entries to `<embedded:ir.data:lgbgen-bootstrap>`, and every `lg` binary decodes that on each start.

The phase-1a comment already states the rule this breaks: `ir.*` has to be excluded *before* the bundle is encoded, because filtering the namespace order alone would not keep its bytecode out of the pool. `ir.data` was the one namespace that rule did not cover.

Nothing outside `ir.*` references `ir.data` at compile time (`rg` over `pkg/rt/core` excluding `ir/` finds no `ir.data` or `ir/` symbol), so the only consumers of the bootstrap are the `ir.*` namespaces compiled for Go lowering, and those already run after `writeBundle`.

## What changed

Phase 0 becomes `bootstrapIRData`, called first inside `compileIRForLowering`. Bundle-only mode no longer compiles `ir.data` at all; `--target=go` and `--target=both` compile it after the bundle is written and before the rest of `ir.*`, the same relative order as before. The bundle, `generated.manifest`, and `generated.sums` are regenerated.

| | before (`a9c183f9`) | after |
|---|---|---|
| `core_compiled.lgb` | 311,302 B | 273,748 B |
| chunks | 937 | 824 |
| consts | 2,611 | 2,253 |
| `lg` binary, darwin/arm64 | 14,228,178 B | 14,195,154 B |

Related: #503 (footprint epic) and the bundle-membership thread on #735, which measured the driver's require closure with these chunks already in the baseline.

## Verification

`make generate` regenerates cleanly and `make check-generated` passes. `make test` passes. `GOOS=linux`, `GOOS=plan9 GOARCH=amd64`, and `GOOS=js GOARCH=wasm` build. `cmd/lgbstat` on the new bundle lists no `ir.data` source, and `(require 'ir.passes.pipeline)` still loads from source on the new binary.

Boot, measured with `scripts/smoke-boot.sh` on bootprobes built from both trees, four interleaved rounds of 25 samples each:

| round | before | after |
|---|---|---|
| 1 | 1.836 ms | 1.605 ms |
| 2 | 1.919 ms | 1.627 ms |
| 3 | 1.875 ms | 1.673 ms |
| 4 | 1.764 ms | 1.545 ms |

The probe's phase split puts decode at most of boot, so a 12% smaller bundle showing up as a 12% faster boot is the expected shape.

Note: Originally opened as a draft because the regenerated artifacts would conflict with anything else that regenerates the bundle. As of 2026-09-10, it's clear.


nooga (approved):
Verified the fix empirically rather than just reading the diff: built worktrees at both the PR's stated base commit and its head, diffed the resulting `core_compiled.lgb` directly (311,302 B -> 273,748 B), and confirmed with `cmd/lgbstat` that the embedded `ir.data` bootstrap chunk is present before and absent after — exactly matching the PR's claim that bundle-only mode no longer needlessly bootstraps it. `make generate` (with the correct Go 1.26 toolchain on PATH) reproduces a byte-identical tree and the manifest digest in the diff matches exactly. `make check-generated` passes including the `gogen_ir`-tagged lowered-tree check, and `./build/lg -e "(require 'ir.passes.pipeline)"` confirms `ir.*` still loads correctly now that its bootstrap moved out of the bundle-only path. Full `go test ./...` passes (the 3 failures seen in an isolated worktree are pre-existing worktree-setup artifacts — missing lowered tree / submodule — reproduced identically on plain `main`, unrelated to this PR).

One nit: the PR description's closing line still says "opened as a draft" but the PR itself isn't a draft — worth a quick edit for anyone reading it later.

Approving — please rebase onto current main and merge when ready.

PR #839, opened 2026-09-09
feat(compiler): honor *unchecked-math* for + - * inc dec

abogoyavlensky:
## Honor `*unchecked-math*` for `+ - * inc dec`

`*unchecked-math*` was an inert compatibility var: `set!` worked and nothing changed. The docstring said let-go has no such compile mode, but the hard half was already solved. `CompileMultiple` compiles and evaluates each top-level form in turn, so an earlier form's compile-time effect reaches later ones (#195). A top-level `set!` is therefore already visible at compile time.

When the var is truthy, a call whose head resolves to the core `+`, `-`, `*`, `inc` or `dec` var and is not lexically shadowed is rewritten to the matching `unchecked-*` call. `tryFastOpcode` declines those symbols in the same mode, so `OP_ADD` cannot silently restore the overflow check.

Arity follows Clojure, since `unchecked-*` is strict 2-ary while `+` and `*` are variadic with identities:

| form | rewrite |
|---|---|
| `(+)` `(*)` `(+ x)` `(* x)` | left alone |
| `(- x)` | `unchecked-negate` |
| `(+ a b)` | `unchecked-add` |
| `(+ a b c)` | `(unchecked-add (unchecked-add a b) c)` |

### Two prerequisite bug fixes

The primitives coerced every argument through `ToInt`, so non-int64 types were silently truncated. Each fix is its own commit and stands alone:

| | before | Clojure |
|---|---|---|
| `(unchecked-add 1.5 2.5)` | `3` | `4.0` |
| `(unchecked-add 1/2 1/2)` | `0` | `1` |
| `(unchecked-add 1M 0.5M)` | `1` | `1.5M` |

Only `Int`/`Int` pairs now wrap; `Float`, `Float32`, `Ratio`, `BigInt` and `BigDecimal` route through the numeric tower. That is also what makes the rewrite safe to apply unconditionally.

Replacement symbols are emitted fully qualified, so a local named `unchecked-add` cannot capture the rewrite. No new opcodes: this reuses the existing primitives, costing a function call in unchecked mode only. An opcode version is a reasonable follow-up if profiling justifies it.

Tests in `test/unchecked_math_test.lg` cover wrapping, every identity arity, floats and the other numeric types, qualified heads, shadowing of both `+` and `unchecked-add`, and that the default still raises.

Motivating consumer is `clojure.test.check.random`, whose splittable RNG depends on int64 wrapping. This does not by itself make test.check load; it still needs `proxy`.


nooga (approved):
Verified the arity rewrite matches real Clojure exactly — identity arities `(+) (*) (+x) (*x)` untouched, `(- x)` to `unchecked-negate`, 2-ary direct, N-ary left-nested — via the 44 assertions in `test/unchecked_math_test.lg`, including qualified-head and lexical-shadowing cases. The prerequisite fix to `unchecked-add/subtract/multiply/negate` (no longer truncating non-`vm.Int` args through `ToInt`) is correct: boxed types now route through `vm.NumAdd/NumSub/NumMul/NumNeg` as real JVM `Numbers.unchecked_*` dispatch would. `go build ./...` clean, full `go test ./test/ -count=1` passes, `go generate ./pkg/rt/` reproduces the committed artifacts byte-for-byte.

One thing worth a doc callout rather than a blocker: `*unchecked-math*` is a genuinely global, non-namespace-scoped compiler var (confirmed a `set!` in one `load-string`'d file leaks into later top-level compiles in the same process) — this is not a let-go bug, it faithfully mirrors real Clojure's own compiler-flag semantics, and the PR's own test file already resets the var defensively. Might be worth a one-line note in the docs for anyone loading third-party `.lg` files.

Approving — please rebase onto main and merge. Same note as #841/#840: textual conflict in `compiler.go` if merged around the same time as #841, trivial to resolve plus a `go generate` re-run.

PR #840, opened 2026-09-09
feat(rt): add missing JVM numeric statics (Math/*, Long/bitCount, ...)

abogoyavlensky:
## Add the missing JVM numeric statics

`Math/*` was absent entirely, so any library reaching for `Math/floor` or `Math/abs` on its `:clj` branch failed to compile.

| namespace | added |
|---|---|
| `Math` (new) | `abs` `ceil` `floor` `pow` `log` `exp` `sqrt` `round` `scalb` `getExponent` |
| `Long` | `bitCount` `reverse` |
| `Integer` | `MAX_VALUE` `MIN_VALUE` |
| `Double` | `POSITIVE_INFINITY` `NEGATIVE_INFINITY` `NaN` `isNaN` |
| `Character` | `isDigit` |
| `Byte` | `MAX_VALUE` `MIN_VALUE` |

`Double/isNaN` is included because NaN is not equal to itself, so `Double/NaN` alone leaves no way to test for it.

### Go's `math` is not a drop-in for `java.lang.Math`

Each of these is pinned by a test:

- `Math/round` returns a long and rounds half toward positive infinity, so `-2.5` is `-2`; Go's `math.Round` gives `-3`. It is not `floor(x + 0.5)` either: that addition perturbs the input, so `0.49999999999999994` would give `1`. Java dropped that formula for the same reason (JDK-6430675). Out of range saturates to `Long.MAX_VALUE`/`MIN_VALUE`, NaN is `0`.
- `Math/pow` follows Java, not IEEE, when the base has magnitude 1: IEEE says `pow(1, y)` is `1` for all `y`, Java returns NaN for a NaN or infinite exponent. A zero exponent is `1.0` even for a NaN base.
- `Math/abs` keeps the JVM overloads, long in / long out and double in / double out.
- `Math/getExponent` has no Go equivalent and reads the exponent bits. Zero and subnormals give `MIN_EXPONENT-1`, NaN and infinities give `MAX_EXPONENT+1`.
- `round` and `getExponent` honour the float overloads: `round(float)` returns an int saturating at `Integer.MAX_VALUE`, `getExponent(float)` uses the single-precision bias, so zero is `-127`.
- `Character/isDigit` accepts a character or an int code point, matching both JVM overloads.

Non-finite inputs propagate rather than raising, so `(Math/log -1)` is NaN.

Tests are in `test/host_jvm_statics_test.go` and assert JVM values on exactly these boundary cases.

Motivating consumer is `clojure.test.check`, which needs `Long/bitCount` and most of `Math/*`. This does not by itself make it load.


nooga (approved):
Verified each JVM-vs-Go numeric divergence directly rather than trusting the diff: `Math/round` half-up + saturating to `Long.MAX/MIN_VALUE` on ±∞, `Math/pow` NaN special-casing at `|base|=1`, `Math/abs` long/double overload preservation, `Math/getExponent` bias for zero/NaN/Inf, and `Character/isDigit` char/codepoint overloads — all match real JVM semantics, confirmed independently against the boundary cases in `test/host_jvm_statics_test.go`. `go build ./...` clean, targeted `pkg/rt` tests and full `go test ./test/ -count=1` pass, CI green including CodeQL. These are real implementations of the divergences, not shims, and I didn't find any duplicate registrations or aliasing issues in `installJVMStatics`.

Approving. Please rebase onto current main and merge — only trivial conflict risk is on `generated.sums` if #841/#839 land first, resolved by a plain `go generate ./pkg/rt/` re-run.

PR #841, opened 2026-09-09
feat(compiler): accept Clojure's canonical (. obj member ...) forms

abogoyavlensky:
## Accept Clojure's canonical `(. obj member ...)` forms

let-go supports the `(.member obj args...)` shorthand and the `..` macro, but not the raw `.` special form both are sugar for. Neither Clojure source shape compiled:

```clojure
(. "abcabc" indexOf "b")     ; Can't resolve indexOf
(. "abcabc" (indexOf "b"))   ; Can't resolve indexOf
(. "abc" toString)           ; Can't resolve toString
```

The internal representation quotes the member, `(. obj 'member args...)`, which is what the `(.member obj ...)` rewrite produces. Both source shapes arrive unquoted, so the member was compiled as an expression and failed to resolve.

The fix normalizes them into that shape before compiling. An already-quoted member is left alone, so the existing rewrite path is untouched, as are the reflection warning for a target whose type is not statically known and the source-info propagation used for error locations.

Tests in `test/dot_special_form_test.lg` cover the three failing shapes, pin the two that already worked, and include a raw `.` nested as an argument inside a `..` chain.

Motivating consumer is `clojure.data.priority-map`, which writes `(.. sc comparator (compare (. sc entryKey e) key))`. This does not by itself make it load; its deftype also needs host interfaces.


nooga (approved):
Reviewed the dot-form normalization directly (both `(. obj member ...)` and `(. obj (member ...))` shapes) and traced the `quotedSymbol` guard to confirm it can't double-rewrite or recurse infinitely. Source-location propagation through the rewrite matches the existing `.member"/`Name.` conventions in the same function. `go build ./...` clean, `test/dot_special_form_test.lg` covers all three previously-failing shapes plus the pre-existing working shapes as a regression guard, and `go test ./test/ -count=1` passes with no regressions. Diff is minimal and scoped exactly to the stated goal.

Note for whoever merges #839 first: both PRs append a helper function at the same point in `pkg/compiler/compiler.go`, so merging this after #839 (or vice versa) will hit a textual (not semantic) conflict — trivial "keep both blocks" resolution, then re-run `go generate ./pkg/rt/` for the generated-artifact digest.

Looks good to merge — go ahead and rebase onto current main and merge when ready.

PR #842, opened 2026-09-10
docs: document the sums merge driver and generated-artifact conflict recovery

jasalt:
## Issue
`docs/regenerating-generated-artifacts.md` documents the happy path
with `make generate` but not one with a conflict, and it has drifted since its last content
edit on 2026-06-20.

## Commit message / changes

`docs/regenerating-generated-artifacts.md` was last touched 2026-06-20 and still describes a two-artifact, one-merge-driver world. Two changes have landed since:

  * #375 (2026-07-06) added the `merge=sums` driver for `pkg/rt/generated.sums`, so `make install-hooks` now registers two drivers, not one.
  * #641 (2026-08-16) added `pkg/rt/generated.manifest`, which appeared nowhere in the docs tree.

That left the digest chain described one level off: the doc calls `generated.sums` "a content digest of every `.lg` + `cmd/lgbgen` source", but it is a digest *of `generated.manifest`* (`pkg/genmanifest/manifest.go` hashes `DepManifestRelPath`); the manifest is what carries the per-source hashes.

Documented, all in the existing doc:

  * the ordered chain sources -> .lgb -> .manifest -> .sums, and that the artifacts can only be refreshed together, in that order;
  * both merge drivers, plus why `generated.manifest` deliberately has none (sorted, one record per line, merges as text);
  * that jj skips merge drivers as it skips hooks -- `make generate` after the rebase;
  * a conflict-recovery recipe. The non-obvious step: the marker-laden manifest must be resolved BEFORE `make generate`, because generate.lg's first staleness query parses it and rejects the marker (`malformed manifest line: "<<<<<<< HEAD"`). `core_compiled.lgb` needs no pre-step -- it is `binary` in `.gitattributes`, so git leaves an intact copy behind.

README's contributing note is pluralized to match. Docs only; no behaviour change.

_____
LLM assisted PR (claude opus)


nooga (approved):
Read the full diff and spot-verified its factual claims against the current repo rather than taking them on faith: `scripts/git-merge-sums.sh`/`git-merge-lgb.sh` and both `.gitattributes` merge-driver entries exist as described, `pkg/rt/generated.manifest` and `generated.sums` are both present, and `cmd/check-generated`'s `-write-manifest`/`-write`/`-o` flags match what the doc says they do. The corrected digest chain (`.lg sources -> core_compiled.lgb -> generated.manifest -> generated.sums`) and the conflict-recovery recipe (resolve the manifest/sums conflict markers before `make generate`, since its first staleness query parses the manifest and rejects a marker as a malformed record) are both accurate and non-obvious enough to be worth documenting. Docs-only, no behavior change.

Approving — please rebase and merge when ready.

nooga:
Already approved with a full line-by-line verification. No new review needed — flagging the current blocker for visibility: `mergeStateStatus` is BLOCKED with an empty `statusCheckRollup` ("no checks reported on the branch"), which looks like the required-CI workflow never triggered for this branch (plausibly external-contributor gating) rather than any content issue. A rebase or a manual workflow trigger should unstick it.

mparrett:
@jasalt Quick rebase and this one can get merged.

nooga (approved):
Docs-only change (README + regenerating-generated-artifacts.md), unchanged since the original commit aside from merge-syncs. Confirmed `go build ./...` still succeeds at the PR head.

Approving.

mparrett:
@jasalt Thanks for your patience. Please rebase when you get a chance and I'll keep an eye on this one.

PR #843, opened 2026-09-10
fix(string): clojure.string/split follows Java Pattern.split semantics

nnunley:
## What

`clojure.string/split` was an alias of the raw `core/split` primitive, which wraps Go's `strings.Split` / `Regexp.Split`. That diverges from Clojure in three ways: trailing empty fields were kept, the result was a list, and there was no limit arity. `clojure.string/split-lines` did not exist.

```clojure
(str/split "a\n" #"\n")   ;; let-go: ("a" "")   Clojure: ["a"]
(str/split "\n"  #"\n")   ;; let-go: ("" "")    Clojure: []
```

Clojure delegates to `java.util.regex.Pattern.split`, whose one-argument form is defined as limit zero: *"the pattern will be applied as many times as possible, the array can have any length, and trailing empty strings will be discarded"* ([String#split(String,int)](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/String.html#split(java.lang.String,int))). Interior empty fields are kept, and an input with no match is returned as its own single field, so `""` gives `[""]`.

## How

The Clojure semantics live in `pkg/rt/core/string.lg`. `core/split` stays the raw primitive and gains an optional third argument, Go's `SplitN` count: `n > 0` yields at most `n` fields with the remainder last and `n < 0` keeps every field, which are exactly Java's positive and negative limit cases, so those pass straight through. Only limit zero needs Lisp-side handling: drop trailing empty fields unless the input had no match. Results are vectors. `split-lines` is `(split s #"\r?\n")` as in Clojure.

Existing callers of `core/split` with a string delimiter (semver, scripts) are unchanged.

## Verification

- `test/string_split_test.lg`: every expected value was produced by Clojure 1.12.5 before being written down.
- `make generate` + `make check-generated`: fresh.
- `go test ./...` untagged and `-tags gogen_ir`: green except two e2e subtests that fail for environmental reasons in a jj workspace (`git ls-files` returns nothing there, so the `TestCustomMain` fork fixture is empty); the same package is fine on a git checkout.
- `TestClojureTestSuite`: 242 pass, no `knownFailing` change.

Spec context: `docs/specs/clojure-test-conformance.md` §4.9 in #793.


nooga (approved):
Reviewed and verified. Solid, well-tested fix.

**What it does:** Replaces `clojure.string/split` (previously a raw alias of `core/split`, which just wrapped Go's `strings.Split`/`Regexp.Split`) with real Java `Pattern.split` semantics: limit-0 drops trailing empty fields (except an unmatched input stays a single-element result), adds the limit arity, returns a vector instead of a list, and adds the missing `clojure.string/split-lines`.

**Correctness check:** Traced the limit-0 logic in `string.lg` by hand against all the test cases and Java's actual rules:
- `count == 1` after a full (`-1`) split is a sound proxy for "no match occurred" (each match adds exactly one field), so the short-circuit correctly gives `"" → [""]` and `"a" → ["a"]`.
- The trailing-pop loop correctly stops at the first non-empty-from-the-end field, matching Java's semantics for interior vs. trailing empties.
- Positive/negative limits pass straight through to Go's `SplitN`/`Regexp.Split`, whose sign conventions already match Java's.
- The Go-side `core/split` primitive's existing 2-arg behavior is unchanged (`n` defaults to `-1`), so the only other caller (`let-go.semver`'s unqualified `split`) is unaffected, as the PR description claims.

**Verification performed:**
- Built and ran the full Go unit suite (`pkg/...`) and the `.lg` integration suite (`test/`, excluding `test/e2e`) on this branch in a clean worktree: all green, including the new `test/string_split_test.lg`.
- Re-ran `go generate ./pkg/rt/` and confirmed zero drift — the committed `core_compiled.lgb`/`generated.manifest`/`generated.sums` are reproducible and match.
- The two failure classes I hit while testing in a `/tmp`-based worktree (`os_paths_test.lg` canonical-path mismatch, and `test/e2e` "repo root not found") are both environmental artifacts of running from a linked worktree — not present when run from a stable path — matching what the PR description itself notes.

**On the CONFLICTING mergeable status:** `git merge-tree` shows the only real conflict is in the binary generated blob `pkg/rt/core_compiled.lgb` (main has moved on with unrelated IR work since this branch was cut). The source diff (`string.lg`, `lang.go`) merges cleanly — this is the usual generated-file drift, resolved by rebasing and re-running `go generate`.

**Minor, non-blocking note:** The PR description links `docs/specs/clojure-test-conformance.md` (from #793), which doesn't exist on `main` or this branch yet — a dangling reference, not a defect in this PR.

Issue #506, opened 2026-07-14
Unify loadPrecompiledBundle and BootCore onto one core-load path


mparrett:
`BootCore` (#499) and `pkg/compiler/eval.go`'s `loadPrecompiledBundle` are parallel implementations of the same core boot: decode the embedded core, run the main chunk, eager-run the baseline and hybrid namespace chunks. #499 kept them separate to stay focused. This tracks collapsing them onto one path.

Two things the unify has to get right, both reasons it's worth doing deliberately rather than a mechanical dedup:

- **Fold in the `*ns*` save/restore, don't adopt current behavior.** `loadPrecompiledBundle` carries the same latent `*ns*` gap `BootCore` fixed in c774c44: it iterates the chunk map without restoring `CurrentNS`. It doesn't surface today only because `api.NewContext` re-establishes `*ns*` before user code, whereas `BootCore` returns straight to the caller. The shared path must save/restore and iterate `unit.NSOrder`.
- **Single point for `reapplyGeneratedPrimitives` (coordinates with #438).** Per nnunley's note on #499, a BootCore'd binary that eager-re-Defs a native-primitive var keeps the bootstrap closure and permanently trips the root-deviation counter, losing the direct-call fast path. Whoever lands second needs to call `reapplyGeneratedPrimitives` (or its then-equivalent) after the eager loop; the unified function is the one place to do it.

Part of #503. Coordinates with #438.

PR #846, opened 2026-09-11
docs: rewrite devlog comments to describe the code as it stands

mparrett:
Follow-up to #836, which adds `scripts/lint.lg`. That PR left the cleanup out so the tool and the sweep review separately.

A devlog comment narrates the change rather than the code: the shape the code had earlier, what a reviewer asked for, what a later change will do to it. It reads as current and goes stale as soon as the change lands.

## What changed

The 13 findings the linter reports on `main`, plus three clauses it does not. Each rewrite keeps the reason the comment was written and drops only the narration.

| file | was |
|---|---|
| `pkg/nrepl/server.go:295` | "same hazard this PR fixes for the CLI REPL" |
| `pkg/resolver/resolver.go:286` | "the source snippet and stack this PR restores" |
| `pkg/ir/lisp_lower_go_test.go:2026` | "the miscompile this PR fixes" |
| `pkg/ir/spike_rpnvm_test.go:128` | "what used to be the Go-side `compactDeadParams`" |
| `pkg/rt/coreload.go:38` | "folds together what were three parallel implementations" |
| `pkg/rt/lang.go:412` | "Shadow-warn suppression moved into…" |
| `pkg/rt/core/ir/ops.lg:33` | "(moved from `typeinfer.lg`'s `infer-one` arms)" |
| `pkg/rt/asbytes_test.go:12` | "the one that used to be impossible" |
| `pkg/vm/box_arg_error_test.go:10` | "used to be discarded by `boxArgForReflect`" |
| `pkg/vm/var_deref_bench_test.go:15` | "Before this change every Deref took…" |
| `pkg/compiler/eval.go:158` | "the hybrid order that used to be BootCore-only" |
| `pkg/cli/wasm.go:130` | "It used to be baked into the module scaffolder" |
| `cmd/lginterop/main.go:82` | "deps.edn used to be the only way in" |

The three `this PR` references are in shipped code, naming a pull request that merged long ago. `compactDeadParams` is a symbol no longer in the tree. The `typeinfer.lg` note is deleted rather than rewritten: it was a forwarding address for four helpers that are private to `ir.ops` and consumed by the `op-infer` facets beside them, so it told a reader of `typeinfer.lg` nothing about `typeinfer.lg`.

## Three the linter does not report

Notes for #835, where the phrase list gets settled:

- `pkg/rt/core/ir/passes/typeinfer.lg:169` uses "moved to". The list carries "moved from" and "moved into".
- `pkg/rt/coreload.go`'s "loadPrecompiledBundle only got away without this" matches no phrase at all, and probably should not — the wording is too ordinary to add without a false-positive cost.
- `pkg/compiler/eval.go`'s "now lives" is on the list, but was masked: a block reports its first matching phrase and stops, and "used to be" won. Fixing the first phrase in a block can expose a second, so re-run the linter after a sweep rather than trusting one pass.

## The manifest refresh

`pkg/rt/generated.manifest` digests the source files it tracks, so editing a comment in `ops.lg` or `typeinfer.lg` changes its entry, and `generated.sums` carries the digest of that manifest. Both refreshed through `make generate`. `core_compiled.lgb` and every other generated artifact are byte-identical: comments never reach the reader's form tree, which is the same reason the linter has to read raw source.

## Verification

`cmd/check-generated` passes, `go build ./...` and `go vet` are clean, and the test suites for the seven affected packages pass. Re-running the linter over `pkg scripts cmd` reports 0.


nooga (approved):
Pure comment/doc rewrite — read the full diff, no logic changes anywhere, manifest/sums regenerated to match. CI is fully green. LGTM.

pkg/cli/wasm.go
```diff
@@ -128,8 +128,8 @@ func letgoDepMeta(info *runtimeDebug.BuildInfo) (version, commit string) {
 }
 
 // wasmModuleName is the module name of the throwaway Go module the WASM build
-// scaffolds around its generated sources. It used to be baked into the module
-// scaffolder; it is passed in now that the AOT native path shares that code.
+// scaffolds around its generated sources. It is passed to the module scaffolder
+// rather than baked in, because the AOT native path shares that code.
```
nnunley: Looks like this just changed the order to still give the dev log style comment. :) But, this is better.

pkg/compiler/eval.go
```diff
@@ -160,12 +160,11 @@ func loadPrecompiledBundle() error {
 	// shared with rt.LoadCore and rt.BootCore. EagerHybrids is true here for the
 	// same reason it is in BootCore: api.NewContext returns straight to user
 	// code, so hybrid vars reachable via qualified symbols (which bypass the
-	// on-demand loader) must be bound before then. The spine also folds in the
-	// *ns* save/restore and NSOrder-deterministic hybrid order that used to be
-	// BootCore-only.
+	// on-demand loader) must be bound before then. The spine also owns the *ns*
+	// save/restore and the NSOrder-deterministic hybrid order.
 	//
 	// Decode diagnostics (LG_DECODE_TAG_STATS, #356) still work: the var-ref
```
nnunley, line 166: So, we missed including PR/Issue ids in comments. This line should have been rewritten.

pkg/ir/lisp_lower_go_test.go
```diff
@@ -2028,8 +2028,8 @@ func TestLowerGoStrictExactDTypeArgStaysConcrete(t *testing.T) {
 // c<nid>_. nid is unique only within one function IR, so two nested closure
 // templates can share a nid; a flat prefix then makes the inner closure's
 // param collide with the captured outer param it shadows — the same
-// "captured name shadowed by block param" miscompile this PR fixes, just one
-// level deeper. (Reported on PR #247.)
+// "captured name shadowed by block param" miscompile, one level deeper.
```
nnunley, line 2031: Same deal here - leaving a trail of IDs for fixes does not go in the comments.

pkg/vm/box_arg_error_test.go
```diff
@@ -7,15 +7,14 @@ import (
...
-// caller saw
+// boxArgForReflect must surface the conversion error unboxMapInto returns.
+// Discarding it and falling through to the Unbox fallback hands reflect.Call a
+// *vm.PersistentMap, so the caller sees
 //
 //	reflect: Call using *vm.PersistentMap as type map[string]interface {}
 //
-// rather than the reason. The careful diagnostics unboxMapInto produces were
-// invisible at the one call site a wrapper author actually hits.
+// rather than the reason, and unboxMapInto's diagnostics stay invisible at the
```
nnunley, line 16: Not sure that this is an improvement - it's mostly taking the most cautious change possible that still mostly allows keeping the hint of what used to be.

nnunley (commented):
So, I'd suggest that the prompt for rewriting needs to be tuned. It's still feels like it's figured out to keep the breadcrumbs, but it's changing the tense of the sentences to present tense; It's changing:

This used to do blah, now it does the new thing.

to 

this does blah, rather than what it used to do

mparrett:
@nnunley Excellent feedback, thank you.

mparrett:
You're right. The two halves need different fixes, so let me take them separately.

**The rewrites.** Three of the thirteen did what you describe, and reviewing them properly turned up two more problems than the one you named.

`coreload.go`'s "loadPrecompiledBundle alone could do without it" is the clean example of your point: the previous state, moved into a subordinate clause.

`wasm.go` is subtler than I first read it. "passed to the module scaffolder rather than baked in, because the AOT native path shares that code" — the first half is the re-tensing you describe, but I initially wrote off the second half as unsupported, and it isn't. `pkg/gomod`'s package doc states that rationale already, naming #596 and the three behaviors, and `TestGenerateModuleNameIsCallerSupplied` pins the parameter. The defect is that it's a second copy, on a constant in `pkg/cli`, of something documented where it belongs.

`var_deref_bench_test.go` I'd characterize differently from the other two. Its lock clause is a self-contained counterfactual, and counterfactuals are fine — a benchmark comment should say what it would catch. What was actually wrong is that it names `bindingsMu`, a symbol that no longer exists, and that the rewrite carried the field names `root` and `curr` when the fields are `root` and `rootBind`. Paraphrasing the old comment preserved a name a rename had already invalidated. That is the sharper version of your point: re-tensing skips reading the code, so it inherits whatever has gone stale.

That file had a second one the linter doesn't flag — `BenchmarkVarDerefDistinctParallel` also cites `bindingsMu`. Fixed in the same commit.

**The prompt.** There wasn't a tuned prompt to point at. The rewrite rule lived in whatever I typed that session, so the only place you could review it was its output — here, and on every cleanup after it.

`scripts/lint.lg` is report-only and says so in its header: it decides what to look at and leaves what to do about it unwritten. That gap is where this went wrong. I've drafted the judgement layer as `.agents/skills/comment-lint/SKILL.md`, in the shape `docs-status` already uses for `scripts/docs_status.py` — a hit means re-read the block rather than delete the phrase, a counterfactual about the code as it stands is fine while a re-tensed history is not, rewrite from the code rather than from the old comment, and every causal claim you keep has to name code you can point at repository-wide. That last rule is there because of the `wasm.go` miss above.

That's #858, separate so this one stays a comment cleanup. It's proposed guidance rather than anything enforced — but I'd rather the rule be a file you can send a diff against than something you have to re-derive from my output each sweep, so it's worth your review more than this cleanup is.

The phrase-list half of your feedback I've put on #835 with measured counts. Short version: "moved to" is a clean add at +2, and "no longer" would be +22 with a real false-positive share, which surprised me.

nooga:
Already approved ("Pure comment/doc rewrite... LGTM") and the re-tensing feedback has been substantively addressed, including the concrete `var_deref_bench_test.go` fix — verified present in the current diff. No new review needed.

Status has moved since approval: `mergeable` is now CONFLICTING/DIRTY for real, not just generated-artifact drift — the base is ~15 commits behind main, and several intervening commits (#843, #853, #854) touch the same files this PR rewrites comments in (`pkg/rt/lang.go`, `pkg/compiler/eval.go`, `pkg/resolver/resolver.go`), so expect textual conflicts in the comment regions themselves on top of manifest/sums drift. Needs an author rebase, not just `make generate`.

mparrett:
Dropped this PR's two edits under `pkg/rt/core/ir/` in `066fc86c`, so that subtree is byte-identical to `main` again. The other twelve files are unchanged.

They were `ops.lg:33` and the forwarding note in `passes/typeinfer.lg`. Nothing was wrong with the rewrites — it is that a comment-only edit is a poor reason to put a third party's change into that subtree while four PRs are open across it, and the IR layer is worked on from a separate branch that main tracks rather than the other way round. @nnunley, if you would rather those two were rewritten, say so and they come back as one commit; they are a `git revert` away and the reasoning is in the message.

Two consequences of the revert, stated rather than left to be found:

- `scripts/lint.lg` now reports **1** finding over `pkg scripts cmd`, not 0. It is `ops.lg:33`, restored on purpose. The PR body's "reports 0" line was true before this commit.
- It reports 1 and not 2 because `typeinfer.lg`'s note says "moved to", which is not on the phrase list. So this branch restores two devlog comments and the linter can only see one. That is the `"moved to"` gap from #835, now visible in a diff rather than argued in the abstract — it adds two findings repo-wide, both true positives.

Still open from your review, and not addressed here: the no-IDs-in-comments rule and the `box_arg_error_test.go` objection. The ID rule reaches further than this PR — 146 comment lines across 72 files carry an issue or PR reference on `main` — so I would rather agree the shape with you than guess. My read is that a bare provenance citation (`(Reported on PR #247.)`) should go, while an ID that names something with no other name (`the native adapters #438 Def'd at init`) earns its place, but that is a proposal, not a position I am attached to.

pkg/compiler/eval.go
mparrett: Agreed. `#356` earns nothing here: `LG_DECODE_TAG_STATS` already names the thing, so the number is pure provenance. Dropping it.

Worth flagging the scope before I sweep further, because it is larger than this PR: 146 comment lines across 72 files carry an issue or PR reference on `main`, and `pkg/ir/lisp_lower_go_test.go` alone has 16 in the form `// Regression for #534:`. I have put the counts on #835 so the rule gets settled once rather than per-PR. My instinct is that a bare citation goes, while an ID that is the only name a thing has, like `the native adapters #438 Def'd at init`, is doing work. If you would rather it were flat "no IDs", say so and I will take the whole 146 in one pass.

pkg/ir/lisp_lower_go_test.go
mparrett: Yes. `(Reported on PR #247.)` is the clearest case of the lot. It says nothing about the test. Removed.

The other 15 in this file are the `// Regression for #534:` shape, which is that file's whole convention, so I have left them for the sweep rather than making this file half-converted.

pkg/vm/box_arg_error_test.go
mparrett: Half right, and the half you are right about is worth fixing. The sentences are the old ones re-tensed, which is the same complaint you made on `wasm.go` and I did not carry it through here.

Where I would push back: what is left is a counterfactual, not history. A regression test's comment should say what breaks without the code, and "the fallback hands reflect.Call a *vm.PersistentMap" is checkable against the function as it stands, with no previous version needed. I would rather keep that and cut the padding around it:

```go
// boxArgForReflect must surface the error unboxMapInto returns. Without that,
// the Unbox fallback hands reflect.Call a *vm.PersistentMap and the caller gets
//
//	reflect: Call using *vm.PersistentMap as type map[string]interface {}
//
// instead of the conversion diagnostic, at the one call site a wrapper author
// actually hits.
```

Two lines shorter, and it drops the "discarding it and falling through" framing that was carrying the old shape. If you read the counterfactual itself as the breadcrumb rather than the phrasing around it, tell me and I will cut it to the bare invariant instead.

pkg/rt/coreload.go
```diff
@@ -44,16 +44,15 @@ type CoreLoadOptions struct {
 // configured loader picks them up on (require ...). Returns the decoded unit so
 // a caller can read its const pool or chunk map.
 //
-// It folds together what were three parallel implementations (#506):
-// compiler.loadPrecompiledBundle, rt.LoadCore, and rt.BootCore. Two invariants
-// that used to live in only some of them are now unconditional here:
+// It is the one implementation behind compiler.loadPrecompiledBundle,
+// rt.LoadCore, and rt.BootCore (#506). Two invariants hold here for every
```
nnunley, line 48: Another pr/issue

pkg/cli/wasm.go
mparrett: (Posted by Claude on Matt's behalf.)

Right on both counts. The wasm.go edit reordered the clauses without losing the devlog shape, which is the same thing you flagged on box_arg_error_test.go. That one is carried through in e315fd37; wasm.go stays as it stands, since you read the result as the better of the two.

Resolving this thread. The other three are resolved against e315fd37, where the two ID citations were dropped and the box-arg comment rewritten.

mparrett:
Status note, so the thread resolutions above are not mistaken for reviewer sign-off.

Both re-review requests are still open. nnunley was re-requested on 2026-09-11 after his review comments, nooga on 2026-09-13 after the fixes landed. Neither has re-reviewed.

The `APPROVED` state on this PR is nooga's review of 2026-09-11 10:45, which predates `e315fd37`. That commit is what addressed the inline feedback: the two PR and issue citations dropped, and the box-arg comment rewritten. This repo does not dismiss approvals when the head moves, so the aggregate reads green against a head the approver has not seen.

The four review threads were resolved on 2026-09-13 against `e315fd37`. Three already carried my replies, written alongside that commit; the fourth, on `wasm.go`, got one at resolve time. The resolutions and that reply were posted by Claude on my behalf.

Nothing here is a claim that the re-reviews happened. Not merging while they are outstanding.

pkg/rt/asbytes_test.go
```diff
@@ -10,8 +10,9 @@ import (
 )
 
 // asBytes backs the binary file/stream sinks (spit, write!). The byte-array case
-// is the one that used to be impossible: a byte-array handed to spit/write! would
-// stringify to its #byte-array[…] repr, so bytes >127 could never be written.
+// is the one that needs it: without the coercion, a byte-array handed to
```
nnunley, line 13: Better, but still editorialized

nnunley (approved):

nooga (approved):
Diffed the actual changes in `eval.go`, `lang.go`, and `resolver.go` — all 15 commits are comment-only, no logic changes anywhere. CI is fully green on the current head.

Flagging for the record: per-commit author dates show real comment-polish commits landed after both prior approvals (nooga 09-11, nnunley 09-13T23:43) — nnunley's approval is now ~12 commits stale. Content confirmed comment-only via direct diff, so treating this as a fresh approval rather than asking for a re-review round.

Approving.

pkg/rt/coreload.go
mparrett: (Posted by Claude on Matt's behalf.)

Dropped in `fa2d6414`. `(#506)` trailed a list that already named all three functions, so it was carrying provenance and nothing else.

Rewriting the block turned up a second problem in it, fixed separately in `98f65783`. The sentence read:

> Two invariants hold here for every caller:

That is false for the second one. The NSOrder-deterministic hybrid replay and its `ReapplyGeneratedPrimitives` call sit inside `if opts.EagerHybrids` (`coreload.go:120`), and two of the four call sites pass false: `pkg/rt/boot.go:33`, which is `rt.LoadCore`, and `cmd/bootprobe/main.go:72`. The sentence names `rt.LoadCore` as one of the callers it covers, so it contradicted itself, and `CoreLoadOptions.EagerHybrids` already documents that a `LoadCore + bytecodeNSLoader` caller can leave it false.

It now reads "Two invariants hold here:", with "for every caller" moved onto the `*ns*` bullet where it holds, and the second bullet scoped to `EagerHybrids`.

pkg/rt/asbytes_test.go
mparrett: (Posted by Claude on Matt's behalf.)

Rewritten in `fa2d6414`. You were right, and there were two things wrong, not one.

The editorializing was "the byte-array case is the one that needs it" — a ranking the code does not make. `TestAsBytes` covers three cases and the function treats two of them the same way.

The second problem showed up on re-reading the callers: the sentence was also false. It said a byte-array handed to `spit`/`write!` stringifies to its `#byte-array[…]` repr. That is true of `write!`, which falls back to `vs[1].String()` at `iort.go:253`, and false of `spit`, which returns `spit expected String or byte-array` at `lang.go:6203` and writes nothing. One sentence asserted one behavior for two callers that do different things.

It now describes the function rather than the callers:

```go
// asBytes converts a String or a byte-kind TypedArray to raw bytes and rejects
// everything else. High bytes survive both conversions; the callers that accept
// binary input (spit, write!, the hash primitives) rely on that.
```

The parenthetical is illustrative rather than a full list; `base64-encode`, `base64url-encode`, and `net.go:92` call it too.

mparrett:
Rebased onto current main (`a044ead1`, #862) and force-pushed; the head is now `651bada2`. That replaces the merge commit the branch briefly carried, so the history is linear again and the CONFLICTING state is cleared.

The rebase is content-preserving, which matters given the approval predates it: the whole diff between the reviewed head `e5ec8478` and `651bada2` is exactly #862's own fourteen files, nothing else. The one added commit refreshes the `generated.sums` digest, which the merge driver rewrites without recomputing.

Both open threads are answered above. Two further fixes went in that no review comment asked for, both from re-reading the code under the comments this PR rewrites:

- `pkg/rt/coreload.go`: the `EagerHybrids` problem described in that thread (`98f65783`, manifest refresh in `958504b2`).
- `pkg/vm/box_arg_error_test.go`: the comment said `boxArgForReflect` "must surface the error `unboxMapInto` returns", unconditionally. It surfaces it only when the Unbox fallback cannot serve the target: `native_func.go` guards the return with `if convErr != nil && !usableAsReflectArg(out, target)`. Twelve lines below in the same file, `TestBoxArgFallbackStillServesABoxedGoMap` pins the other half, and its own comment says reporting eagerly "would break calls that work today". Two tests pinned both halves while the comment asserted one of them unconditionally. Fixed in `5dbadca6`, which also drops "at the one call site a wrapper author actually hits"; it names no call site.

`scripts/lint.lg` over `pkg scripts cmd` still reports the single development-note finding at `pkg/rt/core/ir/ops.lg:33`, restored on purpose in `133adb31`. `cmd/check-generated` is clean at the new head and CI is green.

One thing I left alone deliberately. `EPIC-012` in `pkg/rt/lang.go:418` and `STORY-0059` in `pkg/ir/spike_rpnvm_test.go:128` sit on lines this PR does not touch, and they are a different kind of identifier from the ones we have been dropping: they resolve to neither a let-go issue nor a PR, so a reader here cannot follow them, but they do cluster real work — `EPIC-012` appears eight times across five files, and `STORY-0059` heads `pkg/rt/core/ir/passes/cleanup.lg`. A comment cleanup is the wrong place to settle that, so I have put the question on #835 instead of acting on it here.

mparrett:
Correcting my previous comment. I wrote that the whole diff between `e5ec8478` and `651bada2` is "exactly #862's own fourteen files, nothing else". That is wrong: it is eighteen files, and three PRs rather than one. Alongside #862 it carries #803 (`pkg/api/api.go`) and #798 (`pkg/rt/core/test.lg`, `test/run_tests_test.lg`).

The error was mine and mechanical: I verified the diff from the merge commit the branch briefly carried, then wrote the sentence as though I had verified it from the reviewed head. Different baselines.

What I should have said, and have now checked file by file: none of the eighteen is a file this PR authors. Every comment this PR rewrites is byte-identical between `e5ec8478` and `651bada2` — `git diff e5ec8478 651bada2` restricted to the twelve source files and the skill returns empty. The eighteen are main catching up, plus `generated.manifest` and `generated.sums`, which are generated.

Worth stating plainly rather than leaving implied: no human has reviewed `651bada2`. @nooga's approval is at `e5ec8478` and @nnunley's is at `e315fd37`, so the green `APPROVED` is the usual stale aggregate. The authored content is unchanged since both, but the head is not the one either of you read.

mparrett:
@nnunley Landing this. Seven files changed after your approval at `e315fd37`, all comment-only, and most of them come from your 09-11 note about the rewrites keeping the breadcrumb and only re-tensing it.

The `asBytes` comment was the clearest case. Rewriting it from what the function does, rather than from the sentence already there, turned up that the old claim was true of `write!` and false of `spit`, where the coercion returns an error instead of writing anything. That became a section in `comment-lint/SKILL.md`: passing the re-tense test does not make a sentence true.

nooga re-reviewed at the current content on 09-16. Happy to take any remaining concerns in a follow-up.

Issue #845, opened 2026-09-11
io/line-seq turns every read error into end of file


nnunley:
`makeLineSeq` in `pkg/rt/ions.go` treats every error from `ReadString` as end of file:

```go
line, err := br.ReadString('\n')
if err != nil {
    if len(line) > 0 { ... return last line }
    return nil, nil // EOF → empty seq
}
```

A read timeout, a connection reset, or a cancelled context therefore looks like a clean end of the sequence. A consumer parsing a server-sent event stream cannot tell "the server finished" from "the connection died", and any typed error the reader produced is lost.

Proposed fix: only `io.EOF` ends the sequence; any other error is returned from the thunk so realizing the sequence throws it. A partial line before a non-EOF error is dropped (it was never terminated). Covered by `TestLineSeqPropagatesReadErrors` in `pkg/rt/http_timeout_test.go`: the first line is read, the second read hits a stream_read timeout, and `(first (rest s))` throws `http stream_read timeout after 100ms` instead of returning nil.

Issue #844, opened 2026-09-11
http/request ignores timeouts; add connect, request and stream_read scopes


nnunley:
`http/request` derives its context from the execution scope (scope cancellation works) but honours no timeout option: `:timeout_ms` and `:timeout` are ignored, so a peer that never answers, never sends headers, or stalls mid-stream blocks the caller until the scope is cancelled.

An HTTP client used for LLM streaming needs three scopes, as the unified-llm spec Attractor implements describes: connect (dial), request (to the complete response, or to the headers for a streamed body), and stream_read (longest silence between chunks).

Proposed contract, implemented in a local workspace and covered by tests:

- `:timeout {:connect s :request s :stream_read s}` in seconds (numbers), or a bare number meaning `:request`; `:timeout_ms` keeps meaning request scope.
- connect: a pooled `http.Transport` per connect timeout with a `net.Dialer{Timeout}`; error text `http connect timeout after 100ms`.
- request: `context.WithTimeout` for non-streaming requests (whole cycle including body read); for `:as :stream` a timer that cancels the request if headers have not arrived, then stops. Error text `http request timeout after 100ms`.
- stream_read: the streamed body is wrapped so each `Read` must complete within the gap or the request context is cancelled; the read then fails with `http stream_read timeout after 100ms`. Closing the body releases the context.

Tests: `pkg/rt/http_timeout_test.go` (request scope prompt, bare number and `:timeout_ms`, stream_read fires on a stalled body but not on a slow start, request scope on a stream bounds only the headers, connect scope on an unreachable host), passing with `-race`.

Related: line-seq masks the resulting read error (separate issue).

Issue #833, opened 2026-09-08
eval runs the compiled form in the root execution context, ignoring the caller's bindings and scope


nnunley:
## Summary

`eval` compiles the form and runs it in a frame created with `vm.NewFrame` (nil execution context), so the evaluated code runs against the root context regardless of the caller's dynamic bindings or scope. `with-out-str`, `binding`, and structured-concurrency scope around `eval` therefore do not apply to what is evaluated.

## Reproducer

```clojure
(println (pr-str @(future (with-out-str (eval (read-string "(println :captured)"))))))
```

Expected (Clojure semantics, where eval honours thread bindings): prints `":captured\n"`. Observed on `bdd8268c`: `:captured` goes to the real stdout and the result is `""`.

## Where this bit

strange-lettractor evaluates console input in a hub-owned future and needs the printed output captured for the client; with the current `eval` the output leaks to the hub process's stdout.

## Fix

Make `eval` a context-aware native and run the compiled chunk in the caller's context:

```go
evalFn := vm.NewCtxNativeFn("eval", func(ec *vm.ExecContext, vs []vm.Value) (vm.Value, error) {
    ...
    f := vm.NewFrameIn(c.chunk, nil, ec) // NewFrame + f.ec = ec.orRoot()
    out, err := f.RunProtected()
```

with a small exported `vm.NewFrameIn(code, args, ec)` helper. `load-string` has the same shape and would benefit from the same treatment. Local patch plus `TestEvalRunsInCallerContext` in `test/e2e/scope_test.go` in use; happy to open a PR.

Issue #832, opened 2026-09-08
bound-fn*: bound functions run under the root scope, escaping the caller's structured-concurrency scope


nnunley:
## Summary

`bound-fn*` conveys dynamic bindings, but the wrapper it returns is a context-free native (`vm.NativeFnType.Wrap`) that invokes the target through `vm.NewExecContextFrom(snap)`. That fresh `ExecContext` has a nil scope, which `Scope()` normalises to the root `Goroutines` scope. So every call of a bound fn, and every future/go block spawned inside it, runs under the **root** structured-concurrency scope regardless of which scope the caller is in. Closing the caller's scope does not cancel work inside a bound fn.

## Reproducer

```clojure
(def ^:dynamic *x* :root)
(def f (binding [*x* :caller] (bound-fn* (fn [] (sleep 10000) [*x* (scope-cancelled?)]))))
(def parent (scope-open))
(def seen (promise))
(future (deliver seen (f)))
(sleep 50)
(scope-close! parent 5000)
(println (deref seen 1000 :timeout))
```

Expected: `scope-close!` interrupts the sleep and prints `[:caller true]` promptly. Observed on `bdd8268c`: the drain times out after 5 s with the "with-scope drain timed out" warning, then `:timeout`; the bound fn's sleep was running under the root scope. (`scope-cancelled?` is from #830; the drain-timeout symptom does not depend on it.)

## Where this bit

strange-lettractor submits provider/tool jobs to a per-operation coordinator and wraps them in `bound-fn*` so the caller's bindings reach tool callbacks. With the current wrapper the jobs silently escaped the coordinator's owned child scope and became uncancellable.

## Fix

Make the wrapper context-aware and inherit the invoking execution's scope:

```go
return vm.NewCtxNativeFn("bound-fn", func(callEC *vm.ExecContext, args []vm.Value) (vm.Value, error) {
    c := vm.NewExecContextFrom(snap)
    c.SetScope(callEC.Scope())
    return c.Invoke(fn, args)
}), nil
```

Bindings are still re-established from the captured snapshot per call; only the scope now follows the caller, matching Clojure where `bound-fn` never affects which thread pool or interrupt context the work runs in. Local patch with `TestBoundFnKeepsInvokingScope` in `test/e2e/scope_test.go` is in use; happy to open a PR.

PR #847, opened 2026-09-11
Runtime fixes: HTTP scope cancellation and timeouts, streamed http/serve, scoped bound-fn*/eval, EDN data reader

nnunley:
Runtime fixes found while building an LLM client and coding agent on let-go ([strange-lettractor](https://github.com/nnunley/strange-lettractor)). Each change has an issue filed here; they are bundled because later commits build on earlier ones, and I am happy to split this into one PR per issue if that is easier to review.

## Commits

1. **`fix(http): inherit caller scope cancellation`** — #816. `http/get`, `http/post` and `http/request` derive their request context from the caller's execution scope, so cancelling a scope aborts an in-flight request or streamed body instead of leaving it running.
2. **`feat(rt): scope-cancelled?, streamed http/serve bodies, scoped bound-fn*, eval, data reader`**
   - #830 `scope-cancelled?` predicate, so cooperative loops can observe cancellation without sleeping.
   - #831 `http/serve` streams channel and lazy-seq response bodies, flushing each element.
   - #832 `bound-fn*` captures the caller's scope, so bound work runs under the scope that created it.
   - #833 `eval` runs in the caller's execution context (`vm.NewFrameIn`), so `with-out-str` and scope bindings apply inside it.
   - #801 / #823 `clojure.edn/read-string` and `edn/read-all-string` read with data semantics: metadata attaches, real sets, discards splice nothing, duplicate keys and set elements are rejected. Code reading is unchanged.
3. **`http: add connect, request and stream_read timeouts; line-seq surfaces read errors`**
   - #844 `http/request` honours `:timeout {:connect s :request s :stream_read s}` (seconds; a bare number or `:timeout_ms` means the request scope). Connect uses a pooled transport per dial timeout; request bounds the whole cycle; stream_read cancels the request when a body read stalls longer than the gap. Errors name the scope: `http <scope> timeout after <d>`.
   - #845 `io/line-seq` ends a sequence only on `io.EOF`; any other read error (timeout, reset, cancellation) is thrown when the sequence is realized instead of looking like a clean end of file.
4. **`http: the first streamed chunk is bounded by the request scope, later gaps by stream_read`** — #844 follow-up. A model server may process a long prompt before its first token; that wait belongs to the request scope, and stream_read measures the silence between chunks. Found running a local llama.cpp server, where a 30 s stream_read cut off prompt processing.
5. **`chore: refresh the generated-artifact manifest`** — `make generate` after the source changes above. The bundle and lowered Go tree were already current; `pkg/rt/generated.manifest` and `generated.sums` still recorded the old source digests, so `pkg/genmanifest`'s freshness tests failed.

## Tests

- `pkg/rt/http_cancellation_test.go`, `pkg/api/http_cancellation_test.go`, `test/e2e/scope_test.go`: scope cancellation of requests and streamed bodies.
- `pkg/rt/http_timeout_test.go`: each timeout scope, the first-chunk rule, and `line-seq` error propagation; passes with `-race`.
- `pkg/compiler/reader_test.go`: data-reading mode.
- `go test ./...` passes on the branch.

## Notes for review

- The branch is 28 commits behind `main`; I can rebase if you would like it current before review. The generated artifacts would then need `make generate` again on the new base.
- The downstream project has run on this branch for its full suite and live model runs; its notes on each fix are in `docs/let-go-followups.md` there.


mparrett (commented):
Nothing here blocks. This is an agent-assisted review: a Codex pass, then every finding checked against the diff by hand. Two of its findings did not survive that check and are listed at the end. The repro programs are Codex's; I confirmed the mechanisms by reading the code rather than by running them.

## Findings

**P2 — `pkg/compiler/reader.go:1498`: `^meta` is dropped on symbols, the case that matters most for EDN.** `vm.Symbol` is `type Symbol string` and has no `WithMeta`, so it fails the `vm.IMeta` assertion and the metadata is discarded: `(edn/read-string "^String foo")` yields a bare symbol. The comment above the branch explains the fallback as covering "values without metadata support (maps, vectors)"; but `PersistentMap`, `ArrayVector`, and `PersistentVector` all implement `WithMeta`, and `Symbol` does not. The two named types are the ones that work; the unnamed one is the one that loses data. Give `Symbol` metadata support, or reject `^meta` on values that cannot carry it, and add a test over `^:x sym` and `^String sym`: "metadata attaches" is the headline of #801.

**P2 — `pkg/rt/http.go:609`: a streamed request gets the `:request` budget twice.** `headerTimer` bounds reaching the headers and is stopped when `Do` returns; `gapReader` is then constructed with `first: timeouts.request`, a fresh full budget for the first body read. With `{:request 0.1}`, headers at 70 ms and a first byte 70 ms later, the operation takes 140 ms and succeeds. The comment two lines above says the body is governed by `:stream_read` gaps, which the constructor contradicts.

A second path reaches the same bug: `gapReader.started` flips only when `n > 0`, so a `Read` returning `(0, nil)`, legal under `io.Reader`, restarts the first-chunk timer with the whole duration again. Carry one deadline from request start and pass `gapReader` the remainder; that closes both.

**P2 — `pkg/rt/http.go:109`: a fired request deadline is reported as a connect timeout.** `context.DeadlineExceeded` implements `Timeout()`, and a dial killed by the request context surfaces as `net.OpError{Op: "dial"}`, so with `{:connect 10 :request 0.1}` the connect branch matches first and the caller is told `http connect timeout after 10s`. Decide which deadline fired before classifying: naming the scope is the point of these errors.

**P2 — `pkg/rt/http.go:86`: `clientForConnectTimeout` panics on a replaced transport, and bypasses a configured one.** `http.DefaultTransport` is typed `http.RoundTripper`; the `*http.Transport` assertion discards its `ok`, so an embedding application that replaces it leaves `base` nil and `base.Clone()` panics. The same path ignores `http.DefaultClient.Transport`, so a host's mTLS or proxy configuration is dropped whenever `:connect` is set. Handle the non-`*http.Transport` case and start from the configured client. This one bites embedders rather than scripts.

**P3 — `secondsValue` ignores two numeric types silently.** It handles `vm.Int` and `vm.Float`; `Ratio`, `BigInt`, and `BigDecimal` fall to the `default` and return 0, which reads as "not set", so `{:timeout {:request 1/2}}` means no timeout rather than an error. Either accept them or reject them loudly.

**P3 — `pkg/compiler/reader.go:791`: duplicate detection is quadratic.** The map and set checks compare every pair, so a large `edn/read-string` input pays n²/2 comparisons. Nothing in-tree hits this, so treat it as a note, though the LLM-client use case in the description is where a big map would come from.

## Dropped from the Codex pass

Recorded so the review is auditable.

- **`eval` ignoring `*ns*`** came back as the top finding. `ns := rt.CurrentNS.Deref()` is character-identical on the merge base; the PR changed only `NewFrame` to `NewFrameIn`. Pre-existing, and out of scope here. What the PR does introduce is a comment saying eval "runs in the CALLER's execution context", which holds for dynamic bindings and the scope but not for the namespace compilation resolves against. Narrowing that sentence would keep the next reader from expecting more than it delivers.
- **`describeTimeout`'s `connectOnly` parameter** is `false` at both call sites. Dead weight rather than a defect.

## The merge conflict

`git merge-tree` against current `main` resolves everything except `pkg/rt/generated.manifest`. `reader.go`, `lang.go`, `vm.go`, `core/edn.lg`, `core_compiled.lgb`, and `generated.sums` all auto-merge: the sums driver fires and recompiles the bundle cleanly. The manifest is the one committed artifact with no merge driver, so it conflicts on every rebase whatever the content. A rebase plus `make generate` clears it; your last commit already did that work against the old base.

The 28-behind figure in the description is still accurate: merge-base is `bdd8268c`, and `main` has moved 28 commits since.

nooga:
+1 to @mparrett's review — the P2s look correct and worth fixing before merge; direction of the PR (scope-cancellable HTTP, `eval` in caller context, data-mode EDN reader) is sound. One thing to add: no CI has run on this branch at all (`gh pr checks` reports nothing) — worth approving Actions for it so we get real signal before merge, not just the manual review. Agree with mparrett's read that CONFLICTING is just `generated.manifest` (no merge driver); a rebase once #846 lands should clear it.

nnunley:
Superseded by smaller PRs, one per issue, each branched from current `main` with its own test and regenerated artifacts:

- #848 — #816 http clients inherit the caller's scope cancellation
- #849 — #828 empty `:headers` map
- #850 — #830 `scope-cancelled?`
- #851 — #831 streamed `http/serve` bodies
- #852 — #832 `bound-fn*` keeps the invoking scope
- #853 — #833 `eval` in the caller's context
- #854 — #801 / #823 EDN data reading
- #855 — #845 `line-seq` surfaces read errors
- #856 — #844 HTTP timeouts, stacked on #848

Closing this one in their favour.

PR #848, opened 2026-09-11
fix(http): http clients inherit the caller's scope cancellation (#816)

### LINKED ISSUE #816: http/request does not inherit scope cancellation, leaving held requests and workers alive
## Observed

With `lg dev (bdd8268)` / source HEAD `bdd8268c9cb3acf369f0854bada47af79d1d673d`, closing the scope owning a future blocked in `http/request` does not cancel its HTTP request. After a 100ms scope drain plus a 500ms server-observation window, the future is still pending, `scope-live` is 1, and the loopback server reports one held request handler and zero request-context cancellations.

The fixture drains the request body to EOF (bounded to 1 MiB) before reporting readiness, so Go HTTP/1 disconnect detection is active. This is a real socket check, not a mocked transport. After explicitly releasing the fixture, the worker settles and `scope-live` becomes 0; the fixture server exits normally.

The same issue affects an Attractor native OpenAI request waiting for response headers: the high-level operation returns its `:abort` error but the underlying HTTP worker and server request remain active. Attractor also needs to join its workers; adding that join alone would wait on this uncancellable request.

## Root cause

`pkg/rt/http.go` registers `http/request` through `NativeFnType.Wrap`, creates `http.NewRequest`, and calls `http.DefaultClient.Do`. It never attaches the caller's VM scope context. `pkg/vm/scope.go` documents cooperative cancellation of blocking operations through the scope context.

## Reproduction

The `http-cancellation` branch of `nnunley/strange-lettractor` contains:

- `dev/http_cancellation_server.go`: bounded loopback fixture; prints its URL.
- `dev/http_scope_cancellation_check.lg`: native runtime-only reproduction.
- `dev/http_llm_cancellation_check.lg`: downstream Attractor reproduction.
- `docs/let-go-http-cancellation.md`: evidence and restoration criteria.

Run `go run dev/http_cancellation_server.go`, then in another shell:

```sh
lg dev/http_scope_cancellation_check.lg http://127.0.0.1:PORT
```

Actual final output (exit 1):

```clojure
warning: with-scope drain timed out; 1 goroutine(s) still live
{:result :pending, :after {"accepted" 1, "active" 1, "canceled" 0}, :live 1,
 :before {"accepted" 1, "active" 1, "canceled" 0}}
{:cleanup-settled true, :cleanup-live 0}
```

Use a fresh fixture for each probe. The checks release/shut down in cleanup; the server has independent 10-second request and 30-second process safety bounds.

## Expected / suggested direction

Native HTTP requests should inherit the invoking VM scope cancellation context so closing the scope interrupts waiting for headers and reading bodies, settles owned workers, and closes the underlying connection. Use a context-aware native entrypoint and Go request contexts; no JVM-shaped API is needed. Add held-header and held-body tests that observe server-side cancellation and zero live workers.

Separate capability note: this native implementation also ignores forwarded `:timeout`/`:abort_signal` options. These are Attractor conventions, not a claim that let-go already promises those exact option names. Scope cancellation is the independently reproduced runtime defect; the native deadline API can be designed separately.

No local let-go source changes were made.

nnunley:
Fixes #816.

`http/get`, `http/post` and `http/request` build their requests with the calling execution's context (`http.NewRequestWithContext(ec.Context(), …)`), so cancelling the caller's structured-concurrency scope aborts an in-flight request, including one blocked reading an `:as :stream` body. Before this, the request and the goroutine reading it outlived the scope. The three clients carry `:scope-cancellation true` metadata so callers can detect the capability.

## Tests

- `pkg/rt/http_cancellation_test.go`: cancellation before the headers, during the body and of an unread stream, for each client, plus isolation from a sibling scope.
- `pkg/api/http_cancellation_test.go`: the cancellation surfaces as a catchable error.
- Both fail on `main` without the change.

## Notes

- One of the PRs split out of #847. #856 (HTTP timeouts) is stacked on this one.


nooga (approved):
Reviewed (traced the `ExecContext.Invoke` → `ctxProxy` → `http.NewRequestWithContext` dispatch path, then verified empirically by reverting just `pkg/rt/http.go` in a worktree and confirming all 12 new subtests across both test files fail without the fix and pass — under `-race` — with it).

- `http/get`/`http/post`/`http/request` now correctly thread the calling execution's scope context, so cancelling the caller's scope aborts an in-flight request, including an unread `:as :stream` body.
- Non-VM callers (direct host/reflection invocation) fall back safely to the root context — no regression there.
- The `:scope-cancellation true` var metadata addition is additive; these vars had no prior metadata, so nothing is clobbered.
- No races, leaks, or incorrect error wrapping found.
- Generated-artifacts refresh is properly isolated in its own commit.

No blocking issues. **Approving.** #856 (HTTP timeouts) is stacked on this one and should merge right after.

PR #849, opened 2026-09-11
fix(http): http clients accept an empty :headers map (#828)

### LINKED ISSUE #828: http clients panic on an empty headers map
Native HTTP clients panic when passed an empty headers map.

Reproduction in let-go (a listening loopback server is sufficient):

```clojure
(http/request {:url "http://127.0.0.1:PORT/" :headers {}})
```

Omitting `:headers` succeeds. Supplying `{}` returns a caught native nil-pointer
panic, not the HTTP response. Reproduced in local runtime based on bdd8268c with
the separate scoped-cancellation fix applied. Also reproduced in a permanent
let-go loopback checker in strange-lettractor (context_discovery_http_check.lg).

Root cause: PersistentMap.Seq returns vm.EmptyList for an empty map. The header
loops in pkg/rt/http.go iterate while the sequence is non-nil; its first value is
vm.NIL, which passes the Sequable check, but entry.Seq() is nil. Calling es.First()
then panics. The get/post/request loops all use this pattern.

Expected: empty headers behave exactly like omitted headers. A narrow local fix
skips vm.NIL entries before decomposing them. New tests are written in let-go;
this issue does not request a new Go test-server implementation.

nnunley:
Fixes #828.

Each client walks the `:headers` map entry by entry and assumed every entry was a key/value pair. An empty map yields a nil entry, so `http/get`, `http/post` and `http/request` panicked on `:headers {}`. Nil entries are now skipped.

## Tests

- `pkg/rt/http_empty_headers_test.go` calls each client with `:headers {}` against a local server and expects status 200. It fails on `main` without the change.

## Notes


nooga (approved):
Reviewed, and reproduced the underlying bug directly.

Root cause is slightly more precise than the PR description: `PersistentMap.Seq()` on an empty map returns the non-nil `EmptyList` sentinel rather than a nil `Seq`, so the header-iteration loop runs one spurious pass where `entry == vm.NIL`; since `*Nil` also implements `Sequable`, the type assertion succeeds and the loop eventually calls `.First()` on `Nil.Next()` (a literal Go `nil` interface), which panics. Confirmed this exact panic on the pre-fix commit for all three clients (`get`/`post`/`request`), and confirmed `go build` + all three new subtests pass on the fix.

The fix (`if entry == vm.NIL { continue }`) is minimal and can't mask other malformed-header cases — the existing `Sequable` type-assertion already guards those independently. All three clients got the identical fix. No overlap/conflict expected with #848 despite touching the same functions (disjoint line ranges — request-context wiring vs. header-loop body).

No issues found. **Approving.**

PR #850, opened 2026-09-11
feat(rt): add scope-cancelled? (#830)

### LINKED ISSUE #830: rt: expose scope-cancelled? so Lisp code can observe cancellation of its own scope
## Summary

Lisp code has no way to observe that its own scope has been cancelled. Blocking natives (`sleep`, channel ops, promise deref) return early and silently when `ec.Context()` is done, so a loop such as

```clojure
(loop [] (when (running?) (sleep 1) (recur)))
```

turns into a busy spin after its parent scope is closed: every `sleep` returns immediately and nothing signals the cancellation. Only `scope-open`, `scope-close!`, `scope-live` and `scope?` are exposed.

## Where this bit

strange-lettractor runs one coordinator future per LLM operation. When the caller's scope is closed without any explicit abort signal, the coordinator has to notice and drain its child scope. With no predicate it spins forever on `(sleep 1)`. A timing heuristic ("did `sleep 5` return in under 5ms?") is unsound because `System/nanoTime` is `time.Now().UnixNano()` (wall clock, not monotonic).

## Proposal

Add a context-aware native `scope-cancelled?`:

- `(scope-cancelled?)` → `ec.Scope().Context().Err() != nil`
- `(scope-cancelled? scope)` → same for an explicit scope

A local patch (about 25 lines in `pkg/rt/lang.go` next to `scope-open`, plus `TestScopeCancelledPredicate` in `test/e2e/scope_test.go`) is in use on top of `bdd8268c9cb3acf369f0854bada47af79d1d673d`. Happy to open a PR if the shape is acceptable. An alternative or complement would be for `sleep` to throw on cancellation, as an interrupted `Thread/sleep` does in Clojure; that is a larger behavior change.

Related: #829 (lazy-seq thunks realize under the root context, another reason cancellation is hard to reason about from Lisp).

nnunley:
Fixes #830.

Blocking natives such as `sleep` return early and silently when their scope is cancelled, so Lisp code cannot tell "woke up" from "was cancelled". This adds a predicate:

- `(scope-cancelled?)` reports whether the calling execution's current scope is done.
- `(scope-cancelled? s)` inspects a given scope.

A coordinator loop parked on `sleep` can now check after waking and unwind instead of carrying on after its owner has gone.

## Tests

- `test/e2e/scope_cancelled_test.go`: false for a live scope; true for an explicit scope after `scope-close!`; true from inside a future whose parent scope was closed, with the future's `sleep` interrupted. It fails on `main` without the change.

## Notes

### DIFF
--- a/pkg/rt/generated.manifest
+++ b/pkg/rt/generated.manifest
@@ -141,7 +141,7 @@ pkg/rt/core_compiled.lgb pkg/rt/key_tokenizer.go generator 197eba584a6f0a33fc02b
 pkg/rt/core_compiled.lgb pkg/rt/keysource.go generator f7331de29ceb2055b3be9b1a9f4aee6d07b16b01a336306014dff76d86f5f2cd
 pkg/rt/core_compiled.lgb pkg/rt/keysource_js_wasm.go generator d790cddc0c77a3993b668d205cf38264db312d3cff314f48b92a81d82e08073c
 pkg/rt/core_compiled.lgb pkg/rt/keysource_queued.go generator 16e6acbefe44cd071ab3102fef172ff6887770899656b3b595d05955442fbde9
-pkg/rt/core_compiled.lgb pkg/rt/lang.go generator 120947cec62fb42699ea31708119dccc2985b1464f7ca9756329338956d4f47a
+pkg/rt/core_compiled.lgb pkg/rt/lang.go generator 4527edd5543ff25b5f747904e884bf6eea6f1ac5721a0d9e2ae77bdf4a9a98a8
 pkg/rt/core_compiled.lgb pkg/rt/lg_bind_marker.go generator e9732750420945f868e1653f0ee827a3de55530183df970ac87df931ec795ed0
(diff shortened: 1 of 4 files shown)


nooga (approved):
Reviewed and built locally.

- No-arg `(scope-cancelled?)` correctly resolves the calling execution's *live* scope (`ExecContext.Scope()`) — verified it's inherited correctly into a `future` body, not root and not stale.
- Arity-1 form type-checks cleanly and errors sensibly on a non-scope value; wrong-arity is also rejected. Verified both manually.
- Uses the same `context.Context`/`Err()` signal as `sleep`/`scope-close!`/`future*` elsewhere in the runtime — consistent with existing cancellation semantics, including cascading parent cancellation.
- Ran the new e2e test on the branch (passes) and confirmed it fails on `main` (symbol doesn't resolve).

Worth noting for sequencing, not a blocker: on `main` today, `(scope-cancelled?)` called from inside `bound-fn*` would report `false` even after the enclosing scope is cancelled, because `bound-fn*` currently runs in a scope-less root context — that's exactly what #852 fixes, so the two are correct together once both land. Only the generated-artifact files will conflict across this stack, as already called out in the PR body.

PR #851, opened 2026-09-11
feat(http): stream channel and lazy-seq http/serve response bodies (#831)

### LINKED ISSUE #831: http/serve: stream channel and lazy-seq response bodies instead of buffering
## Summary

`http/serve` writes the response body in one shot: `coerceResponseBody` materializes strings, readers and sequences into a single `[]byte` before `resp.Write`. A handler therefore cannot stream a response (SSE, long poll) or hold a connection open mid-body, which makes it impossible to write let-go loopback fixtures for "response headers arrived, then the body stalled" scenarios. Today that needs a Go fixture.

## Proposal

Stream channel and lazy-seq bodies element by element, flushing after each, and stop when the client disconnects:

```clojure
;; paced SSE from a lazy seq
{:status 200 :headers {"Content-Type" "text/event-stream"}
 :body (lazy-seq (cons "data: a\n\n" (lazy-seq (sleep 100) (list "data: b\n\n"))))}

;; producer-driven from a channel
(let [ch (chan)]
  (future (>!! ch "data: a\n\n") (sleep 100) (>!! ch "data: b\n\n") (close! ch))
  {:status 200 :headers {"Content-Type" "text/event-stream"} :body ch})
```

Strings and readers keep the buffered path, so existing handlers are unchanged. A local patch (`streamResponseBody` in `pkg/rt/http.go`, about 50 lines, on top of `bdd8268c9cb3acf369f0854bada47af79d1d673d`) is in use; a client reading with `:as :stream` and `io/line-seq` observes chunks at the expected offsets (`[["a" 1] ["b" 153] ["c" 305]]` for 150ms pacing). Happy to open a PR.

One limitation of the patch: the Lisp handler still cannot observe the client disconnect (the handler has already returned its response map). A follow-up could expose the request context or an `:on-close` hook.

nnunley:
Fixes #831.

A handler that returns a channel or a lazy sequence as `:body` now has it written element by element, with a flush after each, so a response can be held or paced mid-body (server-sent events, long polls, test fixtures that stall a stream). Strings and readers keep the buffered path. Iteration stops when the client disconnects.

## Tests

- `pkg/rt/http_serve_stream_test.go`: for a channel and for a lazy seq, the handler holds the second chunk behind a gate and the test requires the first chunk to reach the client before the gate opens. The gate is always opened on the way out, so a buffering server fails the test instead of hanging it. It fails on `main` without the change.

## Notes


nooga (approved):
Reviewed, and reproduced both pre-fix defects directly against `main`: a channel `:body` serialized the literal Go pointer string (`<chan 0x...>`) instead of streaming, and a lazy-seq `:body` was fully buffered before any bytes were written, defeating SSE/long-poll use cases.

- `http.Flusher` assertion is the safe comma-ok form, nil-checked at every use site — no panic risk on a `ResponseWriter` that doesn't support flushing.
- Channel path uses `select` against `request.Context().Done()`, so it can't block forever on client disconnect; write errors break the loop rather than spinning.
- Ran the new tests under `-race` — clean.

Two non-blocking gaps worth a note to the author, not blockers: a lazy-seq generator that blocks synchronously mid-generation can't be interrupted by client disconnect until it returns (inherent to the seq protocol, not fixable here), and there's no test specifically for the disconnect/write-error paths (verified correct by inspection only). No conflict with #848/#856 — different function in the same file.

**Approving.**

PR #852, opened 2026-09-11
fix(rt): bound-fn* calls run under the invoking scope (#832)

### LINKED ISSUE #832: bound-fn*: bound functions run under the root scope, escaping the caller's structured-concurrency scope

nnunley:
Fixes #832.

`bound-fn*` re-established the captured bindings in a fresh execution context rooted at the root scope, so work done inside a bound fn escaped the caller's structured-concurrency scope: a `sleep` or child task inside it kept running after the caller's scope was cancelled. The wrapper now conveys only the binding stack, and each call runs under the invoking execution's scope.

## Tests

- `test/e2e/bound_fn_scope_test.go`: a bound fn called from a future inside a scope sleeps for ten seconds; closing the scope must interrupt the sleep while the captured binding stays visible, giving `[:caller :woke]`. Detached from the scope, the sleep runs to completion and the deref times out. It fails on `main` without the change.

## Notes

### DIFF
--- a/pkg/rt/generated.manifest
+++ b/pkg/rt/generated.manifest
@@ -141,7 +141,7 @@ pkg/rt/core_compiled.lgb pkg/rt/key_tokenizer.go generator 197eba584a6f0a33fc02b
 pkg/rt/core_compiled.lgb pkg/rt/keysource.go generator f7331de29ceb2055b3be9b1a9f4aee6d07b16b01a336306014dff76d86f5f2cd
 pkg/rt/core_compiled.lgb pkg/rt/keysource_js_wasm.go generator d790cddc0c77a3993b668d205cf38264db312d3cff314f48b92a81d82e08073c
 pkg/rt/core_compiled.lgb pkg/rt/keysource_queued.go generator 16e6acbefe44cd071ab3102fef172ff6887770899656b3b595d05955442fbde9
-pkg/rt/core_compiled.lgb pkg/rt/lang.go generator 4527edd5543ff25b5f747904e884bf6eea6f1ac5721a0d9e2ae77bdf4a9a98a8
+pkg/rt/core_compiled.lgb pkg/rt/lang.go generator fdb1d0c2587950206390d01c70b19b5a6ae0de7c87833e7cd0a0869b3f0e08e9
 pkg/rt/core_compiled.lgb pkg/rt/lg_bind_marker.go generator e9732750420945f868e1653f0ee827a3de55530183df970ac87df931ec795ed0
 pkg/rt/core_compiled.lgb pkg/rt/math.go generator d5860fe3242df6e2afd6a6b69b0e5224e60ce94041a47d823d98225f878d5ed2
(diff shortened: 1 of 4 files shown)


nooga (approved):
Reviewed and traced the dispatch mechanics.

The fix correctly separates two axes: var bindings are still captured at `bound-fn*` *creation* time (Clojure-faithful, unchanged), while the structured-concurrency scope is now adopted from the *invoking* execution at call time — which is what issue #832 needed. Confirmed via `ExecContext.Invoke`/`NewCtxNativeFn` that the wrapper receives the live caller context, and via `scope-open`/`future*` that the scope reaching it at call time is the one actually live in the caller, not root.

Ran the new e2e test directly: passes in ~1.3s, clean under `-race` ×3 — not a real 10s sleep, since scope cancellation interrupts `sleep` almost immediately, so no flakiness concern. Also confirmed 4 pre-existing `.lg` test failures on `main` are identical with or without this PR (unrelated).

Only cross-PR interaction: generated-artifact hash conflicts with sibling #850/#848 on `pkg/rt/lang.go`, already documented in the PR body as expected (`make generate` re-run needed depending on merge order).

PR #853, opened 2026-09-11
fix(compiler): eval runs in the caller's execution context (#833)

### LINKED ISSUE #833: eval runs the compiled form in the root execution context, ignoring the caller's bindings and scope

nnunley:
Fixes #833.

`eval` compiled a form and ran it in a frame bound to the root execution context, so dynamic bindings active around `eval` (`with-out-str`, `binding`) and the caller's scope did not apply to the evaluated code. This adds `vm.NewFrameIn(code, args, ec)`, which builds a frame with a given execution context installed, and `eval` now runs its frame in the caller's context, as thread bindings apply around Clojure's `eval`.

## Tests

- `test/e2e/eval_context_test.go`: `with-out-str` around `eval` inside a future captures what the evaluated form prints. It fails on `main` without the change.

## Notes

### DIFF
--- a/pkg/compiler/eval.go
+++ b/pkg/compiler/eval.go
@@ -284,8 +284,12 @@ func postCoreInit() {
 	lsVar := coreNS.LookupOrAdd(vm.Symbol("load-string"))
 	lsVar.(*vm.Var).SetRoot(loadStringFn)
 
-	// eval: compile and evaluate a single already-read form in the current namespace.
-	evalFn, _ := vm.NativeFnType.Wrap(func(vs []vm.Value) (vm.Value, error) {
+	// eval: compile and evaluate a single already-read form in the current
+	// namespace. The compiled form runs in the CALLER's execution context, so
+	// dynamic bindings active around eval (with-out-str, binding) and the
+	// caller's structured-concurrency scope apply to the evaluated code, as
+	// thread bindings do around Clojure's eval.
+	evalFn := vm.NewCtxNativeFn("eval", func(ec *vm.ExecContext, vs []vm.Value) (vm.Value, error) {
 		if len(vs) != 1 {
 			return vm.NIL, nil
 		}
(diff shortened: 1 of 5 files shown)


nooga (approved):
Reviewed. Built both the PR branch and its parent commit and confirmed the new `test/e2e/eval_context_test.go` genuinely fails pre-fix and passes post-fix (not a tautology).

- `vm.NewFrameIn` is a minimal, correct addition — mirrors the existing internal `newFrameForBytecodeCall` pattern, doesn't miss any frame field, and the single `ReleaseFrame` call site is unchanged (no double-release/leak risk).
- Confirmed `eval`'s frame now correctly resolves dynamic vars and nested calls against the caller's `ec` rather than root.
- Hand-verified a behavior change worth flagging: a blocking call (e.g. `sleep`) inside `eval` can now be interrupted when the caller's structured-concurrency scope closes. This is intentional and consistent with #852's `bound-fn*` fix, but it's **not covered by any test** in this PR — a regression test for it would be a good low-cost follow-up (no need to block on it; #852 already establishes the same pattern is safe).
- Branch is currently `BEHIND main` — worth a rebase before merging.

PR #854, opened 2026-09-11
fix(edn): clojure.edn reads data with Clojure data semantics (#801, #823)

### LINKED ISSUE #801: Reader compatibility: metadata yields a with-meta list and map-interior #_ discard fails

### LINKED ISSUE #823: EDN reader returns a hash-set call instead of a set for set literals
# Set literals read as executable forms rather than set values

Observed with local `lg dev (bdd8268)` while exercising artifact round trips.
This is distinct from #810's nested sequential equality issue: the reader returns
the wrong value type. #801 covers metadata/discards; this adds a set-literal case.
The current upstream HEAD has not been tested; no local runtime edits were made.

Run with `lg -e` and JVM `clojure -M -e`:

```clojure
(require (quote [clojure.edn :as edn]))
(let [v (edn/read-string "#{:a :b}")]
  (prn {:read v :set? (set? v) :equal (= #{:a :b} v)}))
```

Local let-go: `{:set? false, :read (hash-set :b :a), :equal false}`.
JVM Clojure: `{:read #{:b :a}, :set? true, :equal true}`.
Element print order is immaterial; the list-versus-set distinction is the defect.

Local `pkg/compiler/reader.go`'s `readSet` constructs a list headed by `hash-set`.
That lowering form may be useful to the compiler, but must not escape a data read
as the result of a set literal. Expected: a set value, including nested maps and
vectors, without evaluating the result.

Attractor's artifact store correctly rejects this failed non-evaluating round
trip with `:unsupported-artifact-value`. Do not use `eval` as a workaround.
Current store evidence uses supported scalar/vector/map values; set-valued
artifact conformance remains deferred until the reader is corrected. Restore
coverage for both `clojure.edn/read-string` and core `read-string`, direct and
nested sets, and artifact store/retrieve after that fix.

nnunley:
Fixes #801. Fixes #823.

`clojure.edn/read-string` used the code reader, which returns the call forms the compiler consumes: `^{:a 1} x` reads as `(with-meta x {:a 1})`, `#{1 2}` as `(hash-set 1 2)`, and a `#_` discard inside a map drops the preceding key. This adds a data-reading mode to the reader and uses it for `clojure.edn`:

- metadata attaches to values that support it;
- `#{}` reads as a set;
- `#_` inside a collection splices nothing, so an odd map is an error;
- duplicate map keys and set elements are rejected, as Clojure's reader does.

Two natives expose it, `read-data-string` and `read-all-data-string`, and `clojure.edn/read-string` and a new `clojure.edn/read-all-string` call them. Code reading (`read-string`, `load-string`, the compiler) is unchanged.

## Tests

- `pkg/compiler/reader_test.go` (`TestDataReaderFollowsClojureDataSemantics`): metadata, sets, discards, duplicates, and that the code reader still produces call forms. It fails on `main` without the change.

## Notes

- `pkg/rt/core/edn.lg` changes, so the core bundle is regenerated with `make generate` in its own commit. If another PR from the split merges first, re-run `make generate`; only the generated files should conflict.


nooga (approved):
Reviewed in depth given the size of this one (reader change, 7 files). Verified empirically (not just by reading) that discard-splicing, duplicate-key/element rejection, and set/metadata handling all match Clojure's edn semantics, and confirmed by hand-tracing that the data-reading path is genuinely separate from (not post-processing of) the code reader — it has no `#=`/eval surface, consistent with the rest of let-go's reader.

One real, non-blocking divergence from Clojure worth a note: `^{:a 1} 5` (metadata on a non-`IMeta` value) silently drops the metadata instead of erroring, where real Clojure throws. This matches let-go's existing `with-meta` runtime behavior, so it's not a new bug — but it's untested, and the code comment justifying it (`reader.go` ~L161) actually names the wrong example types (maps/vectors *do* carry metadata in this codebase, as this PR's own passing test proves).

Also flagging for awareness, not a blocker: `test/zz_compat_test.go` doesn't glob `edn_test` from the `clojure-test-suite` submodule, so this change — which is explicitly about matching Clojure's edn semantics — doesn't get the clojure-test-suite cross-check CLAUDE.md calls for. And there's no `.lg`-level integration test through `clojure.edn` itself (Go-level coverage only).

CI fully green (`build` passed at 11m26s, full suite including `TestClojureTestSuite` twice). No blocking issues. **Approving** — the metadata-drop behavior and comment fix would make a good quick follow-up.

PR #855, opened 2026-09-11
fix(io): line-seq surfaces read errors instead of ending the sequence (#845)

### LINKED ISSUE #845: io/line-seq turns every read error into end of file

nnunley:
Fixes #845.

`makeLineSeq` treated every error from the underlying reader as end of file, so a timeout, a connection reset or a cancelled request looked like a clean end of the stream, and a consumer parsing server-sent events could not tell "the server finished" from "the connection died". Only `io.EOF` ends the sequence now; any other error is thrown when the sequence is realized. An unterminated final line before EOF is still returned.

## Tests

- `pkg/rt/line_seq_error_test.go`: a reader that delivers one line and then a reset surfaces the error when the second element is realized; plain EOF still ends the sequence and keeps an unterminated last line. It fails on `main` without the change.

## Notes


nooga (approved):
Reviewed and ran the new test plus a full `go test ./... -count=1` on the branch — clean.

- Correctly preserves "unterminated final line still returned at real EOF" (unchanged code path, only reached when `err == io.EOF`).
- Laziness genuinely demonstrated, not just claimed: `TestLineSeqPropagatesReadErrors` realizes element 0 successfully, and only forcing element 1 (whose underlying read is the one that actually fails) throws — proving the error surfaces via the existing `thrownPanic`/`LazySeq.sval()` machinery at realization time, not eagerly at `line-seq` call time.
- The one uncovered edge case — a reader returning a partial line together with a non-EOF error in the same `Read` call, which drops that partial data — is dropped by design per the linked issue's own stated intent, so it's a documented trade-off rather than a gap.
- No coupling with #856 beyond being its natural consumer (timeouts will produce exactly the non-EOF errors this now surfaces correctly).

CI fully green. No blocking issues. **Approving.**

PR #856, opened 2026-09-11
feat(http): connect, request and stream_read timeouts for http/request (#844)

### LINKED ISSUE #844: http/request ignores timeouts; add connect, request and stream_read scopes

nnunley:
Fixes #844.

**Stacked on #848.** The timeouts derive from the caller's execution context, which `http/request` receives only after #848. Until #848 merges, this PR's diff also shows that commit; please review the commits after it.

`http/request` honours three timeout scopes:

- `:timeout {:connect s :request s :stream_read s}` in seconds; a bare number, or the existing `:timeout_ms`, means the request scope.
- **connect** bounds establishing the connection, using a pooled transport per dial timeout so connections are still reused.
- **request** bounds the whole exchange. For `:as :stream` it bounds the headers and the first chunk, because a model server may process a long prompt before sending its first token.
- **stream_read** bounds each later gap between chunks: when a body read stalls longer than that, the request is cancelled and the read fails.

Errors name the scope that fired: `http <scope> timeout after <d>`. Closing a streamed body releases its context.

## Tests

- `pkg/rt/http_timeout_test.go`: each scope fires promptly; a bare number and `:timeout_ms` mean the request scope; stream_read fires on a stalled body but not on a slow start; later chunks are not cut by the request scope; a slow first chunk within the request scope succeeds and past it fails as a request timeout; connect fires on an unreachable host. Passes with `-race`. It fails on `main` without the change.

## Notes

- One of the PRs split out of #847.


nooga (changes requested):
Reviewed in depth given the size (+663/-20, 6 files) and the number of interacting timeout scopes. **Requesting changes** — for two reasons: a structural merge-order block, and two correctness bugs found during review that should be fixed before this lands (not just deferred).

**Structural**: this branch merge-bases 4 commits on top of `main` — the first two are #848's own commits (still open), the last two are #856's own (timeout feature + generated-artifacts refresh, correctly isolated). This can't merge before #848. It should also land after #849, since both touch `pkg/rt/http.go`'s generated artifacts (no logic conflict — #849's edits are inside the `:headers` loop body, which #856 doesn't touch — just a `make generate` re-run needed depending on order). Recommended sequence: **#848 → #849 → #856**.

**Correctness bugs found** (test coverage is otherwise solid — all six scenarios the PR description lists are tested and pass, including under `-race`):

1. **Scope misclassification**: with both `:connect` and `:request` set, a slow dial that straddles the request-scope deadline gets misreported as a *connect* timeout with the *connect* duration, not the request timeout that actually fired. Untested (the connect test only sets `:connect` alone).
2. **`gapReader.fired` is a sticky struct field**, never reset per `Read()` call — a stale fire from an earlier successful-but-racing read can make a later, unrelated error (including plain EOF) get reported as a fabricated `"http stream_read timeout after 0s"`.

Filed both, plus two lower-priority items worth a maintainer decision rather than a blocking fix (`:stream_read` alone leaves the first chunk unbounded; `http/get`/`http/post` silently ignore `:timeout` entirely), as #857: https://github.com/nooga/let-go/issues/857

Everything else checks out: `transportsByConnectTimeout` correctly reuses/dedupes transports per connect-timeout value (verified it's not creating a new pool per request) rather than defeating keep-alive, and the `gapReader`/`LGReader`/`bufio.Reader` interaction correctly preserves "each visible read is individually bounded" under the tested chunk-timing scenarios.

Please address #1 and #2 above (both are small, surgical fixes — not hang/corruption class) before merge; #3/#4 in the linked issue can be a fast follow.

nnunley:
Thanks for the careful review. Both correctness bugs are fixed in the latest push, each with a regression test that fails on the previous head.

**1. Scope misclassification.** Confirmed as a race rather than a deterministic path: when the request deadline expires during a dial, the transport has both the context error and the dial's own timeout ready and returns either. The request deadline is now recorded as its context's cause (`context.WithTimeoutCause`), and `describeTimeout` checks "did the request deadline fire" before the dial-timeout shape.
- `TestDescribeTimeoutPrefersTheRequestScopeThatFired` drives `describeTimeout` directly with a dial `*net.OpError` timeout while the request deadline has fired, so it covers the race deterministically. On the previous head it reported `http connect timeout after 10s`.
- `TestRequestTimeoutWinsOverASlowDial` covers the end-to-end shape, with `:connect` and `:request` both set against a black-hole address, streamed and not.

**2. Sticky `gapReader.fired`.** Fired state is now per read. Each read knows whether its own timer fired (`!timer.Stop()`), and a timer that fires records its real scope and duration. The results:
- the read still delivers any data it got as the deadline passed;
- `io.EOF` is never turned into a timeout;
- a later read that fails because of that cancellation reports the timeout that actually fired;
- any other error passes through unchanged.

`TestGapReaderReportsOnlyTheTimeoutThatFired` drives a `gapReader` with a scripted body: a racing first read followed by EOF, then by a cancellation, then an unrelated reset. On the previous head both of the first two returned `http stream_read timeout after 0s`.

The new and existing timeout tests pass, including under `-race` three times.

**Items 3 and 4 in #857** (`:stream_read` alone leaving the first chunk unbounded; `http/get`/`http/post` ignoring `:timeout`) are left for your decision there, as suggested. My lean: for 3, a `:stream_read` with no `:request` should also bound the first chunk, because "no byte ever arrives" is exactly what a caller setting a read gap wants caught. For 4, wiring `get`/`post` through the same `timeoutsFromOpts` path is a small follow-up if you want it.

**Merge order:** agreed, #848 → #849 → #856. This branch already sits on #848's commits, and it will be rebased and regenerated after those two land.

nooga (approved):
Both correctness bugs from my earlier review are fixed and independently verified:
- Scope misclassification: `describeTimeout` now correctly checks `requestFired` before the dial-shape check, with `context.WithTimeoutCause` for cause inspection. Covered by `TestDescribeTimeoutPrefersTheRequestScopeThatFired` and `TestRequestTimeoutWinsOverASlowDial`, both passing.
- Sticky `gapReader.fired`: replaced with a per-`Read` local `firedHere` plus a CAS'd `expired atomic.Pointer`, `io.EOF` returned before timeout classification. Covered by `TestGapReaderReportsOnlyTheTimeoutThatFired`, passing.

Locally rebasing onto current main (post #848/#849) applies with zero logic conflicts — the reported CONFLICTING/DIRTY status is generated-artifact drift only (`generated.manifest`/`generated.sums`, the #747 class). Full HTTP/scope/eval/bound-fn suites and `-race` pass.

Approving — please rebase onto current main, push a fresh `make generate` commit, and this is ready to merge.

mparrett (approved):

nooga (commented):
Design and correctness look solid — the two bugs found in first review (scope misclassification when connect+request both fire; sticky `gapReader.fired` flag misattributing later errors) are fixed at head and covered by dedicated regression tests. CI is green and this already carries two approvals.

However, `mergeable` currently shows `CONFLICTING`, and it's not pure generated-file drift this time: `git merge-tree` shows a genuine (if narrow) conflict in `pkg/rt/http.go` at the `http/request` context-construction call, because this branch still carries redundant cherry-picks of already-merged #848 instead of being rebased on top of it. Confirmed the PR's side is a proper superset of main's line (wraps `ec.Context()` in the timeout-aware wrapper), so resolving is a "take theirs" on that one hunk, not a design decision.

Suggested fix: rebase onto current main (drops the now-redundant duplicate commits, e.g. `02dc8643`/`c2969710`, leaving only the genuine new commits), resolve the one `http.go` hunk by keeping this branch's `ec.Context()`-based wrapper, re-run `make generate` for the manifest/sums, push. Should be a same-day mechanical rebase, not a new review round. Will approve again once that's done if needed, but the two existing approvals plus this note should be enough to merge post-rebase.

nnunley:
Rebased onto main@ec18f334 (upstream); the two-commit stack's "chore: refresh generated artifacts" commits (for #816, #844, #857) became empty after rebasing onto current main (this PR touches no `.lg` core sources, so nothing to regenerate relative to fresh main) and were abandoned rather than pushed as no-op commits, per the pre-push gate's rule against empty outgoing commits. The PR's own files (`pkg/rt/http.go`, `pkg/rt/http_timeout_test.go`) are unchanged in substance from the reviewed head 63e69fc0; `pkg/rt/generated.manifest`/`generated.sums` were refreshed once more via `make generate` at the final tip. `make check-generated`, `make build`, and `go test ./pkg/...` all pass. Pushed with SKIP=bench-ratchet (the local timing gate is red independent of this change — #791, #899). Note: main advanced again (to #820, b3fd3bbf) mid-push, so GitHub currently shows this as needing another rebase against that — flagging in case a follow-up rebase is needed before merge.

nnunley:
Rebased again onto main@2e1afcee (after #820 and #861 landed); only pkg/rt/generated.sums/generated.manifest conflicted and were regenerated with make generate; pkg/rt/http.go and pkg/rt/http_timeout_test.go unchanged from the reviewed content. make check-generated, make build, and go test ./pkg/... all pass. Pushed with SKIP=bench-ratchet (#791, #899).

PR #802, opened 2026-09-06
docs: add contributor-workflow.md, track an AGENTS.md pointing at it

mparrett:
## Why

The mechanics between a clone and a green PR are spread across the Makefile, `go.yml`, `.pre-commit-config.yaml`, and five docs. A contributor, or a coding agent, working on the implementation has to reassemble them each time. The rule most often missed, regenerate after editing `pkg/rt/core/**/*.lg`, surfaces only after an edit has silently done nothing. This PR adds one page that maps them and tracks a short `AGENTS.md` that points at it.

Tracking `AGENTS.md` reverses a deliberate choice, so it was settled in the thread before landing rather than assumed: @nooga green-lit it on 2026-09-07. `CLAUDE.md` was gitignored on 2026-03-28 (`cb5e4fc4`), `AGENTS.md` on 2026-04-30 (`37a135d2`), and the 113-line `CLAUDE.md` that was still tracked came out on 2026-05-07 (`9b1d56bd`). The case for revisiting it: the file is 36 lines of pointers into `docs/`, not a second copy of the README; the repo already tracks agent material in `.agents/skills/`; and an untracked file goes stale unseen, since the copy I had locally still said CI could not detect a stale `core_compiled.lgb`. `AGENTS.md` is the name most coding agents read, so it holds the text, and `CLAUDE.md` is a one-line import of it so Claude Code reads the same text without a second copy.

## What changed

- `docs/contributor-workflow.md`: the make targets and `go test` invocations CI runs, the generated-artifact rule, which files it commits and that `pkg/rt/core_go_lowered/` is a gitignored build artifact, with a pointer to `regenerating-generated-artifacts.md`, where tests and docs go, the two native registration paths (`ns.Def` in `lang.go` and `//lg:native` via `lgprimgen`) and the three compile paths (direct, `*ir-compile*`, `-tags gogen_ir`), and a table of every job in `go.yml` with what it guards and its local equivalent. Added to the README topical map.
- `AGENTS.md`: the pointer file described above. `CLAUDE.md` is the one-line `@AGENTS.md` import. The two `.gitignore` lines are removed.

## Effect on existing clones

A checkout that has an untracked `AGENTS.md` or `CLAUDE.md` will have its next `git pull` refused with `untracked working tree files would be overwritten by merge`. Move or delete the local copy first; nothing else changes.

## Verification

[...]

PR #858, opened 2026-09-11
docs(skills): comment-lint — how to act on the devlog-comment report

mparrett:
## Why

`scripts/lint.lg` (#836) is report-only by design: findings do not affect its exit status, and its header calls the phrase list a starting point rather than a settled gate. It decides what to look at and leaves what to do about it unwritten, which puts the judgement in whatever prompt the person running it happens to write that day.

#846 is what that costs. Reviewing it, @nnunley noticed the rewrites were keeping the breadcrumbs and flipping them to present tense — "this used to do blah, now it does the new thing" becoming "this does blah, rather than what it used to do". Three of the thirteen did exactly that. Two further defects only surfaced under closer review: one rewrite carried field names a rename had invalidated (`root` and `curr`; they are `root` and `rootBind`), and one kept a causal claim that was true but already documented on the package it was about.

None of those are linter bugs. Removing the matched phrase without removing the history passes a clean run, so the tool cannot be the specification.

## What changed

`.agents/skills/comment-lint/SKILL.md`, in the shape `.agents/skills/docs-status/` already uses for `scripts/docs_status.py` — a report-only script plus the judgement it deliberately leaves out. It covers:

- A hit means re-read the block, not delete the phrase.
- The test: would the sentence still be true and useful to a reader who never saw an earlier version?
- A counterfactual about the code as it stands is legitimate; a re-tensed history is not. The two look alike, which is how this went wrong.
- Rewrite from the code rather than from the old comment — a comment written as history describes code that no longer exists, so its details can be stale in ways the linter cannot see.
- Every causal claim you keep must name code you can point at, repository-wide, and if it checks out, ask where it is already documented before keeping a second copy.
- Never drop an invariant to satisfy a tool that does not gate anything.

Plus the two operational traps: a block reports one phrase and stops — the first in `devlog-phrases` order, not the first appearing in the comment — so re-run after a sweep; and a comment edit in a `.lg` file that `pkg/rt/generated.manifest` lists needs `make generate`. Git-tracked is not the test there: 56 of the 57 `.lg` files under `pkg/` are inputs, while `scripts/lint.lg` and `cmd/lginterop/lginterop.lg` are not.

`.claude/skills/comment-lint` is the symlink into `.agents/skills/`, matching `docs-status`, so Claude Code discovers it without a second copy.

## Verification

Every claim the skill makes about the tool was checked against `scripts/lint.lg` at `5dea980a`: report-only and unaffected exit status, `pkg scripts` defaults, one phrase per block in list order, and the manifest-input rule against `pkg/rt/generated.manifest`. The `.lg` counts were taken the same day.

## Review note

This is proposed guidance, not anything enforced — it governs a sweep only when whoever runs one follows it. @nnunley, this is the "prompt" from your #846 comment, written down so it can be reviewed as a file rather than inferred from its output. Disagreements about the rules are more useful here than on the cleanup.

Separate from #846 on purpose: that one stays a comment cleanup, and this touches the agent-instruction surface #802 is already negotiating.

### DIFF
--- a/.agents/skills/comment-lint/SKILL.md
+++ b/.agents/skills/comment-lint/SKILL.md
@@ -1,0 +1,138 @@
+---
+name: comment-lint
+description: Act on the devlog/breadcrumb comment findings from scripts/lint.lg — decide per hit whether to rewrite or delete a comment, and rewrite it so it describes the code rather than the change that produced it. Use when running the comment linter, cleaning up flagged comments, reviewing a comment-cleanup diff, or writing a comment that describes why code has its current shape.
+---
+
+# comment-lint — acting on the devlog-comment report
+
+Drives `scripts/lint.lg`, the report-only scan for comments that narrate a
+change instead of describing the code. The script decides *what to look at*;
+everything below is the judgement it deliberately leaves out. Rationale and the
+open phrase-list questions: nooga/let-go#835.
+
+## How to run
+
+From the repository root:
+
+```
+lg scripts/lint.lg                  # default: pkg scripts
+lg scripts/lint.lg pkg scripts cmd  # widen the scan
+```
+
+Report-only. It never edits source, and findings do not affect the exit status
+(a run that finds 13 things still exits 0), so a hit is never a build gate and
+never an instruction on its own.
+
+## What a hit means
+
+**A hit flags a comment block to re-read, not a phrase to delete.** The phrase
+list matches tense markers — "used to be", "moved from", "this PR" — because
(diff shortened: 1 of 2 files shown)


nooga (approved):
New skill doc plus symlink into it — verified the symlink resolves correctly and its two load-bearing factual claims (scripts/lint.lg and cmd/lginterop/lginterop.lg absent from generated.manifest) check out against current main. No code changes, no reviews needed.

Approving.

PR #859, opened 2026-09-12
fix(vm): make Int 64 bits wide on every host

mparrett:
## Why

`Int` was `type Int int`, so its width followed the platform. On a 64-bit target that is int64 and everything behaves like Clojure's `long`. On a 32-bit-int target (linux/386, linux/arm, TinyGo wasm) every integer silently narrowed instead:

- `(bit-shift-left 1 40)` returned 0, `(* 1000000000 1000000000)` trapped with an integer overflow, and `unchecked-multiply` wrapped at 2^32.
- A `uint64` coming back from an interop binding was truncated on boxing.
- `MaybeDowngrade` checked `IsInt64()` and then converted through `int()`, so a BigInt that fit int64 but not int32 came back as a truncated `Int`.

So a program was not portable across the two widths even when it never came near overflow. The case that exposed it is xsofy's determinism layer under TinyGo wasm (#504, nooga/xsofy#192): its seed chain is a 64-bit hash, and the 32-bit build computed a different world for the same seed.

## What changed

`Int` is int64 on every host. The paths that went through the platform `int` now go through int64:

- the checked and unchecked arithmetic in `pkg/vm/numbers.go` and `numbers_promote.go`, with `maxIntValue` and the multiply guard fixed at 64 bits;
- the bit-op primitives in `pkg/rt/lang.go` and the opcode fast paths in `pkg/vm/vm.go`;
- the reader's integer literals (`ParseInt` with bit size 64 instead of `Atoi`), `MaybeDowngrade`, and the ratio downgrade;
- the `Long` compatibility layer, `System/currentTimeMillis` and `nanoTime`, int-array reads, and struct-field mapping, which each narrowed a 64-bit producer through `int`;
- `Int.String` (`FormatInt` instead of `Itoa`).

`Int.Unbox` keeps returning the platform `int`: it feeds Go interop (reflect proxies, struct mapping) whose declared types are `int`, so on a 32-bit host a wide value narrows there. `ToInt64` is the full-width read.

Two additions carry that: `MakeInt64` (the cached constructor taking the full value) and `ToInt64` (`ToInt` at full width, with the same coercions). `ToInt` and `MakeInt` keep their `int` signatures for index and length use.

The opcode fast paths store their `Int` results through the small-int cache rather than a bare `f.stack[i] = r`. On a 32-bit host an 8-byte `Int` no longer fits the interface word, so a bare store allocates on every arithmetic op; going through the cache keeps small results allocation-free. `Float` already pays this on those hosts, so `Int` joins it.

## What it costs

Nothing on 64-bit targets, where `int` already is int64: the suite, lint, and xsofy's 328-test game suite are unchanged, and fib(30) plus a small-int loop and a wide-value loop are at parity against `main`.

On a 32-bit host, values outside the small-int cache now allocate when boxed. Measured under TinyGo wasi: fib(27) went from 0.40 s to 3.3 s with bare stores, and back to 0.51 s with the cache routing in this change. Wide values still allocate per result, as they do on 64-bit stock Go for values above 255.

## How it was verified

- `make test`, `golangci-lint run`, and `go run ./cmd/check-generated` are green; the generated manifest is regenerated because `pkg/vm/int.go` is a generator input.
- Cross-compiles for linux/386, linux/arm, js/wasm, wasip1/wasm, and plan9 succeed, and the `pkg/vm`, `pkg/rt` and `pkg/compiler` tests compile for linux/386.
- `pkg/vm/int_width_test.go` pins the width, the wide-value round trip through `MakeInt64`/`String`/`Unbox`/`ToInt64`, the `MaybeDowngrade` fix, and 64-bit checked and unchecked arithmetic. Those assertions are trivially true on a 64-bit target; they exist to fail on a 32-bit one.
- On a 32-bit host, `lg` built with `tinygo build -target=wasi` now prints the same values as native for `(bit-shift-left 1 40)`, `(* 1e9 1e9)`, a wrapping `unchecked-multiply`, `unchecked-negate` of a wide value, and `(int? Long/MAX_VALUE)`. A pure let-go xxh3-64 implementation (the one xsofy carried before its native binding) passes its 61-test golden suite on that build, bit for bit with native, where it previously failed 55 of them. With #860 stacked on top, xsofy's xxh3-64 golden matrix and determinism tests (68 tests, 597 assertions) pass on the same build. The CI TinyGo runtime boot is unchanged.

## Not in this change

- Lowered Go (`-tags gogen_ir`, opt-in) still typed `:int` as Go `int`, so lowered code narrowed on a 32-bit host. That is #862, which builds on this PR.
- **The width guarantee stops at the VM's arithmetic core.** Collection indexing, the `clojure.math` `*-exact` family, `range` bounds, and `ToInt` still convert through the platform `int`, so on a 32-bit host a value above `math.MaxInt32` wraps at those boundaries: `(get [42] 4294967296)` returns `42` rather than `nil`. Those conversions predate this PR at every site; what changes is that a wide `Int` can now reach them. Filed with reproductions as #867 rather than widened here, so this change stays one subject.
- A linux/386 `go test` lane would be the cheapest permanent guard for these paths, since it needs no TinyGo. Not added here.
- The TinyGo guide's int-width entry is updated; the wiki's value-representation page still says `type Int int`.


nooga (approved):
Verified clean: `go build ./...`, cross-compiled `GOOS=linux GOARCH=386` and `GOOS=js GOARCH=wasm` (both `CGO_ENABLED=0`), `pkg/vm`/`pkg/compiler` suites, and the dedicated int-width tests all pass. The one CodeQL alert on `numcache.go` was already correctly dismissed by the author as a false positive (no narrowing occurs post-widening).

This is the base of the #859/#862/#860 stack — approving so it can land first; #862 and #860 build directly on these two commits.

PR #860, opened 2026-09-12
fix(rt): register the xxh3 namespace on TinyGo wasm builds

mparrett:
## Why

`interop_xxh3.go` is lginterop-generated and boxes through reflect, which TinyGo cannot call, so it is gated `!tinygo`. A TinyGo build therefore has no `xxh3` namespace at all, and any program that hashes with it dies at the first call:

```
FATAL ERROR: TypeError: nil is not a function
```

xsofy's determinism layer is one such program (#504, nooga/xsofy#192).

## What changed

`hash_xxh3_wrap.go` adds Wrap-based adapters for the four seed-hash entry points, `Hash`, `HashSeed`, `HashString`, and `HashStringSeed`, in the shape of `hash_murmur3.go`, with the generated bindings' arity checks. They compile on stock Go, so the suite checks them, and on TinyGo wasm, where `interop_xxh3_tinygo.go` registers them as the `xxh3` namespace. Same names, same values.

The build constraint is `!tinygo || wasm`. zeebo/xxh3's scalar path compiles under TinyGo for wasm and wasi, but TinyGo native arm64/amd64 cannot link its assembly once the large-input path is referenced, and its stubs are not `purego`-gated, so a native TinyGo build must not import the package at all. That lane keeps `murmur3` only, as before.

The values are boxed the way `IntType.Box` boxes a `uint64`: reinterpreted as int64, no masking. That is only meaningful where `Int` is 64 bits wide, so this branch carries #859's commit and merges after it; the diff here is the second commit. The TinyGo guide's `xxh3` entry is rewritten to match.

## How it was verified

- `TestXxh3WrapMatchesLibraryAndBoxing` compares every adapter against the library for five inputs, including an empty one and a 300-byte one that takes the large-input path, across five seeds up to 2^64-1; checks the byte-array and string entry points agree and that the boxing matches `IntType.Box`; and checks an extra argument is rejected.
- `go list` selects the two files on TinyGo wasm, the wrapper plus the generated binding on stock Go, and neither on TinyGo arm64.
- `lg` built with `tinygo build -target=wasi` from this branch, which includes #859, runs xsofy's xxh3-64 golden matrix and determinism tests (68 tests, 597 assertions) bit for bit against native, and a 64-bit `xxh3/HashStringSeed` digest prints identically on both.
- The generated manifest is regenerated for the new files under `pkg/rt`.


nooga (approved):
Confirmed via `go list` that build-tag selection is exactly as documented: wasm+tinygo picks up the new wrap-based xxh3 adapters, stock Go picks up the reflect-based ones, native tinygo (no wasm) stays on murmur3-only. `go build ./...` clean, `TestXxh3WrapMatchesLibraryAndBoxing` passes.

Depends on #859 (int64 widening) landing first — approving now so it's ready to go right after.

PR #862, opened 2026-09-12
fix(ir): lower a let-go int to Go int64 on every host

mparrett:
The lowered Go tree (`-tags gogen_ir`) typed a let-go `:int` as Go `int`, which is host-width. On TinyGo wasm, `linux/386` and `linux/arm` that meant the lowered core computed in 32 bits while the VM held 64.

The first symptom is that it does not compile. `hash-combine` mixes in the golden-ratio constant 2654435769, which exceeds `math.MaxInt32`:

```
pkg/rt/core_go_lowered/core/core.go:8556:14: 2654435769 (untyped int constant) overflows int
```

The deeper one is that where it does compile, lowered code and bytecode can disagree, because the same value is narrower on one path than on the other. That split is latent rather than live: the lowered tree is opt-in, and nobody builds it under TinyGo today. This change is what makes building it there possible.

## What changed

`go-type-spec` maps `:int` to `"int64"`, and the spec comparisons keyed off it follow. Three vocabularies keep `"int"` on purpose, since a native prim may still declare a host-width param: `coercion-supported-types`, `scalar-guard-specs`, and `box-result-from-type`.

Two narrowings were independent of the spec, and would have survived a search-and-replace:

- `unbox-from-value` emitted `int(v.(vm.Int))` directly.
- `unsigned-bit-shift-right` went through Go's `uint`, so on a 32-bit host it shifted the wrong width as well as truncating the result. It is now `int64(uint64(l) >> uint64(r))`.

Five registry entries declare an `int` param: `subs` at arity 2 and 3, `nth` at 2 and 3, and `deref` with a timeout. Those take `int64`, so a lowered call still matches the signature instead of falling back to the trampoline. `Nth` bounds-checks in 64 bits before narrowing to a slice index, and rejects an index above `math.MaxInt` rather than truncating it into range: no in-memory collection is that long, so such an index is out of bounds rather than something to wrap. `Deref3` drops a conversion it was already doing internally.

## Verification

`pkg/ir/lisp_lower_int_width_test.go` covers both narrowings. One test pins the wide constant and an `int64` signature, the other pins the explicit `uint64` shift. Both were checked by reintroducing the bug and confirming they fail.

CI has no TinyGo AOT job, so the result it cannot check is:

```
tinygo build -target=wasi -tags gogen_ir -o lg-aot.wasm .
```

That now succeeds, and the module runs under `wasmtime` with output identical to the bytecode build on a core seq and string benchmark.

Whether AOT is faster on that lane is not measured here. On the machine used for this work, an A/A control running the same binary in both arms swung 17%, wider than the difference being looked for, so any number would have been noise.

## Relationship to #859

This depends on #859, which is what makes the constants fit: the lowerer can only promise 64-bit results if the VM holds them. Now that #859 has merged, this branch is rebased onto `main` and carries only its own three commits.


nooga (changes requested):
Confirmed the fix works — `go build -tags gogen_ir ./...` now succeeds natively and cross-compiled for `linux/386` (the golden-ratio constant overflow this PR targets is gone), and `TestLoweringDeterminism` passes.

But there's a reproducible regression: `go test ./test/e2e/ -run TestNativeEntryASTGate` fails 3/6 subtests (`int_arith`, `loop_carried_closure`, `toplevel_effect`). The golden fixtures `test/native-entry/{int_arith,toplevel_effect,loop_carried_closure}.goexpect.json` still expect `int`-typed lowered signatures (e.g. `func AddTwice(ec *vm.ExecContext, arg0 int, arg1 int) int`), but the new lowerer now correctly emits `int64`. Bisected clean: passes on main and on #859 alone, fails only once this PR's commits are added. Not caught by CI because this test self-skips under `-short` and isn't on either of CI's non-short e2e allowlists.

Fix is mechanical: update those three `.goexpect.json` fixtures from `int` to `int64`, mirroring what this PR already did for `test/direct_call_test.lg`.

Requesting changes: update the three golden fixtures before merge.

mparrett:
Fixed in 61c38dae — thanks for the bisect, that saved me the hunt.

All three fixtures now spell the signatures `int64`: `int_arith` and `toplevel_effect` carry the signature plus its locals, `loop_carried_closure` a single `var i int` loop counter. `TestNativeEntryASTGate` passes 6/6, and the full `go test ./test/...` is green.

I swept the other three fixtures rather than assuming your list was the whole set: `closure_capture`, `conditional_closure` and `vm_backed` contain no `int`-typed declaration at all, which is why they were already passing. So the three you named were exactly the affected set.

On the gap that let it through: my pre-push gate was `make test`, which runs `-short`, and this test self-skips there. Running the non-short e2e in a secondary worktree also needs `test/clojure-test-suite` linked, or `TestJankSuiteDirectABIGeneratedGo` fails on the unmaterialized submodule and looks like a second regression. Both are now in my checklist for anything that changes lowered output. Whether the gate belongs on a CI non-short allowlist is your call; it would have caught this without the manual step.

mparrett:
Heads-up on a `pkg/rt/generated.manifest` collision, since this PR's claim note already tracks that hazard against #859.

#846 (devlog comment cleanup) also regenerates the manifest and sums. I've just rebased it onto `e3c3b0a7` and it is MERGEABLE again at `7b58042b` — approved, CI running. It had been sitting CONFLICTING purely on the manifest.

So there are now two manifest-regenerating PRs in flight besides #859, and whichever lands second eats a rebase plus a `make generate`. No action needed from you — flagging it so it arrives as information rather than as a conflict later.

For what it's worth on ordering: #846 is approved and only needed the rebase, while #862 is still awaiting review and blocked on #859 anyway, so #846 landing first would leave you rebasing once rather than #859 and #846 both landing under you. That is your call on your own PR, not a request.

One thing worth knowing if you do end up rebasing across this: **the `merge=sums` driver can produce a wrong digest with no conflict marker.** It recomputes from the worktree at invocation time, but what it hashes is `generated.manifest` — which has no driver of its own (#747), so it may still be conflicted or merely not-yet-regenerated when the driver fires. Seen twice today on #846: driver wrote `69877c2e8b3c`, the true digest was `ce2ef0196f10`. #748 (validate-before-hash) doesn't catch it, because the manifest is well-formed, just stale. So always run `make generate` after the rebase and diff the digest rather than trusting a clean-looking merge:

```
go run ./cmd/check-generated -o /tmp/x && diff /tmp/x pkg/rt/generated.sums
```

Prior occurrences are on #692 and #748; #747 is the tracking issue.

mparrett:
Rebased onto `main` now that #859 has landed — head is b4451b17, three commits, no carried ones.

The branch went `CONFLICTING` the moment #859 squash-merged, since it had been carrying that PR's two commits. Dropping them and replaying mine onto the new `main` resolved it with no conflicts in the IR sources; only `generated.manifest` needed regenerating, which `make generate` handled.

Re-ran the gates on the rebased head: `go test ./pkg/... ./cmd/...` (22 packages), the full non-short `go test ./test/...` including `TestNativeEntryASTGate` at 6/6, `check-generated`, gofmt, and cross-compiles for linux/386, linux/arm, js/wasm, wasip1/wasm and plan9/amd64.

The golden-fixture fix you asked for is still in, as the last of the three commits.

nooga (approved):
Confirmed the fix: `go-type-spec` now maps `:int` → `int64` instead of host-width `int`, closing a real 32-bit correctness gap in the `gogen_ir` lowered-Go backend. Reproduced the pre-fix failure by regenerating the lowered core tree on main and cross-compiling `GOOS=linux GOARCH=386 -tags gogen_ir`: fails with an overflow exactly as described; the same build against this PR's tree succeeds. The three stale `.goexpect.json` golden fixtures flagged in the last review are updated (commit 981c7a1), and `TestNativeEntryASTGate` now passes all 6 subtests in full (non `-short`) mode. Related narrowings (`unbox-from-value`, `unsigned-bit-shift-right`, and the five `int`-param native prims) are all correctly widened to 64-bit, including a sound `i > math.MaxInt` bounds guard in `Nth`/`Nth3`.

All CI green, requested change addressed and independently verified.

Approving.

mparrett:
Wiki follow-up for this change is open as a draft: nnunley/let-go-wiki#25. Three pages carried host-width integers — `go-backend` (the proven-type list and the direct-call example signature), `native-primitives` (the `Subs` annotation example), and `value-representation` (`type Int int`, stale since #859). It is a draft because the `:int` half only becomes true when this merges.

@nnunley, the `unbox-from-value` change here emits `int64(x.(vm.Int))` where it previously narrowed, which is the emission your nominal-type-tracking note describes in `let-go-agentic-stack` (`memory/semantic/DOMAIN_KNOWLEDGE.md`, §Nominal-type tracking in lower-go); that note names `go-type-spec` and `lower_go.lg` as the touch points, and those are the two files this PR changes. Flagging the overlap in case the direction there is more settled than this PR assumes.

nnunley (approved):

mparrett:
@nnunley TY for the approval.

PR #863, opened 2026-09-12
feat(test): port clojure.test — real report/assert/fixture/runner contract

nnunley:
## Read this first: a deliberate divergence from #852

This branch reverts the *mechanism* of #852 ("bound-fn* calls run under the invoking scope", merged four commits before this branch's base) while keeping the property that PR fixed. That needs explaining up front rather than being discovered in review.

#852 rebuilds the call's execution context from the captured binding snapshot alone, then reattaches the caller's scope. That fixes scope escape, but it means a dynamic var the caller bound, which was *not* captured at `bound-fn*` creation time, silently reverts to its root value. Because `*file*` is bound for the whole duration of loading any file, the snapshot is essentially never empty, so in practice this froze every other dynamic var.

Clojure's `bound-fn*` is `with-bindings*` over `push-thread-bindings`, which **pushes** onto the existing thread binding stack rather than replacing it. Uncaptured vars keep the caller's values. This branch matches that: it overlays the captured vars onto a private child of the caller's context.

Two results, both reproducible:

- #852's own regression test, `TestBoundFnKeepsInvokingScope`, **passes** against this implementation. `ExecContext.Child()` copies the parent's scope (`pkg/vm/exec_context.go:59`), which is what #852 needed an explicit `SetScope` for.
- A probe with a caller-bound-but-uncaptured var **passes** here and **fails** under #852's mechanism.

The child context is also what avoids a separate hazard: `pmapv` shares one execution context across workers, and pushing onto it directly let one worker's pop remove another's live binding frame.

A follow-up issue against #852's binding semantics will be filed separately.

## What this adds

`clojure.test`, ported to the real contract. Before, the dialect had a homegrown shim whose `deftest` and `is` printed PASS/FAIL and set a global flag. After:

- `report` as a dynamic multimethod on `:type`, with `do-report` and the 1.12.5 default methods and their exact printed text
- assertions: `is`, `try-expr`, `assert-expr`, including `thrown?` and `thrown-with-msg?`
- `testing`, `are`, the `deftest` family, `test-var`, per-namespace fixtures
- runners that **return summary maps** rather than setting a global flag, plus `successful?`
- `clojure.test.tap` as a separate reporter namespace

The Go test harness now runs `.lg` files through the public API and reads the returned summary, so three temporary bridges are gone rather than left in place.

## Supporting runtime work

The test contract needs things the dialect lacked. Each is useful on its own:

- namespace metadata via `meta` / `alter-meta!`, with a re-entrancy fix (the lock is released during `f`)
- `*file*`, bound during file loads and `NO_SOURCE_PATH` at the REPL, including the `-bundle` and `-wasm` compile paths
- `&form` and `&env` inside `defmacro` (`&env` is hard-coded nil and documented as unimplemented)
- `def` attaches `:ns` and `:name` var metadata
- `*out*` bound to an `io/buffer` captures writes; `*test-out*` resolves through `*out*`'s root
- `meta` on list forms merges reader source positions
- `ex-message` returns `(str v)` for thrown non-exceptions

## Disclosures

- **Bundle size**: `def` now attaches `:ns`/`:name` to every var, which grows `core_compiled.lgb` by about 12%.
- **`clojure.string/split` and `split-lines`**: this branch is rebased onto #843 and uses it directly; the string module here is byte-identical to upstream's.
- **Known pre-existing issues surfaced but not fixed**: `*print-length*`/`*print-level*` are declared in `core.lg` but the printer never consults them, so printing an unbounded value is unbounded (`pkg/vm/exinfo.go:33-44` is one reachable route). And `test/ns_meta_test.lg` is order-flaky under `-shuffle=on`, because `test-all-vars` iterates interned vars in Go map order. Both predate this branch.

## Verification

All green on the rebased branch: `make generate && make check-generated`, `go test ./test/`, the same with `-tags gogen_ir`, `TestClojureTestSuite`, `go test ./pkg/rt/`, and `go vet ./pkg/...`. A fully green run now prints no `FAIL` text on stdout; deliberate-failure probes capture their own output.

**Not yet validated: the benchmark ratchet.** This branch touches `pkg/vm`, `pkg/compiler` and `pkg/rt`, so the perf ratchet is genuinely relevant here. It has not been run, because the machine was saturated (load average 27, an OS security scan) and a ratchet run under that load reports noise rather than signal in either direction. It needs a run on a quiet machine before merge, and the ~12% bundle growth noted above makes `BenchmarkInitFromLGB` the one to watch.


nooga (changes requested):
The port itself looks solid — build, vet, and the full suite (skip only the unavailable clojure-test-suite submodule) are clean, and the rewritten `test/language_test.go` harness (running .lg files through the public `run-tests`/`successful?` API) didn't surface any silent breakage in my pass.

Two things need to happen before this merges, both flagged in the PR description itself rather than hidden:
1. The bench-ratchet run is outstanding — the PR discloses skipping it because the authoring machine was saturated, despite touching `pkg/vm`/`pkg/compiler`/`pkg/rt` and a self-reported ~12% `core_compiled.lgb` bundle growth against `BenchmarkInitFromLGB`. Per repo policy this needs a real run on a quiet machine before merge.
2. This branch deliberately reverts the *mechanism* #852 used for `bound-fn*` execution-context handling, while preserving the property #852 fixed (both covered by passing regression tests, including a new pmapv-context test). The reasoning reads sound, but reversing a previously-merged PR's approach is a design call I want to make explicitly rather than wave through on green tests alone.

Requesting changes: run the bench-ratchet on a quiet machine and report the result, and speak to why reverting #852's mechanism (vs. building on it) is the right call here — then this should be ready.

nooga (commented):
Followed up on the 09-13 review: rebuilt the PR head, `go build ./...` succeeds, and the full test suite (`go test ./pkg/...`, `go test ./test/ -run TestRunner`) passes clean. The port itself looks technically solid.

The two items from that review are still open and no commits have landed since (last commit predates the review by nearly a day):
1. Run the bench-ratchet on a quiet machine and report the ~12% `core_compiled.lgb` growth result.
2. Explain why reverting #852's `bound-fn*` mechanism (while keeping its fix) is the right call vs. building on it.

Separately: the `CONFLICTING` flag is generated-file drift only — `git merge-tree --write-tree origin/main origin/pr/863` shows the sole conflict is `generated.manifest` (compiler.go/reader.go/generated.sums all auto-merge fine), same #747-class issue as elsewhere in this batch. Not a real blocker once rebased.

Once those two questions are answered and it's rebased, this looks close to mergeable.

nnunley:
Rebuilt on current `main`. The `CONFLICTING` flag was generated-file drift only, so each commit's generated artifacts were recomputed from its own sources rather than merged. `ns_meta_test.lg` now asserts fresh metadata on a throwaway `create-ns` namespace instead of the shared `*ns*`, so it no longer depends on deftest order. `make check-generated`, `make test`, and `go test ./pkg/... ./cmd/...` are green. The bench-ratchet run on a quiet machine will follow separately.

**Why `bound-fn*` here replaces #852's mechanism rather than building on it.** The two differ only for a var bound at call time that was not captured at creation time, and on that case Clojure and #852 disagree:

```clojure
(def ^:dynamic *a* 1)  (def ^:dynamic *x* 1)
(def bf (binding [*a* 2] (bound-fn [] [*a* *x*])))
(binding [*x* 55] (bf))
```

| | result |
|---|---|
| JVM Clojure 1.12 | `[2 55]` |
| `main` (#852) | `[2 1]` |
| this PR | `[2 55]` |

Clojure's `bound-fn` pushes the captured frame via `with-bindings`, so uncaptured vars keep the caller's live binding. This PR overlays the captured bindings onto `callEc.Child()`, which reproduces that, keeps #852's fix, and keeps its scope-ownership property — `Child()` copies the caller's structured-concurrency scope, so no explicit `SetScope` is needed. It's `Child()` rather than `callEc` itself because `pmapv` shares one `ec` across workers and binding pops are LIFO per var, not per goroutine. The probe above is the regression test worth keeping.

pkg/rt/core/core.lg
```diff
@@ -1111,11 +1140,28 @@
...
+                name)
+        ;; &env is NOT implemented: it is hard-coded nil here, unlike &form
+        ;; (which is real, bound from *macro-form*). Clojure's compiler binds
+        ;; &env to a map of the lexical environment visible at the macro call
+        ;; site; this runtime tracks no such map, so there is nothing real to
+        ;; bind. A reader should not assume &env has parity with &form just
+        ;; because both are exposed — this is deliberate and pinned by a test
+        ;; (not a placeholder to fill in casually); implementing real &env is
+        ;; separate future work, not something to do here.
+        wrap-body (fn [b] (cons 'let (cons [(symbol "&form") '*macro-form* (symbol "&env") nil] b)))
```
mparrett: Wrapping the entire body in `let` moves a leading pre/post map out of the position where `defn` recognizes it. On this head, `(do (defmacro only-pos [x] {:pre [(pos? x)]} x) (macroexpand (quote (only-pos -1))))` returns `-1`; on the base it throws the assertion. Please inject the implicit bindings without wrapping the condition map (for example, via the macro argument vector), and add a regression test.

pkg/rt/lang.go
```diff
@@ -4386,6 +4456,12 @@ func installLangNS() {
 		if !ok {
 			return form, nil
 		}
+		if CoreNS != nil {
```
mparrett, line 4466: `mf.PushBinding` writes the root execution context, but the macro is invoked through `ec.Invoke`, so a non-root context cannot see the binding. `(do (defmacro reveal-form [] &form) @(future (macroexpand (quote (reveal-form)))))` returns `nil` on this head. Please bind and pop through `ec` (`ec.PushBinding`/`ec.PopBinding`) so `&form` is visible in the invoking context, with a future/child-context regression test.

mparrett (changes requested):
Requesting changes for two correctness issues introduced by the macro port. Both reproduce at the current PR head.

nnunley:
New head `74bfaa15`: the port rebuilt as one commit on current main (`ebd758a6`) from the branch's net change, aligned to the spec in #793 and reconciled against what landed in between. Where the branch and the spec disagreed, the spec won; the reconciliation points:

- **#798** (namespace restored after `run-tests`): kept — `test-ns` and `run-test-var` enter the namespace and restore the previous one in `finally`.
- **#861** (search-path block in `test/language_test.go`): kept verbatim from main.
- **Spec §12.2**: the branch's "namespace current after load" approximation is replaced by the shared classifier (`test_snapshot` / `tested_namespaces` / `classify_loaded`), so the harness runs every namespace a file defined a `deftest` or `test-ns-hook` in. The deliberate-failure probes in `runners_test.lg` / `tap_test.lg` now attach `:test` only for the driven run, and `run_tests_test.lg` is rewritten off the removed `*registered-tests*`.
- **@nooga's inline findings**: the pre/post map is no longer swallowed by `defmacro`'s `&form` binding, and `*macro-form*` is pushed through the exec context so `&form` works inside a `future`. Fixing the first exposed a latent core bug: `fn`'s multi-arity arm tested `list?`, so a macro-built (`Cons`) clause made the whole `fn` nil — which is why `clojure.core/assert` evaluated to nil. Now `seq?`, pinned by a test.

Verified: `make generate`, `make check-generated`, `make build`, `go vet ./...`, `go test ./pkg/... -count=1` (18 packages) and `go test ./test/ -count=1` all green. DoD coverage: §16.1–16.3 and §16.5–16.12 map to tests; §16.4 and the trace-dependent items of §16.9/16.10 are slice 1; the bridge items are slice 3. Pushed with `SKIP=bench-ratchet` (local timing gate red independent of content — #791, #899).

Still open, for me to answer here rather than in code: (1) @nooga's bench-ratchet ask — the ~12% `core_compiled.lgb` growth needs a quiet-machine run, which I will do separately; (2) the rationale for reverting #852's `bound-fn*` mechanism while keeping its fix; (3) a spec-internal tension — §16.12/16.13 want a deliberately failing `test/tap/tap-example.lg` under `test/`, which §12.2's classifier necessarily runs and fails; either the example lives outside the harness walk or the DoD item changes. Drafted with Claude and checked by me.

nnunley:
@nooga on your second question, why #852's `bound-fn*` mechanism is reverted while its fix is kept: the port pushes the bindings that fix needed — `*macro-form*` for `&form`, and `report` for the harness bridge — through the `ExecContext` (`rt.WithBinding`, Section 4.8 of the spec), which is the seam `with-out-str*` already uses and the one a native can install from Go. Keeping `bound-fn*`'s closure-capture path alongside it would have meant two mechanisms for carrying dynamic bindings across a call boundary, each with its own capture point; the exec-context path covers the cases #852 fixed (pinned by the same tests) and is the one the bridge relies on, so the capture path goes. On the bench-ratchet ask: a quiet-machine run on this head is queued and I will post the numbers here. Drafted with Claude and checked by me.

nnunley:
Spec alignment note: #793 (head `1768cf52`) now marks the reference example's `math-test` with `^:expected-failure`, read only by the Go harness (spec 3.4, 12.2, 12.3). When the harness side lands, `test/tap/tap-example.lg` should carry the same marker and keep its intentional failure rather than being rewritten to pass.

nnunley:
Ran `make bench-ratchet` on head `74bfaa15` (Apple M3, machine idle, caffeinate, lowered tree regenerated; raw run in `docs/perf/.runs/ebd758a6e2c5-20260918T235542Z.jsonl` locally).

**Timing:** 0 regressions over the 5% budget. `BenchmarkIRCompile [gogen_ir]` −11%, `[bytecode]` +4.4% (within budget). Anchor drift +0.6%.

**Deterministic (allocs/bytes):** the ratchet reports `BenchmarkInitFromLGB` at 7,041 → 15,752 allocs, but 7,041 is the stale cross-machine row that #900 corrects. Against main's clean capture from earlier today (`49e2a4b3`, same machine) the real delta is:

| metric | main | #863 | Δ |
|---|---|---|---|
| InitFromLGB allocs/op | 11,900 | 15,752 | +32% |
| InitFromLGB bytes/op | 906 KB | 1,240 KB | +37% |
| InitFromLGB ns/op | ~1.00 ms | ~1.21 ms | +20% |
| IRCompile [bytecode] allocs/op | 105,816 | 105,507 | −0.3% |
| IRCompile [gogen_ir] allocs/op | 143,842 | 143,653 | −0.1% |

So the compile paths are unchanged; the cost is entirely bundle init. `core_compiled.lgb` grows 276 KB → 310 KB (+12%) from the ported `test.lg` (+662 lines) and new `test/tap.lg`, and init allocations grow faster than the bundle does because every new var pays the eager def-metadata cost tracked in #897. That's the same mechanism behind the earlier alloc growth, not something specific to this port; the fix belongs in #897 rather than here. Flagging it so the number is on record before merge.

nnunley:
Correction to the bench numbers above, after profiling `BenchmarkInitFromLGB` on both sides: only about half of the +3,850 allocs/op was bundle growth. The other half was this PR's own `attachVarNSMeta`, which called `Var.Meta()` on every def at bundle init and so materialized the lazy `DefMetaPairs` into a real map for ~680 vars that nothing had asked about yet — main keeps them deferred.

Pushed `41c3806f` (same single commit, amended): `Var.materializeMetaLocked` now adds `:ns` from `v.NSRef()` at the moment the pairs are first materialized, and `attachVarNSMeta` leaves deferred metadata alone. `(:ns (meta v))` is unchanged for readers; an explicit `:ns` in the pairs still wins. Two vm tests pin it.

Same machine, same benchmark, 200 iterations:

| metric | main | #863 before | #863 now |
|---|---|---|---|
| InitFromLGB allocs/op | 11,900 | 15,752 | 14,353 |
| InitFromLGB bytes/op | 906 KB | 1,240 KB | 1,051 KB |

The remaining +2,450 is proportional to what the port adds: ~2,000 decoding a 12% larger bundle (more constants and per-def metadata, the #897 cost) and ~450 executing the new top-level forms (16 `defmethod`s on `report`, the `:macro` meta on 11 macros). Bundle bytes: `test.lg` doubles (26 → 63 defs) plus `test/tap.lg`, and `core` grows ~8% from the 40 `ex-info` conversions and `defmacro` binding `&form`/`&env` in every macro body.

nooga (approved):

nnunley:
@mparrett — both of your 09-17 findings are fixed on the current head `52da99e8` (rebased onto main after #744; only `generated.sums` moved). Re-ran your two probes against it:

- `(do (defmacro only-pos [x] {:pre [(pos? x)]} x) (macroexpand '(only-pos -1)))` → `Assert failed: (pos? x)`. `defmacro` now injects the `&form`/`&env` `let` inside each arity's body *after* a leading condition map, using `fn-expand`'s own recognition rule (`pkg/rt/core/core.lg`, `wrap-body`). Regression test included.
- `(do (defmacro reveal-form [] &form) @(future (macroexpand '(reveal-form))))` → `(reveal-form)`. The binding goes through `ec.PushBinding`/`ec.PopBinding` on the invoking context (`pkg/rt/lang.go:4468`), with a future/child-context test.

nooga approved at `41c3806f`; the changes-requested from your review is what remains. Since then the only other change is the lazy `:ns` var-metadata fix described above.

mparrett (approved):
Agent-drafted, posted by @mparrett. Reviewed head `52da99e811aecec7af1f6f07e4f5e764fcc592ec`, diffed against merge base `7148f4ad`.

Both findings from my 09-17 review are fixed, and I checked the fixes hold by reintroducing each bug rather than by reading the patch.

**The pre/post condition map.** `(do (defmacro only-pos [x] {:pre [(pos? x)]} x) (macroexpand '(only-pos -1)))` now throws `Assert failed: (pos? x)`. Reverting `wrap-body` in `pkg/rt/core/core.lg` to wrap the whole body, and regenerating `core_compiled.lgb` — without which the edit is inert and the test passes on a stale bundle — brings back `-1` and turns `macro-precondition-still-fires` red. `macro-body-that-is-a-map-is-a-value` covers the `(defmacro lone-map [] {:a 1})` edge the hoist could have misread.

**`&form` in a child context.** `@(future (macroexpand '(reveal-form)))` returns the call form. Reverting `pkg/rt/lang.go:4468` to `mf.PushBinding` fails `form-visible-in-a-child-context` with `actual: (not (= (quote (capture-form 1 2)) (quote nil)))`, the same `nil` the original report named.

**The rebase since the approval.** @nooga approved `41c3806f` and the current head is a rebase across #744. Comparing blobs for all 63 files in the PR's net change, only `pkg/rt/generated.manifest` and `pkg/rt/generated.sums` differ, both regenerated artifacts, so the approval still covers the source.

**No silent test loss.** This rewrites the harness's discovery, so I compared per-file deftest counts between main and this head over the 220 `.lg` files present on both. No file runs fewer deftests, and the total is 1475 against main's 1469. The 28 files that reported on main and report nothing here all have zero deftests — they are the load-only files the classifier now passes without a summary. `run_test_var_test.lg` and `thrown_test.lg` looked short until I noticed my main-side count also summed the inner `run-test` calls those two files drive; their deftest names are identical on both sides.

**`bound-fn*`.** The table in your 09-16 comment reproduces here: main gives `[2 1]` for the uncaptured-var case and this head gives `[2 55]`, with `(binding [*a* 99] (bf2))` still `7`, so #852's property survives. `[2 55]` is what the JVM does, since `bound-fn*` captures a frame and `with-bindings*` pushes it, so a var the snapshot never captured keeps tracking the caller. That makes this a fix for a divergence rather than one behavior traded for another.

**Spec conformance**, spot-checked against #793: the summary map and `successful?`, `(with-out-str (run-tests 'x))` returning `""` (10.2, and the Appendix B.3 oracle), `run-all-tests` matching on `re-matches` so `#"my\.test\.ex"` matches and `#"test\.ex"` does not, and `report` rebindable and receiving every event. `go test ./test/` and `go test ./pkg/...` are green in a worktree at this head. On CI every check has passed except `build`, which was still running when I posted this.

No findings. Approving, which clears the changes-requested from my 09-17 review.

One unrelated note, pre-existing rather than from this PR: `test/os_paths_test.lg`'s `(is (= (os/canonical-path (os/cwd)) here))` fails whenever the checkout sits under a symlinked path, because `os/cwd` returns the logical path and `canonical-path` resolves it — `/tmp/x` against `/private/tmp/x` on macOS. It behaves identically on main and on this head; I hit it only because I put a scratch worktree under `/tmp`. I can file it separately unless it is already tracked.

PR #868, opened 2026-09-12
feat(lint): objective comment-abuse rules R1-R6 plus code-verbosity catalog

nnunley:
Base of a 2-PR stack (stacked-by: #869).

7 commits: R1 commented-out-code (.lg only), R2 restatement, R3 duplicated-comment, R4 density-outlier, R5 segregation (--edn/--gate), R6 comment-churn (--churn), plus FP fixes and the code-verbosity catalog (scripts/lint-code-rules.edn).

Design: docs/superpowers/specs/2026-09-12-comment-abuse-rules-design.md. Verified: TestLint green, vet clean.

PR #870, opened 2026-09-12
feat(lint): objective comment-abuse rules R1-R6 plus code-verbosity catalog

nnunley:
Base of a 2-PR gh-stack (stacked-by: PR to follow for quality/complexity-scorer).

7 commits: R1 commented-out-code (.lg only), R2 restatement, R3 duplicated-comment, R4 density-outlier, R5 segregation (--edn/--gate), R6 comment-churn (--churn), plus FP fixes and the code-verbosity catalog (scripts/lint-code-rules.edn).

Design: docs/superpowers/specs/2026-09-12-comment-abuse-rules-design.md. Verified: TestLint green, vet clean.


nooga (approved):
This is a re-open of #868 with zero content changes (diffed byte-for-byte, confirmed identical) — #868 was closed only because gh-stack rejects fork-headed PRs (HTTP 422), not for any review objection; it had zero reviews. `go build ./...` clean, targeted `go test ./test/ -run TestLint -v` passes 10/10. Upstream CI (CodeQL, lint, build, race, wasip1/tinygo/no-http builds, gold-differential) is green.

Approving.

mparrett (dismissed):
Blocking findings:

1. `R4` never measures Go function bodies. `form-end-offset` returns as soon as delimiter depth first reaches zero. For every Go declaration, that happens at the closing `)` in `func F(...) {`, before the opening body brace is visited. `file-definitions` therefore sees each Go function as one line and drops it via `r4-min-def-lines`. I reproduced this with 20 six-line Go functions; lint reports “only 0 definition(s) found.” Please make the scanner language-aware (or use Go AST extents) and add a multiline Go-function regression test.

2. `--gate comment-churn` fails even when a change adds zero comments. `r6-findings` emits a finding for every range with at least one added code line; it never tests the computed multiple against a threshold. Repro: baseline comment/code ratio 0.333, then add one code-only function. The tool reports `multiple=0.0x` and exits 1 under `--gate comment-churn`. Please only emit an R6 finding above a defined abuse threshold, or make this measurement non-gateable, and cover the zero-comment case.

The existing targeted lint suite passes, but it does not exercise either failure mode.

mparrett:
Both findings hold, and I am dropping them as merge blockers.

The reason is scope. `scripts/lint.lg` is not wired into CI, `make`, or the hooks, and this PR adds no wiring; `--churn` and `--gate` are opt-in flags nothing passes today. So R4 reporting nothing on Go files and `--gate comment-churn` exiting 1 on a code-only change are both latent. Neither breaks existing code or tests, which is the bar for blocking.

Tracked as #873, with both reproductions. R6 is the one to fix before anything wires `--gate` into CI, since any change that adds code lines without comments trips it.

The targeted suite passes without exercising either failure mode, so a regression test for each is what keeps them fixed.

mparrett (approved):
Re-reviewed the current `684d7c0` head. Both previously raised issues are fixed: R4 now uses Go AST-backed declaration spans with multiline regression coverage, and comment churn is report-only rather than gateable, with zero-comment regression coverage. Targeted vetting and the full CI suite are green. Approving.

PR #871, opened 2026-09-12
feat(quality): static complexity scorer with cost, dup, score, graph

nnunley:
Stacked on #870 (lint stack). ~40 commits: scripts/quality.lg + scripts/quality/*, cmd/go-callables, test/quality_*.lg suites. Includes :qname in coverage :untested so hints print qualified ns/name.

Verified: TestRunner (quality suites) green, go-callables green, TestLint green.


nooga (changes requested):
Content looks sound and it's cleanly stacked on #870 (no file overlap with the lint.lg cluster). But this PR's own CI never ran the standard pipeline — `statusCheckRollup` shows only the perf/bench jobs (all SKIPPED), no CodeQL/lint/build/race/test-location at all. That's no independent CI signal for a ~7,400-line diff, which I'm not comfortable merging on local verification alone.

Locally, most of the new `quality_*.lg` suite passes, but two failures need a second look before I'd trust them as sandbox-only: `quality_cli_test.lg` (expects `jj` on PATH — not installed here) and `quality_inputs_test.lg` (expects a 40-char commit SHA — this worktree was a shallow clone). Plausible both are artifacts of my sandbox, but worth confirming on a full clone with `jj` installed rather than assuming.

Requesting changes: get the standard CI pipeline actually running on this PR (retrigger or check the workflow trigger conditions), and confirm those two test failures don't reproduce on a normal contributor setup.

mparrett (changes requested):
I agree with the existing review that the standard CI pipeline must run. I also found these direct blockers:

1. Qualified calls create false local graph edges. `references` retains both the qualified symbol and its bare name, and `resolve-ref` applies the current-namespace fallback to both. With `a/foo` and `b/foo` defined, `(a/foo)` inside `b/caller` produces `#{a/foo b/foo}`. That falsely marks `b/foo` reachable and corrupts coverage/dead-code results. Qualified (including alias-rewritten) references should not also resolve through an unrelated local bare name; please add a collision regression test.

2. PR-delta cleanup only runs on the success path. If the head revision cannot be extracted, or either `measure` call throws, `release-rev` is skipped and `.workspaces/quality-rev-*` remains. I reproduced this with `--base HEAD --head definitely-not-a-revision`; exit 2 left `.workspaces/quality-rev-HEAD`. Please wrap acquired checkouts in guaranteed cleanup (`try`/`finally`) and test the failure path.

3. Confirming the `quality_cli_test.lg` failure is not merely “jj missing.” The test hardcodes `tracked-by jj` at line 71, while an ordinary Git checkout correctly reports `tracked-by git`; the targeted run is 97 pass / 1 fail on that assertion. This will fail in a standard GitHub checkout. Make the expectation VCS-neutral (or explicitly exercise each backend).

Vetting: `go vet ./...` is clean; `go test ./cmd/go-callables` and the targeted `TestLint` suite pass.

mparrett (changes requested):
Re-reviewed the current `5f7564e` head. The three findings from my previous review are fixed and the standard CI pipeline is green. Four current-head blockers remain:

1. `measure` derives `git-range` from `inputs/head-commit (checkout-root)` even when `:root` points at an extracted revision (`scripts/quality.lg:505`). Consequently both sides of a PR delta use the live checkout HEAD for commit metadata, defect history, and automatic CI lookup. I measured extracted base `684d7c0` at this head and its result reported commit/range end `5f7564e`. Please pass each resolved revision into `measure` and cover distinct base/head input identities.

2. `validated-segments` compares independently parsed reader values for equality (`scripts/quality/forms.lg:86-89`). Regex literals produce distinct pattern objects, so valid source such as `(def r #"x")` is rejected as `:invalid-source`; real quality scripts then degrade to whole-file fallback instead of touched forms. Please validate the segmentation without object equality and add a regex-literal regression.

3. Revision materialization uses the deterministic directory `quality-rev-<sha12>` and deletes any existing directory before extraction (`scripts/quality.lg:903-946`). Two concurrent runs for the same SHA therefore delete and release each other’s checkout. I reproduced run A losing `scripts/lint.lg` immediately after run B released the same SHA. Please allocate a unique directory/workspace per acquisition and release only that handle.

4. Test coverage does not resolve `:refer` imports. `file-aliases` records only `:as`, while `test-roots` flattens bare symbols and `coverage` resolves them without the test namespace (`scripts/quality/graph.lg:70-106,169-203,256-267`). With `a/foo`, `b/foo`, and a test requiring `[a :refer [foo]]`, coverage reports 0 reached because `foo` is considered ambiguous. That creates a false negative despite the stated upper-bound contract. Please retain and apply per-file refer mappings and add a collision regression.

mparrett:
From a multi-agent review — a separate pass from the four findings above, reconciled against them afterwards. Unrelated to those, and not a blocker.

**`make quality` never builds `go-callables` — `Makefile:579`**

`quality: build $(GO-CALLABLES)`, but `GO-CALLABLES` is assigned eight lines below at `Makefile:587`. Make expands prerequisite lists when it reads the rule, so at that point the variable is empty and the prerequisite disappears: `make -p` reports the target as plain `quality: build`, and `make -n quality` plans no `go-callables` build. On a clean tree `make quality` exits 0 with `build/go-callables` absent.

The recipe still passes `QUALITY_GO_CALLABLES=$(GO-CALLABLES)`, and since that value is non-blank the `go run` fallback in `go-measurements` never fires, so `make quality` produces less Go analysis than running `scripts/quality.lg` directly.

What that costs is the callable-level Go metrics: the top-complexity table comes back with no `.go` rows, and erosion counts `.lg` callables alone. The tokenizer-based metrics are unaffected, with Go files still appearing in the corpus count, the per-file debt table, duplication, and verbosity — which is what makes the gap easy to miss. The erosion line prints `go=UNAVAILABLE`; nothing else reports it.

Moving the `GO-CALLABLES =` assignment above the `quality` target fixes it.

mparrett (changes requested):
Re-reviewed current head `6e47ed5` after the rebase onto `main`. `git range-diff` shows all 62 PR commits replayed byte-for-byte, and all four blockers from the prior review still reproduce at this head:

1. `measure` still derives `git-range` from the live checkout HEAD even when measuring an extracted revision (`scripts/quality.lg:505`). Measuring extracted base `684d7c0` at this head reports commit/range end `6e47ed5`, so base/head defect history, commit metadata, and automatic CI lookup are not revision-pinned.

2. `validated-segments` still compares independently parsed reader values (`scripts/quality/forms.lg:86-89`). Valid source containing `(def r #"x")` is rejected as `:invalid-source` because regex pattern objects are not value-equal, degrading real files to whole-file fallback.

3. Revision materialization still uses and deletes the deterministic `quality-rev-<sha12>` directory (`scripts/quality.lg:903-946`). In a concurrent same-SHA repro, run B releases that directory and run A immediately loses its extracted files.

4. Coverage still does not retain or apply `:refer` mappings (`scripts/quality/graph.lg:70-106,169-203,256-267`). With `a/foo`, `b/foo`, and `[a :refer [foo]]`, the test call is treated as ambiguous and coverage reports zero reached, violating the stated upper-bound behavior.

The rebase itself is clean, all required GitHub checks are green, and local `make build`, `go vet ./...`, `go test ./cmd/go-callables`, and the complete `quality_*.lg` suite pass. Those signals do not exercise these four failure modes, so this remains a changes request.

nnunley:
All four blockers from the 09-14 review are fixed, plus the `make quality` finding.

1. **Each side of a delta is measured against its own commit.** `measure` takes an optional
   `:commit`; `git-range` is built from it, so commit metadata, defect history, and the automatic
   CI lookup (which reads `(:to grange)`) follow the measured revision. `delta` passes `base-sha`
   and `head-sha` to their measurements; the working-tree side keeps the checkout's head. A test
   pins `measure` to an older commit and requires `[:inputs :commit]` to report it.
2. **Regex literals validate.** `validated-segments` compares the printed forms of the whole-file
   and per-segment reads instead of the values, since two reads of `#"x"` are distinct pattern
   objects. Regression: a file with `(def r #"x")` and a `re-find` literal segments cleanly.
3. **Concurrent checkouts of one revision are independent.** Each `extract-rev` gets a unique
   directory and jj workspace name, `quality-rev-<sha12>-<8 hex>`, never deletes an existing
   directory, and `release-rev` removes only its own handle. Regression: two checkouts of the same
   revision get different directories, and releasing one leaves the other intact. The two
   release-on-error tests now require that no `quality-rev-<sha12>-*` entry remains.
4. **`:refer` imports resolve.** `file-aliases` records `{:as {alias ns} :refer {name ns}}`, and a
   bare referred name is rewritten to `ns/name` wherever `:as` aliases already apply (edges, test
   roots, top-level refs, entry roots). Regression: with `a/foo` and `b/foo`, a test requiring
   `[a :refer [foo]]` reaches `a/foo` only.

**`make quality`** now builds `go-callables`: `GO-CALLABLES` is assigned above the target, so
`make -p` shows `quality: build build/go-callables`.

Verified: `TestRunner/quality_*` with `LG_QUALITY_REVISION_ROOT` set; every new assertion passes
and no checkout directory or jj workspace is left behind. Run from a jj secondary workspace, the
PR-mode CLI test and one delta test fail identically with and without this change (`git rev-parse
HEAD` names the primary checkout there); CI's git checkout does not hit that.

mparrett (changes requested):
Re-reviewed current head `8fb651d`. The four prior blockers and the `make quality` dependency issue are fixed; their reproductions now pass. Two current-head blockers remain:

1. The new commit-pinning regression assumes `HEAD~1` exists (`test/quality_delta_test.lg:239`). GitHub uses a shallow checkout, so the required `build` job fails with `cannot resolve revision HEAD~1`. I reproduced the same failure independently in a depth-one clone. Please make the test create or otherwise use a revision guaranteed to exist in a shallow checkout.

2. Require resolution is still incomplete (`scripts/quality/graph.lg:92-100`). Explicit `:refer [foo]` now works, but `:refer :all` is deliberately ignored and `:rename` is not recorded. Both forms are used in this repository and still create direct-call coverage false negatives. Reproductions: with `a/foo` and `b/foo`, a test using `[a :refer :all]` reports 0/2 reached; with only `a/foo`, a test using `[a :rename {foo renamed}]` and calling `renamed` reports 0/1 reached. Please retain and apply both import forms and add collision/rename regressions.

Local `make build`, `go vet ./...`, `go test ./cmd/go-callables`, and the complete `quality_*.lg` suite pass in a full-history checkout. All other required GitHub checks pass; the shallow-checkout failure above is the remaining red check.

nooga (changes requested):
This has converged a lot over several rounds — qualified-call false graph edges, PR-delta cleanup on failure paths, revision-materialization race, `:refer` resolution, `measure` using live HEAD instead of the pinned revision, regex-literal equality, and the concurrent-checkout directory collision are all confirmed fixed as of the latest push (8fb651d).

Two blockers remain, one of which is a currently-failing required check:
1. `build` CI is red right now — a new regression test assumes `HEAD~1` exists, which fails under GitHub's shallow checkout (reproduced independently in a depth-one clone). Needs a same-repo-guaranteed revision (e.g. create a throwaway commit) instead of `HEAD~1`.
2. `:refer :all` / `:rename` import forms are still unhandled in the reference graph, causing false-negative dead-code/coverage reports.

Requesting changes for both — the CI fix is the more urgent one since it's currently blocking merge regardless of review sentiment.

nnunley:
New head `5978a49a`, one commit on current main — the 63-commit history is squashed at my request; the tree is what the fixes below produced, checked identical before the squash.

1. **`HEAD~1` in the commit-pinning test.** The test now creates its own throwaway git repository, commits, and pins `measure` to that commit id (asserting `:commit-source "revision"` and the `:git-range :to`), so it no longer assumes any history in the checkout. Reproduced red in a single-commit checkout of the PR tree (`cannot resolve revision HEAD~1`); green there now, whole `.lg` suite included.
2. **`:refer :all` / `:rename`.** `file-aliases` records both; `:rename` maps the local name to the qualified def, and `:refer :all` resolves a bare name against the corpus — one matching namespace rewrites it, zero or two or more leave it alone, so the existing ambiguity rule still yields no edge. Explicit `:refer`/`:rename` and the file's own namespace win. Applied to edges, test roots, top-level refs and entry roots. Your two reproductions (`[a :refer :all]` with `a/foo` and `b/foo`; `[a :rename {foo renamed}]`) plus collision, precedence, edge and entry cases are pinned; red first, green now.
3. **Revision handling no longer materializes checkouts.** `extract-rev`/`release-rev`, the unique-directory naming, the release-on-error handling and the jj path are gone. `measure` takes a corpus seam; in delta mode the base is the checkout's corpus with the changes since base undone in memory from git blobs (`git diff --name-status`, `git show base:path`), so both sides are still scored in full — every corpus-wide term included — and a test requires the reconstructed base to score identically to that commit's tree on disk. Changed `.go` files get one temp file each for the path-taking analyses. Where git would answer for a different tree (a non-colocated jj secondary workspace, whose git worktree is the primary checkout), the tool refuses by name rather than produce a plausible all-zero delta, and those tests skip with that reason.

Verified: `make check-generated`, `make build`, `go vet ./cmd/go-callables`, `go test ./cmd/go-callables`; the whole suite green in a plain git checkout with nothing skipped, and in a jj workspace with the revision tests skipping; `make -n quality` still plans `build/go-callables`. Pushed with `SKIP=bench-ratchet` (local timing gate red independent of content — #791, #899). Drafted with Claude and checked by me.

mparrett (approved):
Re-reviewed current head `5978a49a`. Both blockers from my 09-16 review are fixed, and I checked each against the head rather than against the description. No blocking findings; approving.

**1. `HEAD~1` under a shallow checkout.** Both remaining `HEAD~1` uses now sit inside repositories the tests create themselves. I fetched the PR head into a depth-one clone, where there is one reachable commit and `git rev-parse HEAD~1` exits 128, and the whole `quality_*` suite is green there. That is the condition that was failing `build`, and `build` is green on the PR now too.

**2. `:refer :all` and `:rename`.** `file-aliases` records both, and `references-aliased` applies them to edges, test roots, top-level refs, and entry roots. My two reproductions are pinned with the right expectations: `[a :refer :all]` against `a/foo` and `b/foo` reaches `a/foo` only, and `[a :rename {foo renamed}]` reaches `a/foo`. Two-namespace ambiguity, explicit-over-wholesale precedence, and self-namespace shadowing are covered alongside them.

## The revision rework

Replacing `extract-rev`/`release-rev` with in-memory base reconstruction is a larger change than the two fixes and it arrived unreviewed, so I read it rather than resting on a green suite. It holds up.

`scoring-the-reconstruction-equals-scoring-the-base-tree-on-disk` makes the central claim falsifiable end to end. Its one gap: the fixture's rename carries no content change, so the `:renamed` branch that reads `:old-text` is never taken. I covered the two missing shapes against the head. A low-similarity rename plus edit, which Git reports as add/delete, and a high-similarity one, which it reports as `:renamed` with `:old-text` present, both restore the base content exactly. I also checked `corpus-scope` against `discover` for scope parity under a non-empty `repo-prefix`, and they agree.

Two non-blocking notes:

- `usage` still advertises `[--head REV]` (`scripts/quality.lg:926`), but `delta` now throws unless that revision resolves to the working copy's own HEAD (`:1116`). The error message is the only place the restriction is written down; please add it to the usage string. Nothing in CI or the Makefile passes `--head`, so no automation is affected.
- `base-corpus`'s `materialize` runs a separate `mkdir -p` subprocess per reconstructed file. Only noticeable on a wide base range.

Dropping jj from revision handling is a real narrowing, and skipping the delta tests with a named reason in a non-primary jj workspace beats failing on the shape of the checkout. It does mean the delta path runs in CI and not in your own workspace, so CI is the only thing exercising it.

Squashing the 63 commits leaves `git range-diff` unavailable to confirm the tree is unchanged apart from the fixes, which is how the 09-15 rebase was checked. I compared content instead.

## What I ran

At `5978a49a`, in a detached worktree at that head: `make build`, `go vet ./cmd/go-callables ./test/...` (clean), and the full `go test ./test/ -run TestRunner` (green). Separately, the `quality_*` suite in a depth-one clone of the same head, and two reconstruction probes for the rename shapes above. I did not run `make quality` against the real corpus, so I have not measured whether `:refer :all` resolution shifts the reported dead-code numbers on let-go itself.

Drafted with Claude, checked by me.

nnunley:
@nooga — head is now `d5bea6af` (`5978a49a` that mparrett approved, plus the `--head` restriction added to the usage string as asked). Against your two reviews:

- **Standard CI on this PR.** It runs now and is green on this head: build, CodeQL, lint, race, test-location, generated-artifacts, gold-differential, gogen-diff, docs-frontmatter, wasip1 and tinygo builds. The bench/comment jobs remain skipped by design.
- **`jj` on PATH / shallow clone.** Both dependencies are gone rather than confirmed-as-sandbox: revision handling no longer calls `jj` at all (base corpus is rebuilt in memory from `git diff` + `git show`; the delta tests skip with a named reason in a non-primary jj workspace), and the `HEAD~1` test now creates its own commit. mparrett verified the `quality_*` suite green in a depth-one clone of the head.
- **`:refer :all` / `:rename`.** Resolved in `file-aliases` / `references-aliased` with tests pinning `:refer :all` reaching `a/foo` only, `:rename`, precedence, and self-namespace shadowing.

Your changes-requested review still stands from `8fb651d6`; a re-review (or dismissal) is what's left before this can merge.

nooga (approved):

mparrett:
Two places point at a design doc that is not in the repo:

- `Makefile:571` — `# Code quality report (docs/superpowers/specs/2026-09-11-code-quality-score-design.md).`
- `scripts/quality.lg:48` — `Design: docs/superpowers/specs/2026-09-11-code-quality-score-design.md`

The path has never existed in the repo's history (`git log --all -- '**/2026-09-11-code-quality-score-design.md'` is empty), and it is not in `let-go-agentic-stack`, `lgx`, `strange-lettractor` or the wiki either. #871 shipped the other four quality specs and plans, so this one looks like it stayed local.

The score is a composite of eight weighted terms, and the weights carry judgement — `verbosity` as a declared lower bound, `dead` as an upper bound, `cost` reported at weight 0 until its loop rule traces iteration counts. The docstrings explain each term on its own; what they cannot give is the argument for the weighting as a whole, which is what anyone proposing a change to it needs.

Could you push the design doc, or should the two references point somewhere else?

PR #874, opened 2026-09-13
docs(design): propose the glplat host-capability seam and backend contract

mparrett:
@nnunley @nooga — this is the contract the 2026-09-07 direction on #392 asked for, plus a proposal for the seam re-scope, written so both can be answered yes or no.

It is a docs-only PR. Neither #392 nor #744 changes here; they are blocked on the decision this doc is trying to close.

## The seam

The direction was to re-scope onto the #572 `surface` host-capability model rather than glplat's registry. Taken literally that means routing backends through `Surface.Present`, which carries one RGBA frame per call — the guest would rasterize in Clojure and the host would blit, which deletes the triangle batching both backends exist to provide. §2.5 says so plainly, because if that reading was the intended one this is a different piece of work.

What the doc proposes instead is that the valuable part of #572 is the binding discipline, not the interface: a capability interface in `pkg/rt`, resolved through a dynamic var so `binding` and per-`Run` `pkg/api` bindings work, with a no-op zero value. glplat's registry is a package-level `var` with a setter, so it fails the last two. The proposal is `rt.Display` bound at `*display*`, installed by `api.WithDisplay`, with both backends as implementations and `internal/registry` retired. `runtime-io-host-decoupling.md` already names graphics as a peer capability in exactly this shape, scoped in #255.

## The contract

Fifteen numbered rules over orientation and pixel format, depth and occlusion, resize, input events and modifiers, and lifecycle — the five areas the direction listed.

They are written from the two backends that exist rather than from first principles, which is why some are corrections rather than documentation. Three that need a decision rather than a nod:

- **C3, straight alpha.** The GL path is non-premultiplied by way of `glTexImage2D`; the Ebitengine backend premultiplies on upload because `WritePixels` requires it. That conversion lives in a comment in one backend today, which makes it the likeliest thing for a third to skip silently.
- **C10, modifiers in the event.** Neither backend carries them — the GL backend's `keyCallback` receives GLFW's `mods` and drops it. Fixing it is a format change affecting both backends and any consumer parsing the strings, so it needs a decision rather than a patch.
- **C4, painter's order.** The GL backend has a depth buffer and Ebitengine does not, which is why the 3D mode mis-sorts on one and not the other. The rule picks the weaker guarantee as the portable one.

C6 and C13 restate the `WindowSize` staleness and post-`Terminate` live-state findings from the #392 list as contract rather than as bugs. The per-binding lifetime in §2 is what makes C13 enforceable instead of conventional.

## The font tag

The second of the three conditions is already on #744: `glplat_fonts`, independent of both backend tags, with `check-default-deps` green.

## The ask

Two decisions, in §5. Adopt the seam or reject it, and adopt the rules or amend them by number. If the seam carries, the doc proposes doing the contract-visible behavior changes first, then the seam, then rebasing the two backend PRs onto it — the remaining safety findings on #392 are independent of either decision and need not wait.


nnunley (approved):

mparrett:
@nooga — @nnunley approved, so this merges under the one-approval rule. Noting it here because I had requested you, and you reviewed both PRs this doc came out of.

Merging lands the proposal in `docs/design/`; it does not settle it. The two asks in §5 are still open: adopt or reject the `rt.Display` / `*display*` seam in place of glplat's registry, and adopt C1–C15 or amend by number. #392 and #744 stay blocked on the first.

If you disagree with any of it, including C3, C4 and C10, the three rules that change behavior rather than describe it, say so and I will revise in a follow-up PR.

mparrett:
The two §5 questions are now being answered on #744, since that is the PR a release cut would merge: https://github.com/nooga/let-go/pull/744#issuecomment-5737376318

PR #876, opened 2026-09-14
docs(ci): report where Actions runner time goes

mparrett:
Adds `scripts/ci-usage.sh`, which reports where this repo's Actions runner time goes, plus `docs/ci-actions-usage.md` recording what it found for the week of 2026-09-07 to 2026-09-14.

The question behind it was whether we run more jobs than we need. Answering it took measurement. Two of the findings want a team decision; one looks like a bug to fix.

### Why a script

There is no REST equivalent of the `/actions/metrics/usage` page for this repo. The Actions usage-metrics API is org- and enterprise-scoped, and `nooga` is a user account, so `repos/nooga/let-go/actions/metrics/usage` 404s. The runs + jobs API is what remains, and it is the better source: it exposes each job's queue wait (`created_at` to `started_at`), which the metrics page does not, and that field distinguishes a shared-capacity problem from a self-inflicted one.

### What it found

The week came to 9,957 runner-minutes over 689 runs. Perf Timeline is 61% of that and `go.yml` is 36%; everything else rounds to noise. Run counts are a misleading proxy: 41% of runs are the label-gated perf workflows, which skipped every job and cost nothing.

**The ubuntu timeline leg is dying at its timeout.** 7 of 28 legs ran exactly 90.0 minutes and were killed at `timeout-minutes: 90`. Successful legs have a p50 of 84 minutes and the slowest finished at 89. A quarter of amd64 snapshots are being dropped at full price, and the margin on the ones that survive is a minute. This is the shape #581 item 6 described and #573/#583 fixed for macOS; the ubuntu leg has since grown into its own ceiling. Raising it is the immediate fix. That the leg's runtime drifts upward with nothing watching it is the more durable problem.

**The queue backlog is the concurrency group, not runner scarcity.** #581 item 6 reasonably read the long macOS waits as a scarce runner pool. The job timings say otherwise: on all 19 legs that were already queued when the previous leg finished, the next started 6 to 18 seconds later, with no idle gaps. `concurrency: {group: perf-timeline-<os>, queue: max}` is one server per OS by construction, and a 133-minute service time means a busy day on `main` outruns it. On 09-11, 13 commits landed between 17:35 and 23:12 and the queue did not drain until 09-12T22:49; the last leg waited 19 hours against GitHub's 24-hour cancellation.

**Per-push cadence is the budget.** Two legs at ~3.6 runner-hours per push to `main`, so weekly cost tracks commit volume: 21 hours in the quiet week of 08-24, 105 in the week just measured. The workflow header already treats gaps as acceptable, since backfill can densify any range, which is what puts the cadence in play at all. #581 item 6's step 2 proposed a scheduled or coalescing trigger and is still open.

### Cadence, and what a PR push costs

Two further sections answer the question the cost numbers leave open.

**Per-commit sampling does not buy resolution on arm64.** Tracking how the median per-benchmark delta grows with separation between snapshots: on amd64 it roughly doubles from lag 1 to lag 20 (1.11% to 2.25%), which is real signal accumulating. On arm64 it goes 11.01% to 12.03%, which is flat. Comparing commits 20 apart tells you almost nothing more than adjacent ones, because the leg's resolution is set by run variance. That is the cadence consequence of the 27.2% you measured on #651, and it lands on the most expensive leg in the repo.

The surrounding ground was already filed and section 4 cites it rather than restating it: #651 for per-runner noise and the multi-key finding, #445 for the single-shot A/B floor, #705 for the uncalibrated 5% budget, #597 for cross-runner comparability. The new part is only the delta-versus-separation test, which asks what that noise implies for *how often* to sample.

**On the PR side**, `go.yml` cost 9,222 runner-minutes over 465 runs (2026-08-15 to 2026-09-14), and one test step is 31% of it: "Expensive lowering e2e" is 74% of the `build` job, about 47 runner-hours a month, and it grew 32% when `TestCustomMain` joined it. Fire rates are in the doc. Two gates have clean records at real cost (`gogen-diff` 25 h/month, `tinygo-wasi-build` 17), but neither is redundant, so the note there is path-gating rather than removal. One negative result worth having: `go.yml` has no path filter, but docs-only PRs were 21 of 465 runs, so adding one buys 4.6% and not more.

### Scope

No workflow changes here. The ubuntu timeout probably wants a one-liner, but the cadence question trades against timeline density and against what the legs are used for, so the numbers belong in the repo before the decision. Extending this to which runs are needed at what frequency is the next pass.

Re-run with `scripts/ci-usage.sh [owner/repo] [since-YYYY-MM-DD]`; it needs `gh` and `jq` and makes roughly one API call per run.


mparrett:
A sample run, so the tables above the fold can be checked against what the tool actually prints. Three days rather than the doc's week, to keep it short:

```
$ scripts/ci-usage.sh nooga/let-go 2026-09-12

=== nooga/let-go since 2026-09-12: 225 runs, 982 jobs ===

=== runner-minutes by workflow (skipped jobs excluded) ===
1321	min	12	jobs	Perf Timeline
1153	min	677	jobs	Go
81	min	49	jobs	CodeQL
19	min	44	jobs	Deploy wasm let-go to Pages
0	min	1	jobs	Perf Backfill Releases

=== runner-minutes by runner OS ===
1763	min	777	jobs	ubuntu
811	min	6	jobs	macos

=== top jobs by total minutes (matrix legs folded) ===
1321	min	n=12	avg=110.1m	Perf Timeline :: profile
484	min	n=53	avg=9.1m	Go :: build
185	min	n=53	avg=3.5m	Go :: gogen-diff
167	min	n=53	avg=3.2m	Go :: generated-artifacts
123	min	n=53	avg=2.3m	Go :: tinygo-wasi-build
81	min	n=49	avg=1.7m	CodeQL :: Analyze
47	min	n=53	avg=0.9m	Go :: lint
43	min	n=53	avg=0.8m	Go :: race
27	min	n=53	avg=0.5m	Go :: no-http-build
26	min	n=53	avg=0.5m	Go :: wasip1-build
19	min	n=53	avg=0.4m	Go :: gold-differential
16	min	n=22	avg=0.7m	Deploy wasm let-go to Pages :: build
15	min	n=53	avg=0.3m	Go :: default-deps
6	min	n=47	avg=0.1m	Go :: docs-status
5	min	n=47	avg=0.1m	Go :: docs-frontmatter

=== wasted: minutes burned by jobs that ended cancelled or failed ===
181	min	n=2	Perf Timeline / ubuntu
22	min	n=12	Go / ubuntu
2	min	n=1	CodeQL / ubuntu

=== queue wait: created -> started ===
=== waits across every workflow => account concurrency cap
=== waits confined to one workflow => that workflow's own concurrency: group
Perf Timeline / macos	n=6	median=0m	max=1147m	>10m:2
Perf Timeline / ubuntu	n=6	median=0m	max=575m	>10m:2
Go / ubuntu	n=677	median=0m	max=1m	>10m:0
CodeQL / ubuntu	n=49	median=0m	max=0m	>10m:0
Deploy wasm let-go to Pages / ubuntu	n=44	median=0m	max=0m	>10m:0
Perf Backfill Releases / ubuntu	n=1	median=0m	max=0m	>10m:0
```

Two things visible here that the doc's week-long window states but does not show. The macos queue-wait max is 1147 minutes against a median of 0 — the backlog is bursty, not steady, which is what a saturated FIFO looks like from outside. And `Perf Timeline / ubuntu` shows 181 minutes burned across 2 cancelled legs in three days, which is the 90-minute timeout kill that #878 raises.

nooga (approved):
`scripts/ci-usage.sh` reads cleanly (`set -euo pipefail`, handles the 1000-run `gh api` pagination cap), and the sample run posted on the PR matches the tool's own claims. Pure measurement/reporting, no workflow behavior changes.

Approving — independent of the rest of the perf-page cluster, can merge any time.

mparrett:
Merged on nooga's approval. @nnunley — you were still listed as a requested reviewer here and hadn't had a chance to look; happy to follow up in a PR if anything wants revisiting. No workflow changes landed, so this is measurement only: the ubuntu timeline timeout and the cadence question are both still open decisions rather than anything this commit acts on.

PR #877, opened 2026-09-14
ci(tinygo): fetch the pinned wasmtime asset instead of curl|bash

mparrett:
`tinygo-wasi-build` went red on an unrelated docs PR (#876) with `wasmtime: command not found` at the "Verify wasmtime is on PATH" guard. The install step above it had exited 0 having installed nothing.

### What happened

A transient 504 from the release CDN, in run 34859297430:

```
Fetching archive for Linux, version v47.0.3
https://github.com/bytecodealliance/wasmtime/releases/download/v47.0.3/wasmtime-v47.0.3-x86_64-linux.tar.xz
curl: (22) The requested URL returned error: 504
Error: Could not download Wasmtime version 'v47.0.3'
```

The step still passed. #682 pinned the version here for a symptom that looked like this one, but the exit-0 half was never the version resolution — it is in `install.sh` itself. `install_version()` ends with:

```sh
  if [ "$?" == 0 ]
  then
      update_profile "$install_dir" && info "Finished" '...'
  fi
```

That is the last command in the function, and the function call is the last command in the script. A false `if` with no `else` returns 0, so the script exits clean after any failed download. Worth noting because the obvious fix is wrong: `pipefail` does nothing here, since the pipeline's last element is that `bash` and it is what returns 0.

So #682's guard is the only reason this surfaced at all, and it did its job — 57 seconds, clear message, rather than a confusing failure two steps later. What it cost was a red PR for a transient network error.

### The change

Fetch the pinned asset directly, the way TinyGo is already installed in the step above, and keep the guard. Two consequences: `curl -f`'s exit status becomes the step's, so a failed download fails the step rather than deferring to the guard; and `--retry 3 --retry-all-errors` rides out the transient that caused this one.

Verified locally: the download-extract-install body produces a mode-755 ELF at `$HOME/.wasmtime/bin/wasmtime`, and substituting a 404 asset URL exits 56 instead of 0.

### Not in this PR

The TinyGo `wget` above has the same CDN exposure. It differs in that `wget` does exit non-zero on a failed download, so `bash -e` already catches it — it is flaky, not silent, and a retry flag would be the whole fix. Happy to fold that in if you would rather have both in one go.

Upstream, `install.sh` returning 0 after a failed install looks worth reporting to bytecodealliance/wasmtime. I have not filed anything there.


nnunley (approved):
The comment is very verbose considering the length of the change.

mparrett:
> The comment is very verbose considering the length of the change.

Agreed, fixed in f3648f2. I'll assume the previous approval stands since it was only a comment amendment. But let me know if you have other concerns and I'll take them as a follow-up. Adding automated checks for this common failure mode in my stack.

PR #878, opened 2026-09-14
ci(perf): lower the ubuntu perf capture cap to 60 minutes, provisionally

mparrett:
This PR set out to raise the ubuntu cap to 120 because the count-3 leg had grown to a p50 of 84 minutes and was dying at 90 11 times a fortnight. Then #882 landed and removed the reason: most of the leg's wall time was `runtime.ReadMemStats` calls from `b.StopTimer()` inside the vector benchmark loops. The first upstream leg after the merge:

| Leg | Last pre-#882 run | First post-#882 run |
|---|---:|---:|
| ubuntu-latest | 84 min | 26 min |
| macos-14 | 133 min | 28 min |

([last pre-#882 run](https://github.com/nooga/let-go/actions/runs/34912167945), [first post-#882 run](https://github.com/nooga/let-go/actions/runs/34928096218).)

### Why lower it instead of closing

A cap that is far above the leg costs nothing until something goes wrong, and then it costs the whole gap. The per-OS queue here is FIFO with `cancel-in-progress: false`, so a benchmark change that puts the leg back in its old shape would hold every following push for the full cap before the first one is killed. At 60 that is caught within the hour and the timeline shows a red leg next to the commit that did it. At 90 or 120 the same regression looks, for two hours, like a slow queue.

### Why 60 and why provisional

It is sized off one leg. 60 is a bit over twice the 26 measured, which is more headroom than the old 90 ever had over its p50, but one sample is not an envelope. Both files say so in the comment next to the number and name this PR; the plan is to revisit once a few weeks of legs are in, and to tighten or loosen from data rather than from the shape of this one run.

`perf-release-baseline.yml` runs the same capture on the same runner class and moves to 60 with it, with the cross-reference kept, since one envelope described in two files is how the two drifted apart before.

macOS stays at 150 from #583. Its first post-#882 leg is 28 minutes, so the same trim applies there once a few more are in, as a one-line follow-up.

### Comments retired

Two comments in the capture step were written against the old leg and now mislead:

- The count-4 note framed the cap as the constraint ("count 4 nearly exhausts the job timeout"). A count-4 leg would now be around 35 minutes. The count question is open again, and the comment now says it needs its own measurement rather than an inference from the cap.
- The reduceSamples note said count 3 lands in a "discard-warmup-then-average-of-2" branch. It does not: at count 3 `reduceSamples` keeps all three reps and takes the median; the warmup discard starts at 4. The comment now says that.

`-count` itself stays at 3.


mparrett:
Reviewed: no functional findings. I verified that both Ubuntu count-3 full-capture workflows now share the 120-minute cap, macOS remains at 150, and the other 90-minute perf workflows run different profiles. `git diff --check` passed, and `actionlint` found no new issues in the changed workflows (the repository still has its pre-existing `queue`/shellcheck diagnostics). CI was green at review time.

I also pushed `2069d0f8` to trim the workflow comments. The cleanup keeps the durable rationale, cross-file synchronization requirement, and issue references while removing timing history already preserved in this PR. It changes comments only; workflow behavior is unchanged.

nnunley (approved):

mparrett:
@nnunley Thanks for the review. The cap needs revision after the last change that shipped — updating shortly.

mparrett:
@nnunley rebased and reworked after #882: the cap now goes down to 60 (provisional, comment says so and why) instead of up to 120, and the two stale capture-step comments are rewritten. Body has the before/after leg numbers. One more look please, since the number you approved is gone.

nnunley (approved):

PR #879, opened 2026-09-14
feat(perf-page): page-wide CPU filter for the timeline views

mparrett:
Adds one CPU selector in the page header that both timeline-driven views share, so the whole timeline half of the page can be cut to a single machine tier.

### Preview

- this branch: https://mparrett.github.io/let-go/preview/cpu-filter/

Four smaller changes were split out of this branch and are open separately as #886, #887, #888 and #889. Each has its own preview, and a shared guide says what to look at in each: https://mparrett.github.io/let-go/preview/guide/

### Why

The page already asks the reader to do this. The Explore blurb says the anchor-relative ratio "only normalizes within a CPU, so compare trends per CPU rather than absolute levels across them" — and then offers no way to do it except a text box on the sparkline table, which matches on "benchmark · CPU" and so filters that one table only. The Explore chart overplots every tier with no way to isolate one.

The `-cpu` build flag already cuts a tier, but globally and only by regenerating the page, so it cannot answer "is this regression on every runner, or only on 9V74?" while reading. That question is the point: with seven tiers in the timeline, a mixed series zig-zags on runner assignment rather than on code.

### What it does

One `<select>` in the header, next to the baseline chips, listing the tiers present in the data. It drives **Explore metrics over time** and **Trend sparklines**, which are the two views rendered client-side from the timeline payload. Selection goes into `?cpu=`, so a per-tier view is a link rather than a description.

Details worth knowing:

- The two views fetch the same payload independently and do not know about each other, so the control mounts on first registration and later registrations only contribute CPUs.
- It hides itself when the timeline has fewer than two tiers, so a single-tier build is not given a one-option filter.
- A `?cpu=` naming a tier absent from the build falls back to All rather than rendering an empty page.

### What it does not cover

The server-rendered sections, which are two different cases rather than one:

- **Timeline charts** iterate the whole timeline, so they *pool every tier* — this is the zig-zag #597 describes. They are baked SVG, so the in-page filter cannot reach them, but `-cpu` does cut them at build time.
- **Largest changes, Recently tightened, Current baseline** come from `loadBaseline` → `selectProfile`, which takes the first profile by sorted key. `-cpu` is applied to `timeline` only and never touches these, so nothing cuts them today — the header chip naming the CPU is the only indication of which machine you are reading.

The second is arguably the more interesting gap, and it was not in #597's original scope; I have since noted it there.

### Testing

Generated against the live 474-snapshot timeline from `perf-data` (seven tiers) and driven in Chromium: the selector lists all seven, the sparkline table goes 766 rows across seven CPUs to 107 rows on one, the Explore chart drops the other tiers, `?cpu=` round-trips through a reload, an unknown `?cpu=` falls back to 766 rows, and the console is clean. `go test ./cmd/perf-page/` and `go vet` pass.

---

Built collaboratively with Claude.


cmd/perf-page/main.go
```diff
@@ -2251,6 +2261,85 @@ const pageTemplate = `<!doctype html>
...
+            const u = new URL(location.href);
+            if (current) { u.searchParams.set("cpu", current); } else { u.searchParams.delete("cpu"); }
+            history.replaceState(null, "", u);
+          }
+
+          function mount() {
+            const host = document.getElementById("perf-cpu-filter");
+            if (!host) return;
+            // One tier means nothing to choose between.
+            if (cpus.length < 2) { host.hidden = true; return; }
```
mparrett, line 2309: Agentic self-review: The single-tier path returns here before validating `current` against `cpus`. If a page generated with `-cpu 9V74` is opened with a stale query such as `?cpu=Apple+M1`, `PERF.cpu.apply` filters out every row in both views while this selector remains hidden, so there is no UI recovery. I reproduced that behavior in Chromium. Please move the invalid-selection check and `sync()` before this early return.

mparrett: Confirmed and fixed in 568a0045 — the check and `sync()` now run before the single-tier return, and the duplicate further down is gone.

Reproduced your case first to be sure I was fixing the right thing: a `-cpu 9V74` build opened at `?cpu=Apple+M1` gave 0 sparkline rows, empty states in both views, and the stale param still in the URL. After the move: 107 rows, no empty state, and the URL cleaned back to no `?cpu=`. The seven-tier page is unchanged — 766 rows across all tiers, 107 when filtered, `?cpu=` still round-tripping through a reload.

Worth noting why I missed it: my original check used `?cpu=NoSuchChip` on the seven-tier page, which exercises the guard on the path where it already worked. The single-tier build is the only shape where the early return shadows it, and I never generated one.

cmd/perf-page/main.go
```diff
@@ -2044,7 +2055,7 @@ const pageTemplate = `<!doctype html>
     <section>
       <div class="section-head">
         <h2>Timeline</h2>
-        <p>{{len .Timeline}} snapshot(s). CI snapshots graph real runs; seed points use committed historical/current JSON until the timeline fills in.</p>
+        <p>{{len .Timeline}} snapshot(s). CI snapshots graph real runs; seed points use committed historical/current JSON until the timeline fills in. These charts are drawn server-side from the whole timeline, so they pool every CPU tier and the CPU chips below cannot reach them; <code>-cpu</code> cuts them at build time. The baseline sections further down are a different case again: they render one machine profile picked at build time, which neither the filter nor <code>-cpu</code> changes.</p>
```
nnunley, line 2058: The charts are not being rendered server-side - they're being pulled in as json and rendered in browser, aren't they? Or is ther ea new rendering rule somewhere that I missed in reviews?

mparrett: @nnunley yeah so this page is hybrid and has been for some time. This PR is a minimal extension of the clientside parts that are already there (keeping the build time svg keyed on -cpu). But I don't know if we want it to be hybrid like this or if we got to the hybrid state because of a stack of quick changes in response to specific issues. I recall locking to -cpu, but I don't recall giving up the filtering. That said, I was doing a lot of similar cross-project tinkering with paserati at the time.

The PR stacked on this one brings brought client-side rendering back so interactive filtering could happen without messing with this page

In any case,I'll take a closer look when I get back to the house this morning. My goal here was to make it a little easier to understand the data and offer an opt-in client rendered exploration page. Mostly diagnostic and for the core folks working on bench/perf.

Let me know if you have strong opinions! Happy to consider any feedback.

nooga (approved):
Verified: the CPU-tier filter seeds from `?cpu=`, falls back to "all" for an unknown tier, and only mounts when ≥2 tiers exist. The self-review catch (stale `?cpu=` on a single-tier build zeroing both views with no recovery UI) is fixed in 568a004 and present at head. nnunley's SSR-vs-client question is answered inline, non-blocking.

Approving. This is the base of the stack — recommend merging this before #880 (which branches off it).

cmd/perf-page/main.go
mparrett: Promised a closer look, so here that is:

So, the page has three rendering modes, not two, and here's how they collide.

- **Timeline charts**are server-side SVG, emitted at **build time** from the whole timeline. They pool every CPU tier.
- **The explorer** is what you're describing: it fetches `explorer.json` and **draws in the browser** with d3 + Observable Plot.
- **The baseline sections** are a third case — one machine profile picked at **build time,** which neither the chips nor `-cpu` touch.

Re: new rendering rule- you didn't miss one in review. The server-side timeline SVG has been there since #173 and the client-side explorer since #362; this PR adds 122 lines and none of them are in the chart rendering.

What it does add is a filter over the client-rendered parts, plus document the boundary- the chips can't reach the server-drawn charts, so only `-cpu` cuts those, at build time.

On why the split exists: `explorer.json` is tens of thousands of rows kept out of the HTML deliberately, so the page stays small. The timeline charts are small enough to inline as SVG, so they were never moved. That's a size tradeoff, not a rendering policy — which is roughly what I meant by "hybrid" and why I couldn't tell you offhand whether it was decided or just accreted.
Having now read back through it: the explorer half was a decision, the timeline half is just where it started.

Whether it should stay split is the question I raised, and why #880 exists separately. It redraws the timeline charts in the browser so a tier filter can actually reach them. If we want to discuss consolidating these two in the future, I'm open to that.

PR #880, opened 2026-09-14
feat(perf-page): add a Timeline explorer page

mparrett:
Stacked on #879. Adds a second page at `/perf/explore/` that redraws the timeline charts in the browser, so a CPU-tier filter can reach them.

### Preview

- this branch: https://mparrett.github.io/let-go/preview/timeline-explorer/
- the explorer page itself: https://mparrett.github.io/let-go/preview/timeline-explorer/explore/

Based on #879, so the preview carries both changes. The four smaller changes split out of #879 are open as #886 through #889, with a shared guide: https://mparrett.github.io/let-go/preview/guide/

image

image

### Why

The summary page's timeline charts are server-rendered SVG built from the whole timeline, so they pool every tier — and `ratio_to_anchor` only normalizes within a CPU model. Pooled, the runner you drew moves the chart more than the code does. Measured on a geomean over a fixed basket of 99 benchmarks, the two dominant tiers sit **17.6% apart** (EPYC 7763 median 413, EPYC 9V74 351), which is the bimodal band #597 describes, on the aggregate rather than one benchmark.

`-cpu` already cuts a tier, but globally and only by regenerating the page, so it cannot answer "is this on every runner, or only on 9V74?" while reading.

### What it carries

Six charts: the four the summary page plots, plus two it does not — the compile+run totals, which show what each mode costs end to end rather than execution-only, and a geomean.

The geomean is geometric rather than arithmetic because these are anchor-normalized ratios: an arithmetic mean of ratios is dominated by whichever benchmark carries the largest number. The basket is fixed to benchmarks present in every snapshot, since a geomean over a growing set moves when the set moves, which would read as a change that never happened.

Controls: tier chips (click to exclude from all, alt-click to solo, so single-tier and multi-tier are the same mechanism), a second row for the Go toolchain, per-chart series toggles on the legend, and a scale lock for when you want the axis to hold still while you toggle. All of them live in the query string, so a view is a link.

The toolchain row exists for the same reason as the tier row. Snapshots have always carried `machine.go_version` and nothing surfaced it, but a bump moves every timing number: the go1.26.4 to go1.26.5 *patch* bump alone shifted the geomean 3-4% per tier, and on EPYC 7763 the two populations barely overlap — go1.26.4 spans 33.9% to -4.0%, go1.26.5 spans 3.8% to -11.3%. Pooling them mixes two populations exactly the way pooling two tiers does. The row hides itself when a timeline carries only one toolchain.

### Look and feel

The page does not restyle anything. `pageStyle` is extracted from the existing template and both pages embed the same const; the explorer adds 20 lines for controls the summary page has no equivalent of, and reuses the real `.chart`, `.chart-head`, `.chart-foot`, `.legend` and `.swatch` classes. Two stylesheets that resemble each other drift; one does not.

The summary page's output is byte-identical after that extraction, checked by diffing a generated page before and after.

### Review fixes

A review found three failures that made the page render nothing, all now fixed with regression tests that fail when the fix is reverted:

- Empty series marshalled as `"pts": null` while the page read `.length` on it, so any sparse timeline — including the historical/current fallback used when no timeline directory exists — produced a TypeError and no charts.
- Timestamps used a non-standard format that Chromium parses and JavaScriptCore rejects. Confirmed with `jsc`: the page rendered nothing in Safari. They are RFC3339 now.
- The tier list was built from the chart specs, so a tier occurring only in the geomean could be plotted but not selected, and "All (n)" counted fewer tiers than it drew.

Also: links name the target file rather than assuming a directory index, so a flat `-out summary.html -viewer-out timeline.html` layout works in both directions.

### Testing

`go test ./cmd/perf-page/`, `go vet`, `gofmt` clean. Each of the four new tests was verified to fail with its fix reverted. Rendered against the live 474-snapshot timeline in Chromium: six charts, seven tiers, 5,798 points, filtering and URL round-trip working, console clean. Timestamp parsing checked directly in JavaScriptCore rather than only in Chromium, which is what missed it the first time.

Payload is 597 KB against the 29 MB `explorer.json`, because a chart needs one metric across a handful of series rather than every metric across 157 benchmarks.

### Not here

The release reference line, which needs the per-tier v1.8.0 capture (#597), and the status pill, whose "worst latest" summary would have to become per-tier too.

---

Built collaboratively with Claude.


nooga (approved):
Verified all four new regression tests exist and pass (`TestBuildViewerDataEmitsNoNilPointSlices`, `TestBuildViewerDataKeepsTimestampsParseable`, `TestBuildViewerDataListsEveryPlottedTier`, `TestBuildViewerDataCarriesTheToolchainOnEveryPoint`), covering the three bugs called out in the description (nil `pts` → JSON `null`, non-RFC3339 timestamps failing in JSC/Safari, geomean-only tiers undercounted). `gofmt`/`go vet`/`go test` all clean.

Approving. This branches off #879 — please retarget/rebase onto main once #879 lands.

PR #882, opened 2026-09-15
bench(vm): hoist per-iteration setup out of the vector benchmark loops

mparrett:
The perf-timeline ubuntu leg spends ~70 min in the pkg/vm job, but the snapshot it produces holds only ~16 min of measured time (sum of iterations × ns/op over all 465 records in `20260914T133639Z-8f9159258439-amd64-amd-epyc-7763`). The gap is `BenchmarkVectorConj`, `BenchmarkVectorAssoc` and `BenchmarkVectorSeq`, which built their input inside the `b.N` loop behind `b.StopTimer`/`b.StartTimer`. Every `StopTimer` runs `runtime.ReadMemStats`, a stop-the-world of tens of microseconds, so a sub-benchmark that reaches millions of iterations spent nearly all of its wall time outside the measured region. Scaled by the iteration counts in that snapshot, the 20 sub-benchmarks account for at least 38 of the 70 minutes, before counting ramp iterations. No other pkg/vm benchmark stops the timer inside its loop.

The macos-14 leg has the same shape with a worse ratio: its pkg/vm job on the same commit ran 128 min for 16 min of measured time, and the 20 entries reach 58M iterations there against 37M on ubuntu, so the per-iteration stop-the-world costs it more. That leg runs 135 min against its 150-min cap and is the largest single consumer of runner time (#876).

Local check at `-benchtime 2s -count 1` on the same machine: the 20 sub-benchmarks took 1749 s of wall before this change and 65 s after. `VectorConj/ArrayVector/10` alone went from 65 s of wall for 2.7 s of measured time to about 3 s.

`Conj` and `Assoc` copy on write for both vector kinds and seqs are immutable (`Next` returns a fresh node), so a single base or seq head built before the loop is exactly what each iteration used to rebuild. The loops use `b.Loop`, because Go starts the timer before entering the benchmark body: with a plain `b.N` loop the hoisted construction would sit inside the timed interval, amortized over N, and the unused results would have no liveness guarantee. `b.Loop` excludes the setup and keeps the results alive, and it also stops the setup re-running on every calibration step (the 20 run in 49 s rather than 65 s).

allocs/op and bytes/op are unchanged against the merge base on all 20 entries, measured on one machine, so nothing in this PR moves a deterministic bar. That is the claim that matters for `VectorConj`, which is in the `pr-fast` profile. It is not the same as passing the committed bars: `docs/perf/baseline.json` already fails its own global-minimum check on 15 of these 20 entries at the merge base (the M3 profile holds `VectorConj/PersistentVector/10` at 22 allocs where every amd64 profile has 32), which is a stale cross-profile minimum, not anything this PR introduces. Filed as #883.

Expect a step in the timeline on these 20 series. The measured ns/op drops along with the wall time (`Conj/ArrayVector/10` went from 2701 to 473 ns/op on the same machine), because the timer restart and the cache state after each stop-the-world were inside the old measurement. That is a benchmark correction, not a VM change. `bench-ratchet update` adopts the lower ratio and resets `best_since_sha` on those entries; `seed-baseline` takes the median of the last five snapshots per tier, so the seeded bar flips once three corrected snapshots exist. The perf page joins same-named points, so the step will render like an optimization unless it is annotated; I would rather leave the names alone and note the SHA here than fork 20 series. Measured on a fork dispatch of this head with publishing disabled (mparrett/let-go run 34925093900): the ubuntu leg ran 26 min end to end and the macos-14 leg 27 min, against 84 and 135 min for the same commit's parent on the upstream timeline. The upstream snapshot for the merge commit will be the record of it.

Two related things this PR leaves alone. The leg's `-timeout 30m` is inert for benchmarks (`testing.M.Run` stops the alarm before `runBenchmarks`), so `timeout-minutes` remains the only guard. And the count-3-versus-4 question from #573 becomes much cheaper to measure once the leg is short.


nnunley (approved):

mparrett:
Provisional leg times from a fork dispatch of this head with publishing disabled (mparrett/let-go run 34925093900), one run per leg:

| leg | parent commit, upstream timeline | this head, fork probe |
|---|---|---|
| ubuntu-latest | 84 min (p50; killed at 90 on a quarter of legs) | 26 min |
| macos-14 | 135 min | 27 min |

The pkg/vm capture step itself went from ~75 min to 23 min on ubuntu. The upstream snapshot for the merge commit is the number that counts; I will add it here when it lands.

mparrett:
Upstream timeline for the merge commit (16a12e29, run 34928096218), one leg each:

| leg | before (parent commit) | after |
|---|---|---|
| ubuntu-latest | 84 min | 26 min |
| macos-14 | 135 min | 28 min |

Same as the fork probe within two minutes. The snapshot is on `perf-data`.

PR #886, opened 2026-09-15
fix(perf-page): skip rows with no bar date in "Recently tightened"

mparrett:
## Preview

- this branch: https://mparrett.github.io/let-go/preview/recent-tightened/

## Why

"Recently tightened" is headed *Most recently lowered ratchet bars*. On the published page it shows the alphabetically first eight benchmarks, each with an empty "Bar set" column.

`BestSinceAt` is parsed with `time.Parse(time.RFC3339, ...)` and the error is discarded. An empty string yields the zero time, so when a profile carries no bar-set dates every row compares equal and the sort falls through to its `FullName` tiebreak. Nothing errors; the list quietly stops meaning what its heading says.

This is the shipping case. None of the five amd64 profiles in `docs/perf/baseline.json` carries `best_since_at` on any entry; only the locally captured `arm64/Apple M3` does, on 140 of 140. And `selectProfile` takes the first sorted key — `amd64/` sorts before `arm64/`, so the profile holding the data is never the one rendered.

## What

Rows without a parseable bar date are skipped before the sort. When that leaves nothing, the section says so instead of rendering a table.

A second commit covers the case one step further on: every remaining row sharing a single date is the same failure with the field populated, since nothing is then more recent than anything else. That one is not hypothetical either — a CI timeline snapshot stamps its own `captured_at` on every entry it holds, and all 163 benchmarks in the amd64 EPYC 7763 snapshots carry one identical `best_since_at` equal to the capture time. Seeding the baseline from snapshots, which is what #651 proposes, would put a date on every row and no recency in any of them.

Populating the field with real history is out of scope here and belongs with that seeding work.

## Why it survived since #173

The failure is invisible in the output. An empty `Bar set` cell reads as one row missing provenance, not as a sort that has degenerated into alphabetical order.

The ordering test passes against the old code as well — with every row carrying a date, both versions agree. Only `TestRecentlyTightenedSkipsRowsWithoutBarDate` fails when the fix is reverted; I checked that it does.

## Testing

`go test ./cmd/perf-page/`, `go vet` and `gofmt` are clean.

Rendered against `docs/perf/baseline.json`, the amd64 profile shows the empty state. An arm64-only baseline renders the table newest-first, the three 2026-08-24 rows above the 2026-07-18 ones, with real dates and SHAs.

## Related

- **#886 (this one)** — "Recently tightened" ordering
- #889 — section chrome


nooga (approved):
Real bug, correctly fixed: `time.Parse` on an empty `BestSinceAt` silently yields the zero time, so unparseable rows compared equal and the table degenerated to alphabetical order (confirmed this is what's shipping today — five amd64 baseline profiles have no `best_since_at` at all). Fix skips unparseable rows and also handles the "every row shares one date" case. Three targeted tests cover skip/order/tie cases and check out.

Approving. Independent of #879/#880 — recommend landing first among the four main-based fixes (886 → 887 → 888 → 889), per the author's own stated order.

mparrett:
Merged as `0cca386d`, on @nooga's approval alone. Recording that here because @nnunley was asked for a review when this opened on 2026-09-15 and had not responded — there was no feedback of his outstanding, and nothing was pushed after the request, so the approval covered the head as reviewed.

Rebased onto `c293934a` before merging; the replay was content-preserving (the same 2 files, +89/-3) and the three tests were re-run by name at the rebased head: `TestRecentlyTightenedSkipsRowsWithoutBarDate`, `TestRecentlyTightenedOrdersByBarDate`, `TestRecentlyTightenedDropsATotalTie`.

Happy to revisit in a follow-up if you see something here, Norman.

PR #887, opened 2026-09-15
fix(perf-page): size the timeline dots to the data, halo the delta labels

mparrett:
## Preview

- this branch: https://mparrett.github.io/let-go/preview/chart-legibility/

The difference only shows at real snapshot density, so compare the two directly: main draws the same 482 snapshots as a hatched mass.

## Why

The timeline charts had become unreadable at the current snapshot count, and the cause is arithmetic. The plot area is 456 units wide, so once the timeline passes a few hundred snapshots each one gets barely a unit while the fixed radius of 3.2 drew it 6.4 across: every dot several deep into its neighbours. At the 426 snapshots on `perf-data` as of 2026-09-15, that was 1.07 units apart against a 6.4-wide dot.

The paper-coloured ring on each dot then punched holes in the dots behind it, which is the hatched texture the charts had taken on. It also broke up the line enough to read as gaps. The "Suite memory" and "IR compile" lines look disconnected on the published page; one of them measures 424 points in 3 subpaths, so the line is continuous and the dots were covering it.

## What

The radius follows the spacing now (spacing x 0.45, capped at 3.2), and the ring is dropped once the dots are closer together than they are wide, since below that it erases more than it separates.

The floor is 1.76 rather than the spacing alone. Sizing purely to spacing gives 0.90 at 426 snapshots, which sits inside the 2.5-wide line: the snapshots stop being visible, and since each dot carries the date/SHA/value tooltip they stop being hoverable too. 1.76 clears the line's half-width.

The per-snapshot delta labels sit on top of the series, so they carry a paper halo via `paint-order` rather than competing with whatever is behind them.

## How it is not obvious

`{{$chart := .}}` before the `Series` and `Points` ranges is load-bearing. `$` is the template root, which is `PageData`, so reaching a `Chart` field through it compiles fine and fails at execution — `can't evaluate field Dense in type main.PageData`.

## Testing

Both branches checked against real data. At 426 snapshots: r 1.76, ring off, line 1.5. At 23: r 3.2, ring on, line 2.5, variance bands legible — the shape the fixed radius was tuned for. `gofmt`, `go vet` and `go test ./cmd/perf-page/` are clean, and the regenerated page loads with no console errors.

## Noticed while here, not fixed

"Suite memory" draws a legend with two series and one visible line. Both variants have identical heap bytes per op on this profile, so `ir_bytecode` sits exactly under `aot_native`. The chart is honest; the legend just cannot say so. Left alone rather than special-cased.

## Related

- **#887 (this one)** — timeline chart legibility
- #889 — section chrome


nooga (approved):
Fixes real overplotting at high snapshot density — dot radius now scales to spacing (min 1.76, capped at 3.2) instead of a fixed 3.2 regardless of spacing, and the paper-colored ring is dropped when dots are packed tighter than their own width. Delta-label halo via `paint-order: stroke` is minimal and correct. Diff matches the stated rationale exactly; verified a clean merge after #886 with `go build`/`go vet`/`go test`/`gofmt` all green.

Approving — land after #886.

PR #888, opened 2026-09-15
feat(perf-page): make the sparkline table readable at a glance

mparrett:
## Preview

- this branch: https://mparrett.github.io/let-go/preview/table-readability/

The sparkline table is most of the page; scroll to "Trend sparklines" in both.

## Why

Four things in the same table made its numbers something to decode.

Large values printed in full, so 2,806,110,000 and 13,908,200,000 were compared by counting commas.

Percentages held past the point where they mean anything: +1012.3% is 11.1x, and the multiple is the number actually being compared.

A family of scaling variants rendered a pill per scale across a `colspan` of three, so 133 of 773 rows broke the alignment of First, Last and Delta to show four numbers that are usually the same number: 10 21.0x, 100 22.0x, 1k 21.6x, 10k 21.2x.

The Trend header carried a full sentence, and `th` is nowrap, so it set the column's minimum width — for a sparkline canvas that is a fixed 160px.

## What

Values compact at 10k and above, to three significant figures: 2.81B, 13.91B.

Deltas read as multiples past a doubling and as percentages below it. paserati's perf page uses the same x idiom against its 1.0x bar.

Family rows fill the same three columns as every other row. First and Last carry the range across scales, collapsing to a single value when the scales agree once formatted. Delta carries the geometric mean and the count behind it — geometric because these are multipliers, and an arithmetic mean of ratios follows whichever scale has the largest one. The per-scale figures move to the hover tip, keyed to the three summary cells, which carry a dotted underline to mark that the detail is there.

The Trend header shortens to "Trend" and the name column takes the slack, which drops truncation to 0 of 80 rows at 1440, 1280 and 1100 wide. What the old header explained is already under the section heading.

## How it is not obvious

A `title` attribute was the first attempt at the per-scale detail and never showed: native tooltips wait out a hover delay and do not appear at all while the pointer is moving across a dense table.

Fixing that surfaced a pre-existing contrast bug. The generic `.good`/`.bad` carry a light pill background for the table, and the tip rule overrode only the colour, so a delta rendered as pale text on a pale block against the dark tip. That affected the trend tooltip that was already there.

## Testing

`gofmt`, `go vet` and `go test ./cmd/perf-page/` are clean. Regenerated against the 426-snapshot timeline and loaded in headless Chromium at 1400x900: 773 rows, no console errors, family rows showing ranges and a geometric delta with the scale count.

## Related

- **#888 (this one)** — sparkline table readability
- #889 — section chrome


nooga (approved):
Compact-number formatting, multiple-based deltas for large percent changes, and collapsed scale-family rows (correctly using geometric mean for aggregating multiplicative deltas rather than arithmetic mean) all check out and match the stated problem.

One non-blocking nit: in the collapsed scale-family row, the aggregate delta cell's `good`/`bad` color class is taken from `d.series[0].good` (the smallest-scale variant) rather than the sign of the geometric-mean delta actually shown in that cell. If per-scale variants disagree in direction, the number and its color could mismatch. Low severity, no test coverage added (consistent with the rest of this client-JS file) — worth a one-line fast-follow (derive `good` from the geomean sign) but not blocking this PR.

Approving — land after #887.

mparrett:
Filed the nit from your review as #893, so it survives this merge: the collapsed scale-family delta cell colors from `d.series[0].good` while printing `geoDelta(d.series)`.

The disagreement is reachable — with `lower_is_better`, a family at scale 8 of −5% and scale 1024 of +40% gives a geometric mean of +15.3%, which renders as `+15.3%` in the green `good` class.

Merging this as-is per your non-blocking call; #893 carries the one-line fix.

PR #889, opened 2026-09-15
feat(perf-page): name the baseline, cap its table, align the section heads

mparrett:
## Preview

- this branch: https://mparrett.github.io/let-go/preview/section-chrome/

Compare the "Current baseline" section in both: heading alignment, the provenance line, the `vs v1.8.0` column header and the capped table.

## Why

Four pieces of section chrome in the same template, each from reading the published page.

`.section-head` bottom-aligned the heading and its description, so a heading that wrapped pushed its description down to meet the last line instead of starting level with it.

"Current baseline" said only that it was sorted by package and benchmark, and its Delta column named nothing to compare against — the reference is stated at the top of the page, a full screen away.

That table is the longest on the page and ran at full length, while the two summary tables beside it were already capped.

Both capped tables then take most of the viewport, and the macOS overlay scrollbar only appears once you are already scrolling, so the region gives no sign that it scrolls.

## What

Heading and description both align to the top. The longer descriptions were also squeezing headings into a ~115px column, so a 150px floor holds the natural two-to-three line wrap.

The baseline description carries the capture date, SHA, CPU model and Go version on their own line, says what the Scale bar plots, and the column header reads `vs <ref>`.

The table scrolls inside a 620px region with a sticky header, the same treatment the sparkline table has.

`::-webkit-scrollbar` gets an explicit 14px width, which opts these two scrollers out of the overlay behaviour.

"Trend sparklines (first → last)" drops the parenthetical; first and last are what a sparkline shows, and the section's own description says so.

## How it is not obvious

No `scrollbar-width` or `scrollbar-color` alongside the pseudo-element rules: setting either makes Chrome ignore them entirely, which is how a first attempt rendered a 2px sliver.

## Testing

`gofmt`, `go vet` and `go test ./cmd/perf-page/` are clean, and the regenerated page loads with no console errors.

The scrollbar is the one thing headless cannot confirm: headless Chromium draws overlay scrollbars regardless and omits them from screenshots. It was checked in Chrome on macOS, where the track holds its space at 14px.

## Related

- **#889 (this one)** — section chrome


nooga (approved):
Verified: top-aligns heading/description (was bottom-aligned, broke on wrapped headings), adds a min-width floor, states the baseline's provenance in the column header using the existing `date`/`shortSHA` template funcs correctly, caps the "Current baseline" table to a scrollable 620px region matching the sparkline table, and adds an explicit `::-webkit-scrollbar` width to opt out of macOS's hidden overlay scrollbar. Small, self-contained, matches the description.

Approving — land last in the stack (after #886, #887, #888). Confirmed the full 886→887→888→889 sequence merges cleanly onto main with `go build`/`go vet`/`go test ./cmd/perf-page/...`/`gofmt` all passing at each step.

PR #890, opened 2026-09-17
docs: give the comment linter a page and link it from the index

mparrett:
`scripts/lint.lg` (#836, #870) and the `comment-lint` skill (#858) both shipped with no documentation, and nothing in the tree references either. Searching `comment-lint` outside the skill's own directory returns nothing, so the only way to learn the rule file exists is to already know.

`docs-status` is the precedent both were modelled on, and it is wired into four places: its own page, a `docs/README.md` row, a `docs/contributor-workflow.md` entry, and `.github/workflows/go.yml`. #858 copied the file shape and none of the wiring. This adds the three that apply.

## What the page commits to, and what it does not

Scoped to what is settled: how to run it, the flags, the rule kinds it reports, that it is report-only and wired into neither CI nor the Makefile, and where the judgement layer lives.

The rules themselves are **not** presented as policy. The page says #835 is where they get decided, and states plainly that some of what the skill carries is one contributor's proposal rather than an agreed rule. That hedging is deliberate: a docs page is harder to qualify than a PR comment, and the ID rule in particular is still open.

Listed under `## Commands` rather than the CI gate table, because it is neither a CI job nor a make target. A gate-table row would be false.

## Verification

`scripts/docs_status.py` passes at this head (51 docs scanned, exit 0), and `docs/comment-lint.md` does not appear in its "missing from README index" list, which is the check the new index row exists to satisfy. The eight findings it does report are all pre-existing and untouched here.


nnunley (approved):
Fills a real gap: nothing in the tree referenced scripts/lint.lg or the comment-lint skill. Scoped correctly to what is settled (how to run it, the rule kinds, report-only, not in CI or the Makefile) and explicitly defers the rules themselves to #835 rather than presenting them as policy. Listing it under Commands rather than the gate table is right, since it is neither a CI job nor a make target. Frontmatter present; docs_status.py clean.

PR #891, opened 2026-09-17
perf(bench-ratchet): targeted Apple M3 timing recapture (anchor + 3 rows)

nnunley:
Targeted timing recapture of the `arm64/Apple M3` profile: the anchor and three rows, with the other 137 retained untouched. Rebuilt as one commit on current main now that #900 has merged; it is no longer stacked.

## Why

The M3 timing profile dated from 2026-08-24 (`dadb8e0b`). Replaying the timing gate per capture against that row, `IRCompile [bytecode]` reads +1.04, +2.10, +2.57, +2.77, +3.53, +3.83, +4.03, +4.05, +5.48, +6.56 % across ten captures — over the 5% budget in 2 of 10, around a true value of about +2.6%. So that row makes the gate flaky rather than failing it outright. The other two rows are 6–12% and 19–27% fast against the stored bar; those are real improvements the ratchet should be holding.

## What

Re-records the anchor and `IRCompile [bytecode]`, `IRCompile [gogen_ir]`, `InitFromLGB [bytecode]` at `main@827ef18b`. Captured under the bench-baton exclusive lease (#795), `caffeinate -dimsu`, on AC, Low Power Mode off, no concurrent builds, lowering wire-up generated with `make lowered` so the `gogen_ir` variant measures the lowered passes.

Timing only: `-force` without `-accept-deterministic`, so the two `IRCompile` bytes/op measurements above the stored bar stay refused and that bar is kept.

## How it was accepted

On a **five-capture window reduced in ratio space**, which is what `docs/perf/ratchet.md` §"Why a window, and why not gate on the anchor" prescribes — not on a single capture, and not on anchor drift, which that section explains is "the obvious design and is wrong".

Leave-one-out coherence offsets in ratio space: **−1.59, +0.37, +0.43, +0.42, −1.07 %**. No capture is mixed, so all five are kept, and the row written is the window-median capture.

Stated plainly, because it bears on how much this row can be trusted: across-capture spread per row is **5.4%, 6.3% and 3.3%** against a corpus worst of 1.75%, so residual uncertainty in these medians is roughly 1–2%. That is per-benchmark sampling noise rather than a whole-capture offset, which is why the coherence screen passes. The anchor reproduces to 1.2% across the window; for reference the stored 2026-08-24 row's own anchor samples span 2.5%.

## What this does *not* do

**It does not unblock the gate,** and an earlier version of this description claimed otherwise. `check` exits 1 if either gate trips, and they are independent — `compareDeterministic` against the machine-independent bar sets `exit = 1` before the timing comparison runs (`cmd/bench-ratchet/main.go`). Four deterministic metrics exceed the newest measured row:

| metric | bar | measured | Δ |
|---|---|---|---|
| `InitFromLGB` allocs/op | 13,995 | 14,353 | +2.6% |
| `InitFromLGB` bytes/op | 1,016,128 | 1,051,381 | +3.5% |
| `IRCompile [bytecode]` bytes/op | 4,941,030 | 5,218,615 | +5.7% |
| `IRCompile [gogen_ir]` bytes/op | 5,998,769 | 6,260,974 | +4.4% |

Those reproduce to within 0.01–1.4% across fifty samples, so they are code movement since `477a5d36`, not capture noise, and a timing-only recapture cannot absorb them by design. After this merges, `jj push` from this machine still exits 1 on them. They need fixing or a deliberate deterministic rebase — separate from this PR. Some of that growth is the eager def-metadata cost tracked in #897.

Refs #791, #899.


docs/perf/baseline.json
```diff
@@ -20,21 +20,27 @@
...
+          "ratio_to_anchor": 1966000.0000000002,
+          "best_since_sha": "9172aa2444659a6f8845cea92e1f70d9449ef8a8",
+          "best_since_at": "2026-09-16T02:08:39Z"
         },
         "github.com/nooga/let-go/pkg/ir.BenchmarkIRCompile [bytecode]": {
           "ns_per_op": 34214631,
-          "allocs_per_op": 113777,
-          "bytes_per_op": 4941030,
-          "ratio_to_anchor": 27481631.325301204
+          "allocs_per_op": 105815,
```
mparrett: This forced update changes more than the stated `InitFromLGB` baseline. The global `IRCompile [bytecode]` bar moves from 103,414 to 105,815 allocs (+2.32%) and 4,941,030 to 5,233,366 bytes (+5.92%); `[gogen_ir]` moves from 5,975,898 to 6,271,989 bytes (+4.95%). Those changes exceed the deterministic gate's 2% tolerance and weaken unrelated guardrails. Please limit the accepted reset to `InitFromLGB`, or explicitly justify and independently measure these IR regressions.

docs/perf/baseline.json
```diff
@@ -4705,8 +4735,8 @@
       }
     },
     "arm64/Apple M3": {
-      "captured_at": "2026-08-24T20:18:19Z",
-      "captured_at_sha": "dadb8e0b54b7317e4425ad61493af3f5bbce6bda",
+      "captured_at": "2026-09-16T02:08:39Z",
+      "captured_at_sha": "9172aa2444659a6f8845cea92e1f70d9449ef8a8",
```
mparrett: This profile was captured at `9172aa24`, but the target branch subsequently merged `a044ead`, which changes `core_compiled.lgb` and the IR lowering pipeline—the exact workloads represented by all three updated rows. The baseline therefore describes code older than what will exist after merge. Please rebase and rerun the capture so the provenance and deterministic bars correspond to the merged code.

mparrett (changes requested):
The generated file and ratchet tests are valid, but this update weakens unrelated deterministic gates and was captured before benchmark-relevant changes already present on the target branch. Please recapture from current main and either preserve the existing IR bars or explicitly justify the additional accepted regressions.

docs/perf/baseline.json
```diff
@@ -4717,349 +4717,349 @@
...
         "github.com/nooga/let-go/pkg/ir.BenchmarkIRCompile [bytecode]": {
-          "ns_per_op": 15137978.000000002,
-          "allocs_per_op": 114255,
-          "bytes_per_op": 4963229,
-          "ratio_to_anchor": 14267651.272384545,
-          "best_since_sha": "dadb8e0b54b7317e4425ad61493af3f5bbce6bda",
-          "best_since_at": "2026-08-24T20:18:19Z",
+          "ns_per_op": 17295854,
+          "allocs_per_op": 105814,
+          "bytes_per_op": 5232922,
```
mparrett: Under #900's selection rule, this freshly stamped M3 entry becomes the deterministic reference. Compared with the pre-891 newest row, `IRCompile [bytecode]` bytes rises from 4,941,030 to 5,232,922 (+5.91%), and `[gogen_ir]` rises from 5,998,769 to 6,269,537 (+4.51%). Both exceed the 2% deterministic budget. #900 explicitly says the bytecode regression remains flagged, while this PR describes the recapture as timing-only; the combined stack instead accepts it. A timing-only force operation must preserve deterministic values and their provenance.

mparrett (changes requested):
Composite review with stacked PR #900: the revised capture resolves my earlier concerns—it changes only the M3 profile, is based after the benchmark-relevant merges, and its stored reductions are internally consistent. However, under #900's newest-row rule this nominally timing-only recapture also accepts deterministic regressions that #900 says should remain flagged.

nnunley:
Agreed, and the fix is in the tool (#900, `d7c1e119`): `-force` now keeps allocs/bytes it measured worse, with their provenance, unless `-accept-deterministic` is passed. This row was rebuilt with that tool from the same capture's raw `.jsonl`, no re-benchmark: every M3 timing equals the capture, `InitFromLGB` (an improvement) is adopted and stamped, and the two bytes regressions you cite are rejected — `IRCompile [bytecode]` stays at 4,963,229 and `[gogen_ir]` at 5,975,898, with their previous provenance. `check` from this head reports both as regressions against the `477a5d36` row (+5.9% and +4.5%), which is the honest state of the code at `3bbde90a`; they are #791's to fix, not this recapture's to accept. Body updated. Still stacked on #900.

docs/perf/baseline.json
```diff
@@ -4705,8 +4705,8 @@
       }
     },
     "arm64/Apple M3": {
-      "captured_at": "2026-08-24T20:18:19Z",
-      "captured_at_sha": "dadb8e0b54b7317e4425ad61493af3f5bbce6bda",
+      "captured_at": "2026-09-18T20:03:51Z",
+      "captured_at_sha": "3bbde90a0513a78a088778424bdeb9afc450887f",
```
mparrett, line 4709: Current main has advanced to `ebd758a` and changed `pkg/rt/core_compiled.lgb`, the exact artifact measured by `InitFromLGB`. A focused run on current main reports 11,902 allocs / 905,996 bytes versus this row's 11,900 / 905,860. The difference is within budget, but this data-only timing baseline no longer represents the code it will merge into. Please rebase after #900 and recapture against current main.

mparrett (changes requested):
Re-review after the stacked #900 fix: this revision correctly preserves the rejected IR byte regressions and its stored reductions are internally consistent. However, main has advanced past the recorded 3bbde90a capture with a benchmark-relevant core bundle change, so the data needs one more rebase/recapture after #900 lands.

nnunley:
Rebuilt with #900's per-metric provenance (`edbf9e71`): adopted `IRCompile` allocs are now stamped at `3bbde90a`; the two rejected bytes rows keep their `477a5d36` reference. Agreed on the recapture against current main — it needs a quiet machine and will follow #900's merge, since the row's stamps depend on that tool. Drafted with Claude and checked by me.

mparrett:
One more for the recapture, since it rewrites this data anyway.

The title, body, and commit message describe a whole-profile re-record: "Re-records the `arm64/Apple M3` profile at `main@3bbde90a`", and "every M3 timing value in the profile equals the run's reduced median". Three of the 140 benchmark rows changed — `InitFromLGB [bytecode]` and the two `IRCompile` variants — plus the anchor. The other 137 keep values whose `best_since` is `f154c7fb`, dated 2026-07-18, and the new rows' samples run 17:11 to 17:13 against an anchor at 17:08: an anchor and three benchmarks.

The profile's `captured_at_sha` moves to `3bbde90a` all the same, so the file asserts a capture commit that 137 of its rows do not share. Nothing reads it today, since every entry in that profile carries a `best_since_*` and `unratchetedProvenance` never reaches the profile stamp. But this is the file whose provenance the stack is about, and let-go squashes, so the sentence outlives the PR. Scoping the title, body, and message to the three benchmarks, and leaving `captured_at_sha` describing what was re-measured, would make it true.

Separately, on the rebase you and I agreed on: I measured current main on an Apple M2 and it should barely move the row. `InitFromLGB` reads 11,902 allocs and 906,007 to 906,266 bytes against this row's 11,900 / 905,860; `IRCompile [bytecode]` reads 105,405 to 105,410 allocs against 105,814, and `[gogen_ir]` 143,553 to 143,559 against 143,839. All inside the 2% budget.

mparrett:
Clarifying my “One more for the recapture” comment above: **#891 is not approved; my changes request remains** until #900 lands, this branch is rebuilt on then-current main, and the affected benchmarks are recaptured on the Apple M3. #900 is approved at `75bf7707` but has not merged yet. My Apple M2 spot-check is reassurance, not a substitute for that M3 capture.

The point about 137 retained rows was **not** an additional blocker or a request to remeasure all 140. Retaining unmeasured entries is intentional; the profile-level `captured_at_sha` identifies the latest run, while per-entry stamps identify retained bars. I overstated what the profile stamp asserts. The non-blocking suggestion is simply to describe the current change as a targeted recapture (anchor plus three benchmarks), rather than a whole-profile re-record, in the PR title/body and eventual commit message. Sorry my earlier wording blurred that distinction.

nnunley:
@mparrett — rebuilt on current main as one commit, `67cda12d`, now that #900 has merged. Retitled and rewritten as the targeted recapture you asked for: anchor plus three rows, `arm64/Apple M3` only, 137 rows retained.

The capture was taken by a second agent session holding the bench-baton exclusive lease (#795, merged this morning) while this session stayed off the machine; I verified its window independently from the raw `.jsonl` files before applying, rather than taking the numbers on trust.

Two corrections to what this PR previously claimed, both of which I got wrong:

**1. The acceptance criterion.** I had proposed gating the capture on anchor sample spread. That is wrong, and `docs/perf/ratchet.md` says so directly — §"Why a window, and why not gate on the anchor" calls it "the obvious design and is wrong", with the EPYC 9V74 case (22.4% off its window's anchor, agreeing on every ratio to 0.1%) as the counterexample. The row is now accepted on a five-capture window reduced in ratio space, leave-one-out coherence offsets −1.59/+0.37/+0.43/+0.42/−1.07 %, all five kept. I've also stated the across-capture spread per row (5.4%, 6.3%, 3.3% against a corpus worst of 1.75%) in the body rather than leaving it implicit, since it bounds how much the row is worth.

**2. "Every push fails the timing gate."** Not true. Replayed per capture, `IRCompile [bytecode]` exceeds the 5% budget in 2 of 10 captures around a true +2.6% — a flake, not a stale row. And the thing that actually fails `jj push` from this machine is the *deterministic* gate, which is independent and trips first (`compareDeterministic` sets `exit = 1` before `compareAndReport` runs). Four allocation metrics are over the newest measured row, reproducible to 0.01–1.4% across fifty samples, so they are real code movement since `477a5d36` and no timing recapture can absorb them. The body now has the table and says plainly that this PR does not unblock the gate.

Unrelated correction to something I posted on #819 earlier today: the zero-record `ClojureTestSuite` capture jobs are **not** a bench-ratchet parsing bug against #863. jj workspaces do not check out the `clojure-test-suite` submodule, so that benchmark cannot run in one — `scripts/link-clojure-test-suite.sh` exists for exactly this. Those six jobs are stored in no profile, so they affect neither this PR nor the gate. My mistake; please disregard that note.

---
Delivery verified by local `pr-claim-gate`: exact PR/validated head `67cda12d4f504c3b9f5c1540dc592eb7cc0eb92f`; relevance claims `docs/perf/baseline.json` occur in both the delivered delta and the PR feature delta; receipt `/Users/ndn/.local/state/pr-claim-gate/nooga-let-go/pr-891.json`.

mparrett (changes requested):
Re-review of current head `67cda12d`: **changes requested**. The rebuild is now a single targeted M3 data change on merged #900, and the selected capture's stored sample medians and ratios are internally consistent. But this nominally timing-only `-force` update silently accepts two deterministic regressions.

Before this PR, the newest machine-independent `InitFromLGB [bytecode]` bar is 13,995 allocs / 1,016,128 bytes from `477a5d36`. The proposed M3 row at `docs/perf/baseline.json:4796-4806` records 14,353 / 1,051,381 (+2.6% / +3.5%) and stamps both metrics at `827ef18b`. They are improvements over the old *M3* row, so `forceRebaseline` adopts them, but they are regressions against the newer *global* bar. `machineIndependentBar` therefore selects the new M3 values and `compareDeterministic` reports **zero** `InitFromLGB` regressions. The two IR byte regressions still make the overall gate fail, but two of the four deterministic failures listed in the PR body disappear.

Focused repro on this exact head: in a test under `cmd/bench-ratchet`, load `docs/perf/baseline.json`, select `machineIndependentBar(b)["github.com/nooga/let-go/pkg/compiler.BenchmarkInitFromLGB [bytecode]"]`, and compare a current entry at 14,353 / 1,051,381 with `allocBudget`. The selected bar is 14,353 / 1,051,381 at `827ef18b`, and `compareDeterministic` returns 0; a timing-only recapture should leave the prior 13,995 / 1,016,128 reference and flag both metrics.

Please have the forced-update path check each measured deterministic metric against the pre-update machine-independent bar, not only this machine's prior row. Without explicit deterministic acceptance, a worse global value must not become the newest stamped reference. Then rebuild the baseline from the same raw captures if they are still applicable, or explicitly justify an intentional acceptance of these regressions. This is a tool/data interaction, not a problem with the selected capture's within-run reduction. I could verify that stored capture, but the other four raw captures are not in the PR, so I did not independently verify the stated five-capture coherence offsets.

mparrett:
@nnunley grain of salt, can't oversee to know if the agent is being pedantic here or missing the big picture.

nnunley:
@mparrett — not pedantic. I reproduced it before answering, and it is a real hole that this PR trips; I'd rather it block than land.

Running your repro against `machineIndependentBar` on both sides, comparing a current entry of 14,353 / 1,051,381 at `allocBudget`:

```
BEFORE (main)  bar = 13995 allocs / 1016128 bytes  since 477a5d36  -> 2 regression(s)
AFTER  (#891)  bar = 14353 allocs / 1051381 bytes  since 827ef18b  -> 0 regression(s)
```

So this nominally timing-only PR raises the global deterministic bar by +2.6% / +3.5% and erases both `InitFromLGB` regressions — including the two that #913 was filed for this afternoon. That is worse than a missed catch: it would have quietly invalidated a freshly filed issue.

The mechanism is exactly as you describe. `forceRebaseline` compares each measured deterministic metric against **this machine's prior row** only. The old M3 value was 26,294 allocs, so 14,353 reads as an improvement, gets adopted, and gets stamped at `827ef18b`. Since #900 made selection newest-provenance-wins rather than global-min, that worse-but-newer stamp then becomes the bar every profile is gated against. #900 created the precondition and this PR is the first to walk into it — my responsibility on both counts, since I shepherded #900 and produced this data.

Your fix is the right one and I'd add one word to it: the forced path should compare each measured deterministic metric against the pre-update machine-independent bar, and when a metric is worse it should keep the prior value *and its stamp* rather than writing its own — the same "only a worse measurement keeps the old stamp" rule #900 established for `ratchetMerge`, just applied against the global bar instead of the local row. `-accept-deterministic` stays the deliberate override.

That is a code change to `cmd/bench-ratchet`, not a data change, so I don't think it belongs in this PR. I'd propose: a separate PR that fixes `forceRebaseline` and lands your repro as a regression test, then rebuild this row from the same five raw captures on top of it and re-verify that the two `InitFromLGB` regressions are still reported. The captures are still on disk. Holding this PR until that lands — it should not merge in its current form.

@nnunley — flagging since mparrett asked for a second opinion: the finding stands on its own, the numbers above are from running it rather than reading it. Your call on whether I open the bench-ratchet fix now or queue it.

---
Delivery verified by local `pr-claim-gate`: exact PR/validated head `67cda12d4f504c3b9f5c1540dc592eb7cc0eb92f`; relevance claims `docs/perf/baseline.json` occur in both the delivered delta and the PR feature delta; receipt `/Users/ndn/.local/state/pr-claim-gate/nooga-let-go/pr-891.json`.

nnunley:
@mparrett — #930 fixes the hole you found. Your repro is the regression test, and it fails on `main` exactly as you described.

Replayed against this PR's own capture with the fix in place, the forced update now refuses both metrics — `kept 13995 (since 477a5d36e25f), measured 14353` and the bytes equivalent — while still writing the timing it was run for. So the two #913 regressions survive the recapture instead of being erased by it.

This PR stays as it is until #930 lands; then I rebuild the row from the same five captures on top of it and re-verify that `check` still reports those two. Holding it rather than merging into the gap.

PR #892, opened 2026-09-17
ci(perf): skip the timeline for .github-only commits

mparrett:
## Why

The push trigger already skips `docs/**` and `**.md`, and the comment above it gives the reason: a commit with byte-identical code would spend a full bench run recording its parent's numbers under a new sha. CI plumbing is the same case and is not on the list.

Measured on `origin/main`: **6 of the last 120 commits touch only `.github/**`**. Two of them have nothing to do with benchmarking, and each started a timeline run:

| commit | subject | timeline run |
|---|---|---|
| `9309c95d` | `ci(tinygo): fetch the pinned wasmtime asset instead of curl\|bash` | 133 min |
| `e344ff23` | `ci(tinygo): pin wasmtime and verify it landed on PATH` | 131 min |

264 minutes of runner time across two commits that changed no measured code.

## The exception, and why this is still the right default

A change to the measuring workflow *can* move the numbers, and two of the six were exactly that — `9172aa24` (#878, lowering the ubuntu capture cap) and `c39c0a93` (freezing a per-release baseline). Those now need a dispatch rather than arriving automatically.

That trade seems right in both directions. The automatic snapshot on a measuring change is not obviously trustworthy anyway: the code is identical to the parent, so any movement it shows is the measurement changing rather than the program, which is worth recording deliberately rather than silently. `workflow_dispatch` is untouched and already takes a ref for exactly this.

## What

One entry in `paths-ignore`, plus a comment naming the exception so the next reader does not widen the list further without noticing it.

## Testing

Verified on my fork, which carries the same workflow: after the change, a push touching only `.github/**` started Go, CodeQL and the Pages deploy, and no Perf Timeline run. Before it, the same shape of push started one.

Sits alongside #882 and #878 in the capture-cost work, and under #876.


nnunley (approved):
Same case as the docs/**/**.md skip, and the comment names the one exception (a change to the measuring workflow itself) and the remedy (dispatch a backfill) in place, so the next reader won't widen the list blindly. Two recent .github-only commits burning ~264 runner-minutes to record their parent's numbers is a concrete cost; losing the automatic snapshot on a measuring change is the right trade since that snapshot measures the instrument, not the program.

PR #896, opened 2026-09-18
feat(rt): net/listen, net/accept and net/local-address — owned TCP listeners

mparrett:
Draft, as a proposal. The patch is @nnunley's: he specified it in https://github.com/nooga/let-go/issues/523#issuecomment-5579712836 and has been carrying the implementation in strange-lettractor as [`runtime-patches/net-listener.patch`](https://github.com/nnunley/strange-lettractor/blob/main/runtime-patches/net-listener.patch), applied to a pinned fork build on every install. The first commit here is that patch verbatim, under his authorship; the second adds a Go test and a guide section. Opening it so the change gets reviewed here instead of living on as a downstream patch.

## What it adds

`net` today is client-only (`dial`, `write!`, `read!`, `close!`, from #526). This adds the three server functions from the #523 contract:

- `(net/listen host port)`: explicit host, port 0 asks the OS.
- `(net/local-address listener)`: `{:host :port}` of the bound socket, so a port-0 listener can be advertised without a free-port-then-bind race.
- `(net/accept listener)`: blocks, returns the same connection value `dial` does, so `write!`/`read!`/`bencode/*` work on it unchanged.
- `net/close!` also takes a listener. Idempotent; wakes a blocked `accept`; does not close accepted connections, which the caller owns.

`unix` already has `listen`/`accept`; this is the TCP counterpart. The js/wasm stubs reject the new names the way they reject `dial`.

## Evidence

- `TestNetListen` (new): port-0 bind, `local-address`, accept of a Go client, `write!`/`read!` over the accepted connection, listener close under a blocked accept, accepted connection surviving it, argument checks. Passes with `-race`.
- Existing `TestNet*` and `TestBencode*` unchanged and passing.
- Builds for linux/amd64, js/wasm, wasip1/wasm, plan9/amd64; `go vet` and golangci-lint clean on `pkg/rt`; `check-generated` OK after `make generate` (only the manifest and sums move).
- Downstream: strange-lettractor's nREPL hub and OAuth loopback run on this patch today, and its `test/probes/net_listener_check.lg` covers two clients, fragmented and coalesced bencode frames, and close wakeups.

## Open questions for review

- Ownership model: these are explicitly owned handles, not scope-cancelled resources. That matches `dial` today. If the scope work in #830/#832 is meant to cover sockets, this is where to say so.
- `local-address` returns a map; `unix` has no equivalent. Fine to diverge, or should `unix` grow one for symmetry?


nnunley (approved):
Approving. The first commit is my own patch from strange-lettractor (`runtime-patches/net-listener.patch`), so this approval is for @mparrett's second commit — the test and the guide section — plus a check that the first commit is that patch verbatim, which it is (line for line; the two diffs only differ in which `}` counts as context).

On the test: it covers the whole #523 contract, including the two edges that matter — a listener close wakes a blocked `accept` with an error, and the accepted connection still works afterwards. No sleeps; the only bound is the 5 s wait for the wake-up. Guide text matches the behaviour.

Two observations, neither blocking:

- A dropped listener stays bound until process exit; there is no finalizer. That matches how `dial`'s connections already behave, so it is consistent rather than a regression.
- `local-address` on a `0.0.0.0` / `::` listener reports that unspecified address as `:host`. Honest, and the guide already tells callers advertising a port to pass an explicit host.

Once this lands the downstream patch goes away and strange-lettractor pins a fork build that already has it.

mparrett:
@nnunley Thanks for the patch and for reviewing the PR. And @nooga, I'll take your 👍🏻 on the review as directional endorsement. A couple of things came up in a final review pass. I'll fix those here momentarily rather than in a follow-up.

mparrett:
@nnunley @nooga pushed one new commit, would appreciate your sign-off before merge.

https://github.com/nooga/let-go/pull/896/changes/6ea143996d326d619ab87a24a62f625c18308b2b#diff-d9f1afaac30dabae644e14200518c065a1074973ec4efbe35bb686b795f3195a

nnunley:
@mparrett — I merged this on the standing approval before reading your two comments above, which is my mistake: you'd asked for sign-off on `6ea14399` first. Apologies. I've reviewed that commit after the fact rather than leave it unreviewed:

- `local-address` keeps the IPv6 zone (`fe80::1%en0`), which is the right call — a link-local host without its scope isn't dialable. The `addrOnlyListener` test pins it without needing a real link-local interface.
- The accept test is now bounded (5 s on the accept channel, client and accepted-connection deadlines) and observes entry into `Accept` through the wrapped listener before closing, so the close-wakes-accept case can't race the goroutine start.

Nothing I'd have asked to change; CI was green on that head. If you'd have wanted anything different in it, a follow-up PR is on me. This was posted with Claude's help; the review and the mistake are mine.

nnunley (approved):

PR #898, opened 2026-09-18
http: stoppable server, empty :headers fix; core: & rest binds nil

abogoyavlensky:
Three changes that surfaced together while building a small integrant + sqlite web app under lgx. Each is a separate commit.

## 1. `[x & more]` destructuring binds `nil`, not `()` (`fix(core)`)

`destructure-vector` expanded the rest binding to `(drop i n)`. `drop` returns an empty lazy seq, which is truthy, so the common `(if-let [[x & xs] s] ...)` loop never terminates. Clojure's `destructure` uses `nthnext`, which yields `nil`.

Concrete casualty: `weavejester/dependency`'s `reachable?` uses exactly that loop, and `dep/depend` runs it on every edge whose target already has an edge, so integrant's `ig/init` hangs on any three-component chain (`server -> handler -> db`), depending on map iteration order.

```clojure
(let [[a & more] [1]] more)   ; before: ()   after: nil
```

The expansion emits `clojure.core/seq` over `drop` (`nthnext` is defined later in `core.lg` than the first `let` that needs the expansion; the qualified symbol keeps a local named `seq` from capturing it). Plain fn rest args were already `nil` and are untouched. Generated artifacts regenerated with `make generate`.

## 2. `http/serve` accepts an empty `:headers` map (`fix(http)`)

`{:status 204 :headers {} :body ""}` panicked `ServeHTTP` (`invalid memory address or nil pointer dereference` in the header loop) and the client got no reply. Same nil-entry guard #849 added to the client loops.

## 3. `http/start`, `http/stop`, `http/wait`; `serve` is `start` + `wait` (`feat(http)`)

`http/serve` was `ListenAndServe` on the calling goroutine: no handle, no stop, a bind error the only way back.

```clojure
(def srv (http/start handler "127.0.0.1:0"))  ; binds now; a bad address throws here
(:port srv) (:addr srv)                        ; resolved, so ":0" works
(http/stop srv)         ; graceful, then Close after 5 s (or (http/stop srv ms))
(http/wait srv)         ; blocks until fully stopped; nil, or the Serve error
```

`start` returns an `http/Server` record (`:addr`, `:port`, `:server`). `wait`/`stop` accept the record or the handle. "Fully stopped" is after `Shutdown` returns (Go's `Serve` returns `ErrServerClosed` before in-flight requests drain) and after the Serve goroutine has exited (so the port is actually released, even when `stop` lands before `Serve` ran). A `Serve` that fails on its own closes accepted connections before `wait` returns. `serve` keeps its blocking contract and is now `start` + `wait`. Both run under the caller's scope, like the clients since #848: cancelling the scope stops the server.

New guide: `docs/guide/http.md` (the namespace had no page).


mparrett:
Verdict: correct. Reviewed head: dfda8509d1548e86c02ca913c5086cab10da03cf. 

Not run locally: CI's race and generated-artifact lanes already cover that.
Test gaps, non-blocking: none material. If anything, the lg-side test only covers the happy path, but the Go tests carry the edge cases.

mparrett (approved):
Looks solid, thanks!

abogoyavlensky:
@mparrett Thanks! Just merged latest main into the branch.

Issue #899, opened 2026-09-18
bench-ratchet: gate timings on an interval from the stored samples, not a point against a 5% budget


nnunley:
`bench-ratchet` reduces every timing to one number and compares two of them with a flat budget. The baseline keeps the samples it was captured from (ten per benchmark on the Apple M3 row) but reduces them to a single `ns_per_op` (median after a discarded warm-up rep); `check` does the same over `-count 4` at one second each. The dispersion on both sides is thrown away, so the gate cannot distinguish a regression from ordinary variance on the machine it runs on.

## Evidence

Seven full captures of the fast gate on one Apple M3 at `main@3bbde90a` on 2026-09-18, `-count 10 -benchtime 2s`, Low Power Mode off, under `caffeinate`, with the power log checked for the run window. Per-run sample spread (max/min − 1) for `pkg/ir.BenchmarkIRCompile [bytecode]`:

| run | interference in window | median | spread |
|---|---|---|---|
| 3 | none | 15.8 ms | 8% |
| 4 | Chrome video wake lock | 20.1 ms | 49% |
| 5 | `mobileassetd` at 94% CPU | 18.0 ms | 34% |
| 6 | none | 17.3 ms | 19% |
| 7 | Chrome video wake lock | 19.1 ms | 24% |

Two runs with no detectable interference differ by 10% on the median, twice the 5% budget. Against the run-6 baseline, `make bench-ratchet` reported 0 regressions with the machine quiet and, ten minutes later with Chrome playing audio, `InitFromLGB` +12.3% and `IRCompile [gogen_ir]` +5.1%. The code was identical.

The consequence is that any local baseline is either loose enough to pass noisy checks (and miss real regressions of the budget's size) or tight enough to fail quiet code on a busy laptop, and choosing which capture to record becomes a judgement call, which is selection on the outcome.

## Proposal

1. **Store dispersion beside the point.** Derive a robust interval from the stored samples at capture time — median absolute deviation, or a 10th–90th percentile band — and record it with `ns_per_op` / `ratio_to_anchor`. The samples are already there; no new measurement is needed.
2. **Gate interval against interval.** `check` takes enough reps to form its own band (`-count 4` gives three usable samples, which is thin) and flags a regression when the current band sits above the baseline band by more than the budget, not when two medians differ by it.
3. **Report a band too wide to gate.** When a baseline entry's own spread exceeds the budget, say so in the report instead of gating at a precision the data does not have.
[...]

PR #900, opened 2026-09-18
fix(bench-ratchet): compare deterministic metrics against the newest row, not the global min

nnunley:
## Why

`bench-ratchet check` gates allocs/op and bytes/op against the global minimum across every machine profile in `docs/perf/baseline.json`. Those profiles are captured at different commits — a tier keeps its last numbers indefinitely once its runner stops reporting — so the minimum compares today's code against whichever code state happened to allocate least. `pkg/compiler.BenchmarkInitFromLGB` surfaced it: the `amd64/Intel Xeon 8370C` row holds 7,041 allocs/op, seeded from that tier's August window (identified as `f3ca5f9b`; the runner has produced nothing since), while every row captured after it holds 13,995–26,428. Current main measures 11,900 — a 15% improvement on the newest row — and the gate reported +69%. `seed-baseline` also stamped no provenance, which is why every amd64 row has `best_since_sha: null`.

## What

- Per benchmark, the deterministic reference is the entry with the newest provenance among the profiles that carry it (entry `best_since` stamp, else profile `captured_at`), ties broken by the minimum so the bar still ratchets within one commit. The regression line names the commit its reference came from.
- `seed-baseline` stamps provenance, so null cannot recur.
- `-force update` writes only its own profile. Replication into other profiles existed to serve the global minimum; under newest-row selection it would record one machine's local capture as every other tier's floor, outside the seed-window discipline.
- `docs/perf/ratchet.md` gains a section on the deterministic gate and drops the global-minimum description.

No schema change; `docs/perf/baseline.json` is untouched here (the M3 recapture is #891, stacked on this).

## Verified

Three new tests for the selection rule (multi-tier mix, same-commit tie, entry-stamp precedence), all red on the old selection; one asserts a benchmark 43% above the newest row is still reported. `TestForceRebaselineWritesOnlyItsOwnProfile` pins that profile B is untouched by a `-force` on A. `go test ./cmd/bench-ratchet/`, `go vet`, `gofmt`, the docs frontmatter check and `scripts/lint.lg` pass. Against the unmodified baseline on an Apple M3, `check` reported `InitFromLGB` +69.0% allocs / +21.9% bytes before and does not report it after; `IRCompile [bytecode]` bytes/op stays flagged at +5.9% since `477a5d36`, because the newest row is also the lowest.

Pushed with `SKIP=bench-ratchet`: the pre-push timing gate on this machine is red against the pre-recapture M3 row (#791) and, as a point comparison run right after the full test suite, on the recaptured row too — that is #899. All other pre-push hooks passed.

Refs #791, #899.


cmd/bench-ratchet/main.go
```diff
@@ -1434,35 +1415,95 @@ func canonicalizeBenchmarks(bm map[string]BenchmarkEntry) map[string]BenchmarkEn
...
+// entryProvenance returns the commit identity and time the deterministic
+// metrics of one entry were measured at: the entry's own best_since stamp when
+// it has one, else the profile's capture identity.
+//
+// The entry stamp is the more precise of the two — within a profile,
+// ratchetMerge keeps entries whose bar was set at an older commit alongside
+// freshly measured ones, so the profile's captured_at describes the run, not
+// necessarily this number.
+func entryProvenance(prof MachineBaseline, e BenchmarkEntry) (sha, at string) {
+	if e.BestSinceAt != "" || e.BestSinceSHA != "" {
```
mparrett: `entryProvenance` treats `BestSinceAt` as when allocs/bytes were measured, but `ratchetMerge` refreshes that single stamp whenever timing, allocs, or bytes improves. A timing-only improvement therefore stamps old pinned deterministic values with the new commit. I reproduced this with a row retaining 10 allocs/100 bytes from January after a March timing improvement: it outranks a February row that actually measured 15/150, and the gate reports the January values as measured in March. This needs separate measurement/provenance for deterministic metrics, or a representation of the latest observed row that is independent of the ratcheted timing stamp.

cmd/bench-ratchet/main.go
```diff
@@ -1434,35 +1415,95 @@ func canonicalizeBenchmarks(bm map[string]BenchmarkEntry) map[string]BenchmarkEn
...
+					AllocsPerOp:  e.AllocsPerOp,
+					BytesPerOp:   e.BytesPerOp,
+					BestSinceSHA: sha,
+					BestSinceAt:  at,
+				},
+				at: at,
+			}
+			cur, ok := best[name]
+			switch {
+			case !ok || cand.at > cur.at:
```
mparrett: The documented same-commit minimum is only applied when timestamps are equal. Two profiles from the same SHA captured at different times do not tie—the later capture replaces the earlier row wholesale—while different SHAs captured in the same second would be merged. A reproduction with the same SHA at two timestamps returns 20/200 instead of the minimum 10/100. Please track the SHA in `barEntry` and tie on matching non-empty SHAs.

mparrett (changes requested):
Composite review with stacked PR #891: the new selection rule relies on provenance that does not identify when the stored deterministic metrics were measured, and its same-commit tie-break compares timestamps rather than commits. Both cases can select or loosen the wrong allocation/byte bar. Focused regression tests reproduce both failures.

nnunley:
Both points taken; fixed in `d7c1e119` on top of the reviewed commit, each with a test that reproduces your case first.

1. **Deterministic provenance is its own stamp.** `BenchmarkEntry` gains `allocs_bytes_since_sha` / `allocs_bytes_since_at`, set only when allocs/bytes are (re)recorded: on capture, on seed, and in `ratchetMerge` only when the deterministic values themselves change. `entryProvenance` reads that stamp first. For an entry that has never been ratcheted — no `best_since_*` at all — its values were written by the seed or capture in the run the profile records, so the profile's `captured_at_sha`/`captured_at` is their exact provenance and is used. An entry with `best_since_*` but no deterministic stamp is unknown and sorts oldest, which is your January/March case; that test fails on the previous commit and passes now.
2. **Ties are by SHA.** `barEntry` carries the SHA; equal non-empty SHAs tie and take the minimum regardless of timestamp (your same-SHA-two-timestamps case now returns 10/100); distinct SHAs rank by timestamp, with the SHA breaking a same-second tie rather than merging.

Also here, because your #891 comment is a tool defect rather than a capture one: **`-force` no longer accepts deterministic regressions.** It keeps any allocs/bytes it measured worse, with their stamp, and prints them as `NOT ACCEPTED`; a new `-accept-deterministic` records them deliberately. `docs/perf/ratchet.md` documents the schema fields, the selection rule and the `-force` change. 39 tests in `cmd/bench-ratchet` pass; vet, gofmt, frontmatter and `scripts/lint.lg` clean. `check` against the committed baseline on this M3 reports exactly one deterministic line, `IRCompile [bytecode]` bytes +5.9% since `477a5d36`, and no longer reports `InitFromLGB`.

cmd/bench-ratchet/main.go
```diff
@@ -570,6 +633,16 @@ func ratchetMerge(existing, current MachineBaseline) (MachineBaseline, RatchetSu
 			BytesPerOp:    minI(cur.BytesPerOp, base.BytesPerOp),
 			RatioToAnchor: minF(cur.RatioToAnchor, base.RatioToAnchor),
 		}
+		// The deterministic pair carries its own stamp, moved only by a run that
+		// actually lowers one of the two. Sharing the timing stamp would date
+		// untouched allocs/bytes to a commit that only ran faster.
+		if merged.AllocsPerOp != base.AllocsPerOp || merged.BytesPerOp != base.BytesPerOp {
```
mparrett: One stamp cannot truthfully date both values after a mixed change. If March improves allocs 10→9 but regresses bytes 100→200, this stores the synthetic pair 9/100 and stamps both as March, resurrecting January's bytes as a current floor; a focused test then selects 100 over a February row that actually measured 150. The inverse occurs in stacked #891: improved allocs 105,814/143,839 have no deterministic stamp because bytes were retained, so the gate still selects 113,777/151,840 from the AMD row. A future regression to 110,000 allocs would therefore pass despite exceeding the M3 improvement by 4%. This needs separate alloc and byte provenance/selection, or storage of only pairs actually measured together.

cmd/bench-ratchet/main.go
```diff
@@ -1434,35 +1509,121 @@ func canonicalizeBenchmarks(bm map[string]BenchmarkEntry) map[string]BenchmarkEn
...
+					BestSinceAt:  at,
+				},
+				sha: sha,
+				at:  at,
+			}
+			cur, ok := best[name]
+			switch {
+			case !ok:
+				best[name] = cand
+			case cand.sha == cur.sha:
```
mparrett: When same-SHA rows are merged, the metrics take their minimum but `cur.at` remains the timestamp of whichever profile key was encountered first. A later distinct SHA is then ranked against that arbitrary timestamp. A focused case with same-SHA captures in January and March plus another SHA in February incorrectly selects the February row. Please update the merged group's timestamp/provenance—normally to the maximum same-SHA timestamp—before comparing it with other commits.

mparrett (changes requested):
Re-review after d7c1e119: the original timing-only provenance and direct same-SHA reproductions are fixed, and the new force behavior correctly rejects deterministic regressions. Two edge cases remain. Most importantly, one stamp for the alloc/byte pair still misdates a pinned metric whenever the other improves; this is already visible in stacked PR #891 and can cause the gate to miss later allocation regressions. Focused regression tests reproduce both findings below.

mparrett (changes requested):
Re-review of current head d7c1e119: the two remaining blockers still reproduce unchanged.

1. Mixed deterministic updates still assign one provenance stamp to values measured at different commits. An alloc improvement paired with retained older bytes creates a synthetic pair and can hide a later regression. This is observable in the stacked #891 data. See https://github.com/nooga/let-go/pull/900#discussion_r4050670134.
2. Same-SHA rows take minimum metrics but retain whichever timestamp is encountered first, so comparison with another commit can select the wrong reference. See https://github.com/nooga/let-go/pull/900#discussion_r4050670150.

I also reread #891. The trailing author comment addresses the earlier force-rebaseline issue, not these later findings. Its capture remains based on 3bbde90a while current main at ebd758a changes pkg/rt/core_compiled.lgb.

The clean package tests pass, but focused reproductions for both selection cases fail on this head.

nnunley:
Both fixed in `d98e9aa1`, each reproduced by a failing test first.

1. **Allocs and bytes carry separate provenance and are selected separately.** `allocs_bytes_since_*` is replaced by `allocs_since_sha/at` and `bytes_since_sha/at`, stamped independently wherever each metric is (re)recorded — capture, seed, `ratchetMerge` for the metric that moved, and the `-force` keep/adopt path per metric. `machineIndependentBar` picks the allocs reference and the bytes reference each from the entry with the newest provenance for that metric, with the SHA tie rule applied per metric. Your mixed case now yields bytes 150 from February and allocs 9 from March; the #891 inverse now stamps the adopted allocs, so a later 110,000 is flagged.
2. **Merged same-SHA groups take the newest member timestamp** before ranking against other commits; your January+March vs February case selects the same-SHA group.

42 tests pass; vet, gofmt, frontmatter clean. `check` against the committed baseline on this M3 still reports exactly the `IRCompile [bytecode]` bytes line since `477a5d36`. #891 is rebuilt from the same capture with this tool (`edbf9e71`): its adopted `IRCompile` allocs are now stamped and become the allocs references (105,814 / 143,839 @3bbde90a), while both bytes bars stay at the `477a5d36` row; the recapture against current main you asked for on #891 will follow once this lands. Drafted with Claude and checked by me.

cmd/bench-ratchet/main.go
```diff
@@ -570,6 +630,20 @@ func ratchetMerge(existing, current MachineBaseline) (MachineBaseline, RatchetSu
...
 			RatioToAnchor: minF(cur.RatioToAnchor, base.RatioToAnchor),
 		}
+		// Each deterministic metric carries its own stamp, moved only by a run
+		// that lowers THAT metric. Sharing one stamp would date a value this run
+		// never produced — the timing stamp dates a run that only got faster, and
+		// a pair stamp dates a kept metric to the run that moved the other one.
+		if merged.AllocsPerOp != base.AllocsPerOp {
+			merged.AllocsSinceSHA, merged.AllocsSinceAt = current.CapturedAtSHA, current.CapturedAt
+		} else {
+			merged.AllocsSinceSHA, merged.AllocsSinceAt = base.AllocsSinceSHA, base.AllocsSinceAt
```
mparrett, line 648: A run that measures allocs equal to the stored bar leaves the stamp untouched, so the measurement is recorded nowhere. On an entry with no `allocs_since_*` yet, which is every entry in the committed `docs/perf/baseline.json`, the same run sets `best_since_*` as soon as timing improves, and `unratchetedProvenance` then returns UNKNOWN instead of the profile's `captured_at`. The row is dated before the run and undated after it, and stays undated until the allocs value itself moves. Replaying `aggregate` runs over the committed baseline with 10% of entries improving per release, the retired `f3ca5f9be1da` tier supplies 13 of 157 allocation references after one release and 85 after six, and `InitFromLGB` returns to 7,041 allocs `since f3ca5f9be1da`. Stamping when `cur.AllocsPerOp <= base.AllocsPerOp`, rather than when the merged value differs, keeps all 157 on the newest release commit over those same six cycles; `TestDeterministicProvenanceSurvivesTimingOnlyImprovement` is then the only failing test, and it asserts the bar is February's 15 where March measured 10. Please take that change, or say why the February row should win.

cmd/bench-ratchet/main.go
```diff
@@ -1434,35 +1510,157 @@ func canonicalizeBenchmarks(bm map[string]BenchmarkEntry) map[string]BenchmarkEn
...
+		sha, at  string
+		selected bool
+	}
+	// consider folds one measurement into the running selection for one metric.
+	consider := func(cur metricBar, value int64, sha, at string) metricBar {
+		cand := metricBar{value: value, sha: sha, at: at, selected: true}
+		switch {
+		case !cur.selected:
+			return cand
+		case cand.sha == cur.sha:
```
mparrett, line 1616: Same-commit rows reduce to their minimum only when they are adjacent in profile order. A row from a different commit whose timestamp falls between them replaces the running selection, dropping the earlier same-commit member before its group is complete: `shaX@Jan (10)`, `shaY@Jun (50)`, `shaX@Sep (30)` selects 30 rather than the documented minimum of 10. Nothing in the current data hits it, since the six profiles carry six distinct SHAs and same-commit rows should agree on deterministic metrics anyway, so this is a gap between the documented rule and the fold rather than a wrong bar today. Collecting the per-SHA minimum before ranking groups would close it; leaving it is also defensible if you would rather not grow the function.

mparrett (changes requested):
The two findings from the previous round are fixed: allocs and bytes now carry separate stamps and are selected separately, and same-commit groups rank by their newest member. One case in the same area still lets the gate drift back to a stale row.

A stamp moves only when its metric changes, so a run that re-measures the stored value at a newer commit records nothing. On an entry that has no `allocs_since_*` yet, which is every entry in the committed `docs/perf/baseline.json`, that same run sets `best_since_*` as soon as timing improves, and `unratchetedProvenance` then reports the row as undated rather than as measured in the run the profile names. The row is dated before the run and undated after it.

Replaying `aggregate` runs against the committed baseline in a test harness, with 10% of entries improving on timing per release and allocs/bytes unchanged, the retired 8370C tier supplies 13 of 157 allocation references after one release and 85 after six. `InitFromLGB` goes back to 7,041 with `since f3ca5f9be1da`: the state #883 describes, now with an attribution line on it. I have not run a real capture, so this is the selection rule replayed over your data rather than an observed CI result.

Stamping when the run measured a value at or below the stored bar, a confirmation rather than only an improvement, holds all 157 references on the newest release commit across the same six cycles. It fails one existing test, `TestDeterministicProvenanceSurvivesTimingOnlyImprovement`, which asserts the bar is February's 15 where profile `a` measured 10 in March.

That is the one blocker. Take the confirmation stamp, or say why February's bar is the one you want there. The second comment is low and does not block.

mparrett (commented):
Re-review at `d98e9aa1`. My previous pass was against `d7c1e119` and predates two of the commits here.

Both of my earlier blockers are addressed. Deterministic metrics now carry their own `allocs_since_*` / `bytes_since_*` stamps, resolved separately by `allocsProvenance` / `bytesProvenance` and selected through independent maps in `machineIndependentBar`; same-commit rows tie on the SHA, and the merged group takes its newest timestamp before ranking. `TestDeterministicProvenanceIsPerMetric`, `TestMachineIndependentBarTiesOnSHANotTimestamp`, and `TestMachineIndependentBarMergedGroupTakesNewestTimestamp` pin all three. The #891 case I raised resolves with them: the improved allocs are stamped, and the bar selects 105,814 / 143,839 rather than the AMD row's 113,777 / 151,840.

Four things below, none of them merge blockers. The first decides whether the new reference is trustworthy for the profiles that currently supply it, so it is the one I would want settled before this lands.

**A seeded value is a median across up to five commits, stamped with one — `cmd/bench-ratchet/seed.go:285`**

`reduceBenchmarks` takes `medianOf` over the seed window, which is five snapshots by default and therefore up to five different commits, and `seedOneMachine` sets `CapturedAtSHA` from the newest surviving snapshot. This PR's new `stampAll(&out)` writes that identity onto every entry's `allocs_since_*` / `bytes_since_*`.

Under the global minimum the imprecision was inert, because nothing read the date. Selecting on provenance makes it load-bearing: a windowed median now outranks a single-commit measurement, and the gate attributes it to a commit that measured something else.

Over the five most recent `perf-data` snapshots for `amd64/AMD EPYC 7763` (`3dc58413` through `ebd758a6`), 10 of the 163 benchmarks present in all five have a median that differs from the newest snapshot:

```
IRCompile [bytecode]    105815 105815 105815 105815 105408  -> stores 105815, stamps ebd758a6
IRCompile [gogen_ir]    143908 143908 143908 143908 143621  -> stores 143908, stamps ebd758a6
InitFromLGB [bytecode]   11900  11900  11900  11900  11902  -> stores 11900,  stamps ebd758a6
```

`arm64/Apple M1 (Virtual)` shows the same 6% over its own window. Stamping the window's oldest snapshot would keep a windowed median from outranking a point measurement; carrying the window span and ranking by its oldest member would do the same while staying legible in the file.

**An old tag's capture is dated by the clock, not by its commit — `cmd/bench-ratchet/main.go:1224`**

`-sha` overrides `CapturedAtSHA`, but `CapturedAt` stays `time.Now()`, and `machineIndependentBar` ranks on the timestamp. A capture taken at an old commit and aggregated today becomes the newest row for every metric it lowers, and the gate then measures current code against the older state — the reduction this PR removes, re-entering through the clock.

A focused case at this head: a row at 100 allocs stamped a week ago, then an aggregate of an old-tag capture at 50 with `-sha` pointing at that tag. The stored entry reads `allocs_since_sha: <old tag>` with today's `allocs_since_at`, the bar becomes 50, and a current run at its true 100 reports +100.0% against a commit from an old release.

This is latent rather than live. `perf-release-baseline.yml` is the path that would do it — its dispatch input is "Release tag to freeze a baseline for (e.g. v1.9.0)" and the tighten step comments that it matters for old tags — but it has never run, and `docs/perf/historical/` still holds only `v1.8.0.json`. On the automatic `v*` trigger the tag is current, so date and commit agree. The combination is in use by hand, though: #891's own capture is pinned to `3bbde90a` via `-sha` while its `captured_at` is the aggregation time. Deriving `CapturedAt` from the provenance commit when `-sha` is passed would close it; the workflow already has the tag.

**The same-commit tie also fires on two unknown provenances — `cmd/bench-ratchet/main.go`, `consider`**

`case cand.sha == cur.sha` compares equal empty strings, so two rows whose provenance is UNKNOWN reduce to their minimum as though they named one commit. That is the cross-profile minimum this PR removes, surviving in the residual case. I asked earlier for a tie on matching *non-empty* SHAs; the SHA is tracked in the bar entry now, which was the other half of it.

Narrow today, since unknown provenance needs a row carrying `best_since` with no metric stamp and only the M3 profile has those. It opens when a second profile acquires them, which a `-force` keep does: the kept branch writes back whatever `bytesProvenance` returned, unknown included.

**`-force` help still describes the removed behavior — `cmd/bench-ratchet/main.go:180`**

"replace current-machine timing and rebase global deterministic metrics" — the rebase into other profiles is what this PR deletes, and this is the help for the flag whose meaning changed.

This also makes the wiki's `concepts/perf-ratchet.md` § Deterministic bars stale, which still describes the bar as the global minimum across every profile.

`go test ./cmd/bench-ratchet/ ./pkg/perfdata/` is green at `d98e9aa1` in a detached worktree. The two focused cases above were throwaway tests in that tree, removed after. I did not run the full suite, `scripts/lint.lg`, or any real capture. Nothing covers the empty-SHA tie or the clock-versus-commit ranking, which is where both cases sit.

I am leaving the request-changes in place for the seed stamp alone; the other three are fine to take as follow-ups if you would rather land this now.

mparrett:
Two reviews from me landed within a minute of each other, drafted by different agents working the same PR. Reconciling them.

The 00:21 review holds the blocker. I checked its central claim: `merged.AllocsPerOp != base.AllocsPerOp` is false when a run re-measures the stored value, so `ratchetMerge` keeps the old stamps, while a timing improvement in the same run sets `best_since_*` — and `unratchetedProvenance` then returns UNKNOWN for a row its profile dated a moment earlier. The confirmation stamp it asks for is the fix I would take.

That also corrects my own comment. I called the empty-SHA tie in `consider` narrow, on the grounds that only the M3 profile carries ratcheted rows with no metric stamp. The 00:21 replay shows the opposite: every timing-only improvement mints another undated row, 85 of 157 references after six cycles. It belongs to that defect rather than standing as a separate narrow one, and the fold-order case at `main.go:1608` is the better-evidenced statement of the same function's problem.

The other three in my comment stand as follow-ups: the seed median and the stamp it carries, `-sha` leaving `CapturedAt` on the wall clock, and the stale `-force` help text.

mparrett (changes requested):
Consolidated re-review of current head d98e9aa1, considering stacked #891 at edbf9e71. Please treat this as my go/no-go assessment in place of the two overlapping reviews and subsequent reconciliation comment.

**NO-GO for #900 as-is.** The previous mixed-metric provenance and same-SHA timestamp findings are fixed. One merge blocker remains: ratchetMerge updates allocs_since_* / bytes_since_* only when the stored minimum changes. A run that actually remeasures the same allocs or bytes on newer code keeps the old stamp. For legacy entries without a metric stamp, a timing improvement in that run sets best_since_* and makes the metric provenance UNKNOWN; the row can then lose to an older, worse profile. A focused case on this exact head remeasures 10 allocs / 100 bytes in March, but the gate selects 15 / 150 from February. The clean package tests pass; the focused reproduction fails. This is the case in https://github.com/nooga/let-go/pull/900#discussion_r4051603776.

Please stamp each metric when the current run confirms its stored value (equal or better), and retain the old stamp only when that metric was measured worse. The existing timing-only test expects February to win despite March measuring the better value, so it needs to change with the fix.

The separate seed-window, historical -sha timestamp, SHA grouping/format, and help-text concerns are follow-ups, not additional reasons for this no-go. #891 remains a separate no-go until its agreed recapture against current main after #900 lands. I am not adding duplicate inline threads.

nnunley:
@mparrett — head `8d8e2a38`, fixing the blocker from your consolidated review.

`ratchetMerge` now stamps each deterministic metric whenever the current run **confirms** it — `cur.AllocsPerOp <= base.AllocsPerOp`, and likewise bytes, judged independently against what the run itself measured rather than against the already-minimized merged value. A run that measures a metric worse leaves that metric's value and stamp alone.

Tests, written first and red before the change:
- `TestDeterministicProvenanceLegacyRowConfirmedByNewerRun` — your case: February row 15/150, legacy row 10/100 with no metric stamp, a March run remeasures 10/100 and improves a timing; the gate now selects 10/100.
- `TestDeterministicProvenanceSurvivesTimingOnlyImprovement` — expectation flipped, as you said it must be: it was asserting February's 15/150 over March's confirmed 10/100.
- `TestRatchetMergeWorseMeasurementKeepsOldStampPerMetric` — a worse allocs measurement keeps the old value and stamp while confirmed bytes restamp.

`docs/perf/ratchet.md` describes the stamp as "most recently confirmed". `baseline.json` untouched. `go test ./cmd/bench-ratchet/... ./pkg/perfdata/...` 44 passed, vet and gofmt clean.

The branch is also squashed to one commit and rebased onto current main, so a range-diff against `d98e9aa1` won't line up; the net change from that head is `main.go` (the two conditions plus comment), `main_test.go`, and `ratchet.md`. Seed-window, historical `-sha` timestamp, SHA grouping and help text are left as the follow-ups you named. #891 gets recaptured against main once this lands.

mparrett:
Follow-up on current head `8d8e2a38` (confirmed unchanged before posting): the equal-measurement fix is in `ratchetMerge`, but the `-force` path still reverses it. At `cmd/bench-ratchet/main.go:528-540`, `stampAll(&current)` initially gives this run a stamp, then the equal-value branches replace it with the previous stamp.

Focused reproduction: a legacy M3 row holds 10 allocs / 100 bytes with no per-metric stamps; a March `-force` timing recapture measures the same 10/100. Its provenance remains unknown, so the deterministic gate selects a worse February row at 15/150. `TestRereviewForcedEqualMeasurementConfirmsProvenance` fails on this head with `bar = 15/150 since febsha; want March-confirmed 10/100`.

Could `-force` treat equality as confirmation, retaining the current per-metric stamp, and preserve the old value/stamp only when that metric measures worse? That would match the new `ratchetMerge` rule and documentation. This is about the timing-recapture workflow used by #891, not a claim that #891's current three changed rows already exhibit this case.

nnunley:
@mparrett — head `75bf7707` (still one commit). You're right, `-force` undid it: `stampAll(&current)` gave the run its stamp and the two `else if cur.X == entry.X` branches in `forceRebaseline` then put the previous one back.

Those two branches are gone. On the forced path a metric now keeps the stored value and its date only when it measured **worse**; equal or better carries this run's stamp, which is the same confirmation rule `ratchetMerge` applies.

Test written first and red on `8d8e2a38` with `provenance = ""/"", want marsha/marsha`: `TestForceRebaselineEqualMeasurementConfirmsProvenance` is your case — a legacy M3 row at 10/100 with no per-metric stamps, a March `-force` recapture measuring the same 10/100, a February profile at 15/150 — and asserts both the row's stamps and that `machineIndependentBar` selects the March-confirmed 10/100. It also carries a mixed row in the same forced run (allocs measured worse, bytes equal): allocs keep `jansha` and are the only rejection reported, bytes restamp to March.

The existing force tests pass unchanged, including `TestForceRebaselineKeepsDeterministicUnlessImproved` and `TestForceRebaselineRecordsProvenanceOfKeptValues`. `docs/perf/ratchet.md`'s `-force` section says equal counts as confirmation. `go test ./cmd/bench-ratchet/...`, vet and gofmt clean; `baseline.json` untouched.

Understood that this is about the recapture workflow #891 will use rather than its current three rows; #891 gets recaptured against main with this rule once #900 lands.

---
Delivery verified by local `pr-claim-gate`: exact PR/validated head `75bf770701b3e373ab4ecec00714d8055d01d166`; relevance claims `cmd/bench-ratchet/main.go` occur in both the delivered delta and the PR feature delta; receipt `/Users/ndn/.local/state/pr-claim-gate/nooga-let-go/pr-900.json`.

mparrett (approved):
Re-review of head 75bf7707: approved for #900. The prior blockers are resolved. Alloc and byte provenance are selected independently; equal measurements now confirm the current commit in both ratchetMerge and the -force recapture path, while worse measurements retain their stored value and date. The focused reproductions that failed on the earlier heads now pass, along with the package tests and vet. All reported checks on this head have completed without failures.

This approval is for the ratchet tool change only. The seed-window, historical -sha timestamp, SHA grouping/format, and help-text points discussed earlier remain non-blocking follow-ups. It does not approve stacked #891, which still needs to be rebuilt and recaptured against current main after #900 lands.

PR #901, opened 2026-09-18
host: Thread/currentThread over scopes; core: format honors %n

abogoyavlensky:
## Why

Running `weavejester/ragtime` 0.11.0 (`core/`) under let-go found two gaps:

1. `ragtime.core` does not load. `migrate-all` checks
   `(.isInterrupted (Thread/currentThread))` between migrations and throws
   `InterruptedException`; `Thread/currentThread` is unresolved, so the whole
   namespace fails to compile.
2. `format` passes Java's `%n` straight to Go's `Sprintf`, which has no such
   verb: `(format "Applying %s%n" "x")` is `"Applying x%!n(MISSING)"`.

### `Thread/currentThread` is the current scope (`pkg/rt/host_thread.go`)

`Thread/currentThread` returns the calling scope dressed as a
`java.lang.Thread`, so the cooperative-cancellation idiom resolves and means
scope cancellation. let-go already had every piece but the object:
`InterruptedException` is a registered exception class, `scope-cancelled?`
reads a scope's context, `Scope.Cancel()` exists.

| Method | Behaviour |
|---|---|
| `.isInterrupted` | `scope.Context().Err() != nil` (the same test `scope-cancelled?` makes) |
| `.interrupt` | `scope.Cancel()`; returns `nil` |
| anything else | `java.lang.Thread .<name> is not supported under let-go` |

`Thread.` stays a loud stub: let-go cannot construct threads.

The one decision worth a look: `.interrupt` really cancels the current scope.
Inside `with-scope` that is bounded to the scope. At the root it cancels every
tracked goroutine's context for the rest of the process (`Cancel` does not
reinstall a generation the way `CancelAll` does, which is what keeps
`.isInterrupted` true afterwards). That is what interrupting the main thread
means, and the docs say so. If this proves controversial, the fallback is a
loud stub for `.interrupt` only.

### `format` honors `%n` (`pkg/rt/lang.go`)

A single rewrite pass turns `%n` into `\n` before both the argument scan and
`Sprintf` see the string, so `%n` never consumes an argument. `%%` is copied
as-is, so `%%n` stays a literal `%n`. Java's `%n` is the platform line
separator; let-go emits `\n` on every platform, same as `println`.

`generated.manifest` / `generated.sums` are refreshed (a Go edit under
`pkg/rt` moves the generator-input digest; no `.lg` changed, so the bundle
is byte-identical).

### Docs

Two bullets under "Behavioral differences" in
`docs/guide/clojure-compatibility.md`.


pkg/rt/host_thread.go
```diff
@@ -0,0 +1,74 @@
...
+			return vm.Boolean(h.scope.Context().Err() != nil), nil
+		}
+	case "interrupt":
+		if len(args) == 0 {
+			// Cancel is terminal for the scope (it does not install a fresh
+			// context generation the way CancelAll does), which is what keeps
+			// isInterrupted true afterwards. Inside with-scope that is bounded;
+			// at the root scope it cancels every tracked goroutine's context for
+			// the rest of the process — what interrupting the main thread means.
+			h.scope.Cancel()
```
mparrett, line 61: **[P1] Avoid permanently cancelling the process-wide root scope.** When `currentThread` is called outside `with-scope`, this is `vm.Goroutines`; `Cancel()` is terminal, so one library call cancels unrelated tracked goroutines and makes all later sleeps/channel operations observe a cancelled context for the rest of a long-lived or embedded process. Java `Thread.interrupt()` only marks the target thread and does not poison all current and future process work. Please keep root `.interrupt` unsupported or model the interrupt flag without terminally cancelling the global scope.

pkg/rt/lang.go
```diff
@@ -6739,7 +6767,7 @@ func CoreFormatf(vs ...vm.Value) (vm.Value, error) {
 	if !ok {
 		return vm.NIL, fmt.Errorf("format expected String")
 	}
-	fmts := string(fmtStr)
+	fmts := formatNewlines(string(fmtStr))
 	args := make([]any, len(vs)-1)
```
mparrett, line 6771: **[P2] Do not forward unused arguments after removing `%n`.** `args` retains `len(vs)-1`, but the scan leaves entries corresponding to unused arguments as `nil`; Go then appends its EXTRA diagnostic. For example, `(format "%n" 42)` currently returns `"\n%!(EXTRA <nil>)"`, whereas Java/Clojure ignores extra arguments. Please pass only the consumed arguments (or otherwise omit unused values) and add this case to the tests.

mparrett (changes requested):
Two correctness issues to address before merging: root Thread.interrupt has process-wide terminal effects, and the %n path still exposes Go's EXTRA diagnostic when unused arguments are supplied. Focused thread/format tests otherwise pass.

mparrett (approved):
Re-reviewed fe6402e. Both requested changes are addressed: root-scope Thread.interrupt is refused without poisoning vm.Goroutines, and format now omits surplus arguments with regression coverage. Focused test/host-thread and builtins tests pass locally, as do pkg/rt and pkg/vm. No further findings.

PR #902, opened 2026-09-19
fix(aot): stop entry-frame binaries from running -main twice

### LINKED ISSUE #796: entry-frame binaries run -main twice: main-chunk replay, then the frame

mparrett:
## Problem

An `lg-compile --entry-frame` binary can run a guarded top-level `-main` twice: once while replaying bytecode and once from the native frame. The earlier fix avoided that by setting `*compiling-aot*` during replay, but that also skipped guarded runtime initialization in required namespaces, as @nooga found.

## Fix

`lg -c` now accepts `-entry-frame-entry namespace/-main` (or `namespace/main`) when producing bytecode for a native entry frame. It omits only the selected top-level entry invocation from that bundle. It preserves surrounding top-level forms and required namespace initialization. Plain `lg -c` retains its existing behavior.

The generated frame uses ordinary `LoadProgramNamespaces` and `RunProgramMainChunk` again, so `*compiling-aot*` stays false at runtime. The native frame invokes the selected entry once after replay. The example build and native-entry tests now use the new compile option.

## Verification

- A real two-namespace binary prints `libx: runtime init ran`, `app: runtime init ran`, and `42`, each exactly once.
- Native-entry AST fixtures and matrix pass.
- Entry-frame language tests: 17 tests, 139 assertions.
- `go test ./pkg/compiler ./pkg/cli ./pkg/rt` passes; `make check-generated` passes.

Fixes #796.


mparrett:
The first fix only covered single-file programs. `f13aca37` covers the rest.

`lg -c` emits an empty NS table only when the program is one file. Add a `:require` and the bundle carries an NS table in which `MainChunk` *is* one of the `NSOrder` chunks, so the top level, and the guard with it, runs inside `LoadProgramNamespaces` — `RunProgramMainChunkForEntryFrame` then no-ops by design. Bracketing the replay alone missed that path, and a two-namespace program built with `--entry-frame` still printed its line twice:

```
NSOrder=[depns mainns]
  mainns: chunk==MainChunk? true
```

The frame now calls `rt.LoadProgramNamespacesForEntryFrame` as well, so both halves of the prologue hold `*compiling-aot*`, with the previous root restored before the entry call. The same two-namespace program now prints once.

`guarded_entry` cannot reach this shape: fixtures in `test/native-entry/` are a single file, so their NS table is always empty. The namespace-load half is pinned by `TestLoadProgramNamespacesForEntryFrameBracketsCompilingAOT` in `pkg/rt`, which runs in both stdlib lanes. Reverting either bracket fails its own test.

mparrett:
Review found that the original guard covered only the main-chunk replay; in multi-namespace bundles, the guarded call runs during namespace loading. The updated PR brackets both phases, and I pushed `b8fccf3c` to correct the `*compiling-aot*` documentation and refreshed the PR description. Focused Go tests in both stdlib lanes, the entry-frame language test, and native-entry fixtures pass.

nooga:
Thanks — the mechanism works and the tests are real falsifiers; I reverted each bracket separately and each one fails on its own.

I don't want to take the semantics change though. The replay in step 2 runs the *whole* top level, including every required namespace, so setting `*compiling-aot*` there skips every `when-not *compiling-aot*` form in the program — not just the entry call. A library that uses the idiom the guide recommends, for something that isn't `-main`, silently stops running in an AOT binary:

```clojure
;; src/libx/core.lg
(ns libx.core)
(defn helper [a] (+ a 41))
(when-not *compiling-aot*
  (println "libx: runtime init ran"))
```
```clojure
;; src/app.lg
(ns app (:require [libx.core :as lib]))
(defn -main [] (println (lib/helper 1)))
(when-not *compiling-aot* (-main))
```

| | `lg app.lg` | entry-frame binary, `cffc09e1` | entry-frame binary, `b8fccf3c` |
|---|---|---|---|
| libx init | ran | ran | **dropped** |
| `42` | ×1 | ×2 | ×1 |

So the trade isn't "doubled work vs. an unusual var read" — it's a slow binary vs. a binary that quietly skips your setup code. The second is worse: you can see a bad number in a benchmark, you can't see an init that didn't happen.

<details>
<summary>full repro script</summary>

```sh
#!/bin/sh
# Usage: LETGO=/path/to/let-go sh repro.sh   (checkout built with `make build`)
set -e
: "${LETGO:?set LETGO to a let-go checkout built with 'make build'}"
LETGO=$(cd "$LETGO" && pwd)
WORK=$(mktemp -d); cd "$WORK"; mkdir -p src/libx out

cat > src/libx/core.lg <<'EOF'
(ns libx.core)
(defn helper [a] (+ a 41))
(when-not *compiling-aot*
  (println "libx: runtime init ran"))
EOF

cat > src/app.lg <<'EOF'
(ns app (:require [libx.core :as lib]))
(defn -main [] (println (lib/helper 1)))
(when-not *compiling-aot* (-main))
EOF

echo "=== on the VM ==="
(cd src && LG_SOURCE_PATHS=. "$LETGO/bin/lg" app.lg)

LG_SOURCE_PATHS="$WORK/src:$LETGO/pkg/rt/core" \
  "$LETGO/bin/lg" "$LETGO/scripts/lg-compile" --entry-frame "$WORK/out" aotexp \
  "$WORK/src/libx/core.lg" "$WORK/src/app.lg" >/dev/null
(cd src && LG_SOURCE_PATHS=. "$LETGO/bin/lg" -c "$WORK/out/program.lgb" app.lg)
printf 'module aotexp\n\ngo 1.23\n\nrequire github.com/nooga/let-go v0.0.0\n\nreplace github.com/nooga/let-go => %s\n' "$LETGO" > out/go.mod
(cd out && go mod tidy >/dev/null 2>&1 && go build -o "$WORK/app.native" .)

echo "=== entry-frame binary ==="
"$WORK/app.native"
```
</details>

Two things before this lands:

1. **Suppress the entry call, not the whole top level.** You're right that the bytecode doesn't record which form is the entry call — but `lg -c` is looking at the source when it builds the bundle, so it could leave that form out when the consumer owns the entry. Then the fact lives in the artifact, `*compiling-aot*` keeps meaning one thing, and nothing else gets skipped.
2. **A test that checks what a two-namespace binary prints.** `guarded_entry` can't reach this shape, since fixtures under `test/native-entry/` are single-file. The unit test asserts the var is set, which is the mechanism rather than the behavior — that's why this got past a green run.

mparrett:
@nooga Thanks for the clear repro. You were right: setting `*compiling-aot*` during replay skipped library runtime initialization. I changed frame-specific `lg -c` compilation to omit only the selected top-level entry call, and restored ordinary replay semantics. A real two-namespace binary test now checks that both initializers and `42` each print once; the native-entry gates and `make check-generated` pass. Would you kindly take another look? Thanks!

nooga (approved):
Reviewed and tested. The rebase's fix (compiler.go's omitTopLevelEntryCall + the -entry-frame-entry CLI flag) properly addresses the *compiling-aot* semantics issue raised earlier: it now syntactically strips only the selected top-level entry call from the compiled bundle instead of setting *compiling-aot* during replay, so required-namespace runtime init still runs. Verified locally: exact two-namespace repro (TestNativeEntryNamespaceReplay) passes, check-generated/check-generated-manifest clean, and the PR's full stated verification suite (pkg/compiler, pkg/cli, pkg/rt, test/e2e) passes. One unrelated e2e failure (TestJankSuiteDirectABIGeneratedGo) reproduces identically on plain main and is not caused by this change.

Issue #867, opened 2026-09-12
Wide Int values narrow at collection, math, and interop boundaries on 32-bit hosts


mparrett:
With `Int` widened to `int64` on every host (#859), the VM's arithmetic core carries full-width values on a 32-bit host. Several consumers still convert through Go's platform `int` before using the value, so a value above `math.MaxInt32` silently wraps at the boundary rather than being rejected or carried.

Each case below is a real divergence between the same build on two hosts, not an inference from reading the code. Reproduced on `tinygo build -target=wasi` under `wasmtime`, where Go's `int` is 32 bits, against the same source run natively on arm64.

| expression | native | TinyGo wasi |
|---|---|---|
| `(get [42] 4294967296)` | `nil` | `42` |
| `(vec (take 3 (range 4294967296 4294967299)))` | `[4294967296 4294967297 4294967298]` | `[0 1 2]` |
| `(clojure.math/add-exact 1099511627776 1)` | `1099511627777` | `1` |
| `(clojure.math/round 1099511627776.0)` | `1099511627776` | `2147483647` |
| `(subvec [42] 4294967296)` | throws | `[42]` |
| `(first (repeat 4294967296 :x))` | `:x` | `nil` |

The `get` case is the one I would fix first: an out-of-range index returns a neighbouring element instead of `nil`, so it reads as a successful lookup.

## Sites

- **`pkg/vm/vector.go:373`** — `!ok || numkey < 0 || int(numkey) >= len(l)` narrows before the bounds comparison, so a wide index wraps into range and indexes the wrong element. The same shape appears in persistent vectors, strings, typed arrays, transient vectors, chunks, and sequence views.
- **`pkg/vm/numbers.go:1172`** — `ToInt` returns `(int(n), true)` for any `Int`, reporting success after wrapping. Callers that treat `ok` as "this index is usable" then proceed with a wrapped value; `subvec` and `repeat` above are two.
- **`pkg/vm/range.go:188`** — `NewRange` stores `int(start), int(end), int(step)`.
- **`pkg/rt/math.go:161`** and its siblings — the `*-exact` family, `round`, `floor-div`, and `floor-mod` return through `vm.MakeInt(int(r))`. `pkg/rt/host_jvm_statics.go:297` narrows `Math/abs` the same way.

Two more I believe are affected but did not reproduce directly:

[...]

PR #906, opened 2026-09-19
fix(rt): keep 64-bit integers across the rt boundary on 32-bit hosts

mparrett:
Refs #867.

#859 widened `vm.Int` to `int64` on every host, so the VM's arithmetic core carries full-width values on a 32-bit host. Several `pkg/rt` functions still convert through Go's platform `int` on the way back out, so a value above `math.MaxInt32` wraps as it crosses the boundary — with no error.

#867 catalogues this class. This PR takes the leaf conversions in `pkg/rt`: the ones where an unbounded 64-bit value is narrowed on its way into a return. It leaves the `pkg/vm` half alone, because #867 is right that the `ToInt` question should be settled before `get`, `subvec`, `range` and `repeat` are touched, and that decision shapes their call sites.

## What was wrong

Measured on `tinygo build -target=wasi` under wasmtime and on a TinyGo browser bundle, against the same program run natively as the oracle:

| expression | native | TinyGo wasm |
|---|---|---|
| `(Math/abs -4000000000)` | `4000000000` | `-294967296` |
| `(hash "let-go release probe")` | `4010233256` | `-284734040` |
| `(math/add-exact 4000000000 1)` | `4000000001` | `-294967295` |
| `(math/increment-exact 4000000000)` | `4000000001` | `-294967295` |
| `(math/negate-exact 4000000000)` | `-4000000000` | `294967296` |
| `(math/floor-div 8000000000 2)` | `4000000000` | `-294967296` |
| `(math/round 4000000000.7)` | `4000000001` | `2147483647` |
| `(parse-long "9007199254740993")` | the value | `nil` |
| `(Long/parseLong "9007199254740993")` | the value | range error |

**`Math/abs` returns a negative number.** No range check precedes the narrowing, so the sign survives the wrap.

**`core/hash` and `Util/hash` disagree inside the same binary**, on the same input. They are the same function; `Util/hash` already returns through `vm.Int(int64(...))` and is correct, `clojure.core/hash` narrows. The two values are the same 32 bits — one unsigned, one sign-extended — so anything that persists, transmits or compares a hash across a stdgo and a TinyGo build sees them disagree.

**`parse-long` returns `nil`, not an error.** `nil` is how it reports unparseable input, so a valid 64-bit literal comes back indistinguishable from malformed input. `Long/parseLong` at least raises.

The `*-exact` family exists to signal overflow rather than wrap. On a 32-bit host it is the family that wraps silently.

## The change

`vm.MakeInt64` already existed in `pkg/vm/numcache.go`. The nine `math.go` sites and `Math/abs` use it instead of `vm.MakeInt(int(...))`; the two `strconv.Atoi` calls become `strconv.ParseInt(s, 10, 64)`.

`Integer/parseInt` now has a separate `strconv.ParseInt(s, 10, 32)` path. It previously shared the `Long/parseLong` helper; widening that shared helper would let `Integer/parseInt` accept values outside the signed 32-bit range.

The `Long/parseLong` comment previously justified `Atoi` as avoiding a CodeQL "incorrect conversion" finding. `ParseInt(..., 64)` into `MakeInt64` has no narrowing conversion at all, so that constraint is satisfied rather than traded away; the comment now records the width reason instead.

`pkg/rt/generated.manifest` and `generated.sums` move because the three edited files are generator inputs.

## What this deliberately does not touch

- **`pkg/vm`'s `ToInt`, `get`, `range`** — per #867, the `ToInt` decision comes first.
- **The `int` coercion paths** in `native_prims.go` and `lang.go` — these range-check against the int32 bounds *before* narrowing, so they are correct as written. Clojure's `int` is supposed to refuse a value that does not fit.
- **`golower_runtime.go:401`** — #867 lists it as suspected and unreproduced, and exercising it needs a `-tags gogen_ir` build on a 32-bit host. Not claimed here.

## Verification

A 69-case differential corpus covering integer arithmetic and shifts, hashing, floats, ordering, records, protocols, multimethods, transients, macros, channels and exceptions, compiled once to bytecode and run on native, `GOOS=wasip1` and TinyGo wasi.

Before: three divergences between the stdgo and TinyGo runtimes. After: none. The comparison is against a stdgo `lg-runtime` rather than the full `lg` CLI, so that the two sides differ only in toolchain and not in which namespaces the binary carries.

Also checked: `go test -short ./pkg/...` clean, `check-generated` OK, and `GOOS=js`, `GOOS=wasip1` and `GOOS=linux` all build. TinyGo native-linux produces the corrected values too.

Follow-up review: `go test -short ./test -run TestJVMStatics -count=1` covers the `Integer/parseInt` bounds and `Long/parseLong` wide values; `make check-generated` passes after the parser split.

## Suggested follow-up

A `linux/386` test lane would catch this whole class without a TinyGo toolchain, as #867 notes. I left it out of this PR as a separate CI decision, and am happy to add it here instead if you would rather it ship together.


mparrett:
Reviewed the updated #906 diff. The full-width return conversions match the PR's 32-bit host goal.

I found that `Integer/parseInt` reused the widened `Long/parseLong` parser, so it could accept values outside the signed 32-bit range. Addressed in eb5156f2: `Integer/parseInt` now has a 32-bit parser, while `Long/parseLong` keeps 64-bit parsing. I added tests at both boundaries and verified them with `go test -short ./test -run TestJVMStatics -count=1` and `make check-generated`.

The separate, pre-existing `math/multiply-exact` overflow case for `-1 * Long.MIN_VALUE` remains outside this PR's width-conversion fix.

nooga (approved):
Reviewed: narrow, well-tested boundary fixes (Math/abs, hash, parse-long, the *-exact family, Integer/parseInt split). Scope matches the PR description — pkg/vm's ToInt/get/range correctly left for #867. Generated artifacts refreshed, CI green, no findings.

Issue #547, opened 2026-07-17
TinyGo native: upstream the Darwin link fix (libc syscall(2) shim + term sysdep seam)


mparrett:
Part of the TinyGo lane (#504).

#504 lists "native TinyGo doesn't link on Darwin" as an open gap. A local spike closed it: native TinyGo on `darwin/arm64` now links and runs (`lg` ~4.3 MB, xsofy plays natively). This issue tracks turning that into a reviewable, build-tag-gated PR.

## The blocker / proposal

TinyGo can't compile the `.s` trampolines behind `syscall.Syscall` and x/sys' darwin libSystem dispatch (tinygo-org/tinygo#4794), so native macOS died at link on three fronts: `chzyer/readline`'s `Syscall6` ioctls, `x/sys/unix`'s `syscall_syscall`, and `term.go`'s direct `SYS_IOCTL`.

The spike:
- Routes `syscall.Syscall`/`Syscall6` through libSystem's variadic `syscall(2)` (present in TinyGo's macos-minimal-sdk stub list) via a CGo gateway (`syscall_tinygo_darwin.go`, errno via `__error()`).
- x/sys' per-function trampoline address vars can't be `go:linkname`-overridden (symbol multiply defined), so the x/term and `unix.Poll` touchpoints move behind a terminal sysdep seam: `term_sysdep_default.go` keeps today's x/term path for stock builds, `term_sysdep_tinygo_darwin.go` reimplements raw-mode/size/isatty/poll over raw darwin syscall numbers.

## Remaining gaps (from the spike)

- readline REPL nil-deref.
- A `reflect` `.WriteString` wall.
- SIGWINCH untested.

## Scope

If Darwin is in scope for the TinyGo lane (preferred), turn the spike into a PR: darwin-only, build-tag-gated (no effect on stock or Linux builds), plus a link-check CI job like `wasip1-build`.

Spike branch for reference (fork): https://github.com/mparrett/let-go/tree/spike/tinygo-darwin-native (`f3cc4ab`).

Refs: #522, tinygo-org/tinygo#4794.

PR #909, opened 2026-09-19
ci: move the TinyGo gate to 0.42.0

mparrett:
`tinygo-wasi-build` installs TinyGo 0.41.1, released 2026-04-22. 0.42.0 has been out since 2026-09-01, so the gate has been validating against a toolchain one release behind for the better part of a month. A pinned version is the right call for a gate — this is a bump, not a switch to `latest`.

## What changes for us

0.42.0 carries #5401, which implements file-level `//go:linkname`. That removes one of the three link failures on native darwin:

```
0.41.1                                      0.42.0
readline term_bsd.go  _syscall.Syscall6     readline term_bsd.go  _syscall.Syscall6
x/sys    zsyscall_*   syscall_syscall       (gone)
pkg/rt/term.go        _syscall.Syscall      pkg/rt/term.go        _syscall.Syscall
```

The remaining two are tinygo-org/tinygo#4794, still open, so native macOS still does not link and `docs/guide/tinygo.md` stays accurate. #547's planned Darwin work is the thing this affects: the x/sys half of that workaround is already unnecessary on 0.42.0.

## Verification

On darwin/arm64 with 0.42.0, against the stdgo `lg-runtime` as oracle:

- `tinygo build -target=wasi ./cmd/lg-runtime` succeeds, and the module boots under wasmtime past the reflect shim — the two steps this job runs.
- A 69-case differential corpus (integer arithmetic and shifts, hashing, floats, ordering, records, protocols, multimethods, transients, macros, channels, exceptions) is identical to both the stdgo runtime and to 0.41.1. No semantic change from the LLVM 20.1.1 → 22.1.4 jump.

Two numbers worth having on record, both from a single local build rather than a controlled measurement:

- the wasi module grows about 3%, 9,379,333 → 9,676,053 bytes
- the build ran in roughly half the wall time, 113s → 47s

CI on linux/amd64 is the real check for both; I measured on darwin/arm64.

## Not included

0.43 will carry tinygo-org/tinygo#5612, which fixes the `//go:cgo_import_dynamic` trampolines that made x/sys calls segfault at runtime on darwin even once they linked. That is the other half of the #547 story and worth a look when 0.43 ships, but it is not in any release yet.


nooga (approved):
Clean, minimal TinyGo gate bump (0.41.1 -> 0.42.0) with verification data (differential corpus match, wasm size/build-time deltas). No concerns.

PR #911, opened 2026-09-19
build(bench-ratchet): wrap ratchet runs in $BENCH_WRAP, caffeinate by default on macOS

nnunley:
## Why

A machine that idle-sleeps or throttles during a capture moves the calibration anchor, and the ratchet then reports timing regressions no commit caused. `docs/perf/ratchet.md` already tells contributors to capture on a quiet machine; nothing enforced the "stays awake" half of that.

## What

Every Makefile recipe that runs `cmd/bench-ratchet` is prefixed with `$(BENCH_WRAP)`: `bench-ratchet`, `-update`, `-show`, `-full`, `-full-update`, `ratchets`, `ratchets-update` and `perf-snapshot`. It defaults to `caffeinate -is` where `caffeinate` exists and to nothing elsewhere, so Linux CI is unaffected. The pre-push gate calls `make bench-ratchet`, so it inherits the wrapper.

The name uses `_` rather than this file's usual `-` so the environment can set it; a Make variable with a hyphen cannot be exported from a shell.

```sh
make bench-ratchet                                                 # caffeinate -is (macOS default)
BENCH_WRAP= make bench-ratchet                                     # run bare
BENCH_WRAP="systemd-inhibit --what=idle:sleep" make bench-ratchet  # Linux counterpart
BENCH_WRAP="taskset -c 0-7" make bench-ratchet                     # pin cores instead
```

`systemd-inhibit` is deliberately not the automatic Linux default: CI runners and headless servers do not idle-sleep, so it would be a no-op there, and taking an inhibitor lock goes through polkit, which can prompt or be refused over SSH or in a container. A default that can break `make bench-ratchet` is worse than no default.

## Scope — Make targets only

`BENCH_WRAP` reaches the Make targets, including the pre-push gate. It does **not** reach the two-phase `go run ./cmd/bench-ratchet ... capture` invocations documented just below it in the guide, nor `scripts/recapture-release-baseline.sh`, both of which build and run the tool directly. `docs/perf/ratchet.md` now says so, and tells you to prefix those by hand on a machine that might sleep.

## Verification

`make -n` for each mode, confirming the recipe that would run:

| invocation | command |
|---|---|
| default | `caffeinate -is go run ./cmd/bench-ratchet check` |
| `BENCH_WRAP=` | `go run ./cmd/bench-ratchet check` |
| `BENCH_WRAP="nice -n 0"` | `nice -n 0 go run ./cmd/bench-ratchet check` |

`python3 scripts/docs_frontmatter_hook.py --check docs/perf/ratchet.md` passes.

This does not claim to make captures reproducible on its own — measured on an Apple M3 with a live desktop session, a five-capture window still spreads 3.3–6.3% per row against a corpus worst of 1.75%. It removes one specific, avoidable source of anchor movement.

## Review corrections (mparrett, 2026-09-20)

Both points are addressed in `0708fa20`, and both were correct:

- The earlier description claimed the guide documented `systemd-inhibit` and `taskset` for Linux. **It did not** — that was an inaccurate description of a commit that only showed `BENCH_WRAP=`. The guide now carries both examples, so the claim is true rather than removed.
- "Every local capture" overstated the scope. Corrected above and in the guide: Make targets only, with the direct `go run` path and the recapture script called out as unwrapped.

Thanks for checking that `caffeinate -is` propagates a nonzero exit and that all eight recipes carry the prefix.


mparrett (commented):
Reviewed commit `474a012`.

- **Documentation finding (nonblocking):** `docs/perf/ratchet.md` (new wrapper paragraph, around line 111) only shows `BENCH_WRAP=`, which disables the wrapper. The PR description says the guide documents `systemd-inhibit` and `taskset` for Linux users, but neither appears there. Please add the Linux override example to the guide (and the optional `taskset` guidance if intended), or correct the PR description.
- **Scope note:** `BENCH_WRAP` applies to the Make targets, including the pre-push gate. The direct `go run ... capture` example immediately below the new paragraph and `scripts/recapture-release-baseline.sh` remain unwrapped, so the PR description's “every local capture” wording is broader than the change. Please make that scope explicit in the description or guide.

I found no blocking issue in the Makefile changes. I checked that macOS `caffeinate -is` propagates a wrapped command's nonzero exit status and that all eight Makefile recipes invoking `cmd/bench-ratchet` carry the prefix.

Issue #914, opened 2026-09-19
perf/test: compile-skipped suite files are invisible to the ratchet and to CI


mparrett:
## What happened

#863 broke compilation of 47 of the 242 jank suite files (fix: #915). Nothing reported it:

- `TestClojureTestSuite` records a compile failure as `t.Skip`, so the Go lane stayed green. The only trace was the `TOTALS:` line in the `-v` log moving from `files=242 ... compile=0` to `files=195 ... compile=47`.
- `BenchmarkClojureTestSuite` and `BenchmarkClojureTestSuiteCompileAndRun` report `files`, `pass`, `fail`, `skips`, `compile_ms`, and `run_ms` as custom metrics, but the timeline snapshot keeps only `ns_per_op`, `bytes_per_op`, `allocs_per_op`, and `ratio_to_anchor`. The perf page therefore showed the 21% smaller workload as a speed-up: on the EPYC 7763 tier the bytecode suite went from 459 ms to 365 ms per pass at cffc09e1, and the B/op step it also showed (373 MB to 475 MB) understated the real one (529 MB with all files running).

The ratchet cannot tell that the workload changed, so it judged both as timing and memory changes on a fixed one.

## Proposal

Two gates, both tighten-only in the style of `docs/perf/ir-stress-baseline.edn`:

1. Record the suite benches' `files`, `pass` and `skips` in the timeline snapshot and treat them as deterministic rows. #900 already judges deterministic metrics by the newest evidence; a drop in `files` or `pass` against that evidence should fail the ratchet the way a B/op regression does, and the perf page can show the counts next to ns/op so a reader can tell a smaller workload from a faster one.
2. Give `TestClojureTestSuite` a checked-in skip baseline (`compile`, `panic`, `runtime` counts) and fail when any count exceeds it. Lowering the baseline after a fix is the normal path; raising it needs a human decision, as with the IR stress gate.

Both fit `cmd/bench-ratchet` and `test/zz_compat_test.go` as they stand. I have not started on either. I can take (1) after the release cut unless it is already on your list with the `clojure.test` work.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

mparrett:
[...]

PR #915, opened 2026-09-19
fix(test): expand one-arity `is` directly to `try-expr`

mparrett:
## Problem

#863 wrote the one-arity `is` the way Clojure does:

```clojure
(defmacro is
  ([form] `(is ~form nil))
  ([form msg] `(try-expr ~msg ~form)))
```

let-go's syntax-quote qualifies a symbol only when it already resolves at the point the macro is read. `is` is not interned yet inside its own definition, so the expansion is a bare `is`, and it only resolves in a caller namespace that happens to refer `is`. `are` expands to `test/is` with one argument, so a namespace that refers `are` but not `is` fails to compile:

```
caused by: compiling function position
caused by: Can't resolve is in this context
```

47 of the 242 files in the jank `clojure-test-suite` have exactly that ns shape (`associative_qmark`, `count`, `get`, `update`, and most of the `*_qmark` predicate files). Since #863 merged they fail to compile and are skipped. CI stayed green because `TestClojureTestSuite` reports a compile failure with `t.Skip`; the `TOTALS:` line in the `-v` log went from `files=242 ... compile=0` to `files=195 ... compile=47`.

## Effect on the perf timeline

The suite benchmarks run the same corpus, so they ran 21% fewer assertions (6311 to 4962) from cffc09e1 on. The step the perf page shows at that commit is the missing work, in both directions:

| amd64 EPYC 7763, `BenchmarkClojureTestSuite [bytecode]` | 9d32c1c4 | cffc09e1 onward |
|---|---|---|
| ns/op | 459 ms | 365 ms |
| B/op | 373 MB | 475 MB |

With this fix the benches run all 242 files again. Local A/B on an M1, bytecode mode, two interleaved rounds each, same 6311 passing assertions:

| | 9d32c1c4 (pre-#863) | main + this fix |
|---|---|---|
| `run_ms` | 244 to 281 | 204 to 233 |
| `compile_ms` | 344 to 443 | 571 to 731 |
| B/op | 373 MB | 529 MB |
| allocs/op | 7.6 M | 10.5 M |

Execution is flat to slightly faster. Compile time is roughly 60% higher and heap per suite pass 42% higher than before the port. That is the cost of the real `clojure.test` expansion, where each `is` becomes a `try`/`catch` around `assert-expr` plus a report map; it predates this PR, which only makes the timeline show it. Expect the deterministic B/op and allocs rows for the two suite benchmarks to move up against the rows recorded since #863; those rows measured a truncated suite. `BenchmarkIRCompile` and `BenchmarkInitFromLGB` are unaffected.

Follow-up on why nothing noticed the 47 skips: #914.

## Change

- `pkg/rt/core/test.lg`: the one-arity `is` expands to `(try-expr nil ~form)`, the same form the two-arity case produces, instead of recursing through its own name. Behaviour is identical for callers that refer `is`.
- `test/are_without_is_refer_test.lg`: a namespace that refers `are` and `testing` but not `is`, using `are` at top level and nested under `testing`. Fails on main with the error above, passes with the fix.
- Regenerated `core_compiled.lgb`, `generated.manifest`, `generated.sums`.

## Verification

On this branch: `make generate`, `make check-generated`, `go build ./...`, `go vet ./pkg/...`, `go test ./test/` (both with and without `-tags gogen_ir`), `go test ./pkg/rt/` (with and without `-tags bootstrap`), `go test ./pkg/compiler/ ./pkg/vm/`. `TestClojureTestSuite` totals: `files=242 assertions: pass=6311 fail=0 | skipped: compile=0 panic=0 runtime=0`.


mparrett (commented):
Self-review: no actionable findings.

The one-arity `is` expansion now names the already-defined `try-expr` macro directly, matching the two-arity expansion. The new regression test fails on the base commit with “Can’t resolve is in this context” and passes on this PR (5 assertions). The existing assertion reporting and form metadata tests pass, including a `gogen_ir` run. `make check-generated` and `make check-generated-manifest` both pass in an isolated checkout.

nooga (approved):
Reviewed and tested. Minimal, well-targeted fix: the one-arity is macro recursed through its own (unqualified at read-time) name, so it only resolved in namespaces that happened to refer is; expanding directly to (try-expr nil ~form) matches the two-arity case and fixes the 47/242 jank-suite files broken since #863. Verified locally: check-generated/check-generated-manifest clean, TestClojureTestSuite totals match exactly (files=242 pass=6311 fail=0 skipped=0), and go test ./test/ passes both with and without -tags gogen_ir.

nnunley (approved):

PR #916, opened 2026-09-19
fix(genmanifest): pin the manifest toolchain from go.mod, not GOTOOLCHAIN=local

nnunley:
## Why

`goModuleBuildInputs` fixes `GOOS`, `GOARCH`, `GOAMD64`, `CGO_ENABLED`, `GOFLAGS`, `GOWORK` and `GOENV` so that a committed manifest describes **one reference graph on every host**. It also fixed `GOTOOLCHAIN=local` — which reads like a pin but is not one.

`local` means *"whichever toolchain this host's `go` binary happens to be"*. That leaves the package graph host-dependent in exactly the dimension every other variable is pinned against: two contributors on different Go patch releases run the query against different toolchains.

It has a second, louder failure. Once `go.mod` requires a newer Go than the `go` first on `PATH`, `local` forbids the switch that would satisfy it:

```
go: go.mod requires go >= 1.27 (running go 1.26.5; GOTOOLCHAIN=local)
```

and that takes `make generate` and `make check-generated` down with it. A contributor whose version manager pins an older Go cannot regenerate at all — not a hypothetical, it is what #910 (Go 1.27) runs into locally.

## What

Pin the toolchain the module itself names. `goModToolchain` reads `go.mod`'s `toolchain` line and passes it as `GOTOOLCHAIN`; a module with no `toolchain` line falls back to `auto`, which honors the `go` line. Ambient `GOTOOLCHAIN` is still stripped, as before.

Every host now agrees on one toolchain, and a host whose `go` is older fetches it instead of failing.

## Verification

- `go test ./pkg/genmanifest/` — passes, including `TestDependencyManifestFresh`
- New `TestGoModToolchainPinsFromGoMod` covers all three cases: `toolchain` line wins, no `toolchain` line → `auto`, unreadable `go.mod` → `auto`
- New `TestGoModToolchainOnThisRepo` asserts this repo never silently degrades to `auto`
- `make check-generated` — every artifact in lockstep
- **Reproduction, then fix**: in a Go 1.27 worktree with `go1.26.5` first on `PATH`, `make generate` failed with the error above; with this change the same worktree regenerates every artifact successfully.

The change is package-graph neutral — regenerating moves only `depmanifest.go`'s own content hash in `generated.manifest`, and no dependency edges.

## Ordering

Independent of #910 and useful on its own, but it is what makes #910 usable for contributors on an older local Go, so landing it first avoids a window where regeneration is broken for them.


nooga (approved):
Reviewed and tested. Fixes a real bug I hit myself this session: GOTOOLCHAIN=local in the manifest generator's dependency query means whichever toolchain the host's go binary happens to be, not a pin, and it fails outright once go.mod requires a newer Go than what's on PATH. goModToolchain() reads go.mod's toolchain line and passes that as GOTOOLCHAIN (falling back to auto with no toolchain line), so every host agrees on one graph and an older local Go fetches what's needed instead of failing. Verified locally: clean rebase onto main, check-generated-manifest and check-generated both clean, and go test ./pkg/genmanifest/... passes including the new TestGoModToolchainPinsFromGoMod (all three cases) and TestGoModToolchainOnThisRepo.

Issue #873, opened 2026-09-13
`lint.lg`: R4 skips Go function bodies; `--gate comment-churn` fails with no added comments


mparrett:
Two defects in the R4 and R6 rules added by #870. Neither blocks that PR: `scripts/lint.lg` is not wired into CI, `make`, or the hooks, and `--churn` and `--gate` are opt-in flags nothing passes today. Both should be fixed before anything wires `--gate` into CI.

## R4 never measures Go function bodies

`form-end-offset` returns as soon as delimiter depth first reaches zero. For a Go declaration that happens at the closing `)` of `func F(...) {`, before the opening body brace is visited. `file-definitions` therefore sees every Go function as one line long and drops it via `r4-min-def-lines`.

Reproduced with 20 six-line Go functions: lint reports `only 0 definition(s) found`.

R4 is a no-op on Go sources, a silent false negative rather than a wrong answer, which is why nothing downstream notices it.

Fix: make the scanner language-aware, or take extents from the Go AST. Add a multiline Go-function regression test.

## `--gate comment-churn` fails even when a change adds zero comments

`r6-findings` emits a finding for every range with at least one added code line. It never tests the computed multiple against a threshold.

Reproduced from a baseline comment/code ratio of 0.333, adding one code-only function: the tool reports `multiple=0.0x` and exits 1 under `--gate comment-churn`.

Any change that adds code lines without comments hits this, so a CI job wired to the gate would fail on ordinary commits.

Fix: emit an R6 finding only above a defined abuse threshold, or make the measurement non-gateable. Cover the zero-comment case.

## Test coverage

The targeted suite in `test/lint_lg_test.go` passes and exercises neither failure mode. Both fixes want a regression test, or the next change to `form-end-offset` has nothing to signal that R4 is blind.

mparrett:
Agent-drafted, 2026-09-19, continuing this issue rather than opening a second one.

A third defect of the same shape, found while reviewing the comments on #926. It is a silent false negative like the two above, and it sits in a **gateable** rule, so it reaches `--gate` in a way the other two do not.

## R1 only sees a block that is entirely a code form

`comment-blocks` joins every run of contiguous comment lines into one block and concatenates their text. `commented-out-code?` tests that joined text with `whole-form?`, which requires the string to open with a delimiter and close with its match. One line of prose anywhere in the run fails that test, so a code form beside it is never examined.

Reproduced at `6462badb`, the same form in four positions:

```clojure
(defn alone [x]
[...]

PR #917, opened 2026-09-19
build: run the comment linter from make lint

nnunley:
> **Stacked on #910**, now merged. This PR's own delta is the three files below.

## Why

`scripts/lint.lg` has been report-only and run by hand — `docs/contributor-workflow.md` said as much — which in practice means it ran when someone remembered it. Wiring it into `make lint` makes it run as part of linting, which is what a linter is for.

There is a second benefit: the comment linter is itself a let-go program that walks the whole `.lg` and Go corpus. Running it routinely turns a runtime regression that breaks it into a visible failure to run, rather than silence.

## What

`make lint` now runs both linters. `make lint-go` and `make lint-comments` run either alone.

**This costs CI nothing.** The `lint` job invokes the `golangci-lint-action`, not this target, so the addition is local-only. The CI-gates table in `docs/contributor-workflow.md` is corrected to say so — its local equivalent is now `make lint-go`.

## It stays report-only, and what gating would actually require

Two things gate, and only one of them is the findings.

**Only four kinds are gateable at all.** The driver's `gateable-kinds` is `#{:commented-out-code :restatement :duplicated-comment :comment-density-outlier}`, and `--gate` intersects against it, silently ignoring any other name. Dividers, devlog phrases, churn and the entire code-pattern catalog are advisory: clearing them enables nothing.

| kind | findings | gateable |
|---|---|---|
| `commented-out-code` | 3 | yes — but see R1 below |
| `restatement` | 54 | yes |
| `duplicated-comment` | 116 | yes |
| `comment-density-outlier` | 306 | yes |
| `comment-divider` | 36 | no — advisory |
| `devlog-comment` | 3 | no — advisory |
| code-pattern catalog | 88 | no — advisory |

**R1 undercounts, so its 3 findings are a floor rather than a count.** `comment-blocks` joins consecutive comment lines into one block, and `whole-form?` requires the *joined* text to begin with an opener and end with its matching closer. One prose line above a commented-out form makes the block start with a word, so the form is never reported. Reproduced on identical input:

```clojure
;; (def dead-thing {:a 1 :b 2})        ; reported
(defn unrelated [q] (* q 7))

;; Kept for reference while the migration lands.
;; (def dead-thing {:a 1 :b 2})        ; NOT reported — same form
(defn unrelated [q] (* q 7))
```

Clearing the three visible hits would therefore not make `--gate commented-out-code` enforce the rule. Fixing or covering that case is a prerequisite to gating R1; both the guide and the Makefile comment now say so, referencing #873.

## Verification

- `make lint-go` — 0 issues
- `make lint-comments` — reports, exits 0
- `make lint` — runs both, exits 0
- `python3 scripts/docs_frontmatter_hook.py --check` — passes on both touched docs

## Review corrections (mparrett, 2026-09-20)

Both P2 items are addressed in `bd1a5f14`, and I verified both against the source before changing anything — `gateable-kinds` is indeed four kinds, and the R1 false negative reproduces exactly as described.


Makefile
```diff
@@ -401,10 +401,26 @@ ifneq (,$(wildcard .cache))
...
+
+lint-go:
 	@mkdir -p $(GOLANGCI-LINT-CACHE)
 	GOLANGCI_LINT_CACHE=$(GOLANGCI-LINT-CACHE) go tool golangci-lint run
 
+# Report-only, deliberately. Every rule currently fires somewhere (3 findings
+# for the narrowest, 306 for the broadest), so there is no kind that could be
+# gated today without first clearing it; `--gate KIND` is how to gate one once
+# it reaches zero. Running it here keeps it honest in a second way: the linter
+# is itself a let-go program, so a runtime regression that breaks it shows up
```
mparrett: [P2] Qualify the R1 gate plan while its false negatives remain. Issue #873 documents that `comment-blocks` joins adjacent prose and a commented-out `.lg` form, while `r1-findings` tests the entire joined block with `whole-form?`; the form is then missed. Clearing the three reported `commented-out-code` hits and adding `--gate commented-out-code` would not enforce the intended rule. Please make fixing/covering that case a prerequisite to the proposed first gate (or state that limitation here).

docs/comment-lint.md
```diff
@@ -19,18 +19,25 @@ It never edits source.
...
-hooks — you run it by hand. `--gate` is the only way to make it fail, and it
+finds things still exits 0. `make lint` runs it alongside golangci-lint, and
+`make lint-comments` runs it alone; it is in neither CI nor the git hooks, so
+it can never fail a build. `--gate` is the only way to make it fail, and it
 fails only on the rule kinds you name.
 
+Gating waits on the findings, not on the wiring: every rule fires somewhere
+today, from 3 findings for the narrowest kind to 306 for the broadest, so there
+is no kind that could be gated without first clearing it. Clear one and
+`--gate` it, one kind at a time.
```
mparrett: [P2] `--gate` cannot be enabled for every kind after its findings are cleared. The driver’s `gateable-kinds` contains only `commented-out-code`, `restatement`, `duplicated-comment`, and `comment-density-outlier`; it explicitly ignores `comment-divider`, `devlog-comment`, `comment-churn`, and the code-pattern catalog. This paragraph and the new Makefile comment imply any kind can be gated one at a time. Please distinguish the four eligible kinds from advisory-only reports and avoid suggesting that clearing a non-gateable kind is sufficient.

mparrett (changes requested):
Requesting changes to the proposed gating guidance. The current text suggests clearing any reported kind is enough to enable --gate, but the driver supports only four gateable kinds; it also names commented-out-code as the first candidate despite the R1 false negative tracked in #873. I left the specific corrections inline. The report-only Makefile wiring itself looks sound.

Issue #921, opened 2026-09-19
scope: a goroutine's failure reaches nothing — no exit signals, no restarts


nnunley:
## What happens

A scope supervises *lifetime* — it cancels work and waits for it. It does not supervise *failure*. When spawned work throws, nothing about it reaches the scope: the scope is not notified, `with-scope` returns normally, and no handler runs.

Where the error goes depends on which form spawned it, and the three disagree (all verified on current main):

| Spawned by | On a throw | What the caller can observe |
|---|---|---|
| `(agent …)` + `send` | agent enters a failed state | `(agent-error a)` returns it — as Clojure does |
| `(go …)` | `ExecutionError` printed to **stderr** | the result channel yields `nil`, indistinguishable from a block that returned `nil` |
| `(future …)` | **nothing printed at all** | `@f` is `nil`, now and on every later deref — see #805 |

Only the agent path is recoverable in code.

## Why it matters

The type is called a supervision tree and is described in `pkg/vm/scope.go` as the Go-native cousin of an Erlang supervision tree / Trio nursery / Loom `StructuredTaskScope`. All three of those deliver a child's failure somewhere a supervisor can act on it; this one does not. A reader who takes the analogy at face value will assume a failed worker is noticed, and it is not.

Concretely missing:

- **Exit signals.** No mechanism delivers a child's failure to its scope, so there is nothing for a supervisor to act on and no way to make sibling failure policy (fail-fast, fail-soft) expressible.
- **Restarts.** No restart strategies; a failed goroutine is not restarted and its siblings are not told.

Both are gaps rather than deliberate boundaries.

## Notes toward a design

- Trio's nursery re-raises a child's exception from the nursery block itself, which is the closest fit to `with-scope`'s shape and would give fail-fast for free.
- Delivery is the prerequisite for the rest: a per-scope handler, a failure channel, or re-raising at the boundary all need the failure to arrive somewhere first. Restart strategies are a layer on top and can follow separately.
- Interacts with #920: if cancellation becomes a condition absorbed at the scope boundary, that boundary is also the natural place for a child's *failure* to surface, and the two want one mechanism rather than two.
- The three spawn forms should end up agreeing, which likely means #805 is fixed as part of this rather than independently.

## Related

- #805 — `future` deref silently returns nil when the worker throws.
- #920 — cancelled blocking ops drop `ctx.Err()`.
- Documented as known gaps in `docs/guide/structured-concurrency.md`.

Issue #920, opened 2026-09-19
scope: cancelled blocking ops drop ctx.Err() instead of surfacing it


nnunley:
## What happens

When a scope is cancelled, blocking runtime operations return normally with `nil` and discard the cancellation. `sleep` (`pkg/rt/lang.go`):

```go
// Cancellable sleep: ... Matches Clojure where an interrupted sleep aborts.
t := time.NewTimer(d)
defer t.Stop()
select {
case <-t.C:
case <-ec.Context().Done():
}
return vm.NIL, nil
```

The comment states the intended semantics — an interrupted sleep aborts, as it does in Clojure — but the cancel arm returns `nil` with no error. Channel take does the same (`pkg/rt/async.go`): `case <-ctx.Done(): return vm.NIL`, dropping `ctx.Err()` outright.

The native signature is already `func(ec *vm.ExecContext, vs []vm.Value) (vm.Value, error)`, so the error slot exists and is unused on exactly the path that should use it. Per the callback error contract in `docs/contribution-policy.md` §4, a failure with an error slot available should route through it.

## Why it matters

A cancelled take is indistinguishable from a channel that legitimately yielded `nil`, so cancellation is unobservable except through `scope-cancelled?`. The practical consequence is that a loop bounded by time or iteration count does **not** stop on teardown — it runs to completion at full speed, because every `sleep` now returns instantly:

```clojure
(def n (atom 0))
(with-scope [s]
  (go (loop [i 0] (when (< i 100) (swap! n inc) (sleep 20) (recur (inc i)))))
  (sleep 50))
@n  ;=> 100    ;; with a (not (scope-cancelled?)) guard: 3
```

## Proposed

Cancellation becomes a condition that blocking natives raise through the error slot they already have, with three parts that need to land together:

1. **A distinct class**, outside `Exception`, in the spirit of Java's `Throwable → {Error, Exception}` split. let-go already has registered ancestry (`pkg/rt/hierarchy.go`) to express it.
[...]

Issue #919, opened 2026-09-19
with-open / finally: a failing release replaces the in-flight exception instead of being suppressed


nnunley:
## What happens

A cleanup that throws destroys the failure it was cleaning up after. Reproduced on current main:

```clojure
(try (with-open [r :not-a-channel] (throw (ex-info "body-failure" {})))
     (catch Throwable e (ex-message e)))
;=> "expected IOHandle, got let-go.lang.Keyword"
```

The body threw `body-failure`. What reaches the handler is the *release's* error. The original failure is not chained and not attached — it is gone.

The same holds for bare `try`/`finally`, which is what `with-open` expands to:

```clojure
(try (try (throw (ex-info "inner" {}))
          (finally (throw (ex-info "from-finally" {}))))
     (catch Throwable e (ex-message e)))
;=> "from-finally"
```

## Why it matters

Resource handling and recovery handling are meant to be separate channels: `finally` / `with-open` release things, `catch` decides what to do about a failure. Today the release channel can silently rewrite what the recovery channel sees, so the diagnosis you get is from the cleanup rather than the fault. It is worst exactly when it hurts most — a body that failed *because* the resource was bad will produce a cleanup failure too, and only the second one survives.

## Prior art

Java hit this and fixed it in try-with-resources: a close failure is **suppressed** — attached to the primary exception and reachable through `getSuppressed()` — rather than substituted for it. Clojure inherits that via interop, but its own `with-open` has the same replacement hazard let-go has.

## Proposed

- If an exception is in flight when a release throws, keep the primary and attach the release failure to it rather than replacing it. `ex-info`'s data map is the natural carrier (a `:suppressed` vector), since there is no `getSuppressed` to mirror.
- If nothing is in flight, the release failure is raised as primary, as it is today.
- Applies to `with-open` first; whether bare `finally` should do the same is a larger language-semantics question worth deciding explicitly rather than by accident.

## Related

- #918 — a `Closeable` protocol, which is what makes `with-open` reachable for user resources in the first place.

Issue #805, opened 2026-09-06
future: deref silently returns nil when the worker throws


nnunley:
## Reproduction

Confirmed with local let-go at `bdd8268c9cb3acf369f0854bada47af79d1d673d` while implementing StrongDM Attractor.

```sh
lg -e '(let [f (future (throw (ex-info "future-probe" {:probe true})))] (try (println {:returned (deref f 1000 :timeout)}) (catch Object e (println {:caught (ex-data e)}))))'
```

Actual: `{:returned nil}`. The exception is lost, indistinguishable from a successful nil result.

JVM Clojure comparison (also executed):

```sh
clojure -M -e '(let [f (future (throw (ex-info "future-probe" {:probe true})))] (try (println {:returned (deref f 1000 :timeout)}) (catch Throwable e (println {:caught (.getName (class e)) :cause-data (ex-data (.getCause e))})))) (shutdown-agents)'
```

Expected: dereference throws; JVM reports `java.util.concurrent.ExecutionException` with cause data `{:probe true}`. Exact Java wrapping need not be copied, but the failure and its data must remain observable.

## Implementation seam

`pkg/rt/lang.go`'s `futureStar` receives `v, err := child.Invoke(fn, nil)` but explicitly delivers `vm.NIL` when `err != nil`. `vm.Promise` currently stores only a value. Native `pmapv` already preserves worker errors and returns `vm.NIL, err` after joining, so this is specifically the future asynchronous boundary, not an absence of Go error-pair support.

## Acceptance

- Both ordinary and timed deref propagate a completed future's error.
- Successful nil remains valid and distinct from failure; pending timed deref retains its timeout value.
- Repeated derefs retain the error and exception data.
- Scope ownership/cancellation and successful future behavior remain intact; cover interpreted and AOT use.

Downstream currently catches inside the worker and transports a tagged error explicitly. That should be a removable workaround. Related #804 concerns `Func.Unbox()` callback conversion, a different boundary.

PR #922, opened 2026-09-19
docs(guide): document the scoped supervision framework

nnunley:
## Why

The scoped supervision framework shipped without user-facing documentation. The only tracked mention is `docs/design/exec-context-threading.md`, which argues the `Scope`/`ExecContext` merge as a design decision rather than teaching the feature — so there is currently no way to learn `with-scope` short of reading `pkg/vm/scope.go` and the tests.

## What

Adds `docs/guide/structured-concurrency.md` and its `docs/README.md` row. Every example in it was run against a build of current main; none is written from reading the source.

It covers `with-scope`, the cooperative nature of cancellation, nesting, the drain timeout, the full lg API, the Go embedding API, and how a scope travels on the `ExecContext` rather than the goroutine.

Two things it deliberately makes prominent, because both surprise people:

- **Cancellation is not an abort.** A cancelled take returns `nil` and the code after it still runs.
- **The consequence:** a loop bounded by time or iteration count does *not* stop on teardown unless it also tests `scope-cancelled?` — it runs to completion at full speed, because every `sleep` now returns instantly. Measured: 100 of 100 iterations without the guard, 3 with it.

## Gaps disclosed

Writing it turned up four unintended gaps. Rather than describe the framework as if it supervised failure, the guide states what actually happens and points at the tracking issue for each:

| | |
|---|---|
| A failed `future` loses its error entirely; a `go` block prints to stderr but its result channel also yields `nil`. Only agents surface a failure, via `agent-error`. | #805 |
| No exit signals and no restarts — a goroutine's failure reaches nothing that could act on it. | #921 |
| Cancellation is not observable as a failure: the cancel arms discard `ctx.Err()` though the natives have an error slot, and `sleep`'s own comment says it should abort as Clojure's does. | #920 |
| A failing cleanup replaces the in-flight exception instead of being suppressed, so release code can hide the fault it was cleaning up after. | #919 |

The guide also says plainly, in its opening, that a scope supervises *lifetime* and not *failure* — the word "supervision" otherwise implies a guarantee that is not there.

## Verification

`python3 scripts/docs_frontmatter_hook.py --check` passes on both files. Docs-only; no code changes.


docs/guide/structured-concurrency.md
```diff
@@ -0,0 +1,291 @@
...
+status: active
+last-verified: 2026-09-19
+authoritative-for:
+  - structured-concurrency
+  - scope-lifecycle
+---
+
+# Structured concurrency: scopes
+
+Every goroutine let-go spawns — a `go` block, a future, an agent send, a channel
```
mparrett: This overstates what a child scope owns. `async/pipe` (and several other async helpers) spawn with `vm.Goroutines.Go`, and timeout entries use `vm.Goroutines.Context()`, so they remain in the root scope even when created inside `with-scope`. For example, creating a parked `async/pipe` inside `with-scope` leaves `(scope-live s)` at 0; closing `s` neither cancels nor drains that worker. Please qualify the opening guarantee and identify which forms actually follow the current scope.

docs/guide/structured-concurrency.md
```diff
@@ -0,0 +1,291 @@
...
+supervise *failure*. Nothing about a throw inside spawned work reaches the
+scope: the scope is not notified, its live count is unaffected once the
+goroutine exits, and `with-scope` returns normally.
+
+Where the error goes depends on which form spawned the work, and the three
+differ:
+
+| Spawned by | On a throw | What the caller can observe |
+|---|---|---|
+| `(agent …)` + `send` | Agent moves to a failed state | `(agent-error a)` returns the error. The agent's value is unchanged. |
```
mparrett: `agent` and `send` are synchronous atom-backed placeholders in `core.lg`, not spawned goroutines. A throwing action stores an error for `agent-error`, but the agent does not enter a failed state: a subsequent `(send a inc)` still runs and updates its value. Please describe that behavior here and correct the later claim that agents behave as Clojure does.

mparrett (changes requested):
Two user-facing descriptions of scope and agent behavior need correction before merge. I verified both against the PR head and runtime probes.

nnunley:
@mparrett — both correct, both mine, fixed in `57bc390a`. I reproduced each before changing anything.

**Scope ownership.** You're right, and my opening sentence was an overgeneralisation from the one case I probed. Measured inside a `with-scope`:

```
go        -> scope-live 1
future    -> scope-live 1
pipe      -> scope-live 0
timeout   -> scope-live 0
async/map -> scope-live 0
```

`go*` and `future` spawn through `ec.Scope()` (`lang.go:2681`, `:3633`); the `async` helpers go through `vm.Goroutines` directly — 15 such sites in `async.go`, covering `pipe`, `map`, `reduce`, `split`, `pub`, `alts!` and `timeout`. The opening now says a scope is a supervision node without claiming every spawn joins it, names which forms do and which don't, carries that measured table, and states outright that closing a scope neither cancels nor drains an `async/pipe` started inside it. It's also listed under Known gaps, since it's a gap rather than a design choice.

**Agents.** Also right, and my "behave as Clojure does" was the sloppier error — I probed `agent-error` returning the error and generalised from that without checking what happens next:

```
(send ag (fn [_] (throw ...)))  -> @ag 0, agent-error set
(send ag inc)                   -> @ag 1, agent-error still set
```

The agent keeps accepting actions where Clojure refuses until `restart-agent`. The failure table now says `agent-error` is set but the agent does not stop, it's a Known gap in its own right, and the note that they're synchronous atom-backed placeholders rather than spawned workers is in there — that's the part that explains the rest.

I also swept for anything else the corrections contradicted and fixed two: "only the agent path is recoverable" now says the error is *readable* but the agent doesn't stop, and the "everything defaults to the root" gap distinguishes `go`/`future` from the `async` helpers.

Every number in the guide was re-run against a build of this head after the edits.

---
Delivery verified by local `pr-claim-gate`: exact PR/validated head `57bc390a22e98ddc9f6751b10ed464896a69b207`; relevance claims `docs/guide/structured-concurrency.md` occur in both the delivered delta and the PR feature delta; receipt `/Users/ndn/.local/state/pr-claim-gate/nooga-let-go/pr-922.json`.

docs/guide/structured-concurrency.md
```diff
@@ -0,0 +1,315 @@
...
+
+# Structured concurrency: scopes
+
+A **scope** is a node in a supervision tree: it owns a cancellation context
+derived from its parent's, and it knows how many goroutines are live beneath it.
+Cancelling a scope cancels its whole subtree and leaves its siblings running.
+
+**Not every spawn joins the current scope.** `go` blocks and futures do — they
+spawn through the calling execution's scope. The `async` namespace's own
+helpers do **not**: `pipe`, `map`, `reduce`, `split`, `pub`, `alts!` and
```
mparrett: `async/alts!` is incorrectly grouped with root-scope workers. It does not spawn a goroutine: `altsf` calls `reflect.Select` synchronously and, without `:default`, selects on `ec.Context().Done()`. Thus an `alts!` parked inside a scoped `go` block is released by closing that child scope. I verified `(go (reset! got (async/alts! [(chan)])))` inside `with-scope`: `got` was `:parked` before close and `nil` afterward. Please remove `alts!` from this root-spawn list and the Known gaps list, and describe it as a context-aware blocking operation.

mparrett (changes requested):
The two findings from my first review are fixed in this head. One new factual error remains in the expanded async scope description.

nnunley:
@mparrett — right again, fixed in `38233ec1`. And the way I got it wrong matters more than the line itself.

I probed `pipe`, `timeout` and `async/map` directly and got 0 for each. For `reduce`, `split`, `pub` and `alts!` I did **not** probe — I attributed `vm.Goroutines.Go` call sites to enclosing functions with an `awk` heuristic and wrote the result down as fact. The heuristic put `alts!` in the list because a spawn site sits nearby; `altsf` itself spawns nothing.

Confirmed your reading: it builds `reflect.SelectCase`s over its channels plus `ec.Context().Done()` (`async.go:78`) and calls `reflect.Select` synchronously (`:82`). Your probe reproduces here — inside `with-scope`, `scope-live` is 1 (the `go` block, not `alts!`), `got` is `:parked` before close and `nil` after.

So `alts!` is removed from the root-spawn list and from Known gaps, and now sits with `sleep` and channel take/put under cancellation, described as selecting over its channels *and* the current scope's `Done()`, released on close, yielding `nil`.

Having found my method unreliable I re-probed the rest rather than fixing only what you caught. `pipe`, `map`, `reduce`, `split`, `pub` and `timeout` are all 0 inside a scope — those five claims stand, now measured rather than inferred.

Three findings, three real errors, all from the same habit: generalising from one probe instead of testing each claim. The remaining statements in this guide are now each backed by a run I can point at.

---
Delivery verified by local `pr-claim-gate`: exact PR/validated head `38233ec12fa69c2a8990b5fa400ec951dbe59c0e`; relevance claims `docs/guide/structured-concurrency.md` occur in both the delivered delta and the PR feature delta; receipt `/Users/ndn/.local/state/pr-claim-gate/nooga-let-go/pr-922.json`.

mparrett (approved):
Reviewed current head `38233ec`. The three factual issues from my prior reviews are addressed: scope ownership is qualified, agent failure behavior is accurate, and `alts!` is correctly described as a context-aware blocking operation. I checked the relevant runtime paths and the reported CI checks are green.

Issue #913, opened 2026-09-19
bench-ratchet: InitFromLGB bar sits above #838's improvement, which #863 then overshot (residual is #897)


nnunley:
## Summary

`BenchmarkInitFromLGB` reads +2.6% allocs / +3.5% bytes against the `477a5d36` bar, which makes it look like steady drift. It is not. #838 improved the metric substantially, the bar was never moved down to follow, and #863 then overshot that improvement. The residual above the improved level is the eager per-def metadata cost already tracked in #897, measured *after* the fix that landed in #863 itself.

## Attribution

Bisected over `477a5d36..827ef18b`, with `pkg/rt/core_compiled.lgb` blob sizes alongside:

| commit | allocs/op | B/op | bundle |
|---|---|---|---|
| `477a5d36` (#778, the bar) | 13,995 | 1,016,164 | 311,302 |
| `ebd9a7bd` (#826) | 13,970 | 1,016,529 | 311,302 |
| **`83931029` (#838)** | **11,769** | **897,372** | **273,748** |
| `ebd758a6` (#898) | 11,902 | 906,080 | 276,152 |
| **`cffc09e1` (#863)** | **14,354** | **1,051,491** | **309,987** |
| `827ef18b` (main) | 14,354 | 1,051,532 | 309,987 |

## Reading

**#838 ("bootstrap ir.data after the bundle is written")** takes 2,201 allocations and 118,792 bytes out, alongside the bundle shrinking 37,554 bytes. This is a build-topology change — it alters what goes *into* the bundle — so it moved work rather than eliminating it, and an honest bar should say so.

**#863 (clojure.test port)** adds 2,452 allocations and 145,411 bytes on a bundle that grows 33,835 bytes. Allocation growth (+20.6%) outruns content growth (+12.3%) by about four points. That gap is the #897 signature: eager per-def metadata materialisation scaling with the number of defs, and clojure.test adds many. Note this is already *post*-fix — the initial figure was ~15,752 allocs, reduced to 14,354 within #863 by materialising `:ns` lazily.

So the "+2.6% since the bar" reading exists because the bar sits above a real improvement, not because the code steadily degraded.

## Why it matters now

`check` exits non-zero from the deterministic comparison before timing is considered (`cmd/bench-ratchet/main.go:460-470`), so this blocks unrelated pushes — including #910, a two-line `go.mod` version bump. See #912 for the other half (the `IRCompile` bytes bar, attributable to #764).

## Suggested resolution

[...]

Issue #912, opened 2026-09-19
bench-ratchet: IRCompile bytes/op bar is stale against #764's array-map representation trade


nnunley:
## Summary

The deterministic (`bytes/op`) bar for both `BenchmarkIRCompile` variants predates #764 and has blocked every local push since. The movement is entirely #764's insertion-ordered array-maps, it is a deliberate representation trade rather than a regression, and it *improves* allocation count. This asks for the bar to be rebased on its merits, which needs an owner's sign-off — a different action from "fix the regression".

## Attribution

Bisected over `477a5d36..827ef18b`. Both variants step at `638b4a6aa1ab` (#764) and nowhere else:

| commit | `[bytecode]` allocs/op | `[bytecode]` B/op | `[gogen_ir]` allocs/op | `[gogen_ir]` B/op |
|---|---|---|---|---|
| `3aa9f4ea` (#800) | 113,772 | 4,940,707 | 151,771 | 5,977,136 |
| **`638b4a6a` (#764)** | **105,855** | **5,232,425** | **143,855** | **6,269,244** |
| `827ef18b` (main) | 105,467 | 5,218,067 | 143,613 | 6,260,721 |

Allocation count falls 7,917 and 7,916; bytes rise 291,718 and 292,108.

## Why this is #764's representation change and not bundle content

`pkg/rt/core_compiled.lgb` is **byte-identical across `638b4a6a` and its parent** — 311,302 bytes on both sides, and #764 is not among the commits that touch the bundle. The decoder's input is unchanged, so the content contribution is exactly zero and the movement is runtime structure.

The mechanism predicts the measured shape: insertion-ordered array-maps replace many small per-entry nodes with one contiguous backing slice, so allocation count should fall and bytes should rise. Both variants agree to within one allocation, which is difficult to explain as anything but a single shared representation change.

## Why it matters now

`cmd/bench-ratchet/main.go:460-470` — `check` sets `exit = 1` from `compareDeterministic` against `machineIndependentBar` *before* the timing comparison runs, and `-force` refuses allocs/bytes measured worse unless `-accept-deterministic` is passed. So a timing rebaseline (#891) cannot clear this, and the gate currently blocks unrelated work: #910 is a two-line `go.mod` version bump that touches no allocation behaviour and cannot pass the pre-push gate.

## Suggested resolution

[...]

PR #926, opened 2026-09-20
fix(lower-go): keep statement-position `try` regions in the generated Go

### LINKED ISSUE #923: TestJankSuiteDirectABIGeneratedGo is red on main since #863, and no CI lane runs it
Fallout from the `clojure.test` port rather than anything wrong with it — the port is a big change and this is the kind of thing a large one shakes loose. Filing it because the gate is genuinely red and nothing is currently watching it.

`TestJankSuiteDirectABIGeneratedGo` fails on `main` at `f36a8c0b`, in a primary checkout with `test/clojure-test-suite` materialized:

```
inline Go AST oracle mismatch: expression target "TestIdenticalQmark": found 0, want exactly 1
--- expected
+++ generated
-ec.Invoke(rt.CachedVarFn(&__v_clojure_core_identical_QMARK_, "clojure.core", "identical?"), []vm.Value{x, y})
+ec.Invoke(rt.CachedVarFn(&__v_clojure_test_test_var, "clojure.test", "test-var"), []vm.Value{rt.LookupVar("clojure.core-test.identical-qmark", "test-identical?")})
```

It surfaced on #902, where it was correctly set aside as reproducing on plain `main` and unrelated to that change.

## Bisected to #863

| commit | result |
|---|---|
| `9d32c1c4` — parent, pre-#863 | pass |
| `cffc09e1` — #863, the `clojure.test` port | fail |

One caveat for anyone reproducing: `pkg/rt/core_go_lowered/` is generated rather than tracked, so a plain checkout reuses the previous commit's tree and the result means nothing. Both arms above were run with `rm -rf pkg/rt/core_go_lowered && make generate` first. My first attempt without that reported a spurious "missing package" failure on the parent, which is worth knowing before anyone repeats it.

## Not covered by #915

#915 fixes the other half of the same fallout, the one-arity `is` self-reference that left 47 of 242 jank-suite files failing to compile. This one is separate: at #915's head `c1c9cd76`, with a clean regenerate, the failure is byte-identical to the above. Flagging it only because "#915 clears the #863 fallout" is a reasonable thing to assume, and this would quietly stay behind.

## Why it went unnoticed

`make native-entry-gate` runs this test, and no workflow invokes that target. `.github/workflows/go.yml` runs `./test/e2e/` under explicit `-run` lists, and this test is in none of them — which the workflow itself warns about, in the comment above the expensive-lowering step: a test not named in the list "would never run in CI at all".

So the gate that proves the direct-ABI claim only ever runs by hand. That is the half I would most like to fix, since it is what let this sit rather than the change that tripped it.

## The open question

Whether the generated AST is now correct. After the port, `deftest` may well route through `clojure.test/test-var` by design, in which case the fixture's expectation is simply stale and the fix is a fixture update plus a CI lane. If that routing is not intended, the lowering regressed and the fix is larger. @nnunley, that one is yours — I did not want to guess at it, since the answer changes the size of the fix. Happy to do the fixture-and-CI version if that is the call.

## Reproduce

```sh
git checkout f36a8c0b
git submodule update --init test/clojure-test-suite   # primary worktree
rm -rf pkg/rt/core_go_lowered && make generate
go test ./test/e2e/ -run TestJankSuiteDirectABIGeneratedGo -count=1
```

Found while auditing the v1.13.0 cut (#718), where it belongs in the known-open-items until it is settled.

### LINKED ISSUE #925: lower_go: sibling `is` assertions of the same shape lower to a single block
## Symptom

Two `clojure.test/is` assertions with the same predicate shape in one fn body lower to Go with only one of them present. Each `do-report` call below is one site in the generated Go, counted on `main` at 261991ed with a clean `make generate`:

| body | `do-report` sites | expected |
|---|---|---|
| any single `is` | 3 | 3 |
| `(is (= x y))` then `(is (not (identical? x y)))` | 6 | 6 |
| `(is (identical? x x))` then `(is (identical? y y))` | 3 | 6 |
| the four assertions of `identical_qmark.cljc` | 3 | 12 |

In the two-assertion case the surviving block is the second one (its quoted `:expected` form is `(identical? y y)`). The four `PushBinding` calls for `*assertion-position*` are all still emitted, so the `binding` wrappers survive while their bodies do not.

## Repro

```clojure
(require 'ir.passes.pipeline)
(ns probe (:require [clojure.test :refer [is]]))
(def form '(defn g [] (let [x (hash-map) y (dissoc (hash-map :a 1) :a)]
                        (is (identical? x x))
                        (is (identical? y y)))))
(eval form)
(let [go (ir.passes.pipeline/lower-ns-to-go "probe" 'probe [form])]
  (println (count (re-seq #"\"do-report\"\)" go)) "do-report sites, want 6"))
```

```sh
bin/lg -source-paths . probe.lg
```

## What it is not

- Not the shared auto-gensyms: renaming every `__G__` symbol in the second expansion changes nothing.
- Not `binding` or `try` on their own: hand-written analogues (two `binding`+`try`+`let`+`if` blocks over `(list x x)` and `(list y y)`, calling ordinary fns) keep both blocks.
- Not the IR: the jank suite passes all 6311 assertions under `ir_bytecode`, and the same body run through `*ir-compile*` as bytecode reports every assertion. The loss is on the `lower_go` side of the pipeline.

## Where it shows

`TestJankSuiteDirectABIGeneratedGo` (#923). Once its harness selects the `:test` fn, the lowered fixture builds, then fails at run time with `invoke of nil`, which may be the same defect or #783. An `lg compile` of a namespace with two same-shaped assertions likely goes through the same path.

I am looking into this now and will post what I find; filing so the repro is on record and #923's fixture fix has something to wait on.

mparrett:
## The lowering defect (#925)

A `try` whose value nothing reads was dropped from the generated Go, along with everything only it used. `liveness/reachable-nids` seeds the DCE read-set from value-reading terminators and `:call` instructions; `:try` was not a root. A `let`-local captured only by the try's body closure was therefore dead, `try-assign-stmts` could not materialise the closure and returned nil, and `lower-block-stmts` treated that nil as "not live, skip" rather than as a failure. Nothing fell back to bytecode; the statement was gone.

Two `try`s in a row whose bodies each capture a different `let`-local lower to Go with zero call sites on `main`:

```clojure
(defn t2 [] (let [x (hash-map) y (hash-map)]
              (try (h x) (catch Throwable e nil))
              (try (h y) (catch Throwable e nil))
              :done))
```

`clojure.test/is` expands to exactly this shape (a `binding` around a `try`), which is how the jank direct-ABI gate found it after #863: the four assertions of `identical_qmark.cljc` lowered to a single block.

Two changes, each sufficient for its own half:

- `passes/liveness.lg`: `:try` seeds the read-set like `:call`. The region runs for its effects whether or not its value is read.
- `lower_go.lg` `lower-block-stmts`: a nil from lowering a `:try` is a failure that forces bytecode fallback, the same as for `:call`. With the liveness fix this arm is not reached by the repro, but a future shape that cannot lower will now fall back instead of vanishing.

`test/ir_lower_go_statement_try_test.lg` pins the repro: it fails on `main` (0 call sites, want 2) and passes here.

## The gate (#923)

`TestJankSuiteDirectABIGeneratedGo` needed three changes to run the port's `clojure.test`, none of which are about the lowering:

- The harness selected the `def`'s fn, which since #863 is the `test-var` wrapper. It now selects the `:test` fn from the `alter-meta!` form.
- The oracle expected a direct `identical?` call. `is` expands a predicate to `(apply pred (list args…))` so the report can carry the values, so no assertion calls a core predicate directly. The oracle now targets the body's `dissoc`, a direct core call the fixture still makes, with the falsifier swapped to `not-dissoc`.
- The generated runner followed the old shim (`set!` on a map of counters). The port binds `*report-counters*` to a ref and `inc-report-counter` commutes it, so the runner now binds a ref and reads it back. It also boots with `rt.LoadCore` instead of `rt.BootCore`: `BootCore` discards the bundle's chunk map, so the later `RequireNS("test")` found no chunk for the lazily marked namespace and returned its placeholder, leaving `report` and `test-var` nil (`invoke of nil` at run time).

The gate joins the expensive-lowering e2e lane in `go.yml`; it was in no `-run` list, which is why it sat red from cffc09e1 until #923.

## Verification

On this branch: `make generate`, `make check-generated`, `go vet ./pkg/...`, `go test ./test/` with and without `-tags gogen_ir`, `go test ./pkg/rt/` with and without `-tags bootstrap`, `go test ./pkg/compiler/ ./pkg/vm/ ./pkg/ir/`, `make native-entry-gate` (both halves pass, including the gate's own structural and semantic mutants), `make gogen-diff`, `make ir-stress-gate`.

Closes #925. Closes #923.


pkg/rt/core/ir/lower_go.lg
```diff
@@ -2635,7 +2635,10 @@
                              ;; an un-lowerable side-effecting call (e.g. push-binding!
                              ;; with a (var X) arg) — dropping it would change behavior,
                              ;; so the whole function must fall back instead.
-                             (if (and (not= op :call)
+                             ;; :call and :try run for their effects: a nil from their lowering is
+                             ;; a failure that must surface as bytecode fallback, never a silently
+                             ;; skipped statement (#925).
+                             (if (and (not (contains? #{:call :try} op))
```
mparrett, line 2640: [P2] Mirror this `:try` guard in `block-inst-stmts` as well. `lower-body` tries the structured `lower-tree` path first; its instruction loop at line 3191 still treats nil from any non-`:call` op as skippable when the value is dead. A statement-position `:try` has an unread value. In an isolated probe, I made the registered `:try` lowerer return nil and `lower-ns-to-go` emitted `ProbeTry` with the `try` body omitted, rather than falling back. The changed guard here runs only in the goto emitter. Applying the same `#{:call :try}` check to the structured loop, with a regression for the nil path, would complete the safety fix.

mparrett (commented):
Self-review: the new statement-position regression and direct-ABI gate pass, but I found one remaining silent-drop path in the fallback guard. Generated manifest provenance also passes.

pkg/rt/core/ir/lower_go.lg
mparrett: Fixed in 6462badb. `block-inst-stmts` now forces fallback when `:try` lowering returns nil, matching the goto emitter. Added a regression that forces this nil path; it failed before the fix and passes in default and `gogen_ir` modes. `make check-generated`, `make check-generated-manifest`, and `TestJankSuiteDirectABIGeneratedGo` pass.

mparrett:
Perf note for review. `make bench-ratchet` on this branch matches current `main` within run-to-run jitter: suite B/op 529.9 vs 530.5 MB, allocs 10.52 vs 10.54 M, `BenchmarkIRCompile` and `BenchmarkInitFromLGB` identical to four digits. The lowered core tree grows 4 KiB: one `try` region in `test-ns` that the old liveness rule dropped is back. That copy is not on an active path (a `-tags gogen_ir` `lg` runs `run-tests` correctly on `main` and here), so there is no behaviour change from it.

Two local gates are red on this branch, and identically red on plain `main` at 369e2a69: `make bench-ratchet` flags four deterministic rows against the checked-in `baseline.json` row 477a5d36, which predates #863 (the deltas are #863's own, see #912 and #913), and `make fanout-ratchet` reports 4819 vs a 4393 baseline whose module list is stale (`main` reports 4818). Neither runs on PR CI. Mentioning it so a local run does not read as this PR's doing.

PR #930, opened 2026-09-20
fix(bench-ratchet): gate a forced update on the global deterministic bar too

nnunley:
Unblocks #891. Found by @mparrett reviewing that PR; the repro here is theirs.

## Why

A timing-only `-force` update could silently **raise** the bar every profile is gated against, erasing a real regression.

`forceRebaseline` compared each measured deterministic metric against *this machine's* prior row only. A value can be an improvement over a stale local row and still be a regression against the newest row any profile carries — and since #900 made selection newest-provenance-wins, adopting and stamping it makes the worse value the bar. The regression then stops being reported by the gate that exists to report it.

Measured on #891's own data, before this change:

```
BEFORE (main)   bar = 13995 allocs / 1016128 bytes  since 477a5d36  ->  2 regressions
AFTER  (#891)   bar = 14353 allocs / 1051381 bytes  since 827ef18b  ->  0 regressions
```

Those are the two `InitFromLGB` regressions #913 was filed for this afternoon. A timing recapture would have retired the issue without fixing anything.

## What

Each deterministic metric is now judged against **both** bars, in one decision per metric so a violation is reported once. Either violation keeps the stored value and its date, and the rejection names whichever bar was exceeded.

A benchmark this machine has no prior row for is handled separately: the existing loop only visits benchmarks already in the local profile, so a new one would be published by `stampAll` as the newest provenance unchecked. There is no local value to fall back to there, so a metric over the bar keeps the bar's own value and date.

A rejection keeps the **local** stored value, never the bar's — writing another machine's number into this row would claim this machine measured it. What matters for correctness is only that the measured value is not adopted and not stamped as newest.

`-accept-deterministic` still records a regression deliberately, unchanged.

## Verification

`go test ./cmd/bench-ratchet/...` green, `go vet` and `gofmt` clean. All existing force tests pass unchanged, including `TestForcedKeepStampsAdoptedAllocsSeparately`, which pins that a metric better than both bars is still adopted and stamped with this run.

New: `TestForceRebaselineGatesDeterministicOnGlobalBar` is @mparrett's case — stale local row at 26294, global bar 13995 since `477a5d36`, a recapture measuring 14353 — asserting the value is not adopted, not stamped, and that `compareDeterministic` still reports 2 regressions afterwards. It fails on `main` with `worse-than-global values were adopted`.

Replayed against #891's actual capture (`clean-h.jsonl` at `827ef18b`): the forced update now refuses

```
! InitFromLGB [bytecode]  allocs/op  kept 13995   (since 477a5d36e25f), measured 14353
! InitFromLGB [bytecode]  bytes/op   kept 1016128 (since 477a5d36e25f), measured 1051381
```

while still writing the timing it was run for (`ns_per_op` 1501696 → 1128733), leaving 3 changed rows, `arm64/Apple M3` only, and no new stamp on the refused metrics. `baseline.json` is untouched by this PR.

`docs/perf/ratchet.md` states the two-bar rule in the `-force` section.


cmd/bench-ratchet/main.go
mparrett, line 1: Nonblocking test gap: could we add a regression test for this no-local-row branch? The new test covers a stale row on this machine, while this branch handles both a new machine profile and a new benchmark on an existing profile. I exercised both cases locally with regressed allocs/bytes; the global bar kept its prior value and provenance. A committed test would protect that behavior.

mparrett (commented):
I found no functional issue in the forced update path. The existing-row case keeps the local value and provenance when either bar is exceeded; the no-local-row fallback preserved the global bar in a local probe. `go test ./cmd/bench-ratchet/... -count=1`, `go vet ./cmd/bench-ratchet/...`, `gofmt`, and `git diff --check` passed on the PR patch. One nonblocking test coverage suggestion is attached.

mparrett:
@nnunley This looks ready. Anything by else needed?

nnunley:
I think this one is pretty solid.

mparrett (approved):
Approving. Previous ask is non-blocking and can be treated as a follow-up.

PR #933, opened 2026-09-20
fix(homebrew): migrate deprecated cask postflight metadata

nnunley:
Homebrew warns when loading the let-go cask because `postflight` is deprecated. Migrate it to declarative `postflight_steps`, using `run`, a macOS guard, and the `{{staged_path}}` template to preserve the existing quarantine-removal action.

Verified Ruby syntax and loaded the identical change in the local tap with `brew info --cask nooga/let-go/let-go` and its JSON output: no deprecation warning, and the serialized action retains the macOS guard and expected arguments.

The pre-push benchmark hook reports the existing deterministic allocation regressions tracked in #912/#913. The user explicitly approved bypassing that hook for this cask-only commit (23e7b305).


mparrett (approved):
I found no actionable issues in [[PR #933](https://github.com/nooga/let-go/pull/933)](https://github.com/nooga/let-go/pull/933). The new Homebrew step keeps the macOS guard, runs the same `xattr` command, and expands `{{staged_path}}` at install time, as documented in the [[Cask Cookbook](https://docs.brew.sh/Cask-Cookbook)](https://docs.brew.sh/Cask-Cookbook).

Ruby syntax passes, and Homebrew 7.0.6 serializes the step with the expected guard and arguments. I did not run a full cask installation.

Issue #897, opened 2026-09-18
perf(bytecode): decode def metadata lazily — readDefMetaPairs is 42% of InitFromLGB allocations


nnunley:
`BenchmarkInitFromLGB` spends about 42% of its allocations (≈5,024 of 11,900 allocs/op at `3bbde90a`) in `readDefMetaPairs`, materializing var metadata for the ~800 core defs on every bundle load. The leaves are interface boxing of repeated string-table keywords and symbols in `readValueV2`.

## History

| commit | PR | allocs/op | bytes/op |
|---|---|---|---|
| 1f281df4 | #654 | 7,391 | 762,860 |
| 885ae648 | #668 (var metadata survives the .lgb round-trip) | 25,669 | 1,559,355 |
| b0397f63 | #781 (DefMetaPairs) | 13,894 | 1,010,622 |
| 6fcadb2e | #859 | 11,900 | 906,016 |

Bisected on an Apple M3 (allocs are deterministic; one run per point). Bundle growth is a minor term: the bundle at tip is smaller than at #780.

## Proposal

Decode def metadata lazily: keep the raw bytes and string-table indices per var and materialize the map on first `doc` / `source` / `meta` access, the same way `NewLazySourceMapRaw` defers source-map decoding. Expected to remove nearly all of the 5,024 allocs/op and the corresponding bytes; a `vm.Var` accessor change is required so readers go through the lazy path.

Memoizing the boxed `vm.Value` per string-table index for keywords and symbols in `pkg/bytecode/decoder.go` was prototyped as a cheaper alternative: 11,900 → 9,857 allocs/op (−17%) at +1.8% bytes/op. Extending it to strings reached 9,497 but cost +3.9% bytes, over the ratchet's 2% bar. Lazy decoding supersedes both, so the prototype is not proposed.

Related: #791 (timing regressions since the #780 baseline); the ratchet's deterministic bar compares against the global minimum across machine rows captured at different commits, which is what surfaced this — tracked separately.

PR #941, opened 2026-09-20
Reduce IR compiler allocation and generated AOT code duplication

nnunley:
Reduce IR compiler allocation and emitted bytecode size, add quality reporting for generated AOT Go, and remove closure bodies duplicated across guarded native calls. Direct calls and Var-redefinition fallback remain intact.

The bytecode compiler uses enum-derived scalar type lookups and simpler use counting and emission. Export discovery now excludes declarations from function lowering, eliminating the misleading `not lowered (nil)` warning while preserving real failures. `make quality-aot` measures freshly generated Go independently of handwritten source.

Measured results:

- Emitted seven-function bytecode corpus: 676 → 644 bytes.
- VM IR compiler allocation: 5,237,051 → 4,598,179 B/op; 105,489 → 89,300 allocs/op.
- Generated Go: 159,775 → 143,331 lines; duplicated lines: 14,321 → 11,494.
- Generated erosion: 0.973 → 0.966; `BackEdges` cyclomatic complexity: 470 → 68. Source erosion remains approximately 0.592; substantial erosion work remains.

Validation passed: full short Go suite, native IR suite, generated-artifact freshness and native dispatch, native-entry/Jank gates, and execution tests for captured closures under normal and redefined dispatch. The normal pre-push gate also passed Go tests, Clojure compatibility, lowering coverage, and artifact freshness.

Benchmark exception: the normal push was rejected only by inherited `BenchmarkInitFromLGB` memory regressions against `477a5d36`: 13,995 → 14,352 allocs/op (+2.6%) and 1,016,128 → approximately 1,051,365 B/op (+3.5%). These are attributed to the earlier clojure.test/metadata changes in #913, with lazy metadata decoding tracked in #897. The user explicitly approved publishing exact revision `48c5aa34a4685e86ae9da54578cba79444aed6c4` with this failure disclosed. No baseline was relaxed. IR compilation passed its deterministic checks; standalone anchor-normalized compiler time improved 10.9% for VM execution and 25.1% for native execution. Six suite rows were new and had no comparison baseline.

This branch includes the standalone Jank lowering fix from #935, which should land first. No merge is requested by this publication. Handwritten Go is limited here to AST identifier traversal and regression tests; the compiler changes remain in let-go, consistent with the self-hosting direction.


mparrett (changes requested):
I found two reproducible issues that should be addressed before merge:

1. **[P2] Handle nil ASTs in `ident-names`** — `pkg/rt/gogen.go:317`

   `unboxNode(nil)` validly returns a nil node, but `ast.Inspect` requires a non-nil root and panics here. This is reproducible with:

   ```text
   build/lg -e '(do (require (quote gogen)) (gogen/ident-names nil))'
   ```

   Analogous AST query helpers return an empty result for nil. Please return an empty vector before calling `ast.Inspect`.

2. **[P2] Count the measured generated corpus, not marker-tagged files** — `scripts/quality.lg:890`

   In `--generated` mode every discovered file is measured with source-only terms omitted, but `:generated-included` counts only files carrying the generated-code marker. For an explicitly supplied unmarked corpus, the report can say `19 files tracked-by generated, 0 generated files included`. Please use the measured corpus count (for example, `(count loaded)`) or relabel the field as marker-tagged files.

Verification otherwise passed:

- `go test -short ./...`
- `make check-generated`
- shared-closure native guard/redefinition integration test
- generated lowered-tree compilation and native dispatch checks

nnunley:
Addressed both review comments in the latest push:

1. **Handle nil ASTs in `ident-names`** (`pkg/rt/gogen.go:317`):
   Guarded `cIdentNames` when unboxing a nil AST node to return `vm.EmptyList` (zero-allocation singleton), avoiding panic on `(gogen/ident-names nil)` while satisfying Clojure collection equality with `'()` and `[]`. Verified with `TestCIdentNamesNilAndEmptyNode` and `test/gogen_test.lg:ident-names-nil`.

2. **Count measured generated corpus** (`scripts/quality.lg:890`):
   Updated `:generated-included` to report `(count loaded)` in `--generated` mode instead of filtering on generated marker comments. Verified with `test/quality_generated_test.lg:generated-corpus-counts-unmarked-files-correctly`.

All gates (`make test`, `make check-generated`, `make test-gogen-diff-gate`) pass cleanly.

mparrett (approved):
Thanks for the follow-up. Nothing left on my side. Looks good to merge.

PR #948, opened 2026-09-21
fix(compiler): compile every value type a form can carry, and refuse the rest

### LINKED ISSUE #943: compileForm silently emits nothing for PersistentVector forms (no case, no default) — macro-folded vectors >32 elements panic at call time
`compileForm`'s type switch has no case for `vm.PersistentVectorType` and no `default:`, so a `PersistentVector` appearing as a form to compile emits **no bytecode at all**, silently. The compile succeeds; the function crashes when it is called.

Found from the consuming side in [nooga/xsofy#204](https://github.com/nooga/xsofy/issues/204) — a macro folding a constant into its expansion works for short inputs and panics for long ones, with nothing in between to explain why.

## Repro

let-go 1.13.0 (`369e2a6`), go1.27.1 darwin/arm64. Core only, no third-party code:

```clojure
(defmacro small [] (reduce conj [] (range 32)))   ; 32 -> stays ArrayVector
(defmacro big   [] (reduce conj [] (range 33)))   ; 33 -> promotes to PersistentVector

(defn f [] (small))
(defn g [] (big))

(println "32 elements (ArrayVector):     " (count (f)))
(println "33 elements (PersistentVector):" (count (g)))
```

```
32 elements (ArrayVector):      32
error: runtime error: index out of range [-1]
```

One element of difference. The `try`/`catch` form does not catch it — it is a Go panic that takes the process down.

## It is the type, not the size

`conj` promotes `ArrayVector` to `PersistentVector` above 32 elements, which is what makes this look like a size threshold. It is not:

- a **500-element `ArrayVector`** constant compiles and runs fine
- a **33-element `PersistentVector`** constant crashes
- the two 33-element vectors are `=` to each other and have identical element types (`let-go.lang.Int`)

```clojure
(def av (vec (range 33)))              ; ArrayVector
(def pv (reduce conj [] (range 33)))   ; PersistentVector
(= av pv)                              ;=> true
((eval (list 'fn [] av)))              ;=> works
((eval (list 'fn [] pv)))              ;=> index out of range [-1]
```

## Root cause

`compileForm` switches on `o.Type()` at [pkg/compiler/compiler.go:578](https://github.com/nooga/let-go/blob/369e2a69/pkg/compiler/compiler.go#L578):

| line | case |
|---|---|
| 579 | scalars |
| 583 | `vm.SymbolType` |
| 632 | `vm.ArrayVectorType` |
| 662 | `vm.MapType` |
| 696 | `vm.SetType` |
| 717 | `vm.ListType` |

No `vm.PersistentVectorType`, and **no `default:`**. An unhandled type falls straight out of the switch having emitted nothing, so the call site's stack is one value short and the VM eventually indexes at `-1`.

The reason this bites vectors and nothing else is that `PersistentVector` is the only persistent collection that reports a *distinct* `ValueType`:

```
pkg/vm/persistent_map.go:801    func (m *PersistentMap) Type() ValueType { return MapType }          // shared
pkg/vm/persistent_set.go:47     func (s *PersistentSet) Type() ValueType { return SetType }          // shared
pkg/vm/persistent_vector.go:104 func (v PersistentVector) Type() ValueType { return PersistentVectorType }  // distinct
```

So the `MapType` and `SetType` cases already cover their persistent twins for free — the vector case does not, and nothing said so.

## Same failure for the sorted collections

`SortedMapType` and `SortedSetType` are also distinct and also absent from the switch:

```clojure
(def sm (sorted-map :a 1 :b 2))
((eval (list 'fn [] sm)))   ;=> index out of range [-1]
```

So this is a class of bug, not one missing case. Any value type not enumerated in that switch is a silent miscompile.

## Suggested fix

Two parts, the second mattering more than the first:

1. Add `vm.PersistentVectorType` to the vector case (and the sorted types to their map/set cases). The existing case already walks elements and emits `(vector e0 e1 ...)`; it just needs to reach the elements through `Seq`/`Unbox` the way the `MapType` case already does for `PersistentMap`.

2. Add a `default:` that returns a compile error naming the type. [`parseBindingsVector`](https://github.com/nooga/let-go/blob/369e2a69/pkg/compiler/compiler.go#L1681) is already written this way — both vector representations handled, `default:` erroring — so the pattern is established in the same file. Without it, the next `ValueType` added anywhere in `pkg/vm` inherits the same silent failure.

A regression test asserting that a macro expanding to a 33-element `conj`-built vector compiles and runs would pin part 1; a test that an unknown `ValueType` produces a compile error rather than a panic would pin part 2.

## Why xsofy needs this

xsofy's `with-folded-salt` (`xsofy/hash.lg`) is a compile-time constant folder: it runs the salt-key encoder at macroexpansion time and splices the resulting byte vector into the emitted form. The encoder builds that vector with `conj`, so **every folded salt-key encoding to 33 bytes or more expands to a PersistentVector and crashes at runtime.**

That is not an exotic size. The macro's own docstring example, `[:combat :hit attacker target]`, encodes to 33 bytes the moment its variable terms are literals:

```clojure
(defn f [seed] (with-folded-salt seed [:combat :hit 12 34]))
(f 42)   ;=> index out of range [-1]
```

Shorter keys work, which is why the existing test — `with-folded-salt-equals-combine-fully-literal`, covering `[:combat :hit]` at 15 bytes — passes and has always passed. The fully-literal path is the one the macro exists to provide, and it is unusable above a threshold that nothing documents.

xsofy can work around it by forcing `vec` on the folded bytes before splicing, and will if this is not a quick fix, but the workaround is a cast whose only justification is this bug. Happy to send a PR for either part if you would like to point me at the shape you prefer.

nooga:
## Why

`compileForm` switched on `o.Type()` with no case for `vm.PersistentVectorType` and **no `default:`**, so a value whose type was not enumerated emitted no bytecode at all and returned no error. The compile succeeded and the call blew up much later indexing a stack one slot short, with nothing pointing back at the compiler.

`conj` promotes `ArrayVector` to `PersistentVector` past 32 elements, which is why a macro folding a conj-built vector into its expansion worked at 32 and panicked at 33 — a threshold that reads as a size limit but is a missing case. Reported from the consuming side in nooga/xsofy#204, where a compile-time constant folder emits a byte vector that crosses 33 bytes for realistic inputs.

`PersistentMap` and `PersistentSet` were never affected because they report the shared `MapType` / `SetType`. The types that report a *distinct* `ValueType` are the ones that fell through, and there were more than the vector: `SortedMapType`, `SortedSetType`, `RangeType`, `RepeatType`, `IterateType`, `SequenceType`, `TypeType`, plus every opaque runtime value (atom, volatile, delay, promise, transients, queue, ex-info).

## What changed

- **Vectors.** `ArrayVectorType` and `PersistentVectorType` share the case; both representations reach their elements through one `vectorFormElements` helper, replacing the inline `ArrayVector`/`MapEntry` unwrap. A vector form is not a constant — its elements are subforms — and that now holds for the promoted representation too.
- **Maps and sets.** The sorted variants join their case and rebuild with `sorted-map` / `sorted-set`, so both the type and the ordering survive a compile. A custom comparator is not expressible as any form, so it is **refused** rather than quietly rebuilt with the default ordering — that would be the same class of silent wrong answer. This is what the two new `UsesDefaultComparator` predicates in `pkg/vm` are for.
- **Seqs.** `RangeType`, `RepeatType`, `IterateType` and `SequenceType` join `ListType`. A seq is a call form; a lazy seq already behaved that way by reporting `ListType`, so this makes the seq representations agree instead of three of them miscompiling.
- **Type values** join the constant case. The symbol path already emitted host classes as constants, so this closes the gap between reaching a type through a symbol and splicing the value in.
- **`default:`** returns a compile error naming the type. This is the part that matters beyond today's list: the next `ValueType` added anywhere in `pkg/vm` now produces a compile error instead of inheriting the silence. `parseBindingsVector` was already written this way in the same file.

Also removed a commented-out "const vectors could be pushed as constants" block in the vector case, which the rewrite made unreachable.

## The other two compiler paths

- **`*ir-compile*`** — unaffected and already correct for the vector: `ir/build.lg`'s `build-form` dispatches on predicates (`vector?`, `map?`, `list?`), which already cover the persistent representations, and it already ends in an `:else` throw rather than falling through. Verified directly: an IR-compiled `defn` whose body is a 33-element conj-built vector returns 33 both before and after this change.
- **`-tags gogen_ir`** — builds on the IR path, and `make check-generated` exercises the lowered tree under that tag. Unchanged and passing.
- One **pre-existing** IR divergence surfaced while checking: an IR-compiled `defn` whose body is a spliced `SortedMap` returns a plain `let-go.lang.Map` — `ir/build.lg`'s `build-map` drops the sorted type. That is a separate bug in a separate path with its own generated-artifact footprint, so it is not in this PR. Worth its own issue.

## How it was verified

- `pkg/compiler/form_value_types_test.go` asserts the underlying contract rather than a list of symptoms: **compileForm either returns an error or leaves exactly one value on the stack**, checked over one representative value per form-reachable type (scalars, tagged literals, both vector representations, all three map and set representations, every seq representation, and the opaque runtime values). A refusal is an accepted outcome as long as it names the type. **16 of its subtests fail on the parent commit**, including the reported vector, both sorted collections, the seqs and every opaque value.
- A locally defined fake `ValueType` pins the `default:` arm, so the guard keeps working for types that do not exist yet.
- Alongside it: the issue's repro at 0/1/31/32/33/64/200 elements, element evaluation inside a promoted vector, sorted round-trips keeping type and order, the comparator refusal, and seq-as-call.
- `test/compile_form_types_test.lg` carries the same ground end to end from the language side; it fails on the parent commit with the reported `index out of range [-1]`.
- `make test` (exit 0, no failures), `make lint` (0 issues), `make check-generated` and `make check-generated-manifest` all pass. `core_compiled.lgb` is unchanged — only the manifest and sums move, because the Go sources this change touches are recorded generator inputs.

Closes #943


mparrett (changes requested):
The new generalized seq branch can hang the compiler indefinitely. `IterateType` is always unbounded, and one-argument `repeat` plus zero-argument `range` are unbounded too, but `compileForm` realizes every non-list seq until `Next()` returns nil (currently `pkg/compiler/compiler.go:753`).

Confirmed repro:

```clojure
(defmacro infinite-form [] (iterate inc 0))
(defn consume-infinite-form [] (infinite-form))
```

Compilation timed out with the goroutine stuck in that realization loop. Please refuse unbounded seq representations rather than realizing them—e.g. exclude `IterateType` and guard infinite `Repeat`/`Range`, accepting only known-finite sequences. This should be fixed before #948 merges.

PR #949, opened 2026-09-21
fix(ir): keep a sorted map sorted through the IR and Go lowering paths

nooga:
> **Stacked on #948** — base is `fix/compile-form-value-types`, because the bytecode-side fallback this relies on, and the `UsesDefaultComparator` predicates, land there. Rebase onto `main` once #948 merges.

## Why

A sorted map spliced into a macroexpansion came out of the IR path as a plain map — same entries, no ordering, no error. `=` does not catch it either, so the only symptom is a collection that stopped being sorted somewhere between the macro and the call.

```
(require 'ir.passes.pipeline)
(defmacro sm [] (sorted-map :c 3 :a 1))
(binding [*ir-compile* true] (eval '(defn b [] (sm))))
(println (b) (type (b)))
;; before: {:a 1, :c 3} let-go.lang.Map
;; after:  {:a 1, :c 3} let-go.lang.PersistentTreeMap
```

## Three layers, not one

The first read pointed at `build-map`, but fixing only that would have moved the loss one step downstream — each layer flattens the map independently:

| layer | what it did | what it does now |
|---|---|---|
| `pipeline/expand-all` | rebuilt every walked map with `(into {} …)`; `{}` is an array-map, so the type was gone **before** the IR builder saw the form | rebuilds into `(empty form)` |
| `build-map` | rebuilt a map with non-literal entries as an `(array-map …)` call | declines a sorted map it would have to rebuild |
| `lower_go/boxed-value-expr` | rendered any `map?` const with `vm.NewPersistentMap` — a plain map baked into an AOT binary | returns `nil`, the "not a literal" signal the `boxed-*-expr` helpers already use |

`expand-all` was the one causing the reported symptom; the other two only became reachable *because* of the first fix, so leaving them would have relocated the bug rather than removed it.

**Why decline rather than rebuild.** There is no call that reproduces a `sorted-map-by` comparator, so a sorted map that needs rebuilding cannot be emitted faithfully. Declining hands the fn to the bytecode compiler, which (#948) rebuilds sorted maps as sorted maps and refuses a custom comparator outright. An all-literal sorted map still takes the `:const` arm and carries the real collection, comparator included — so the common case stays fully IR-compiled, and the three outcomes match the bytecode path exactly:

| form | IR | result |
|---|---|---|
| all-literal sorted map | `:const` | type, order and comparator preserved |
| sorted map + subform | declines | bytecode rebuilds it as a sorted map, subform evaluated |
| `sorted-map-by` + subform | declines | bytecode refuses, naming the comparator |

## The `empty` fix in pkg/vm

`SortedMap.Empty` / `SortedSet.Empty` returned the shared default-ordered singleton, dropping the receiver's comparator:

```clojure
(let [m (sorted-map-by > 1 :a 3 :c 2 :b)]
  (into (empty m) (seq m)))   ;=> {1 :a, 2 :b, 3 :c}  — ascending, comparator lost
```

That is a bug on its own (Clojure's `empty` preserves it), and it is exactly what makes the `expand-all` rebuild faithful for `sorted-map-by`, so it is fixed here rather than worked around.

## Sets are deliberately untouched

`build-form` has no `set?` arm at all, so a set form throws and the fn falls back to the bytecode compiler — which is why sorted **sets** already came out right, and why `expand-all` having no `set?` branch is currently invisible. Giving the IR builder real set support is a feature with its own coverage and lowering consequences, not part of this fix.

## How it was verified

- `pkg/ir/lisp_sorted_map_const_test.go` covers each layer separately: `expand-all` keeps the type, the comparator, and still expands entries; `build-map` emits an intact sorted const and declines the rebuild; `lower_go` declines the const instead of rendering `NewPersistentMap`. Each has a plain-map control asserting the untouched path still behaves.
- `test/ir_sorted_map_test.lg` covers the same ground end to end through `*ir-compile*`, plus the `empty`-preserves-comparator contract on both sorted collections.
- **Three of the four Go tests and three of the five `.lg` deftests fail on the parent commit.** The fourth Go test characterises the `:const` arm the decline depends on, so it passes either way by design.
- `make test` (exit 0, no failures), `make lint` (0 issues), `make check-generated` and `make check-generated-manifest` all pass. `core_compiled.lgb` is unchanged — the `ir.*` namespaces load from source on `require` rather than from the bundle — so only the manifest and sums move.


mparrett (commented):
Reviewed #949 together with its stacked base #948. I found no additional correctness issue in #949: the comparator-preserving IR changes and focused/language suites pass. However, #948 has a blocking compiler-hang finding in its new generalized seq handling, so this PR should remain stacked until #948 is fixed and merged, then be rebased onto `main`.

nnunley (approved):
lgtm

PR #952, opened 2026-09-21
build(ratchet): require the perf gate only for changes that can move the numbers

nnunley:
## Why

The pre-push performance ratchet ran on every push (`always_run: true`). The stated rationale was sound — a Go runtime change or a regenerated bundle can regress the metrics with no benchmark file in the diff — but it has a failure mode the rationale did not anticipate.

The gate compares a fresh capture against a **committed** bar. So a bar that has gone stale fails *every* push from that machine, regardless of what the push contains. Today that blocked a two-line test addition (#930's review follow-up), a VCS fix, and a documentation change — none of which touch the IR or the VM — behind a baseline rebaseline they had nothing to do with. And the rebaseline itself could only be delivered through the same gate it existed to fix. #912 and #913 are that instance.

That is a structural fragility, not a one-off: any gate that blocks all work on a shared, drifting reference will eventually block work unrelated to the reference.

## What

Require the ratchet when the pushed range can change **what the ratchet measures**: the IR, the VM, and everything they transitively depend on.

Transitivity is computed, not approximated. `cmd/ratchet-scope` expands the benchmarked packages (`pkg/vm`, `pkg/ir`, `pkg/compiler`, `test` — mirroring the job list in `cmd/bench-ratchet`) with `go list -deps -test`, under **each build-tag configuration the ratchet captures**, since the `gogen_ir` build compiles a different file set and therefore a different closure. It then asks whether any pushed file falls inside that closure.

- `pkg/rt/lang.go` → **RUN**. Not a benchmark root, but the benchmarks compile it.
- `docs/`, `cmd/bench-ratchet`, `docs/perf/baseline.json` → **SKIP**. Nothing in the measured binaries compiles them.

A path regex could not express this without going stale the first time an import changed.

Inputs that imports cannot see are matched by path instead:

- `pkg/rt/core/`, `pkg/rt/core_compiled.lgb`, `pkg/rt/core_go_lowered/`, `pkg/rt/generated.*` — the runtime loads these; `InitFromLGB` decodes the committed bundle, so its content is an input to the measurement.
- `go.mod`, `go.sum`, `mise.toml` — the module graph and toolchain change generated code beneath every package. (Measured today: allocation counts were unchanged across go1.26.5 → go1.27.1 for these benchmarks, but that is a finding, not a guarantee.)

### Failure direction

Every ambiguity runs the ratchet: an undeterminable range, an empty change set, a build failure, an unrecognised decision. A false run costs one benchmark; a false skip defeats the gate.

The same reasoning drove two implementation choices worth noting in review:

- Matching is on the **directory tree**, not the exact package directory. An earlier version read `pkg/vm/testdata/fixture.edn` as out of scope; if a measured package embeds or reads that data, that is an under-trigger.
- The scope decision reads the **working copy**, not the last snapshot, so it describes the tree actually being pushed.

## Verification

End-to-end against a stubbed `make`:

- tooling/docs-only range → skips, exits 0, no benchmark
- a file under `pkg/vm` → invokes `make bench-ratchet`
- fresh checkout with no `build/` directory → works

**This PR's own push is the live test**: it was delivered through the modified hook, which reported `benchmark performance ratchet (IR/VM changes only)....Passed` without running benchmarks, while `go test`, the Clojure compatibility suite, `ir-stress-gate` and the generated-artifact check all still ran.

24 unit tests cover the decision table, path-boundary matching (`testdata/` must not match the `test` package), longest-match reporting, and a live `go list` assertion that the closure contains `pkg/rt` and excludes the gate tooling — that last one is the guard against `benchmarkRoots` silently narrowing if a benchmark moves.

## Deliberately not included

`ir-stress-gate` still runs unconditionally. It has the same `always_run` rationale and plausibly the same problem, but scoping it is a separate change with its own risk surface; the docs say so explicitly rather than leaving it implied.


mparrett (changes requested):
I found two gate-bypassing false skips and one range-selection issue that should be addressed before merge.

1. **[P1] Deleted dependency files are classified as `SKIP`.**

   `matchedDepFile` checks only the post-change dependency closure. A deleted compiled or embedded file is no longer present there. I reproduced this by deleting `pkg/vm/bench_ratchet_anchor_test.go` and passing that changed path to the scope tool; it returned `SKIP`, even though the change removes the ratchet's anchor benchmark. Please compute and union the dependency closures at the base and destination revisions, or conservatively run for deletions. A regression test covering deletion of a compiled/test file would pin this behavior.

2. **[P1] Direct benchmark inputs are missing from `pathTriggers`.**

   The Clojure suite benchmarks execute the vendored `test/clojure-test-suite` corpus, and `runSuiteOnce` directly compiles `test/compat/clojure/core-test/portability.lg`. Neither input is visible to `go list`. The new scope tool returns `SKIP` for both the portability shim and the submodule gitlink, even though either can change the measured workload. Please include these runtime-read inputs in the explicit triggers and add corresponding real-closure/decision tests.

3. **[P2] The hook inspects the current checkout, not necessarily the refs being pushed.**

   `changed_paths` ignores `PRE_COMMIT_FROM_REF` / `PRE_COMMIT_TO_REF` and instead diffs `@` or `HEAD`. A standard Git pre-push can push another ref or multiple refs, and a dirty working tree can differ from the pushed commit. In those cases the hook can classify an unrelated docs-only checkout and skip a VM-changing pushed ref. Please prefer the hook-provided ref pair (or raw pre-push updates) and fail closed when the actual pushed range cannot be resolved.

The added Go tests, `bash -n`, and `git diff --check` pass; these findings are behavioral gaps not caught by the current suite.

PR #953, opened 2026-09-21
fix(vcs): derive the recorded commit from the jj workspace, not the shared git HEAD

nnunley:
> **Stacked on #952.** The diff below includes #952's commit; review only the second commit, or read this after #952 lands. It is stacked because the pre-push ratchet is unconditional without #952, so this branch could not be delivered on its own.

## Why

A colocated jj repository keeps **one git HEAD for the whole repo**, tracking whatever the default workspace last exported. Every `jj workspace` therefore reports the same value from `git rev-parse HEAD` no matter which commit it is on. Observed here: a workspace sitting clean on `main` (`6cdeb6d0`) reported `efcd30925663` — a bookmarkless local commit that is not an ancestor of `main@upstream` and never will be.

The failure is silent and the recorded sha looks plausible, which is why it survived in three separate places.

### The one that ships

`Makefile:180` stamped `COMMIT` into `-X main.commit`, which becomes `let-go.semver/current-commit`, which is what SHA-pinned `require-letgo` constraints prefix-match against. The Makefile comment states the purpose explicitly — the stamp exists *so those pins can fire on `make` builds*.

So every `lg` binary built from any jj workspace has carried the wrong commit, and SHA-pinned constraints have been evaluating against a commit the binary was not built from. That is a correctness bug in shipped binaries, not a provenance wart. `PERF-SNAPSHOT` filenames inherited it.

`bench-ratchet` has the same defect in capture provenance, where the deterministic gate *selects its reference* by those stamps — so a wrong one attributes a measurement to code that did not produce it.

## What

`scripts/current-commit.sh` and `cmd/bench-ratchet` both ask jj for **this workspace's** commit: `@` when the working copy differs from its parent, the parent when it does not, since a clean checkout's `@` is an empty commit that describes nothing and is rewritten on every snapshot.

The result is an ordinary **git** sha, not a jj-only id — jj stores commits in the git object store, so `git cat-file` resolves them. That is the property a non-jj user needs in order to check the commit out and reproduce a run, so it is verified rather than assumed, with a `jj git export` and re-check first.

Resolvability and reachability are different questions: jj keeps local commits alive under `refs/jj/keep/<sha>`, which is never pushed. A run whose commit is not reachable from any remote-tracking ref now warns, because such a baseline cannot be reproduced by anyone who has not fetched it.

### jj is detected, never assumed

let-go's contributors and users do not use jj. Every jj path sits behind a marker or a `command -v` check with a git fallback, and plain git checkouts behave exactly as before. Verifying *that* claim turned up three defects in `scripts/quality`:

- **`os/sh` throws on a missing executable**, and the `sh`/`run` helpers only checked the exit code — so a checkout carrying a `.jj` marker on a machine with no `jj` **crashed** instead of falling back to git, the one configuration the fallback exists for. Both helpers now catch.
- **`head-commit` ignored its `root` argument** for the git path, answering for the process's working directory. Latent because `quality.lg` passes an ancestor of the cwd; both branches are now pinned with `-R` / `-C`.
- **`git-range`'s one-arity** defaulted to a bare `git rev-parse HEAD`, bypassing the very function that fixes this. Only a test used that arity, so no shipped report was mispinned, but it was a trap for the next caller. It now delegates, rooted by walking up for the nearest `.jj` **or** `.git` marker — `--show-toplevel` is wrong here, because a jj workspace has `.jj` but no `.git` and it walks past to the main checkout.

`head-commit` resolves `@-` unconditionally where the other two resolve `@` when dirty. That divergence is deliberate — a report's `:to` doubles as a `git rev-list` endpoint and wants a committed boundary — and is now documented rather than incidental.

## Verification

Four VCS configurations, since this checkout has both markers and both tools and can exercise neither fallback: git-only with jj installed; git-only with `jj` absent from `PATH`; a `.jj` marker with no `jj` binary (previously a crash); and a real jj workspace. The regression test builds a throwaway git repo for that reason.

The Makefile stamp moves from `07549b0e0` (stale shared HEAD) to the workspace's own commit. A test pins `current-commit.sh` against `jjWorkspaceSHA`: the Makefile cannot call Go, so the rule necessarily exists twice, and two copies drift silently — both would keep returning a plausible sha.

`make quality`: `issues 3255 -> 3255`, zero findings added or removed, every weighted metric unchanged at two significant figures. 51 tests in `cmd/bench-ratchet`, 14 in the quality suite.


mparrett (changes requested):
Blocking: the new resolvers assume plain `git` commands work from every jj workspace. A standard secondary workspace created outside the primary checkout has only `.jj`; it has no `.git` and no Git-owning ancestor. In an isolated colocated repo, I ran `jj workspace add /private/tmp/let-go-pr953-jj-ws`, moved that workspace to this PR commit, and reproduced both failures: `scripts/current-commit.sh` prints `none` (so `make build` leaves `let-go.semver/current-commit` unknown), while `go test ./cmd/bench-ratchet -run TestJJWorkspaceSHAResolvesThisWorkspace` fails because `jjWorkspaceSHA` returns empty. `jj log` itself correctly returns the workspace commit, and `jj git root` points to the primary checkout Git directory. Please resolve the underlying Git repository through `jj git root` and run the `cat-file`, export verification, reachability query, and Git fallback with that repository context in both `cmd/bench-ratchet/main.go` and `scripts/current-commit.sh`. A regression test should create the secondary workspace outside the primary checkout so this valid layout is covered.

PR #955, opened 2026-09-21
test(parity): compare which tests failed, not just how many

mparrett:
`gogen-parity.sh` reduces each suite to counts before comparing engines, so two engines can fail a **different set** of tests and be reported as PARITY. This adds a per-test identity comparison alongside the existing summaries, using data the logs already contain.

## The flaw today

The bucket half of the ir-stress comparison currently verifies nothing on a green corpus:

```
untagged   pass=842 fail=0 buckets=d41d8cd9  43.04s
gogen_ir   pass=842 fail=0 buckets=d41d8cd9  43.61s
```

`d41d8cd9` is the first eight characters of `md5("")`. With no failures the hashed section is empty, so both sides hash the empty string and that half of the oracle compares nothing to nothing — indistinguishable from a working check. This is the same class `require_ran` already guards one level up, where the comment calls it "the structural fix for the class of bug where success is inferred from the absence of output."

The counts half has a quieter version of the same problem: `jank_totals` greps the single `TOTALS:` line, so a file that skips for `compile` under one engine while a **different** file skips for `compile` under the other leaves TOTALS byte-identical.

## What changed

Two extractors and a comparison, wired into `check()` so identity is compared even when the summaries agree — agreeing counts being the case it exists to second-guess.

- `jank_identity` — each suite file is a `t.Run` subtest, so `go test -v` already prints one PASS/FAIL/SKIP line per file. SKIP is included: a file skipped for `compile` under one engine and `panic` under the other only shows in TOTALS if the counters happen to differ.
- `ir_identity` — the `<file>: <ok>/<total> ok` lines ir-stress already prints, compared in full rather than filtered to failures, so a fixture scoring 3/5 under one engine and 4/5 under the other is caught even though both are "a failure" by count.
- `compare_identity` — prints the symmetric difference on a mismatch, and reports **NO DATA** as a distinct state when both listings are empty rather than treating it as agreement.

No suite changes and no extra execution: the data was already being produced and discarded.

## Verification

`--selftest` falsifies the oracle rather than exercising it. It builds log pairs whose counts agree and whose per-test outcomes differ, then asserts that `compare_summaries` **is** fooled — the blind spot, demonstrated against the real function — while `compare_identity` is not. It also asserts that identical listings still agree, so the checker cannot pass by being always-red. No build or corpus needed; runs in about two seconds.

```
$ scripts/gogen-parity.sh --selftest
  PASS  counts agree on a real divergence (the blind spot, as designed)
  PASS  identity caught the differing failing test
  PASS  identity caught the per-fixture score swap
  PASS  two empty listings report NO DATA, not agreement
  PASS  identical listings agree (checker is not always-red)
selftest: all cases passed.
```

End-to-end on the real corpus, `make parity-check` now reports:

```
  jank: PARITY
  jank identity: 242 tests agree
  lower-go: PARITY
  lower-go identity: 9 tests agree
```

## What this does not find

Both engines are currently at zero failures, so this surfaces no divergence today. It is insurance that stays meaningful as failures appear, and it makes the "nothing was compared" state visible instead of hiding it behind a constant hash.

One residual limit worth stating: ir identity is per-fixture-file with its score, not per-assertion, so a divergence inside a single fixture that preserves that fixture's score would still pass. Finer than the current aggregate, not exhaustive.


nnunley (approved):
I have minor concerns about the test protocol. (the FAIL/PASS prefixes). Is it supposed to be managed with the same test harness as the rest of let go?

mparrett:
@nnunley Good call-out. Addressed this by running `--selftest` through the standard Go test/CI harness:
`test/gogen_parity_harness_test.go`

Exact amendment: `git diff 841126de f5d96ce9`

nnunley (approved):

